PR #10, opened 2022-10-10
feat: update to runtime version 0.2.2

Oleksiivanov:
Update to runtime version 0.2.2

Related:
https://github.com/camunda/team-connectors/issues/153


menski:
FYI we should also update the README with such changes, see https://github.com/camunda/connectors-bundle/commit/3d3e55c562760f327b61b6eb02696e300cd1c9f6

PR #12, opened 2022-10-12
feat: add Connector templates download script

### LINKED ISSUE #11: Download script for element-templates of all connectors in the latest version

As a Desktop Modeler user I want a convenient way to download the element-templates for all connectors in the latest version.
It is cumbersome to navigate to the currently 7 download locations and download the element-templates one by one, then place them into the model-elements folder

the connector bundles repository could host scripts so all element templates can be added by running it.
(Ultimately the modeler should offer a built-in installation feature)

Created own script:
`wget -O - https://raw.githubusercontent.com/rob2universe/model-element-templates/main/download.sh | bash`

tmetzke:
## Description

Provides a script that downloads all latest element templates matching the versions of the runtimes in the Docker image.

## Related issues

closes #11

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![CI](https://github.com/camunda/connectors-bundle/actions/workflows/CI.yml/badge.svg)](https://github.com/camunda/connectors-bundle/actions/workflows/CI.yml)
 
-The connectors bundle contains all out of the box connectors for Camunda 8. It's an easy way to try them out in your local setup or in k8s.
+The Connectors Bundle contains all out-of-the-box Connectors for Camunda 8. It's an easy way to try them out in your local setup or in k8s.
 
 The bundle contains the following components
 
@@ -60,6 +60,11 @@ docker build \
          --build-arg SLACK_VERSION=0.5.0  \         # Overwrite slack connector version
          -t camunda/connectors-bundle:${VERSION} .
 ```
+
+# Connector templates
+
+You can use the provided [download script](./download-templates.sh) to fetch the latest Connector templates of all out-of-the-box Connectors.
+Adjust the versions in the top of the script if you don't want to use the latest versions.
 
 # License
 
--- a/download-templates.sh
+++ b/download-templates.sh
@@ -1,0 +1,15 @@
+AWS_LAMBDA_VERSION=0.3.0
+GOOGLE_DRIVE_VERSION=0.4.0
+HTTP_JSON_VERSION=0.9.0
+SENDGRID_VERSION=0.11.0
+SLACK_VERSION=0.4.0
+SQS_VERSION=0.2.0
+
+wget -i - << EOF
+https://github.com/camunda/connector-aws-lambda/releases/download/$AWS_LAMBDA_VERSION/aws-lambda-connector.json
+https://github.com/camunda/connector-google-drive/releases/download/$GOOGLE_DRIVE_VERSION/google-drive-connector.json
+https://github.com/camunda/connector-http-json/releases/download/$HTTP_JSON_VERSION/http-json-connector.json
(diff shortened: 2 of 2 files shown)


menski:
Hey thanks for the script, but to be honest I'm not 100% sure if this is a good solution. What is our goal we want to achieve here? That the modelers don't provide the templates out of the box I assume, to be honest I think the solution should be that they do. So is this just a quick fix for a while? Then I would ask my self is this the place where people expect it, why would I check in this repository? Also who maintains this script, it adds a lot of duplication now where the versions and the list of connectors is maintained. Also as a user, what am I doing with the files? I would have to know where to put them correct?  

> all latest element templates matching the versions of the runtimes in the Docker image

Also this is easily inconsistent if a user uses an already released version of the connectors bundle, maybe an older one, and the script is used from the main branch which might have developed further. What I mean it's likely the script downloads the latest templates but it has no direct connection with docker image I run locally.

Sorry if it was to much of a brain dump, if we say this is the way forward for now I'm okay to merge it. I personally think it shouldn't be a long term solution that we promote to users. Feels like a very band aid like solution to me.

tmetzke:
> Hey thanks for the script, but to be honest I'm not 100% sure if this is a good solution. What is our goal we want to achieve here? That the modelers don't provide the templates out of the box I assume, to be honest I think the solution should be that they do. So is this just a quick fix for a while? Then I would ask my self is this the place where people expect it, why would I check in this repository? Also who maintains this script, it adds a lot of duplication now where the versions and the list of connectors is maintained. Also as a user, what am I doing with the files? I would have to know where to put them correct?
> 
> > all latest element templates matching the versions of the runtimes in the Docker image
> 
> Also this is easily inconsistent if a user uses an already released version of the connectors bundle, maybe an older one, and the script is used from the main branch which might have developed further. What I mean it's likely the script downloads the latest templates but it has no direct connection with docker image I run locally.
> 
> Sorry if it was to much of a brain dump, if we say this is the way forward for now I'm okay to merge it. I personally think it shouldn't be a long term solution that we promote to users. Feels like a very band aid like solution to me.

Thanks for those thoughts, @menski. All valid, I think. Especially the missing connection between the Docker image I might be running and the download script I am using. 👍 

I am also fine with not providing something like this at all here but rather pushing for a convenient solution within our modelers themselves. 

In the end, I think there is a certain need for a convenient method to fetch all templates at once, especially for trial users. Happy for any other ideas here. Or a better place to provide this (without any guarantees but rather as a shortcut people can/have to maintain themselves for now).

menski:
In my mind I have two random ideas:

1. Bundle the element templates in the docker image itself, i.e. while fetching the connector jars also fetch the corresponding element templates. The user could the use a `docker cp` command to extract the templates for the running container, has the disadvantage that the container has to run.
2. Modify the script in a way that is is always usable from the main branch, but instead of having the connectors listed explicitly only contain the version of the connectors bundle, which can also be set when run the script, and the the script fetches the Dockerfile for the given connectors bundle version and extracts connectors and version from the Dockerfile.

For both solution it would be really nice if the jar file contains the element template, but I assume that is not the case right?

tmetzke:
> For both solution it would be really nice if the jar file contains the element template, but I assume that is not the case right?

Correct, not contained in the JARs currently.

menski:
@tmetzke how we want to continue here? Just making progress and merge? Or thinking about different solutions?

tmetzke:
I am not all that certain anymore that this will really help more than it harms in this place here. I still think there is a need for it. But maybe not here. Let's discuss this in the team meeting maybe.

menski:
@tmetzke should we close this in favor of #14

tmetzke:
Closed as superseded by #14

PR #17, opened 2022-10-13
chore: publish templates with `-version` qualifier

nikku:
## Description

Ensures we publish templates with a `-$version` qualifier.

Example: If `slack-connector.json` has a `.version` field _THEN_ we rename it to `slack-connector-$version.json`.

Gracefully handles missing version (does not rename).
Gracefully handles templates packed as arrays (we should get rid of these).

## Related issues

Related to https://github.com/camunda/team-connectors/issues/170

### DIFF
--- a/bundle-templates.sh
+++ b/bundle-templates.sh
@@ -20,6 +20,25 @@ https://github.com/camunda/connector-slack/releases/download/${SLACK_VERSION}/sl
 https://github.com/camunda/connector-sqs/releases/download/${SQS_VERSION}/aws-sqs-connector.json
 EOF
 
+tag_version() {
+  local file=$1
+  local version=$(cat $file | jq '. | if type == "array" then .[0] else . end | .version')
+
+  local base_name="${file%%".json"*}"
+
+  if [[ "null" != "$version" ]]; then
+    echo "tag_version: Renaming $file -> $base_name-$version.json"
+
+    mv "$file" "$base_name-$version.json"
+  else
+    echo "tag_version: Keeping $file (unversioned)"
+  fi
+}
+
+for file in *.json; do
+  tag_version $file
+done
+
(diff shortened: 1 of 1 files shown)


.github/workflows/RELEASE.yaml
```diff
@@ -69,6 +69,9 @@ jobs:
           git tag -a "${tag}" -m "${message}"
           git push origin "${tag}"
 
+      - name: Install element template bundling dependencies
+        run: sudo apt install jq
+
```
tmetzke, line 74: I assume you tried that, but `jq` should already be provided in the ubuntu image this runs in. Just wanted to make we don't do this unnecessarily.

nikku: Is jq provided? I did not find documentation on it on the internets. I could of course double check with a GitHub action.

nikku: I can confirm that it is. Handy! https://github.com/nikku/testtest-checks/actions/runs/3248960228/jobs/5330799932

nikku: Will remove this bit.

tmetzke: People seem to use this: https://stackoverflow.com/questions/72752290/using-jq-to-dynamically-build-json-for-use-with-github-actions-strategy-matrix

nikku: Fixed via https://github.com/camunda/connectors-bundle/pull/17/commits/2ec8f7d70ec5b608f542fe4150c39679e3a65109.

tmetzke (approved):
Looks good to me 👍 
Let's try it and otherwise fix it :shipit:

nikku:
We gotta somehow get a `version` into our connectors. Anyone @camunda/connectors-dev wants to take a stab at it? :wink:

PR #21, opened 2022-11-14
Add Maven build for out-of-the-box connectors bundle

berndruecker:
Related to https://github.com/camunda/connector-sdk/issues/205

TODO: Check why this is not based on main as intended


berndruecker:
Replaced by https://github.com/camunda/connectors-bundle/pull/22

PR #22, opened 2022-11-14
Add Maven build for out-of-the-box connectors bundle

berndruecker:
https://github.com/camunda/connector-sdk/issues/255

As soon as the Maven build is executed, the resulting jar can be used for the bundle Dockerfile.
A Dockerfile bundling via classpath could be kept for documentation purpose for people that do not want to go down the Maven path


berndruecker:
@tmetzke Looks good from my end for now - I think you had feedback / change requests to it as well?

tmetzke:
Yes, I had, didn't get to that yet, unfortunately. It's on my list, trying to achieve that today. I'll try to have it done by Monday lunch time at the very latest, if that is fine? We are doing the "live" update next Tuesday together using this, right?

.github/workflows/CI.yml
```diff
@@ -31,6 +40,34 @@ jobs:
...
+        server-password: MAVEN_PASSWORD
+
+    - name: Deploy Maven Artifact
+      run: mvn -B compile generate-sources source:jar javadoc:jar deploy
+      if: github.ref == 'refs/heads/main'
+      working-directory: bundle/mvn/
+      env:
+        MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTORY_USR }}
+        MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTORY_PSW }}
+
```
tmetzke, line 63: Do we need to deploy this already? What are we using this for? If there is no need for this, let's not do this right now.
```suggestion
```

bundle/mvn/camunda-saas-bundle/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+                  <resource>META-INF/spring.tooling</resource>
+              </transformer>
+              <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/spring-configuration-metadata.json</resource>
+              </transformer>
+              <transformer implementation="org.apache.maven.plugins.shade.resource.AppendingTransformer">
+                  <resource>META-INF/io.netty.versions.properties</resource>
+              </transformer>
+              <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                <mainClass>io.camunda.connector.runtime.ConnectorRuntimeApplication</mainClass>
```
tmetzke, line 81: This needs to be the `SaaSConnectorRuntimeApplication`, right?
```suggestion
                <mainClass>io.camunda.connector.runtime.SaaSConnectorRuntimeApplication</mainClass>
```

tmetzke:
Looks good for a start, I think. I added some minor suggestions. Please rebase locally first, there are some updates to the main branch this needs to incorporate, I think. Please be careful to not lose any upstream changes.

Open questions in my head:
- Do we need to keep the `bash` variant of the bundle? Who will use this? Is this just for testing in the end?
- We still need to incorporate the new images into the release workflow, right? It's not necessary for the dev rollout probably but we shouldn't forget this.
- The is [jib-maven-plugin](the https://github.com/GoogleContainerTools/jib/tree/master/jib-maven-plugin) to containerize maven builds, maybe that is even easier than our current shading/exec whatever Frankenstein build 😄 But this might also be a future cleanup task for someone else. Just wanted to mention this as a variant here.

.github/workflows/CI.yml
berndruecker: Where is the problem to deploy a SNAPSHOT of something we plan to have? Better start doing it now and get into the rhythm? 
I think that you are right, that even a `install` should be sufficient as we need the Docker image right now, not yes the Maven artifact. I would still keep it as I find it correct :-)

berndruecker: I commented it for now and created https://github.com/camunda/connectors-bundle/issues/31

berndruecker:
Totally valid question with the bash bundle - created https://github.com/camunda/connectors-bundle/issues/32 for now

> We still need to incorporate the new images into the release workflow, right? It's not necessary for the dev rollout probably but we shouldn't forget this.

Yes, but I am not really sure what this takes exactly. Do we have a checklist for releases where we should add it?

tmetzke:
> > We still need to incorporate the new images into the release workflow, right? It's not necessary for the dev rollout probably but we shouldn't forget this.
> 
> Yes, but I am not really sure what this takes exactly. Do we have a checklist for releases where we should add it?

For now, I only meant we need to add this to the [RELEASE GHA](https://github.com/camunda/connectors-bundle/blob/main/.github/workflows/RELEASE.yaml) in this repository 🙂

PR #33, opened 2022-12-05
Connectors mono repo

### LINKED ISSUE #31: Re-enable `mvn deploy` for bundle
__What should we do?__

- Until we are sure we keep the Maven artifacts as designed, we did not want to deploy them on Artifactory, so we only `mvn install` locally for the Docker build
- Reintroduce a roper `mvn deploy` as commented in the Github workflow

__Why should we do it?__

- The artifact should be available via Artifactory & Maven Central

chillleader:
## Description

Moves the following connectors to the monorepo:
- aws-lambda
- sns
- sqs
- google-drive
- rabbitmq
- sendgrid

### Proposed Maven project structure

- connectors-bundle-parent
  - connector-runtime-bundle-parent
    - connector-runtime-bundle
    - connector-runtime-bundle-saas
  - connector-function-parent
    - connector-aws-lambda
    - connector-sqs
...

All projects have the same version which is managed in a single place - in the root `connectors-bundle-parent` POM in a `revision` property.

### Proposed updated build workflows

- `TEST` - on push to any branch, run tests & verify
- `DEPLOY` - on push to main:
  - Build & deploy snapshot artifacts
  - Build & deploy snapshot docker images
- `RELEASE`
  - Build & deploy release artifacts
  - Set new snapshot version
  - Optional: release docker images

### Progress tracker
- [x] Restructure project
- [x] Add connectors preserving the commit history
- [x] Adjust builds (in progress)
- [x] Documentation and licensing

## Related issues

related to camunda/team-connectors#243
closes #31


.github/workflows/DEPLOY.yaml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Lint Connector Bundle Dockerfile
+        uses: hadolint/hadolint-action@v2.0.0
+        with:
+          dockerfile: docker/bundle/bash/Dockerfile
+
+      - name: Lint Connector Bundle Dockerfile
+        uses: hadolint/hadolint-action@v2.0.0
+        with:
+          dockerfile: bundle/mvn/camunda-saas-bundle/Dockerfile
```
tmetzke, line 74:
```suggestion
          dockerfile: docker/bundle/mvn/camunda-saas-bundle/Dockerfile
```

.github/workflows/DEPLOY.yaml
```diff
@@ -0,0 +1,156 @@
...
+        with:
+          distribution: 'temurin'
+          java-version: '17'
+          cache: 'maven'
+          server-id: camunda-nexus
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+
+      - name: Deploy Maven Artifact
+        run: mvn -B compile generate-sources source:jar javadoc:jar install # TODO : Switch back to deploy
```
tmetzke, line 107: Since we have the Connectors in here now as well, let's switch to deploy 👍 
```suggestion
        run: mvn -B compile generate-sources source:jar javadoc:jar deploy
```

.github/workflows/RELEASE.yaml
```diff
@@ -43,6 +47,114 @@ jobs:
...
+
+      - name: "Build Changelog"
+        if: steps.get-tags.outputs.count > 1
+        id: changelog
+        uses: Requarks/changelog-action@v1
+        with:
+          token: ${{ github.token }}
+          tag: ${{ github.event.inputs.version }}
+          writeToFile: false
+          excludeTypes: build,docs,other,style,ci
```
tmetzke, line 148: We can actually adjust this, it was only needed for new repositories for new Connectors, and this will not be necessary anymore 👍
```suggestion
      - name: "Build Changelog"
        id: changelog
        uses: Requarks/changelog-action@v1
        with:
          token: ${{ github.token }}
          tag: ${{ github.event.inputs.version }}
          writeToFile: false
          excludeTypes: build,docs,other,style,ci
```

.github/workflows/RELEASE.yaml
```diff
@@ -43,6 +47,114 @@ jobs:
...
+
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v1
+        with:
+          body: ${{ steps.changelog.outputs.changes }}
+          tag_name: ${{ github.event.inputs.version }}
+          files: |
+            **/THIRD_PARTY_NOTICES
+            **/element-templates/*.json
```
tmetzke, line 157: There already is something like this at the bottom. We need to consolidate. Question to be answered: How do we add the collected element templates to the release? This is done by the `Bundle element templates` step. This needs some adjustment, I think. So we actually simply fetch the templates from the function modules. Currently, we base this on the `bash/Dockerfile` which doesn't make sense anymore.

connector/aws-lambda/README.md
```diff
@@ -0,0 +1,78 @@
...
+See also the [:lock:Camunda Cloud Connector Run-Time](https://github.com/camunda/connector-runtime-cloud) on how your function
+is run as a Google Cloud Function.
+
+## Element Template
+
+The element templates can be found in the [element-templates/aws-lambda-connector.json](../../element-templates/aws-lambda-connector.json) file.
+
+## Build a release
+
+Trigger the [release action](.githuborkflows/RELEASE.yml) manually with the version `x.y.z` you want to release and the next SNAPSHOT version.
```
tmetzke, line 78: I think we can remove this from the Connectors' READMEs

connector/aws-lambda/THIRD_PARTY_NOTICES
```diff
@@ -0,0 +1,22 @@
+Camunda AWS Lambda Connector
```
tmetzke, line 1: We are missing the THIRD_PARTY_NOTICE GitHub workflow for the Connectors. Question to be answered here: Do we want to have one file for all Connectors or one per Connector? I think the latter would be nice but we need to be able to handle this in a GitHub workflow as well.

pom.xml
```diff
@@ -0,0 +1,65 @@
...
+  </repositories>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
```
tmetzke, line 63: We should probably not add this here, this will run the shade plugin on EVERY child module, I think. Which is not what we want. The Connector function modules will add this themselves.

pom.xml
```diff
@@ -0,0 +1,65 @@
...
+  <description>Parent POM for out-of-the-box Connectors</description>
+  <inceptionYear>2022</inceptionYear>
+
+  <modules>
+    <module>connector</module>
+    <module>docker/bundle/mvn</module>
+  </modules>
+
+  <properties>
+    <revision>0.14.0-SNAPSHOT</revision>
```
tmetzke, line 27: Could you detail why we are doing this via a property?

.github/workflows/RELEASE.yaml
```diff
@@ -43,6 +47,114 @@ jobs:
...
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
```
tmetzke, line 160: Some of the steps below also need to be adjusted (cannot comment there). This was not properly adjusted unfortunately with the latest Docker image restructuring and we didn't create a follow-up. Basically, the paths need to be adjusted and different types of bundles need to be incorporated, just as in the other workflows (like TEST). If you'd like help with that, let me know. 🙂

tmetzke:
Can you please add a root README to the repository as well? 🙂

pom.xml
chillleader: The version plugin still works, we can use its `update-property` goal to set the revision. 

Maybe I'm wrong about the version plugin, my assumption was that we need to run it multiple times to achieve the same result as with this `revision` approach:

1. Bump the root pom version (`set` goal)
2. Bump the parent versions in each of the child poms (`update-parent` goal)

If that's the case, we'll need to adjust our build whenever we add a new module. With `revision`, on the other hand, we only need to manipulate the version in one place.

Please correct me if I'm wrong and we can do it in a single goal for the whole project :)

tmetzke: Usually, for multi-module projects, you simply run `mvn versions:set -DnewVersion=0.4.0` and Maven adjusts the parent and all child modules directly in one go. This is what we do in the SDK as well.

.github/workflows/RELEASE.yaml
```diff
@@ -1,11 +1,14 @@
...
 on:
   workflow_dispatch:
     inputs:
       version:
         description: 'Semantic version number of release: ^[0-9]+.[0-9]+.[0-9]+(-[a-zA-Z0-9.-]+){0,1}$'
         required: true
+      nextVersion:
+        description: 'Semantic version number to use as next SNAPSHOT version: ^[0-9]+.[0-9]+.[0-9]+(-[a-zA-Z0-9.-]+){0,1}$'
+        required: true
       release-runtime:
```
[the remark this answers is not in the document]
tmetzke: Putting more thought into this, I really think it would be best to remove the Connector Runtime Docker image from this mono repository. It has no relation to the other things here, actually. For now, we'd have the Connector Runtime Docker repository and this monorepo for Connectors. At some point, we can merge the Connector Runtime Docker repository into Spring Zeebe (currently not possible to release the Docker image to Docker Hub from the Community Hub repository).

pom.xml
chillleader: Oh ok. Then I agree that we don't actually need the revision approach.

tmetzke: Saves us to add the flatten plugin unnecessarily. We simply don't have a use case for the CI-friendly approach here, I believe. But I might also not understand the benefits entirely yet.

.github/workflows/RELEASE.yaml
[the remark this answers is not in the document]
tmetzke: https://github.com/camunda/connectors-bundle/issues/34

connector/aws-lambda/THIRD_PARTY_NOTICES
chillleader: Eventually I came up with the following configuration (see the new workflow):

- First per-connector THIRD_PARTY_NOTICES files are generated 
- Then an aggregate THIRD_PARTY_NOTICES file is generated for the whole project. This one then gets uploaded with the GitHub release.

.github/workflows/RELEASE.yaml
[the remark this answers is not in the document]
tmetzke: Connector Runtime repository is ready to go. Runtime can be removed here.

connector/aws-lambda/THIRD_PARTY_NOTICES
tmetzke: Cool, I'll have a look 👍

berndruecker:
Hey @chillleader: Just a maybe small remark - but I find it a bit confusing that the Runtime Bundle Maven project is beyond `docker` as a directory:

[grafik](https://user-images.githubusercontent.com/1217596/206137554-e6ba06a6-5d28-4926-8cd8-c297e24a370a.png)

But probably still the easiest for the moment 🤔

tmetzke:
> Hey @chillleader: Just a maybe small remark - but I find it a bit confusing that the Runtime Bundle Maven project is beyond `docker` as a directory:
> 
> [grafik](https://user-images.githubusercontent.com/1217596/206137554-e6ba06a6-5d28-4926-8cd8-c297e24a370a.png)
> 
> But probably still the easiest for the moment 🤔

What would you prefer here?

root
  - connector
  - bundle
    - bash
    - mvn
      - default-bundle
      - camunda-saas-bundle

I guess with the runtime removed, we could do something like that.

licenses/third-party.ftl
```diff
@@ -0,0 +1,48 @@
...
+  <#elseif license == "CDDLv1.0">
+    <#return "https://opensource.org/licenses/CDDL-1.0">
+  <#elseif license == "CDDLv1.1">
+    <#return "https://spdx.org/licenses/CDDL-1.1.html">
+  <#elseif license == "Bouncy Castle">
+    <#return "https://www.bouncycastle.org/licence.html">
+  <#else>
+    <#return "no known URL">
+  </#if>
+</#function>
```
tmetzke, line 30: There are some more licenses we already support, let's add their URLs here. At some point in the future, we should maybe even move this out of here completely and consume it from the SDK.

```suggestion
<#function getUrl license>
  <#if license == "Apache-2.0">
    <#return "https://opensource.org/licenses/Apache-2.0">
  <#elseif license == "MIT">
    <#return "https://opensource.org/licenses/MIT">
  <#elseif license == "BSD-3-Clause">
    <#return "https://opensource.org/licenses/BSD-3-Clause">
  <#elseif license == "EPL-2.0">
    <#return "https://www.eclipse.org/legal/epl-2.0/">
  <#elseif license == "EPL-1.0">
    <#return "https://www.eclipse.org/legal/epl-v10.html">
  <#elseif license == "GPLv2 with Classpath Exception">
    <#return "https://www.gnu.org/software/classpath/license.html">
  <#elseif license == "GNU Lesser General Public License, Version 2.1">
    <#return "http://www.gnu.org/licenses/lgpl-2.1.html">
  <#elseif license == "CDDLv1.0">
    <#return "https://opensource.org/licenses/CDDL-1.0">
  <#elseif license == "CDDLv1.1">
    <#return "https://spdx.org/licenses/CDDL-1.1.html">
  <#elseif license == "Bouncy Castle">
    <#return "https://www.bouncycastle.org/licence.html">
  <#elseif license == "Go License">
    <#return "https://go.dev/LICENSE">
  <#else>
    <#return "no known URL">
  </#if>
</#function>
```

.gitignore
```diff
@@ -1,4 +1,64 @@
...
-.idea
-target
+*.tar.gz
+*.rar
+
+# virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
+hs_err_pid*
+
+HELP.md
+connector/aws-lambda/target/
```
tmetzke, line 26:
```suggestion
target
```

connector/aws-lambda/THIRD_PARTY_NOTICES
tmetzke: Looks good to me. It's a bit sad that we lost the header for each individual Connector but it's also fine, I think. Having both individual files to link to and an overall file to distribute is nice, I think. 👍 

Can you create a follow-up to adjust the public docs at https://docs.camunda.io/docs/reference/dependencies once we're done here? We need to adjust the links to point to the repository here. I think we can remove the individual Connectors from that page even, linking only to the overall file and hinting to look at individual files in the subdirectories.

tmetzke:
> root
> 
> * connector
> * bundle
>   
>   * bash
>   * mvn
>     
>     * default-bundle
>     * camunda-saas-bundle
> 
> I guess with the runtime removed, we could do something like that.

@chillleader, can you incorporate this? If you have any questions, happy to help 🙂

.github/workflows/DEPLOY.yaml
```diff
@@ -0,0 +1,149 @@
...
+        env:
+          MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTORY_USR }}
+          MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTORY_PSW }}
+
+      - name: Build Bundle Artifact
+        run: mvn -B compile generate-sources source:jar javadoc:jar deploy
+        working-directory: docker/bundle/mvn/
+        env:
+          MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTORY_USR }}
+          MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTORY_PSW }}
```
tmetzke, line 51: Should we simply build all modules at once and deploy?
```suggestion
      - name: Build Artifacts
        run: mvn -B compile generate-sources source:jar javadoc:jar deploy
        env:
          MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTORY_USR }}
          MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTORY_PSW }}
```

.github/workflows/DEPLOY.yaml
```diff
@@ -0,0 +1,149 @@
...
+          server-id: camunda-nexus
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
+
+      - name: Deploy Maven Artifact
+        run: mvn -B compile generate-sources source:jar javadoc:jar deploy
+        if: github.ref == 'refs/heads/main'
+        env:
+          MAVEN_USERNAME: ${{ steps.secrets.outputs.ARTIFACTORY_USR }}
+          MAVEN_PASSWORD: ${{ steps.secrets.outputs.ARTIFACTORY_PSW }}
```
tmetzke, line 111: I might be wrong here, but I think we can scratch this, as we're deploying in the `deploy-snapshots` job already, aren't we?

tmetzke: Ah, we might not have the `target` folder for the MVN-based Bundles then, right?

connector/aws-lambda/THIRD_PARTY_NOTICES
chillleader: https://github.com/camunda/team-connectors/issues/256

tmetzke:
I think everything's done so far, right? 

Can you check the Connectors you already integrated so far to make sure we don't have anything left open in there that we need to take over here as well? Version updates of the SDK and Cloud Runtime to `0.4.0` alphas in the original OOTBs can be ignored.

I can have a final look after lunch again.

As a follow-up, we'll have to look into the workflows again probably, especially the deploy and release workflows.
But we can fix them after the general merge to unblock this here and make progress.

chillleader:
> Can you check the Connectors you already integrated so far to make sure we don't have anything left open in there that we need to take over here as well? Version updates of the SDK and Cloud Runtime to 0.4.0 alphas in the original OOTBs can be ignored.

Fortunately nothing new there apart from version bumps 😄

tmetzke:
> Fortunately nothing new there apart from version bumps 😄

Awesome! People seem to have listened 😄 
Any open PRs that are relevant (dependency updates don't count)?

chillleader:
Just for the record: connector repositories were consolidated by merging them into this repository and then relocating them to their new corresponding directories. Therefore the original commit history is retained, however, not all software will show this history correctly by default.

For instance, @tmetzke reported that his git client doesn't show the original history of the connectors files. It turned out that a specific option called _Follow renames_ must be enabled to display the original history.

On the other hand, IntelliJ IDEA handles this correctly by default, as well as GitHub web UI. But if you seem to be missing the original history, try to look for a similar config option in your git client.

chillleader:
> @chillleader, are all upstream changes from `main` incorporated here? I saw that Bernd adjusted some things for the SaaS Bundle, we must not lose those.

Yes, it's 2 commits behind but these are non-conflicting and will be merged correctly

Issue #37, opened 2022-12-07
Validate workflows for monorepo


chillleader:
__What should we do?__

- Make sure workflows are correct and work as expected
- Do an alpha release
- Update team documentation

tmetzke:
Regarding documentation, we also need to fix this repository regarding links to the Connector Runtime. It currently points to the now-removed `runtime` module. Can you tackle that? 🙂

tmetzke:
By the way, can we also adjust https://github.com/camunda/connector-runtime-docker/blob/main/.github/workflows/RELEASE.yaml in the same way, maybe? Just to be consistent.

PR #43, opened 2022-12-09
fix: release workflow

chillleader:
## Description

* Handle pre-release versions automatically (mark as pre-release in GitHub, skip uploading images with latest tag)
* Clean up `bundle-templates.sh`
* Remove extra input flags: `release-runtime`, `release-bundle`, `push-latest`
* Handle missing license URLs in THIRD_PARTY_NOTICES

## Related issues

related to #37

### DIFF
--- a/.github/workflows/RELEASE.yaml
+++ b/.github/workflows/RELEASE.yaml
@@ -173,6 +173,24 @@ jobs:
           push: ${{ github.ref == 'refs/heads/main' }}
           tags: camunda/connectors-bundle-saas:latest
 
+    # Update README in Dockerhub
+
+      - name: Push README to Dockerhub - bundle-default
+        if: ${{ steps.check_prerelease.outputs.isPreRelease == 'false' }}
+        uses: christian-korneck/update-container-description-action@v1
+        with:
+          destination_container_repo: camunda/connectors-bundle
+          provider: dockerhub
+          readme_file: bundle/README.md
+
+      - name: Push README to Dockerhub - bundle-saas
+        if: ${{ steps.check_prerelease.outputs.isPreRelease == 'false' }}
+        uses: christian-korneck/update-container-description-action@v1
+        with:
+          destination_container_repo: camunda/connectors-bundle-saas
+          provider: dockerhub
+          readme_file: bundle/README.md
+
     # Create GitHub release
 
       - name: Bundle element templates
--- a/.github/workflows/RELEASE.yaml
+++ b/.github/workflows/RELEASE.yaml
@@ -1,4 +1,4 @@
-name: Build and Push Release Artifacts and Docker Images
+name: Release a new version
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/RELEASE.yaml
```diff
@@ -101,6 +89,16 @@ jobs:
...
+      - name: Check if new version implies pre-release
+        id: check_prerelease
+        run: |
+          if [[ ${{ github.event.inputs.version }} =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            PRE_RELEASE=false
+          else
+            PRE_RELEASE=true
+          fi
+          echo "isPreRelease=${PRE_RELEASE}" >> $GITHUB_OUTPUT
+
```
tmetzke, line 101: I like that, we should maybe just use that in the SDK as well. Currently, we are explicitly asking for alpha release there. This here makes more sense 👍

tmetzke (approved):
Awesome, I think this should work! 👍 

I have one additional suggestion that we can try:
Can we add something like this as well:
```yaml
      - name: Push README to Dockerhub - bundle-default
        if: ${{ steps.check_prerelease.outputs.isPreRelease == 'false' }}
        uses: christian-korneck/update-container-description-action@v1
        with:
          destination_container_repo: camunda/connectors-bundle
          provider: dockerhub
          readme_file: bundle/README.md

      - name: Push README to Dockerhub - bundle-saas
        if: ${{ steps.check_prerelease.outputs.isPreRelease == 'false' }}
        uses: christian-korneck/update-container-description-action@v1
        with:
          destination_container_repo: camunda/connectors-bundle-saas
          provider: dockerhub
          readme_file: bundle/README.md
```
This will (hopefully) adjust the descriptions of the Docker hub repositories with a release as well. Right now, they are quite outdated already. I couldn't test this yet, unfortunately. We could also do it as a follow-up, would be fine with me. 🙂 

Regarding the `bundle-templates` script, I can't really say if this will work as expected. It might be, but I actually don't fully understand it yet. Happy to take your explanation on it if you've fully understood it. 😄

chillleader:
> Regarding the bundle-templates script, I can't really say if this will work as expected. It might be, but I actually don't fully understand it yet. Happy to take your explanation on it if you've fully understood it. 😄

The [result](https://github.com/camunda/connectors-bundle/releases/tag/0.14.0-alpha2) seems to be correct  😄 

Currently it just creates a temporary directory, copies all element templates to it (I changed it in the previous PR - it used to download them from GH releases of each connector repo), then adds the template version to file names and packs it into an archive.

tmetzke:
> The [result](https://github.com/camunda/connectors-bundle/releases/tag/0.14.0-alpha2) seems to be correct 😄

Works for me 👍

.github/workflows/RELEASE.yaml
chillleader: Created https://github.com/camunda/connector-sdk/issues/305

tmetzke: Thank you!

tmetzke:
We can try out the README update from this branch here (removing the `if`s) or merge and try it on main, I don't mind.

chillleader:
> We can try out the README update from this branch here (removing the ifs) or merge and try it on main, I don't mind.

The README itself is not release-ready and will need to be adjusted a bit, so I'd prefer to try it out later and fix it on the spot if it breaks.

PR #87, opened 2022-12-13
ci: deactivate dockerhub readme updates

tmetzke:
## Description

Deactivate Dockerhub readme updates, as login does not work yet (cannot reuse docker/login action).

## Related issues

n/a


igpetrov (approved):
Once we release, what do we need to do with it?

tmetzke:
We need to adjust the login or remove it completely

PR #96, opened 2022-12-14
fix: update for bearer token sent with quotes - rest connector

mbiciin:
## Description

Update for bearer token sent with quotes - rest connector


igpetrov (approved):
let's test it in SM and then merge

tmetzke:
I assume this is something we can conveniently unit test as well, right?
I cannot see any test that actually verifies the specific payload or headers that will be put into the requests when using OAuth.

igpetrov:
Ah, sorry, @tmetzke, saw your comment after I clicked merge.
Added requirement for unit test at the ticket https://github.com/camunda/connectors-bundle/issues/101

PR #97, opened 2022-12-14
chore(logging): Fixed logging dependencies and added test for Spring …

berndruecker:
The new test will create a Spring context to find obvious problems during build time. 
Closes https://github.com/camunda/team-connectors/issues/251.

Also ironed out the logging profile - `dev` will log to local console - everything else use GCP stuff as default.


berndruecker:
@chillleader I would merge this if OK for you?

tmetzke:
Do we need to adjust the logback version still or does this come with the Spring Boot BOM now already?

igpetrov (approved):

PR #98, opened 2022-12-14
Prometheus

berndruecker:
Add prometheus support via Spring Boot Actuator, so that Metrics can be scraped in Prometheus format under http://localhost:8080/actuator/prometheus or http://localhost:9080/actuator/prometheus depending on the port configured

### DIFF
--- a/bundle/mvn/default-bundle/src/test/resources/application.properties
+++ b/bundle/mvn/default-bundle/src/test/resources/application.properties
@@ -2,3 +2,7 @@
 server.port=9898
 zeebe.broker.gateway-address=localhost:26500
 zeebe.client.security.plaintext=true
+
+management.server.port=9080
+management.context-path=/actuator
+management.endpoints.web.exposure.include=metrics,health,prometheus
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![CI](https://github.com/camunda/connectors-bundle/actions/workflows/DEPLOY.yaml/badge.svg)](https://github.com/camunda/connectors-bundle/actions/workflows/DEPLOY.yml)
 
-This repository manages the [out-of-the-box Connectors](./connectors) provided by Camunda. 
+This repository manages the [out-of-the-box Connectors](./connectors) provided by Camunda.
 For more information on those Connectors, refer to the
 [Camunda 8 documentation](https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/available-connectors-overview/).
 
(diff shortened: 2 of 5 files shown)


berndruecker:
As part of adding the micrometer prometheus dependency, I wondered if adding the Spring Boot POM (https://github.com/camunda/connectors-bundle/pull/98/commits/d36d4be3b14a566d55657efa714395b3d0e8c25a) is a big risk - or actually something we should even push to the bundle root - so that a lot of common versions are harmonized across all connectors kind of automatically. Any thoughts @tmetzke @chillleader ?

bundle/mvn/default-bundle/src/test/resources/application.properties
```diff
@@ -2,3 +2,7 @@
 server.port=9898
 zeebe.broker.gateway-address=localhost:26500
 zeebe.client.security.plaintext=true
+
+management.server.port=9080
+management.context-path=/actuator
+management.endpoints.web.exposure.include=metrics,health,prometheus
```
chillleader, line 8: Should we also apply these changes to `saas-bundle/src/main/resources/applicaiton.properties`? Otherwise it will only work for the local runtime, I think.

tmetzke:
I don't see too much risk with this. In the end, the BOM is only managing versions we want to be compatible to the Runtime anyways. Could be helpful actually and provides a set of versions tested by Spring. And you can still override versions if you need something else and are sure it works. 👍

tmetzke (approved):
Looks good to me 👍 🙂 
We'll also see adjusted versions for all Connectors when the third party notices job ran.

bundle/mvn/default-bundle/src/test/resources/application.properties
tmetzke: Awesome catch ⭐

chillleader: It's actually almost identical already, just missing the `prometheus` endpoint. Not sure if it's important.

tmetzke: It is, yes, otherwise that endpoint wouldn't be available

berndruecker: It is - I add it there - thx!

chillleader (approved):

berndruecker:
I moved pushing Spring Boot dependency management to root level to an own PR - I agree that this should make sense and might avoid some headache: https://github.com/camunda/connectors-bundle/pull/100

PR #114, opened 2022-12-16
Revert "fix(deps): update dependency com.google.cloud:libraries-bom t…

chillleader:
## Description

Revert broken dependency bumps (we can investigate after the patch release)

## Related issues

related to camunda/team-connectors#272

### DIFF
--- a/bundle/mvn/camunda-saas-bundle/pom.xml
+++ b/bundle/mvn/camunda-saas-bundle/pom.xml
@@ -11,7 +11,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <version.spring-cloud-gcp-starter-logging>3.4.1</version.spring-cloud-gcp-starter-logging>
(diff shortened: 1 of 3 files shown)


tmetzke:
Are those really broken? The job that fails is rather flaky as it cannot assign the renovate bot as assignee to the PR which makes sense. We can simply fix the workflow.

chillleader:
Oh ok, I just had an issue with the local build and I thought that it's the same in GitHub. Must be my local issue then, sry.

tmetzke:
If you need help figuring out what's wrong, let me know 👍

tmetzke:
@chillleader, can you create a follow-up to prevent renovate from auto-merging updates? For serious projects, I think that's not the best idea. We should decide ourselves when to update dependencies.

chillleader:
Created https://github.com/camunda/team-connectors/issues/278 in team-connectors, since we probably want to check other repositories as well

PR #125, opened 2022-12-19
feat(github): Added Github webhook element-template

berndruecker:
Moved content from https://github.com/camunda/connector-github to fit in MonoRepo structure

### DIFF
--- a/connectors/github/README.md
+++ b/connectors/github/README.md
@@ -1,0 +1,4 @@
+# GitHub Connector
+
+Connector to start process instances on GitHub events. This connector is based on 
+inbound webhooks.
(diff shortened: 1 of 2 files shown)


tmetzke:
I am wondering if this needs a tighter integration into the Maven module structure. 🤔 
We'd probably need to think about releases here: How do we release template-only Connectors in general?

berndruecker:
I agree - but this is on the radar around the whole discussion how to distribute / install element-templates anyway (I hope)

tmetzke:
Not sure it has been already, it should be now, though 😁 
@chillleader and @markfarkas-camunda, please either create an issue on "How to install/release template-only Connectors" to tackle in the (near) future. Goes for Outbound and Inbound alike 🙂

PR #165, opened 2022-12-28
chore(build): add maven shade plugin where missing

chillleader:
## Description

Maven Shade plugin config is missing in most connectors, which leads to failures during classpath bundling (without Maven)

## Related issues

related to https://github.com/camunda/team-connectors/issues/258

### DIFF
--- a/connectors/aws-lambda/pom.xml
+++ b/connectors/aws-lambda/pom.xml
@@ -40,21 +40,4 @@
     </dependency>
   </dependencies>
 
-  <build>
-    <plugins>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-shade-plugin</artifactId>
-        <configuration>
-          <relocations>
-            <relocation>
-              <pattern>com.fasterxml.jackson</pattern>
-              <shadedPattern>connectorawslambda.com.fasterxml.jackson</shadedPattern>
-            </relocation>
-          </relocations>
-        </configuration>
-      </plugin>
-    </plugins>
-  </build>
-
 </project>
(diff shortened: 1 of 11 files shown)


tmetzke:
Why exactly is the classpath bundle failing? If all connectors and the runtime use the same Jackson version, it should just work fine, shouldn't it? Might rather point out a version mismatch issue we have somewhere else. 🤔

igpetrov (approved):

chillleader:
> Why exactly is the classpath bundle failing? If all connectors and the runtime use the same Jackson version, it should just work fine, shouldn't it?

There's still a mismatch between the versions used in `connector-runtime` (and transitively `connector-sdk`) and the connectors in this repository. Specifically, runtime `0.4.x` brings `jackson 2.13.x` which overrides what we have in this repo. Connector jars with dependencies, however, already contain `jackson 2.14.x`.

We may also fix this by releasing the new version of the SDK, but I think it will also be nice to keep jackson shaded, otherwise we'll still face this issue in the future. Besides, since SDK, runtime and bundle all have different version semantics, someone will inevitably try to use the mismatching versions in a Docker bundle and catch this error.

chillleader:
On the other hand, we can create one shaded dependency for the whole bundle, e.g. `connectors.com.fasterxml.jackson` instead of `connectoraws.com.fasterxml.jackson`, `connectorsns.com.fasterxml.jackson` - because all connectors of the same version are now guaranteed to have the same version of jackson.

tmetzke:
Ok I see, makes sense. Also provides a good practice template for others maybe. I was wondering if we should exclude this shading when we build the default Maven Bundle. It shouldn't be necessary there and might actually create some overhead we don't need. But just a thought, might also be wrong.

chillleader:
Good point. As far as I can see by inspecting the resulting JARs, this shade is not applied when building the default bundle  - it's only active for individual `connector-with-dependencies` artifacts.

tmetzke:
Yeah, makes sense actually, now that I think about it. The bundle uses the "plain" artifacts of the Connectors as dependencies. Thanks for looking into it.

I think the idea with a common relocation for all Connectors is really nice. 👍

PR #166, opened 2022-12-28
feat(monorepo): add archetype for quick creation of new connectors

chillleader:
## Description

Adds archetype for quick creation of new connectors.

The idea is to have a shell script that does all the routine work. I implemented it by creating a Maven archetype for new connector, which is later used in the actual script.

Progress:
- [x] Add template/archetype
- [x] Automate connector creation
- [x] Automate changes in parent pom
- [ ] ~Automate changes in bundle ~ (Decided to skip, feature is not worth the effort & solution is not portable)
- [x] Document the changes

## Related issues

related to https://github.com/camunda/connectors-bundle/issues/129


connectors/README.md
```diff
@@ -38,3 +38,27 @@ is run as a Google Cloud Function.
...
+## Development guidelines
+
+### Create a new Connector
+
+To create a new Connector, simply use the [script](../add_new_connector.sh) or generate a new
+project from the [Maven archetype](../connector-archetype-internal).
+
+Execute from the repository root directory:
+```shell
+chmod +x add_new_connector.sh
```
igpetrov, line 51: I think, this step could be done in git. [Ref](https://git-scm.com/docs/git-update-index#Documentation/git-update-index.txt---chmod-x).

PR #177, opened 2023-01-05
fix: encrypted emoji in chat message and channel message

### LINKED ISSUE #170: Sending an emoji in channel message from MS teams connector is displayed encrypted
__Describe the Bug__

When we try to create a channel message with emoji in MS teams connector we could see that the message is displayed encrypted in the teams but we could still add an emoji in the conversation that we directly create from MS teams app.

__Steps to Reproduce__

1. Create a MS teams connector
2. Select Conversation Type->Channel
3. Select Method->Send message to channel
4. Add the Channel ID and Group ID
5. Add a message with emoji to send to the channel
6. We could see the emoji as an encrypted message in the Teams App

4c240cbaa9d235ffbadf97ce4b1a4950

We could also see another channel message with emoji that has been created from the MS teams app itself

__Expected Behavior__

Add support to add emojis in the message

__Environment__

- OS: [ Windows 10]

mbiciin:
## Description

Fix for encrypted emoji in chat message and channel message

closes #170

### DIFF
--- a/connectors/microsoft-teams/pom.xml
+++ b/connectors/microsoft-teams/pom.xml
@@ -43,6 +43,10 @@
       <groupId>org.danilopianini</groupId>
       <artifactId>gson-extras</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-text</artifactId>
+    </dependency>
   </dependencies>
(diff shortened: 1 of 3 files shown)


markfarkas-camunda (approved):
Looks good to me

PR #185, opened 2023-01-09
fix(ms-teams): add chat type value and link to members format docs

Oleksiivanov:
## Description

Add chat type value and link to `members` format docs

## Related issues
https://github.com/camunda/camunda-platform-docs/issues/1596

closes #

https://github.com/camunda/connectors-bundle/issues/181


markfarkas-camunda (approved):
Looks good to me!

PR #191, opened 2023-01-10
added exemplary RabbitMQ inbound connector (subscription)

berndruecker:
Quick first stab at a RabbitMQ inbound connector (subscription) for discussion.

Requires https://github.com/camunda-community-hub/spring-zeebe/pull/311 to be built successfully

### DIFF
--- a/connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/RabbitMqInboundConnectorSubscription.java
+++ b/connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/RabbitMqInboundConnectorSubscription.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.rabbitmq.inbound;
+
+import com.rabbitmq.client.Channel;
+import com.rabbitmq.client.Connection;
+import com.rabbitmq.client.ConnectionFactory;
+import com.rabbitmq.client.DeliverCallback;
+import io.camunda.connector.api.annotation.InboundConnector;
+import io.camunda.connector.api.inbound.InboundConnectorContext;
+import io.camunda.connector.api.inbound.InboundConnectorProperties;
+import io.camunda.connector.api.inbound.SubscriptionInboundConnector;
+import io.camunda.zeebe.client.api.response.ProcessInstanceEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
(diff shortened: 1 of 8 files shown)


connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/RabbitMqInboundConnectorSubscription.java
```diff
@@ -0,0 +1,84 @@
...
+
+    return context.zeebeClient()
+            .newCreateInstanceCommand()
+            .bpmnProcessId(connectorProperties.getBpmnProcessId())
+            .version(connectorProperties.getVersion())
+            .variables(variables)
+            .send()
+            .join();
+  }
+
```
chillleader, line 71: Hey @berndruecker I like the general approach!

My 2 cents:
1. I would prefer the concrete connector implementations to be abstracted away from communication with Zeebe and process-specific details like `processId`, `version` and other potential target process identifiers.
So if we follow this idea, the connectors would ideally just call a pre-defined method when the inbound event occurs, and the SDK/runtime code will do the rest of the job correlating the message.

2. Why don't we use process messages for all inbound correlations? We can start a process with a message, we can correlate it to an already running process, so this would be uniform for all inbound targets.

berndruecker: I wonder if we already feel confident enough that an abstraction does not hinder us from doing what we need to? 
At the same time, it might be valuable that all inbound connectors follow the same schema when it comes to selecting the right process/version and/or message (like for example there was some logic in the webhook to clarify what happens if a second process version with the same/other context was deployed: https://github.com/camunda-community-hub/spring-zeebe/blob/master/connector-runtime/src/main/java/io/camunda/connector/runtime/inbound/registry/InboundConnectorRegistry.java#L92). 

For 2: As written in https://github.com/camunda/team-connectors/issues/305#issuecomment-1378750080 I think supporting messages only is a bit more challenging, especially as no exception is provided if a message cannot be correlated. At the same time we cannot influence the version of a process instance started - which might be a good or a bad thing - depending on the viewpoint :-)

chillleader: I agree, we shouldn't really overengineer the abstractions so early on 😄 
But one way or another, I'm trying to imagine how it fits together with using inbound connectors as intermediate events. Some sort of abstraction will be necessary, so that the connector code can work against both types of events.

Regarding 2, we'll still have to face these challenges to some extent while implementing the intermediate events, right? But let's not rush with this idea, I'm sure we can generalize the approach later on if it starts to make sense.

sbuettner:
@berndruecker Can we close this one as @chillleader is working on https://github.com/camunda/connectors-bundle/pull/337

PR #199, opened 2023-01-11
fix(archetype): remove xml descriptor

igpetrov:
fix(archetype): remove xml descriptor


chillleader (approved):
Not sure if it changes something or not. But still nice to have these uniform 👍

PR #208, opened 2023-01-12
fix(ms-teams): add refresh token authentication type

Oleksiivanov:
## Description

Added refresh token authentication type

## Related issues

closes #
https://github.com/camunda/connectors-bundle/issues/150


connectors/microsoft-teams/src/main/java/io/camunda/connector/model/authentication/RefreshTokenAuthentication.java
```diff
@@ -0,0 +1,98 @@
...
+        + clientId
+        + "'"
+        + ", tenantId='"
+        + tenantId
+        + "'"
+        + ", clientSecret='"
+        + clientSecret
+        + "'"
+        + ", token='"
+        + token
```
igpetrov, line 94: Let's hide `clientSecret` and `token` as `"[REDACTED]"` or equivalent

igpetrov (approved):

Issue #210, opened 2023-01-13
Create new version of custom template for uipath


mbiciin:
__What should we do?__

We need to support two type of request in the custom template "Add queue item" and "Get item status". We will ad a dropdown list with thse 2 values. The input values from the custom template will have a display condition depending on the type of request selected from the dropdown list. Prefilled values will be used just for hidden fields. We will add notes/docs links as guideline for the user. 

__Why should we do it?__

https://github.com/camunda/product-hub/issues/190

mbiciin:
@bastiankoerber, we forgot, for the request "Get item status", we need an item id for filtering the results. We should ask the user to add it manually? In the prototype diagram, we extracted the item id from the response of request "add queue item". We added some expressions there: https://modeler.cloud.dev.ultrawombat.com/diagrams/1798be99-cf98-4b1e-88b3-cb5cde3b6a5d--prototype-bpmn-uipath-bpmn?v=660,377,1 


![Image](https://user-images.githubusercontent.com/114916619/212298831-9724e9c6-dec8-40df-ae54-90b5ff13356e.png)

![Image](https://user-images.githubusercontent.com/114916619/212298903-c6c9836b-9aac-4faf-85a2-3129d390493a.png)

mbiciin:
The user can build a similar diagram, but using the same template. Is this what we need?

bastiankoerber:
btw: is it possible to have the itemid as an outputfield. Something like: ItemID and i enter the variablename ?

If this is not possible, i would prefill the result expression (like you did).

In the "get item status" i would add two input fields: Status (as dropdown) and item id (as input). The output can be prefilled if there is no possibility to have an "output field"

mbiciin:
I think you can see this diagram from int: https://modeler.cloud.ultrawombat.com/diagrams/d71bcde7-2c2b-4a37-8b3d-eebdb690bf3e--prototype-uipath-diagram?v=660,377,1

bastiankoerber:
> I think you can see this diagram from int: https://modeler.cloud.ultrawombat.com/diagrams/d71bcde7-2c2b-4a37-8b3d-eebdb690bf3e--prototype-uipath-diagram?v=660,377,1

just to make sure, that we are on the same page. The link shows our "old" templates right and not the production ready templates ([Release UI Path Connector#296](https://github.com/camunda/team-connectors/issues/296))

mbiciin:
[...]

PR #212, opened 2023-01-16
uipath: add new version for custom template

mbiciin:
## Related issues

#210

### DIFF
--- a/connectors/uipath/element-templates/uipath-custom-template.json
+++ b/connectors/uipath/element-templates/uipath-custom-template.json
@@ -1,0 +1,592 @@
+{
+  "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
+  "name": "UIPath Connector",
+  "id": "585cc4f7-90e1-4975-aea9-111990be952c",
+  "description": "Orchestrate your UIPath Bots with Camunda. You can create new queue items and get the result from it",
+  "icon": {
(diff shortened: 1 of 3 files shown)


bastiankoerber (changes requested):
i changed a lot.

mbiciin:
The template is not finished. The last part concerning the input fields "Item id" and "Item status" is work in progress. As disscussed, we should let the user type the name of the variable in the input field "Item Id" and bind it with the value return by the resultExpression {itemId: response.body.id}. We don't know if it's possible yet. I'll have a meeting with a collegue from web-modeler team today.

bastiankoerber:
@mbiciin is this the final version?

connectors/uipath/element-templates/uipath-custom-template.json
```diff
@@ -0,0 +1,575 @@
...
+  },
+  "documentationRef": "https://docs.camunda.io/docs/components/modeler/web-modeler/connectors/available-connectors/rest/",
```
bastiankoerber, line 9: link to the new documentation

connectors/uipath/element-templates/uipath-custom-template.json
```diff
@@ -0,0 +1,575 @@
...
...6.1,74.1c-44.1,0-76.1-30-76.1-74.1 c0-45.1,32.1-76.1,76.1-76.1C932.5,233.4,964.6,264.4,964.6,309.5z M826.4,442.7h124.2v487.8H826.4V442.7z M1784,517.9 c0,133.2-88.1,212.4-223.4,212.4h-112.2v200.3h-124.2v-621h236.4C1697.8,309.5,1784,389.6,1784,517.9z M1657.8,517.9 c0-68.1-39.1-108.2-110.2-108.2h-99.2v221.4h99.2C1618.7,631.1,1657.8,591,1657.8,517.9z M2201.7,442.7h124.2v487.8h-124.2v-55.1 c-33.1,39.1-85.1,61.1-158.3,61.1c-132.2,0-225.4-104.2-225.4-249.4c0-144.2,91.2-250.4,225.4-250.4c71.1,0,125.2,25,158.3,67.1 V442.7z M2201.7,687.1c0-84.1-50.1-142.2-129.2-142.2c-79.1,0-128.2,55.1-128.2,142.2c0,81.1,44.1,140.2,128.2,140.2 C2148.6,827.4,2201.7,772.3,2201.7,687.1z M2629.4,824.4h49.1v106.2h-65.1c-118.2,0-169.3-56.1-169.3-171.3V547.9H2386V442.7h58.1 V309.5h124.2v133.2h109.2v105.2h-109.2v212.3C2568.3,803.3,2581.3,824.4,2629.4,824.4z M3214.3,647.1v283.5h-124.2V661.1 c0-74.1-38.1-122.2-111.2-122.2s-117.2,50.1-117.2,133.2v258.4h-124.2V279.5h124.2v213.4c31.1-37.1,80.1-57.1,147.2-57.1 C3124.2,435.7,3214.3,518.9,3214.3,647.1z M3200,279.5h21.7v4.2h-8.5v25.8h-4.8v-25.8h-8.5V279.5z M3261,279.5v30h-4.8v-22.8 l-9.7,22.8h-3.6l-9.7-22.8v22.8h-4.8v-30h6l10.3,23.4l10.3-23.4H3261z'/%3E%3C/g%3E%3C/svg%3E%0A"
```
bastiankoerber, line 7: add UIPath logo

connectors/uipath/element-templates/uipath-custom-template.json
```diff
@@ -0,0 +1,575 @@
...
+      },
+      "condition": {
+        "property": "operationType",
+        "oneOf": [
+          "addQueueItem"
+        ]
+      }
+    },
+    {
+      "label": "Specific Content for UIPath Job",
```
bastiankoerber, line 428: for me it was just working with an feel expression. if it is like this, then set feel as required

bastiankoerber (changes requested):
please change url to documentation, add expected dateformat and check specific content with feel expression

bastiankoerber (approved):
awesome, looks good to me :-)

nikku:
@bastiankoerber I investigated together with @mbiciin one requirement that you had:

> As a user I want to map a specific part of the execution result back to the process (for tracking, further processing)

This is absolutely possible using on board facilities:

* Define an input variable mapping for `resultVariableName`

* Use that variable to dynamically construct a response mapping context
  
  ```
  if 
    resultVariableName = null
  then 
    {}
  else
    put({ }, resultVariableName, response.body.Id)
  ```

Two general comments of what @mbiciin mentioned: 

* We may defined default output mappings, but we should never do that magically (i.e. if `resultVariableName` is not provided, _always_ write a `itemID` variable). This is dangerous and violates the no surprises property: https://github.com/bpmn-io/design-principles#no-surprises

At this stage you cannot map dynamically constructed outputs _and_ offer the default mapping:

image

My suggestion is to keep the default output mapping facility in place as it allows users to not only extract default data, but also access additional diagnostics data that we don't want to offer via explicit template propertiesin the UI (`Item ID NAME`). 

In the future we'll offer dedicated editor tooling for dynamic mapping making that even simpler: 
image

mbiciin:
@nikku, I forgot to tell you in our meeting that we need to use this itemId in "Filters", which is a parameter for query 'getItemStatus'. The expression still evaluetes to {"$Filter":null} which is not correct. For example, I should have {"$Filter":"Id eq 436141352"}. 

Screen Shot 2023-01-20 at 3 22 35 PM

Here is the filter:
```
{
      "label": "Filters",
      "description": "Map of query parameters to add to the request URL",
      "group": "input",
      "type": "Hidden",
      "value": "={\"$Filter\":\"Id eq \"+string(itemId)}",
      "binding": {
        "type": "zeebe:input",
        "name": "queryParameters"
      },
      "condition": {
        "property": "operationType",
        "oneOf": [
          "getItemStatus"
        ]
      }
    }
```

More info about it here: https://github.com/camunda/connectors-bundle/issues/210#issuecomment-1387421948

PR #214, opened 2023-01-17
chore(ms-teams): remove GsonSupplier.java

Oleksiivanov:
## Description

removed GsonSupplier.java . In this connector, we use ObjectMapper for parsing JSON because Object mapper better supports OffsetDataTime (this field exists on the response from Graph API)


igpetrov (approved):
Please, do a regression test once this is merged.
Otherwise, LGTM.

PR #253, opened 2023-01-27
fix(archetype): fix resources dir of the archetype

### LINKED ISSUE #228: Fix automatic creation of new connector
__What should we do?__

Fix the `connector-archetype-internal` project inside the bundle. When we run the add_new_connector.sh it fails to copy `io.camunda.connector.api.outbound.OutboundConnectorFunction` file under `resources/META-INF/services` folder.

The file is present in the `connector-archetype-internal` project, so hypotethically it should work.

__Why should we do it?__

When we run the add_new_connector.sh it will create the new connector but without the resources/META-INF/services folder, which is an issue, because without this folder's content the connector is not usable.

chillleader:
## Description

Fixes the issue with the resource directory in the archetype.

NB: To test locally or use this change before it's merged, don't forget to run `mvn clean install` on the archetype project. This will make the fixed archetype visible to Maven.

## Related issues

closes #228


markfarkas-camunda (approved):
Works like a charm! 🥇

PR #257, opened 2023-01-29
feat(http-json): add CustomAuthentication

Oleksiivanov:
## Description

Add Custom Authentication to REST connector; refactoring

## Related issues
https://github.com/camunda/team-connectors/issues/241

closes #
https://github.com/camunda/connectors-bundle/issues/244

short demo how it works: 

- we have an API where we need to put an authentication header for the successful response.
- in our template (here in the rest template, but in another version it will be another connector that extends REST connector, and not necessary fields we can hint) we put auth URL, METHOD, auth body, and `outputHeader` : 
   - in outpuHeaders we put` {"X-Authorization":"asObject:token.asString"}`,  `asObject:token.asString` it is parsing pattern, by this pattern connector extract token from auth response and put it to next request as header value for `X-Authorization ` header key 
- next, we put all required data for request, without auth header.

https://user-images.githubusercontent.com/108869886/215579292-69997dad-135a-4b11-9f49-32e8e212cb16.mov


connectors/http-json/element-templates/http-json-connector.json
```diff
@@ -74,6 +74,10 @@
         {
           "name": "OAuth 2.0",
           "value": "oauth-client-credentials-flow"
+        },
+        {
+          "name": "Custom Authentication",
```
igpetrov, line 79: `Custom Authentication` is too generic name - it's applicable for every non-standard types of authentication.
---
Consider something like `Credentials in body`.
---
Also, we don't append word `Authentication` in type because it's already mentioned in the field.

What do you think @Oleksiivanov @markfarkas-camunda? Any better suggestions for naming?

connectors/http-json/element-templates/http-json-connector.json
```diff
@@ -74,6 +74,10 @@
...
+          "value": "customAuth"
```
igpetrov, line 80: Once naming changed, we'll also need to rename these references.

connectors/http-json/src/main/java/io/camunda/connector/http/HttpJsonFunction.java
```diff
@@ -91,20 +53,13 @@ public HttpJsonFunction(String proxyFunctionUrl) {
...
-        GsonComponentSupplier.gsonFactoryInstance(),
         proxyFunctionUrl);
   }
 
   public HttpJsonFunction(
-      final Gson gson,
-      final HttpRequestFactory requestFactory,
-      final GsonFactory gsonFactory,
-      String proxyFunctionUrl) {
+      final Gson gson, final HttpRequestFactory requestFactory, String proxyFunctionUrl) {
```
igpetrov, line 60:
```suggestion
      final Gson gson, final HttpRequestFactory requestFactory, final String proxyFunctionUrl) {
```

connectors/http-json/element-templates/http-json-connector.json
igpetrov: I am not in favor of any. `Custom`, `specific`, `unregular`, etc do not explain what and how exactly it is implemented and sending.

When we say `Basic` (authentication) we know that we refer to specific RFC of basic authentication which is always `base64(login+:+password)` function. When we say `OAuth 2.0` we always know the mechanism as well. But when we say `Custom` I have no idea what is it supposed to mean.

If in the next iteration we find another mechanism of authentication, say, for BluePrism how are we going to call it? That's why we should explicitly name the authenctication type to let user understand what's going on without looking into code.

My suggestion is `Credentials in body` with description that `Login and password will be sent in body as JSON object prior to calling resource endpoint blah-blah`. But since I am a bit out of context, maybe there's better name.

connectors/http-json/src/main/java/io/camunda/connector/http/auth/CustomAuthentication.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+import java.util.Objects;
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+public class CustomAuthentication extends Authentication {
+
+  @NotNull @Valid private HttpJsonRequest request;
+
+  @Secret private Map<String, String> outputBodyMap;
```
markfarkas-camunda, line 31: I think this should be `outputBody` instead of `outputBodyMap`. In the element template I was able to find outputBody but not outputBodyMap binding, and when I tested this was always null.

connectors/http-json/element-templates/http-json-connector.json
```diff
@@ -284,6 +372,38 @@
...
+      "binding": {
+        "type": "zeebe:input",
+        "name": "authentication.outputBody"
+      },
+      "condition": {
+        "property": "authenticationType",
+        "equals": "credentialsInBody"
+      },
+      "optional": true
+    },
```
markfarkas-camunda, line 406: Just a note: if I add some value here for `Authentication Output Body`, and then select `GET` as method in the _HTTP Endpoint_ section, it will try to execute a GET method with a body which will result in a `java.lang.IllegalStateException: Apache HTTP client does not support GET requests with content.` exception. As far as I know we cannot add constraints to dropdown values, but if there is a way to do it, maybe it would be something worth considering. We could create a ticket about this, as a suggestion for future development for the template engine. Do you also think it is something notable? Do you know any way to add a constraint to a specific dropdown value?

markfarkas-camunda: If we do not have a solution for this right now, all changes are good from my side.

PR #258, opened 2023-01-30
fix(gdrive): forced parent folder ID as optional

igpetrov:
## Description

Forced parent folder ID as optional.

## Related

- https://forum.camunda.io/t/google-drive-connector/42163


markfarkas-camunda (approved):
This solves the issue, thank you 👍

PR #321, opened 2023-02-22
fix(slack): fix invite user to private channel

markfarkas-camunda:
## Description

Inviting user to private channel was not working, because by default Slack API only lists the public channels. The fix was just specifying the types of conversations to include private channel type as well.

## Related issues
https://github.com/camunda/connectors-bundle/issues/320

closes #
https://github.com/camunda/connectors-bundle/issues/320


igpetrov (approved):

connectors/slack/src/main/java/io/camunda/connector/slack/utils/DataLookupService.java
```diff
@@ -15,6 +15,7 @@
 import com.slack.api.methods.response.users.UsersListResponse;
 import com.slack.api.methods.response.users.UsersLookupByEmailResponse;
 import com.slack.api.model.Conversation;
+import com.slack.api.model.ConversationType;
 import com.slack.api.model.User;
 import java.io.IOException;
 import java.util.*;
```
igpetrov, line 21: Can we please expand wildcard import?

chillleader (approved):

PR #340, opened 2023-03-02
chore(webhook): rename webhook template type

chillleader:
## Description

Webhook type was renamed from `webhook` to `io.camunda:webhook:1` to match the pattern we have for outbound.


igpetrov (approved):

igpetrov (approved):
Also, please don't forget to update the same at the `web-modeler` repository

PR #344, opened 2023-03-06
chore(blue prism): simplify BP element template

### LINKED ISSUE #343: Blue Prism: element template enhancements
__What should we do?__

Simplify data field to send a single element instead of an array.

igpetrov:
## Description

Simplify BP element template

closes #343


igpetrov:
Also https://github.com/camunda/web-modeler/pull/4029

PR #363, opened 2023-03-13
feat(msteams): add the ability to send HTML message 

Oleksiivanov:
## Description

add the ability to send HTML message 

closes #
https://github.com/camunda/connectors-bundle/issues/233

### DIFF
--- a/connectors/microsoft-teams/element-templates/microsoft-teams-connector.json
+++ b/connectors/microsoft-teams/element-templates/microsoft-teams-connector.json
@@ -693,6 +693,34 @@
       }
     },
     {
+      "label": "Content type",
+      "description": "The type of the content. Possible values are text and html.",
+      "group": "data",
+      "type": "Dropdown",
+      "value": "TEXT",
+      "choices": [
+        {
+          "name": "Text",
+          "value": "TEXT"
+        },
+        {
+          "name": "HTML",
+          "value": "HTML"
+        }
+      ],
+      "binding": {
+        "type": "zeebe:input",
+        "name": "data.bodyType"
+      },
+      "condition": {
+        "property": "channelMethod",
+        "oneOf": [
+          "sendMessageToChat",
+          "sendMessageToChannel"
+        ]
+      }
+    },
+    {
       "label": "Message ID",
       "description": "The message ID",
       "group": "data",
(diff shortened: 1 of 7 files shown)


connectors/microsoft-teams/element-templates/microsoft-teams-connector.json
```diff
@@ -692,6 +692,34 @@
...
+      "type": "Dropdown",
+      "value": "TEXT",
+      "choices": [
+        {
+          "name": "TEXT",
+          "value": "TEXT"
+        },
+        {
+          "name": "HTML",
+          "value": "HTML"
```
igpetrov, line 708: My suggestion would be here and everywhere else where users will read labels, to avoid capitalized labels or texts.

To me it looks strange. @bastiankoerber WDYT?

```suggestion
      "value": "TEXT",
      "choices": [
        {
          "name": "Text",
          "value": "TEXT"
        },
        {
          "name": "HTML",
          "value": "HTML"
```

connectors/microsoft-teams/src/main/java/io/camunda/connector/model/request/chat/SendMessageInChat.java
```diff
@@ -55,16 +70,28 @@ public boolean equals(final Object o) {
...
 
   @Override
   public String toString() {
-    return "SendMessageInChat{" + "chatId='" + chatId + "'" + ", content='" + content + "'" + "}";
+    return "SendMessageInChat{"
+        + "chatId='"
+        + chatId
+        + "'"
+        + ", content='"
+        + content
```
igpetrov, line 90: I think, we should not display `content` and redact messages instead. Otherwise, we might face with GDPR regulations.

Applicable here and other similar places.

connectors/microsoft-teams/src/test/resources/requests/success-chat-request-test-cases.json
```diff
@@ -199,7 +199,8 @@
     "data": {
```
igpetrov, line 199: Can you also add test with HTML content?

igpetrov (changes requested):
LGTM overall, requesting some changes.

connectors/microsoft-teams/element-templates/microsoft-teams-connector.json
bastiankoerber: i agree here

igpetrov (approved):

PR #364, opened 2023-03-13
chore(rest): refactoring, add test for ResponseParser

Oleksiivanov:
## Description

At this moment code has already been refactored, and needed classes were moved to the common module. So now I add only a few tests.

## Related issues

https://github.com/camunda/connectors-bundle/issues/265

closes #
https://github.com/camunda/connectors-bundle/issues/265


connectors/connectors-common-library/src/test/java/io/camunda/connector/common/utils/ResponseParserTest.java
```diff
@@ -0,0 +1,102 @@
...
+  private static final String INNER_OBJECT_VALUE_KEY = "inner_object_value_key";
+
+  private static final String ARRAY_VALUE = "value in auth array response";
+  private static final String OBJECT_VALUE = "object value";
+  private static final String INNER_OBJECT_VALUE = "inner object value";
+
+  private static final String KEY_NAME = "keyName";
+  private static final int NUMBER_OF_ELEMENT_IN_ARRAY = 0;
+
+  private final Gson gson = new Gson();
```
igpetrov, line 43: If we have this in `GsonSupplier` or similar - let's pull it from there to maintain consistency.

igpetrov:
Let's wait for @markfarkas-camunda review as well.

markfarkas-camunda (approved):
Overall it looks good to me. I still think [my proposal](https://github.com/camunda/connectors-bundle/pull/257#discussion_r1091771273) could work, but if we do not want to get the value in any other format then it is not necessary and would just complicate things. We can always say we only support string values, and it is always a next step to convert it to other data format if necessary. To conclude: I would leave it as it is now. Works well 👍

PR #383, opened 2023-03-21
OpenAI / ChatGPT Connector

chillleader:
## Description

ChatGPT & Moderation API Connector.

Implemented with https://github.com/MaxTru/camunda-connector-openai-moderation during [Connectors Team hackathon](https://github.com/camunda/team-connectors/issues/363).

Element template (ChatGPT operation selected)
Screenshot 2023-03-21 at 17 46 41

Element template (Moderation selected)
Screenshot 2023-03-21 at 17 47 34


connectors/openai/element-templates/openai-connector.json
```diff
@@ -0,0 +1,330 @@
...
+      "id": "authenticationType",
+      "value": "bearer",
+      "type": "Hidden",
+      "binding": {
+        "type": "zeebe:input",
+        "name": "authentication.type"
+      }
+    },
+    {
+      "label": "OpenAI API Key",
```
igpetrov, line 60: For how long does it last? If it's a typical bearer token that lasts only 3600 seconds, maybe it's worth to take a look whether client credentials flow is supported by OpenAI

igpetrov (approved):

connectors/openai/element-templates/openai-connector.json
chillleader: It's a [static API key](https://platform.openai.com/docs/api-reference/authentication) - but for some reason they call it a bearer token.

BTW I just noticed that it makes sense to add [organization header](https://platform.openai.com/docs/api-reference/requesting-organization) support.

igpetrov: Thanks for confirming. That's weird, but I see it as well now, so no issues here.

PR #394, opened 2023-03-23
Camunda Operate Connector

chillleader:
## Description

Camunda Operate Connector

Screenshot 2023-03-23 at 18 11 26
Screenshot 2023-03-23 at 18 11 51

### DIFF
--- a/connectors/operate/element-templates/operate-connector.json
+++ b/connectors/operate/element-templates/operate-connector.json
@@ -462,6 +462,7 @@
       "label": "Result Expression",
       "description": "Expression to map the response into process variables. Details in the <a href=\"https://docs.camunda.io/docs/components/connectors/use-connectors/#result-expression\" target=\"_blank\">documentation</a>",
       "group": "output",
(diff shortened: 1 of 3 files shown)


connectors/operate/element-templates/operate-connector.json
```diff
@@ -0,0 +1,506 @@
...
+      "description": "Name of variable to store the response in. Details in the <a href=\"https://docs.camunda.io/docs/components/connectors/use-connectors/#result-variable\" target=\"_blank\">documentation</a>",
+      "group": "output",
+      "type": "String",
+      "binding": {
+        "type": "zeebe:taskHeader",
+        "key": "resultVariable"
+      }
+    },
+    {
+      "label": "Result Expression",
```
markfarkas-camunda, line 462: Just a thought: Does it make sense to add predefined values in the Result Expression for each operations as a suggestion or for convenience?

markfarkas-camunda (approved):
Awesome! I added only one comment about a minor thing, but this is perfect! 🥇

PR #406, opened 2023-03-28
chore(msTeams): show only available methods depend on auth type

Oleksiivanov:
## Description

Show only available methods depending on auth type. For client credentials auth type is not available some methods : 
- send message to channel
- Send message in chat
- List chats        

In demo you can see that when we set different auth types we have different available methods :

https://user-images.githubusercontent.com/108869886/228317636-92330100-3b0e-4d10-9908-50a76659690b.mov

## Related issues

https://github.com/camunda/connectors-bundle/issues/168
closes #

https://github.com/camunda/connectors-bundle/issues/168

### DIFF
--- a/connectors/microsoft-teams/element-templates/microsoft-teams-connector.json
+++ b/connectors/microsoft-teams/element-templates/microsoft-teams-connector.json
@@ -237,8 +237,68 @@
         "notEmpty": true
       },
       "condition": {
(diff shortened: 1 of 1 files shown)


igpetrov (approved):
LGTM. Don't forget to update web-modeler afterwards.

PR #473, opened 2023-04-17
feat(rest): Add the ability to work with different content types

Oleksiivanov:
## Description

Add the ability to work with different content types for HTTP-JSON connector, like 'text/plain', 'application/x-www-form-urlencoded' etc.
Add the ability to return string format in the response body.

If the content type is not set, set 'application/json' content type as the default content type

## Related issues

https://github.com/camunda/connectors-bundle/issues/402

closes #

https://github.com/camunda/connectors-bundle/issues/472


connectors/connectors-common-library/src/main/java/io/camunda/connector/common/services/HTTPService.java
```diff
@@ -104,15 +112,39 @@ public <T extends CommonResult> T toHttpJsonResponse(
...
+      if (bodyString != null) {
+        if (isJSONValid(bodyString)) {
+          Object body = gson.fromJson(bodyString, Object.class);
+          connectorResult.setBody(body);
+        } else {
+          try {
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            DocumentBuilder builder = factory.newDocumentBuilder();
+            Document doc = builder.parse(new InputSource(new StringReader(bodyString)));
+            connectorResult.setBody(doc);
```
igpetrov, line 130: What does this block do? Parsing XML?
If so, please it's perfect place to leave an explanation comment in code. Also, rename the method, since it's no longer does `toHttpJsonResponse`

igpetrov (changes requested):
Looks good, some more changes requested + more unit tests with content type please.

PR #478, opened 2023-04-18
chore(docker os): added focal image to release train

igpetrov:
chore(docker os): added focal image to release train

Since, customers have confirmed that `focal` image works for them, adding a stable implementation of the release procedure.

⚠️ I will perform a release `0.18.2` to make sure everything is still working as expected.

Ref: https://github.com/camunda/connectors-bundle/issues/442


chillleader (approved):
LGTM
My general feeling is that the release pipeline is getting too big - maybe it makes sense to try splitting it into a few independent steps, so that failures in one step don't affect the other ones. But it will obviously be a non-trivial task - we'd better do it separately.

igpetrov:
> LGTM My general feeling is that the release pipeline is getting too big - maybe it makes sense to try splitting it into a few independent steps, so that failures in one step don't affect the other ones. But it will obviously be a non-trivial task - we'd better do it separately.

I was thinking about the same today. Created this task: https://github.com/camunda/connectors-bundle/issues/479

PR #529, opened 2023-05-03
fix: use spring-boot-3-style autoconfiguration

chillleader:
## Description

Minor adjustments to support the new auto-configuration registration pattern defined in Spring Boot 3 (see [migration guide](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes#auto-configuration-registration)). The auto-configuration needs to be defined in a different file now. I also kept the old file to keep it potentially compatible with spring boot 2.


igpetrov (approved):
LGTM

PR #533, opened 2023-05-04
chore:(sqs)  add tests for inbound sqs connector

Oleksiivanov:
## Description

add tests for inbound sqs connector

closes #
https://github.com/camunda/connectors-bundle/issues/507

### DIFF
--- a/connectors/sqs/src/test/java/io/camunda/connector/inbound/SqsExecutableTest.java
+++ b/connectors/sqs/src/test/java/io/camunda/connector/inbound/SqsExecutableTest.java
@@ -34,6 +34,7 @@ import java.util.ArrayList;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
@@ -92,8 +93,9 @@ class SqsExecutableTest {
     sqsExecutable.activate(spyContext);
     // then
     assertThat(consumer.isQueueConsumerActive()).isTrue();
-    Thread.sleep(200);
     consumer.setQueueConsumerActive(false);
+    executorService.shutdown();
+    executorService.awaitTermination(1, TimeUnit.SECONDS);
     verify(spyContext).replaceSecrets(properties);
     verify(spyContext).validate(properties);
     verify(spyContext, atLeast(1)).correlate(message);


connectors/sqs/src/test/java/io/camunda/connector/inbound/SqsExecutableTest.java
```diff
@@ -0,0 +1,166 @@
...
+        .thenReturn(sqsClient);
+
+    executorService = Executors.newSingleThreadExecutor();
+
+    isQueueConsumerActive = new AtomicBoolean(true);
+  }
+
+  @ParameterizedTest
+  @MethodSource("successRequestCases")
+  @MockitoSettings(strictness = Strictness.LENIENT)
```
markfarkas-camunda, line 88: Why do we need lenient strictness level here? Is it because of Unnecessary Stubbing exception? If so, can't we resolve that, instead of bypassing it?

connectors/sqs/src/test/java/io/camunda/connector/inbound/SqsExecutableTest.java
```diff
@@ -0,0 +1,165 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
markfarkas-camunda, line 53: Do we need this?

connectors/sqs/src/test/java/io/camunda/connector/inbound/SqsQueueConsumerTest.java
```diff
@@ -0,0 +1,136 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.mockito.junit.jupiter.MockitoSettings;
+import org.mockito.quality.Strictness;
+
+@ExtendWith(MockitoExtension.class)
+@MockitoSettings(strictness = Strictness.LENIENT)
```
markfarkas-camunda, line 36: Do we need this?

connectors/sqs/src/test/java/io/camunda/connector/inbound/SqsQueueConsumerTest.java
```diff
@@ -0,0 +1,142 @@
...
+    Thread thread =
+        new Thread(
+            () -> {
+              consumer.run();
+              verify(sqsClient).receiveMessage(any(ReceiveMessageRequest.class));
+            });
+    thread.start();
+    // then
+    CountDownLatch waiter = new CountDownLatch(1);
+    waiter.await(1, TimeUnit.SECONDS);
```
markfarkas-camunda, line 138: We can eliminate the waiting time entirely. Since you use a do-while loop, that will execute at least once regardless of the value of isQueueConsumerActive, you can set that to false at the beginning to execute your do-while loop only once. Now you only have to wait for the thread to finish. Luckily we have [Thread.join()](https://docs.oracle.com/javase/8/docs/api/java/lang/Thread.html#join--) method.
So you can write this:
```
    thread.start();
    isQueueConsumerActive.set(false);

    // then
    thread.join();
    verifyNoInteractions(context);
```
This way we do not block the execution of the unit tests. 
Related: I am still not convinced using a global boolean variable is the best approach to stop the loop. see: https://github.com/camunda/connectors-bundle/pull/533#discussion_r1196483966

markfarkas-camunda (changes requested):
Please consider changing consumeRun_withNoResults method based on my comment.

markfarkas-camunda (approved):
I added a small fix to the test. Please check it, and if you think it is okay, then we can merge it from my point of view. 👍

PR #539, opened 2023-05-04
fix: connector result not propagated to process variables

chillleader:
## Description

Adds the connector output variable setter call (was lost during the runtime migration).


igpetrov (approved):
LGTM, thank you!

PR #584, opened 2023-05-12
fix(default bundle): returned to port 8080

igpetrov:
fix(default bundle): returned to port 8080

### DIFF
--- a/bundle/default-bundle/src/main/resources/application.properties
+++ b/bundle/default-bundle/src/main/resources/application.properties
@@ -1,12 +1,9 @@
-
-server.port=9898
+server.port=8080
 zeebe.broker.gateway-address=localhost:26500
 zeebe.client.security.plaintext=true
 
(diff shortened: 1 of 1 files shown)


chillleader (approved):
Thank you

bundle/default-bundle/src/main/resources/application.properties
```diff
@@ -1,13 +1,10 @@
-
-server.port=9898
+server.port=8080
 zeebe.broker.gateway-address=localhost:26500
 zeebe.client.security.plaintext=true
 
 camunda.connector.polling.enabled=true
 camunda.connector.polling.interval=5000
 
-camunda.connector.webhook.enabled=true
```
chillleader, line 9: Why do we need to remove this?

chillleader: Got it, thanks!

PR #593, opened 2023-05-16
fix(config): remove default properties in the starter

chillleader:
## Description

If you use the starter in your project and define a custom YAML config, these default properties will override your custom config.

See discussion https://github.com/camunda-community-hub/spring-zeebe/issues/429#issuecomment-1549336734


markfarkas-camunda (approved):
Great! The extra notes are really helpful.

PR #598, opened 2023-05-16
feat(runtime): add element id to inbound connector properties

chillleader:
## Description

Adds `elementId` to the API endpoint to enable filtering in the web modeler.

⚠️ Depends on https://github.com/camunda/connector-sdk/pull/466

## Related issues

related to https://github.com/camunda/connectors-bundle/issues/591

### DIFF
--- a/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/ActiveInboundConnectorResponse.java
+++ b/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/ActiveInboundConnectorResponse.java
@@ -20,20 +20,30 @@ import java.util.Map;
 import java.util.Objects;
 
 public class ActiveInboundConnectorResponse {
+
   private final String bpmnProcessId;
+  private final String elementId;
   private final String type;
   private final Map<String, Object> data;
 
   public ActiveInboundConnectorResponse(
-      final String bpmnProcessId, final String type, final Map<String, Object> data) {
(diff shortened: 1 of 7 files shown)


sbuettner (approved):

spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/ActiveInboundConnectorResponse.java
```diff
@@ -20,14 +20,20 @@
...
+  private final String elementId;
   private final String type;
   private final Map<String, Object> data;
 
   public ActiveInboundConnectorResponse(
-      final String bpmnProcessId, final String type, final Map<String, Object> data) {
+      final String bpmnProcessId,
+      final String elementId,
+      final String type,
+      final Map<String, Object> data) {
```
igpetrov, line 33: Instead of generic `data` maybe consider more precise name, e.g. `connectorPayloadData` - just a thought.

igpetrov (approved):

spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/ActiveInboundConnectorResponse.java
chillleader: Initially, I wanted to keep this name generic on purpose because we may want to provide all kinds of data in the future. But maybe @sbuettner will also want to change it in his health checks PR.

PR #618, opened 2023-05-22
feat(experimental): SNS connector on Kotlin

igpetrov:
Experimental kotlin usage.

This actually works


igpetrov:
Decision. Will not introduce Kotlin now.

PR #625, opened 2023-05-23
fix(sqs inbound): correlate message body as object if it possible

Oleksiivanov:
## Description

Parse received string property to object if is a JSON type.

now we can use activation condition for message body :

Screenshot 2023-05-23 at 18 16 21

## Related issues

closes #

https://github.com/camunda/connectors-bundle/issues/623

### DIFF
--- a/connectors/sqs/src/main/java/io/camunda/connector/inbound/model/message/SqsInboundMessage.java
+++ b/connectors/sqs/src/main/java/io/camunda/connector/inbound/model/message/SqsInboundMessage.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.inbound.model.message;
+
+import java.util.Map;
+import java.util.Objects;
+
+public class SqsInboundMessage {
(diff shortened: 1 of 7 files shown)


connectors/sqs/src/main/java/io/camunda/connector/inbound/model/message/SqsInboundMessage.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+package io.camunda.connector.inbound.model.message;
+
+import java.util.Map;
+import java.util.Objects;
+
+public class SqsInboundMessage {
+  private String messageId;
+  private String receiptHandle;
+  private String mD5OfBody;
```
igpetrov, line 15: This looks weird. Consider the following?

```suggestion
  private String bodyMD5Checksum;
```

connectors/sqs/src/main/java/io/camunda/connector/inbound/model/message/SqsInboundMessage.java
```diff
@@ -0,0 +1,130 @@
...
+
+  public String getReceiptHandle() {
+    return receiptHandle;
+  }
+
+  public void setReceiptHandle(final String receiptHandle) {
+    this.receiptHandle = receiptHandle;
+  }
+
+  public String getmD5OfBody() {
```
igpetrov, line 37: If that works out, consider renaming getters/setter

igpetrov (approved):
LGTM, minor suggestions.

PR #626, opened 2023-05-23
feat(health): Report inbound health for Kafka & RabbitMQ

sbuettner:
## Description

Report Kafka inbound connector health.

## Related issues

relates to #621

### DIFF
--- a/connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
+++ b/connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
@@ -12,28 +12,17 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.argThat;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.timeout;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
 import static org.mockito.Mockito.when;
 
 import com.google.gson.Gson;
 import io.camunda.connector.kafka.outbound.model.KafkaTopic;
 import io.camunda.connector.test.inbound.InboundConnectorContextBuilder;
 import io.camunda.connector.test.inbound.InboundConnectorPropertiesBuilder;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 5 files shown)


chillleader (approved):
Looks great as a first iteration - although we need to double check how we handle the exceptions in the runtime and make the contract between runtime and connectors clear.

connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
```diff
@@ -96,27 +83,36 @@ public void setUp() {
...
-    doNothing().when(mockConsumer).assign(any());
-    doAnswer(new AnswersWithDelay(100, new Returns(new ConsumerRecords<>(new HashMap<>()))))
-        .when(mockConsumer)
-        .poll(any());
     KafkaExecutable kafkaExecutable = getConsumerMock();
 
+    when(mockConsumer.partitionsFor(any())).thenReturn(topicPartitions);
+
+    // Return and stop looping
+    when(mockConsumer.poll(any()))
```
markfarkas-camunda, line 91: Nice adjustment to get rid of the delay, thank you 👍

connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
```diff
@@ -12,28 +12,17 @@
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.argThat;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.timeout;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.*;
```
markfarkas-camunda, line 15: Just a small thing: we have an agreement that we don't use wildcard imports. IntelliJ automatically uses wildcards but you can turn this off: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

markfarkas-camunda (approved):
Looks good to me, just a minor thing with the imports.

connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
sbuettner: @markfarkas-camunda Good point. I think it depends on the number of imports but if that has been decided, shouldnt we then reflect this in our spotless config?

sbuettner: Fixed

PR #632, opened 2023-05-25
feat: support multiple secret providers

chillleader:
## Description

This PR adds support for the new multiple secret providers features from the Connector SDK.
It is based on the following ideas:

- Secret Providers can be defined as Spring beans
- Secret Providers can also be defined using the ServiceLoader pattern
- Strategy to select the secret provider can be customized by implementing `SecretProviderAggregator`. It can also be defined as a Spring bean, overriding the default simple implementation.

## Related issues

see also: [previous PR in the SDK](https://github.com/camunda/connector-sdk/pull/472)

closes https://github.com/camunda/connector-sdk/issues/425


igpetrov (approved):
LGTM

PR #648, opened 2023-05-31
Demo: change parser initialisation

### LINKED ISSUE #606: [Google Connectors family] Change Gson parser to Jackson one
Currently, we need to add two different annotations for enum in order to parse it. Find the way how to add only one.
Example: ```io/camunda/google/model/AuthenticationType.java```

mrebrii:
## Description

Draft PR is created in order to decide with approach apply.

## Related issues

closes #606

### DIFF
--- a/connectors/google/google-drive/src/main/java/io/camunda/connector/gdrive/GoogleDriveFunction.java
+++ b/connectors/google/google-drive/src/main/java/io/camunda/connector/gdrive/GoogleDriveFunction.java
@@ -8,6 +8,7 @@ package io.camunda.connector.gdrive;
 
 import com.google.api.client.json.JsonParser;
 import com.google.api.client.json.gson.GsonFactory;
+import com.google.gson.Gson;
 import io.camunda.connector.api.annotation.OutboundConnector;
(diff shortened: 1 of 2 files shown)


connectors/google/google-drive/src/main/java/io/camunda/connector/gdrive/GoogleDriveFunction.java
```diff
@@ -49,6 +50,7 @@ public Object execute(final OutboundConnectorContext context) {
   }
 
   private GoogleDriveRequest parseVariablesToRequest(final String requestAsJson) {
+    GoogleDriveRequest googleDriveRequest = new Gson().fromJson(requestAsJson, GoogleDriveRequest.class);
```
mrebrii, line 53: just add as an example how parsing will be

igpetrov: I am wondering if [this one](https://camunda.slack.com/archives/C0350959WG1/p1685552122703619) is applicable. Let's have this discussion tmr on daily meeting.

PR #652, opened 2023-05-31
chore: restructure connector runtime

### LINKED ISSUE #525: Separate `spring-boot-starter-connectors` from the actual runtime code
__What should we do?__

We should deliver the `spring-boot-starter-connectors` as a separate artifact. Currently the runtime code lives in the spring boot starter, but we should extract it.

__Why should we do it?__

It is considered best practice for spring boot starters to be separate thin artifacts.

chillleader:
## Description

⚠️ No need for an in-depth review of every file: this is mostly either files moved between packages in this repo, or between SDK and this repo. Here is a suggested plan for review:
- Please check out the project locally
- Validate the new project structure
- It only makes sense to check `spring-boot-starter-camunda-connectors` in detail (contains new auto configurations)

___

This PR aims to achieve the following:
- Restructure the Connector runtime by separating the runtime code from Spring boot starter
- Remove the `runtime-util` module from the SDK repo and move it here instead.

### New repository structure

```
connectors-bundle/
├── connectors/                                    (untouched)
│   ├── slack/
│   ├── http-json/
│   └── ...
├── connector-runtime/
│   ├── connector-runtime-core/                     (former runtime-util from the SDK)
│   ├── connector-runtime-spring/                   (normal connector runtime code, moved from the starter)
│   └── spring-boot-starter-camunda-connectors/     (now a thin layer that only adds auto-configurations)
├── bundle/                                        (untouched)
|   ├── default-bundle/                            
|   └── saas-bundle/
```

Previously, the runtime code was split between this repo and the Connector SDK repo. After this refactoring, the whole runtime is now part of this mono repo, but the modular structure is retained:
- `connector-runtime-core` is a set of basic utilities that can help build a custom runtime
- `connector-runtime-spring` is the actual code of our connector runtime, can be embedded into any Spring application
- `spring-boot-starter-camunda-connectors` is a Spring Boot starter that provides auto-configurations for the Spring-based runtime code.

## Related issues

closes #525


sbuettner:
@chillleader Would it be possible to migrate even more classes from the `connector-sdk/core` project over into the runtime?
Thinking about: 
```
core/src/main/java/io/camunda/connector/impl/context
core/src/main/java/io/camunda/connector/impl/secret
core/src/main/java/io/camunda/connector/impl/config
core/src/main/java/io/camunda/connector/test/inbound
core/src/main/java/io/camunda/connector/test/outbound
core/src/main/java/io/camunda/connector/impl/ConnectorInputException.java
core/src/main/java/io/camunda/connector/impl/Constants.java
core/src/main/java/io/camunda/connector/impl/ReflectionHelper.java
```

sbuettner:
@chillleader Thank you for the investigation. Lets tackle the other artifacts at a later stage when we refactor the core sdk.

PR #670, opened 2023-06-05
642 unique webhook ids

### LINKED ISSUE #642: Unique Webhook IDs

Connectors currently allow to register multiple webhooks pointing to the same `Webhook ID` e.g. URL/endpoint. This allows customers to invoke multiple inbound connectors with a single webhook request. This behaviour doesnt provide any transactional safety as one of the inbound connectors could fail while others were executed successfully.

To provide a clearer execution model for customers, the runtime should limit the usage of a single `Webhook ID` to one inbound connector. The endpoint should also return a non-successful HTTP status code if that connector wasnt executed successfully.

As the runtime cant guarantee uniqueness on the zeebe/operate level it has to do the uniqueness check itself. To allow customers to react to this uniqe constraint violation the runtime should report the inbound webhook connector as `DOWN` via the health report facilities. This way customers can see errors in the web modeler and refactor their BPMN models accordingly.

The runtime keeps an in-memory index of all inbound webhook connectors. The first inbound connector for a `Webhook ID` will get activated while others that use the same `ID` will be reported as `DOWN`.

Keeping the existing multiplexing strategy doesnt provide more value while leading to a confusing transactional model. Therefore the decision was made to only support a 1:1 mapping of an endpoint to an inbound webhook connector.

sbuettner:
## Description

Prohibit duplicate webhook ids.

## Related issues

closes #642


chillleader (approved):
Nice 👍

PR #678, opened 2023-06-06
feat: inbound metrics

chillleader:
## Description

In this PR I added new metrics for inbound connectors and slightly refined the existing metrics.

- Count inbound activations, deactivations, triggers as well as trigger outcomes, plus imported process definitions.
- The semantics of inbound metrics were changes:
  - Previously `activated` metric related to connector trigger events.
  - Now it is related to connector instance activations/deactivations, and triggers are recorded separately.

⚠️ Depends on https://github.com/camunda-community-hub/spring-zeebe/pull/440

Grafana dashboard will follow in another PR.

## Related issues

closes https://github.com/camunda/team-connectors/issues/290

### DIFF
--- a/bundle/default-bundle/src/main/resources/application.properties
+++ b/bundle/default-bundle/src/main/resources/application.properties
@@ -12,6 +12,5 @@ camunda.operate.client.username=demo
 camunda.operate.client.password=demo
 camunda.connector.webhook.enabled=true
 
-management.server.port=9080
 management.context-path=/actuator
 management.endpoints.web.exposure.include=metrics,health,prometheus
--- a/bundle/default-bundle/src/main/resources/application.properties
+++ b/bundle/default-bundle/src/main/resources/application.properties
@@ -11,3 +11,7 @@ camunda.operate.client.url=http://localhost:8081
 camunda.operate.client.username=demo
 camunda.operate.client.password=demo
 camunda.connector.webhook.enabled=true
+
+management.server.port=9080
+management.context-path=/actuator
+management.endpoints.web.exposure.include=metrics,health,prometheus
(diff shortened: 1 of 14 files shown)


bundle/default-bundle/src/main/resources/application.properties
```diff
@@ -11,3 +11,7 @@ camunda.operate.client.url=http://localhost:8081
 camunda.operate.client.username=demo
 camunda.operate.client.password=demo
 camunda.connector.webhook.enabled=true
+
+management.server.port=9080
```
igpetrov, line 15: FYI: we will also need to update Helm charts: https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform/templates/connectors/deployment.yaml#L131

Created a task [here](https://github.com/camunda/connectors-bundle/issues/680) so that we don't forget.

I'll fix it closer to release or when I have time. Alternatively, you can fix that too, and I can help you with testing.

igpetrov (dismissed):

bundle/default-bundle/src/main/resources/application.properties
sbuettner: @chillleader Do we have to introduce this change?

igpetrov: @chillleader ok, just maybe don't commit on much time. For me it will take ~2-3 hours.

Also, `camunda-platform` will also need this change reflected. https://github.com/camunda/camunda-platform/blob/main/docker-compose.yaml#L151

chillleader: I removed this port change altogether so that we don't have to change the Helm charts.

igpetrov (approved):

PR #701, opened 2023-06-08
fix(sqs inbound): add try catch for correlate method

Oleksiivanov:
## Description

add Try catch to correlate method

### DIFF
--- a/connectors/sqs/src/main/java/io/camunda/connector/inbound/SqsQueueConsumer.java
+++ b/connectors/sqs/src/main/java/io/camunda/connector/inbound/SqsQueueConsumer.java
@@ -12,6 +12,7 @@ import com.amazonaws.services.sqs.model.ReceiveMessageRequest;
 import com.amazonaws.services.sqs.model.ReceiveMessageResult;
 import io.camunda.connector.api.inbound.InboundConnectorContext;
 import io.camunda.connector.api.inbound.InboundConnectorResult;
+import io.camunda.connector.impl.ConnectorInputException;
 import io.camunda.connector.inbound.model.SqsInboundProperties;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


connectors/sqs/src/main/java/io/camunda/connector/inbound/SqsQueueConsumer.java
```diff
@@ -38,28 +39,37 @@ public SqsQueueConsumer(
...
     do {
       receiveMessageResult = sqsClient.receiveMessage(receiveMessageRequest);
       List<Message> messages = receiveMessageResult.getMessages();
       for (Message message : messages) {
-        InboundConnectorResult<?> correlate =
-            context.correlate(MessageMapper.toSqsInboundMessage(message));
-        if (correlate.isActivated()) {
+        try {
+          correlate(message);
           sqsClient.deleteMessage(properties.getQueue().getUrl(), message.getReceiptHandle());
```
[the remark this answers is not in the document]
sbuettner: @Oleksiivanov We dont delete messages that failed correlation because of an expception but we delete them in case of failed acitvation. If you read the comments from @chillleader you can see that we dont requeue messages that were not activated:
> NACK (no requeue) - inbound event not correlated
getChannel().basicReject(envelope.getDeliveryTag(), false);

In the case of a failed correlation because of an exeption we do requeue: 
> NACK (requeue) - failed to correlate event", e);
getChannel().basicReject(envelope.getDeliveryTag(), true);

connectors/sqs/src/main/java/io/camunda/connector/inbound/SqsQueueConsumer.java
```diff
@@ -38,28 +39,36 @@ public SqsQueueConsumer(
...
-          LOGGER.debug(
-              "Inbound event correlated successfully: {}, and message with ID {} was deleted",
-              correlate.getResponseData(),
-              message.getMessageId());
-        } else {
-          LOGGER.debug("Inbound event not correlated: {}", correlate.getErrorData());
+        } catch (ConnectorInputException e) {
+          LOGGER.warn("NACK (no requeue) - failed to parse SQS message body: {}", e.getMessage());
+        } catch (Exception e) {
+          LOGGER.debug("NACK (requeue) - failed to correlate event", e);
```
igpetrov, line 55: One exception statement says `no requeue`, the other one `requeue` but they both do nothing. Is that expected?

connectors/sqs/src/main/java/io/camunda/connector/inbound/SqsQueueConsumer.java
```diff
@@ -38,28 +39,36 @@ public SqsQueueConsumer(
...
     LOGGER.info("Stopping SQS consumer for queue {}", properties.getQueue().getUrl());
   }
 
+  private void correlate(final Message message) {
+    InboundConnectorResult<?> correlate =
+        context.correlate(MessageMapper.toSqsInboundMessage(message));
+    if (correlate.isActivated()) {
+      LOGGER.debug("Inbound event correlated successfully: {}", correlate.getResponseData());
+    } else {
+      LOGGER.debug("Inbound event not correlated: {}", correlate.getErrorData());
```
igpetrov, line 68: It was correlated but not activated.

igpetrov (approved):

igpetrov:
Let's wait for @sbuettner for one more review round

connectors/sqs/src/main/java/io/camunda/connector/inbound/SqsQueueConsumer.java
```diff
@@ -38,28 +39,36 @@ public SqsQueueConsumer(
   @Override
   public void run() {
     LOGGER.info("Started SQS consumer for queue {}", properties.getQueue().getUrl());
+
     final ReceiveMessageRequest receiveMessageRequest = createReceiveMessageRequest();
     ReceiveMessageResult receiveMessageResult;
     do {
       receiveMessageResult = sqsClient.receiveMessage(receiveMessageRequest);
       List<Message> messages = receiveMessageResult.getMessages();
       for (Message message : messages) {
```
sbuettner, line 48: Please include the `receiveMessage` in the try catch block. Otherwise the connector stops working when there is a network error for example.

sbuettner (changes requested):
Please change the scope of the `try ... catch`.

sbuettner (approved):

PR #731, opened 2023-06-13
feat(error):  Feel expression error information

### LINKED ISSUE #730: Refine FEEL expression error handling
__What should we do?__

We currently return an empty response in case of a failed feel expression. Which is good from a security endpoint perspective but we should at least show that an expression error occured.

__Why should we do it?__

We should do it to allow customers to debug their webhook endpoints.

sbuettner:
## Description

Provide feel expression error information without context variables to simplify debugging of webhooks.

## Related issues

closes #730


markfarkas-camunda (approved):
Looks good to me 👍

PR #734, opened 2023-06-13
fix(aws-lambda): Fix how AWS region is resolved

### LINKED ISSUE #733: Fix broken AWS Lambda contract
__What should we do?__

The current implementation is not compatible with previous version of the outbound connector.

__Why should we do it?__

Introduce a compatibility layer that fetches the region from the previously used property location.

sbuettner:
## Description

Fix how AWS region is resolved in a backwards compatible way.

## Related issues

closes #733


markfarkas-camunda (approved):
LGTM 👍

PR #738, opened 2023-06-13
733 missing secret annotation spotless

### LINKED ISSUE #733: Fix broken AWS Lambda contract

sbuettner:
## Related issues

closes #733


markfarkas-camunda (approved):
All good!

PR #747, opened 2023-06-15
feat(webhook): added HTTP method to properties

### LINKED ISSUE #610: Webhook: add method to element template / connector
TBD: https://github.com/camunda/product-hub/issues/1253

__What should we do?__

We need to add method selection for inbound connector tasks.

__Why should we do it?__

Customer request. Allows Camunda to serve a sort of API gateway.

igpetrov:
## Description

Added HTTP method to properties

closes #610


igpetrov:
> LGTM, small comment , I think that need to add `method` property to `equals` and `hashcode` methods in `WebhookConnectorProperties`

Good catch. Updated.

PR #756, opened 2023-06-16
feat: get and verify jwt token in webhook connector

markfarkas-camunda:
## Description

Added support for JWT token.

1. Get the JWT from the Authorization header
2. Verify it using the given well-known url of the JWKs
3. Extract roles using the FEEL engine
4. Verify the required roles are present

closes #
https://github.com/camunda/connectors-bundle/issues/629


connectors/webhook/element-templates/webhook-connector-start-event.json
```diff
@@ -140,6 +140,76 @@
...
+        },
+        {
+          "name": "Disabled",
+          "value": "disabled"
+        }
+      ],
+      "binding": {
+        "type": "zeebe:property",
+        "name": "inbound.shouldValidateJwt"
+      }
```
sbuettner, line 163: What about something more generic like "authorizationType" and we map the string to an enum like `None|Jwt`?

connectors/webhook/src/main/java/io/camunda/connector/inbound/authorization/JWTChecker.java
```diff
@@ -0,0 +1,87 @@
...
+    final String jwtToken = JWTChecker.extractJWTFomHeader(payload);
+    DecodedJWT verifiedJWT = JWTChecker.verifyJWT(jwtToken, jwkProvider);
+
+    // parse this as json and expose this FEEL to extract roles, and test that against the
+    // expression
+    byte[] decodedBytes = Base64.getDecoder().decode(verifiedJWT.getPayload());
+    String decodedPayload = new String(decodedBytes);
+    JsonNode jsonNode = objectMapper.readTree(decodedPayload);
+    // FeelParserWrapper.parseIfIsFeelExpressionOrGetOriginal(jsonNode);
+    FeelEngineWrapper feelEngine = new FeelEngineWrapper();
```
sbuettner, line 49: Do we have to create the `FeelEngineWrapper` everytime we verify a token?

connectors/webhook/src/main/java/io/camunda/connector/inbound/authorization/JWTChecker.java
```diff
@@ -0,0 +1,87 @@
...
+    List<String> roles = feelEngine.evaluate(webhookConnectorProperties.getJwtRolePath(), jsonNode);
+    if (!roles.containsAll(webhookConnectorProperties.getRequiredPermissions())) {
+      return false;
+    }
+    return true;
+  }
+
+  public static String extractJWTFomHeader(final WebhookProcessingPayload payload) {
```
sbuettner, line 57: We could use an `Optional` for the result type.

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
```diff
@@ -23,6 +26,9 @@ public class WebhookConnectorProperties {
   @Secret private String hmacSecret;
   @Secret private String hmacHeader;
   private String hmacAlgorithm;
+  private String jwkUrl;
+  private String jwtRolePath; // e.g.: roles or role
```
sbuettner, line 30: What about `jwtRoleExpression`?

connectors/webhook/src/main/java/io/camunda/connector/inbound/authorization/JWTChecker.java
markfarkas-camunda: This has been changed based on our discussion on Friday

connectors/webhook/src/main/java/io/camunda/connector/inbound/authorization/JWTChecker.java
markfarkas-camunda: Good point, corrected it

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
markfarkas-camunda: Changed it, thanks

connectors/webhook/pom.xml
```diff
@@ -41,7 +41,27 @@
       <artifactId>guava</artifactId>
       <version>${version.guava}</version>
     </dependency>
-
+    <dependency>
+      <groupId>com.auth0</groupId>
+      <artifactId>java-jwt</artifactId>
+      <version>4.4.0</version>
```
igpetrov, line 47: Shall we generalize versions one level above? The same as we have for `guava` for example?

markfarkas-camunda: Good thinking, but the reason I put it here is because this is very specific for this connector. It is highly unlikely we will ever use this dependency in any other connector. WDYT?

connectors/webhook/element-templates/webhook-connector-start-event.json
```diff
@@ -140,6 +148,76 @@
...
+    {
+      "label": "Required roles",
+      "description": "List of roles to test JWT roles against",
+      "type": "String",
+      "group": "authorization",
+      "feel": "required",
+      "optional": true,
+      "binding": {
+        "type": "zeebe:property",
+        "name": "inbound.requiredPermissions"
```
sbuettner, line 214: Can we move the JWT related properties to `inbound.jwt` leading to`inbound.jwt.requiredPermissions`.

connectors/webhook/pom.xml
sbuettner: @markfarkas-camunda Agree, we can move it if its needed somewhere else.

igpetrov: Good point

chillleader: I'm also good with defining the version in this pom. But I would still recommend to define the version in a property to follow the general code style

connectors/webhook/element-templates/webhook-connector-start-event.json
sbuettner: I would recommend to use the `authorizationType` as a discriminator for a generalized `Auhtorization` sealed class similar to: https://github.com/camunda/connectors-bundle/blob/2b6dcce6252fdb5eccd20bb7fb21704647a634dd/connectors/aws/aws-dynamodb/src/main/java/io/camunda/connector/aws/dynamodb/model/AwsInput.java

bundle/default-bundle/src/main/resources/application.properties
```diff
@@ -1,5 +1,5 @@
 server.port=8080
-zeebe.broker.gateway-address=localhost:26500
+zeebe.client.broker.gateway-address=localhost:26500
```
sbuettner, line 2: 👍

connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
```diff
@@ -120,5 +131,14 @@ public void activate(InboundConnectorContext context) throws Exception {
     }
     props = new WebhookConnectorProperties(context.getProperties());
     context.replaceSecrets(props);
+
+    // jwk url must be specified in the element template for this to work
+    if (props.getJwkUrl() != null) {
```
sbuettner, line 136: We should only do this if `WebhookConnectorProperties.AuthorizationType.JWT.equals(props.getAuthorizationType()` right?

connectors/webhook/src/main/java/io/camunda/connector/inbound/authorization/JWTChecker.java
```diff
@@ -0,0 +1,157 @@
...
+
+public class JWTChecker {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(JWTChecker.class);
+
+  public static boolean verify(
+      final WebhookProcessingPayload payload,
+      final WebhookConnectorProperties webhookConnectorProperties,
+      final JwkProvider jwkProvider,
+      final ObjectMapper objectMapper) {
```
sbuettner, line 42: Instead of passing the `webhookConnectorProperties` why not create a record object that encapsulates all these properties without the need for the dependencies. Examples:
* Instead of `webhookConnectorProperties.getRequiredPermissions()` pass the requiredPermissions as part of the record object.
* Instead of `JWTChecker.extractJWTFomHeader(payload)`  just pass the headers via the record object.

This will allow you to test this class in a more straight forward way.

connectors/webhook/src/test/java/io/camunda/connector/inbound/authorization/JWTIntegrationTest.java
```diff
@@ -0,0 +1,338 @@
...
+            } catch (NoSuchAlgorithmException | InvalidKeySpecException e) {
+              throw new RuntimeException(e);
+            }
+          }
+
+          @Override
+          public String getPrivateKeyId() {
+            return null;
+          }
+        };
```
sbuettner, line 310: Can we extract this into a Utility class?

sbuettner (approved):
Looks great. Some refactoring suggestions.

sbuettner (approved):

connectors/webhook/src/test/java/io/camunda/connector/inbound/authorization/JWTIntegrationTest.java
```diff
@@ -0,0 +1,282 @@
...
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+@Disabled // to be run manually
```
chillleader, line 42: Just curious: does it take too much time to run in a CI environment?

chillleader (approved):
LGTM ⭐

connectors/webhook/src/test/java/io/camunda/connector/inbound/authorization/JWTIntegrationTest.java
markfarkas-camunda: Not at all, only a few seconds. Reason I disabled is because I would like to avoid to lunch a mock API server when we run the tests in GitHub actions. We have to specify to port on which we would like this server to listen and this port might be occupied already which would lead to false test fails.

sbuettner: @markfarkas-camunda I think its great to have this test. Lets try to enable it. You can use https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/test/context/SpringBootTest.WebEnvironment.html#RANDOM_PORT to use a random port for the server and you can use 
```
@LocalServerPort
private Integer port;
``` to inject it.

For wiremock you can use `.dynamicPort()` to use a random port for the mocking server: https://wiremock.org/docs/configuration/

sbuettner (approved):
Great work!

igpetrov (approved):

PR #762, opened 2023-06-19
fix(webhook inbound controller): allow GET PUT DELETE for /inbound/**

Oleksiivanov:
## Description

Allow trigger webhook connector with GET, PUT, DELETE methods


igpetrov (approved):
Good catch!

PR #783, opened 2023-06-26
fix(deps): Use recommended docker base images.

sbuettner:
## Related issues

closes camunda/team-connectors/issues/430

### DIFF
--- a/bundle/camunda-saas-bundle/Dockerfile
+++ b/bundle/camunda-saas-bundle/Dockerfile
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:17.0.7_7-jre
+FROM eclipse-temurin:17.0.7_7-jre-jammy
 
 RUN mkdir /opt/app
 # Download connectors from maven central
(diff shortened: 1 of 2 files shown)


igpetrov (approved):
Should we also make this explicit at the [release action](https://github.com/camunda/connectors-bundle/blob/main/.github/workflows/RELEASE.yaml)?

sbuettner:
@igpetrov What do you mean by `explicit` in this context? And where exactly?

igpetrov:
> @igpetrov What do you mean by `explicit` in this context? And where exactly?

Sorry, I meant tags [here](https://github.com/camunda/connectors-bundle/blob/main/.github/workflows/RELEASE.yaml#L146).

We currently have `<image>` and `<image>-focal`. Maybe we should have `<image>-jammy` and `<image>-focal`, and `<image>` leading to `<image>-jammy`. Or do you think, it's too much?

sbuettner:
Good question. Lets discuss in architecture session and leave the concept as is for now.

PR #815, opened 2023-06-29
Switch to simplified input mapping where possible.

sbuettner:
First draft for simplified object mapping and secret handling.

## Description

The idea is to unify json mapping, secret replacement and validation.

## Related issues

relates to camunda/team-connectors/issues/392

# Info
Please checkout the SDK and `mvn:install` the snapshot of the SDK from https://github.com/camunda/connector-sdk/pull/501


connectors/connectors-common-library/src/main/java/io/camunda/connector/common/auth/Authentication.java
```diff
@@ -16,34 +16,19 @@
...
 
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = BasicAuthentication.class, name = "basic"),
+  @JsonSubTypes.Type(value = NoAuthentication.class, name = "noAuth"),
+  @JsonSubTypes.Type(value = CustomAuthentication.class, name = "credentialsInBody"),
+  @JsonSubTypes.Type(value = OAuthAuthentication.class, name = "oauth-client-credentials-flow"),
+  @JsonSubTypes.Type(value = BearerAuthentication.class, name = "bearer")
+})
 public abstract class Authentication {
```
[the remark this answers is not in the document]
sbuettner: @chillleader Yes, totally. Would love to see that.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -41,12 +44,16 @@ public OutboundConnectorManager outboundConnectorManager(
...
+      ObjectMapper objectMapper,
       MetricsRecorder metricsRecorder) {
     return new OutboundConnectorManager(
         jobWorkerManager,
         connectorFactory,
         commandExceptionHandlingStrategy,
         secretProviderAggregator,
+        validationProvider,
+        objectMapper,
         metricsRecorder);
```
igpetrov, line 57: This constructor is getting bigger, which indicates that this class does to many things. I don't have a good proposal now how we can split/refactor the logic but we should definitely think about it. No need to take actions in this PR.

connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/model/request/impl/AddValues.java
```diff
@@ -60,54 +58,4 @@ public Object getValue() {
...
-        && Objects.equals(type, addValues.type);
-  }
-
-  @Override
-  public int hashCode() {
-    return Objects.hash(spreadsheetId, worksheetName, cellId, value, type);
-  }
-
-  @Override
-  public String toString() {
```
igpetrov, line 96: Shall we keep `equals`, `hashCode`, `toString`?

igpetrov (approved):
LGTM. Left several suggestions but they might be not applicable to this PR.

---

Good that we have a lot of unit tests :)

connectors/connectors-common-library/src/main/java/io/camunda/connector/common/auth/Authentication.java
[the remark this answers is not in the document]
sbuettner: Great idea. Lets do this in a follow-up.

connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/model/request/impl/AddValues.java
sbuettner: @igpetrov Good point. Please take another look.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
```diff
@@ -35,29 +36,54 @@ public class JobHandlerContext extends AbstractConnectorContext
...
 
-  @Override
-  public Map<String, String> getCustomHeaders() {
-    return Optional.ofNullable(job.getCustomHeaders()).orElse(Collections.emptyMap());
+  private String getJsonReplacedWithSecrets() {
+    if (jsonWithSecrets == null) {
+      try {
+        jsonWithSecrets = getSecretHandler().replaceSecrets(job.getVariables());
+      } catch (Exception e) {
+        throw new ConnectorException("SECRET_MAPPING", "Error during secret mapping.");
```
chillleader, line 70: nit: Should we include the message at least in some cases? e.g. if it's a not found error, I would appreciate a hint which secret I forgot to add in the console.

chillleader (approved):
LGTM - but it's not good if we lose the secret provider error messages. I'd leave them as before, I think we don't expose anything sensitive in those messages, and they are really helpful.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
sbuettner: @chillleader Yes it would but that information is afaik not provided by the SDK as of now as `null` is a supported return value from a `SecretProvider` implementation.

sbuettner: I could imagine that we return a result as part of the `replaceSecrets` operation that contains the mapped result but also reports maybe on missing values along the way that we can then make use of here. I could see this as an additional improvement and will add it to the architecture session.

sbuettner: @chillleader Agree, would you be ok to throw a dedicated `SecretException` in `https://github.com/camunda/connector-sdk/blob/f7a2d15ce83acd6d9140611d7a9ab4ecd6ab4dfa/core/src/main/java/io/camunda/connector/impl/secret/SecretUtil.java#L68` in your PR and map this exception here in your PR as well?  As this requires an SDK update as well it might be leaner this way. WDYT?

chillleader: Works for me

connectors/google/google-base/src/main/java/io/camunda/google/model/AuthenticationType.java
```diff
@@ -6,14 +6,17 @@
...
 package io.camunda.google.model;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.api.client.util.Value;
 import com.google.gson.annotations.SerializedName;
 
 public enum AuthenticationType {
+  @JsonProperty("bearer")
   @Value("bearer")
   @SerializedName("bearer")
```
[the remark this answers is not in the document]
sbuettner: Will take care of as part of the cleanup PR.

PR #828, opened 2023-07-04
fix: restore local connector runtime in the default bundle

chillleader:
## Description

Adds an alternative simple way to run the runtime locally, as previously used main class was deleted.

## Related issues

closes #


igpetrov (approved):
Thank you!

sbuettner (approved):

PR #830, opened 2023-07-05
Simplified input mappings for inbound / Deserializing inbound properties as JSON

### LINKED ISSUE #821: Deserialize inbound connector properties to JSON

Inbound connector properties are, unlike variables for outbound connectors, plain strings. They are not handled by Zeebe to evaluate FEEL expressions and they are not provided in JSON format. Compare the two formats below:

```
# Inbound connector properties, as they come from BPMN XML
authentication.username=user
authentication.password=pass
```
```
# Outbound connector variables, as provided by Zeebe
{
  "authentication": {
    "username": "user",
    "password": "pass"
  }
}
```

This creates complications, as we cannot apply the same property/variable handling mechanisms to both outbound and inbound connectors. 

When we parse the BPMN XML and read the inbound connector properties, we should convert the flat format above to JSON. This allows us to reuse the variable handling mechanisms from outbound connectors implementation.

Maintain a custom approach to deserialize properties, we already have something like this.

chillleader:
## Description

This PR introduces the simplified input mappings to inbound connectors.

With these changes, we transition to deserializing inbound connector properties to JSON.

Please review together with the SDK PR: https://github.com/camunda/connector-sdk/pull/504

## Related issues

closes #821


chillleader:
I'd like to extract the custom Jackson deserializer for FEEL expressions into a separate PR to merge these changes faster, WDYT @sbuettner?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundPropertyHandler.java
```diff
@@ -0,0 +1,101 @@
...
+        traverse(path, wrapped, value);
+      }
+    } catch (RuntimeException e) {
+      throw new RuntimeException("Detected malformed Connector properties: " + e.getMessage(), e);
+    }
+    return wrapped;
+  }
+
+  @SuppressWarnings("unchecked")
+  private static void traverse(List<String> path, Map<String, Object> currentRoot, String value) {
```
igpetrov, line 74: (optional) Consider to be more explicit:
```suggestion
  private static void traverseContextMap(List<String> path, Map<String, Object> currentRoot, String value) {
```

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -158,11 +120,42 @@ public Health getHealth() {
...
+    return Objects.hash(definition);
   }
 
   @Override
   public String toString() {
-    return "InboundConnectorContextImpl{" + "properties=" + properties + '}';
+    return "InboundConnectorContextImpl{" + "definition=" + definition + '}';
+  }
+
+  private Map<String, Object> propertiesWithFeel;
```
igpetrov, line 131: nit/styling: maybe it's worth to put this and all below lines above `equals()`.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDefinitionImpl.java
igpetrov, line 1: What's your opinion on adding new `responseBodyExpression` here?

chillleader: I think I would prefer to have only common properties here that are applicable to every connector, and everything else that is connector-specific can be made accessible via `InboundConnectorContext#getProperties` / `InboundConnectorContext#bindProperties`.

And later on, when we add custom a deserializer for `Function` type, we can use that to "compile" the `responseBodyExpression` into a function that will than be evaluated after correlation.

igpetrov: Got it. Thanks for the explanation

igpetrov (approved):
Few minor suggestions

sbuettner:
@chillleader Agree, the FEEL/Function binding support is out of scope and can be done in a separate pull request.

PR #832, opened 2023-07-05
feat(webhook): introduce response body expression

igpetrov:
feat(webhook): introduce response body expression

Related to https://github.com/camunda/connector-sdk/pull/507


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseManager.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.ResponseEntity;
+
+public class WebhookResponseManager {
```
sbuettner, line 35: Maybe `WebhookResponseMapper` would be less generic.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseManager.java
```diff
@@ -0,0 +1,73 @@
...
+  public static final String PROCESS_DATA_KEY = "processData";
+
+  private final FeelEngineWrapper feelEngine;
+
+  @Autowired
+  public WebhookResponseManager(final FeelEngineWrapper feelEngine) {
+    this.feelEngine = feelEngine;
+  }
+
+  public ResponseEntity produceResponse(
```
sbuettner, line 48: `mapResponse`?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseManager.java
```diff
@@ -0,0 +1,73 @@
...
+      final InboundConnectorResult correlatedResult,
+      final InboundConnectorProperties properties,
+      final Object context) {
+    Map evaluatedResponseBody =
+        Optional.ofNullable(evaluateResponseBodyExpression(properties, context))
+            .orElse(Collections.emptyMap());
+    Map response = new HashMap(evaluatedResponseBody);
+    response.put(PROCESS_DATA_KEY, correlatedResult);
```
sbuettner, line 56: I thought the idea was to allow users to decide whether they want to include the `correlatedResult` into the response or not based on their `responseExpression`. We dont need to add it here. Maybe only if the `responseExpression` is null.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseManager.java
```diff
@@ -0,0 +1,73 @@
...
+  private final FeelEngineWrapper feelEngine;
+
+  @Autowired
+  public WebhookResponseManager(final FeelEngineWrapper feelEngine) {
+    this.feelEngine = feelEngine;
+  }
+
+  public ResponseEntity produceResponse(
+      final InboundConnectorResult correlatedResult,
+      final InboundConnectorProperties properties,
```
sbuettner, line 50: Do we need access to the whole `InboundConnectorProperties` here? Otherwise lets only pass `responseExpression`.

igpetrov: While we don't need the whole map of properties here, I would like to encapsulate all response body expression logic here. Because on the other hand, we need to do an extraction of the response body expression in the REST controller. WDYT?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseManager.java
igpetrov: Alright, that makes sense. Changed and added some unit tests.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseManager.java
igpetrov: But we call this from the REST controller. I explicitly created this object to avoid parsing there :)

Or maybe I didn't get your point? 🤔

igpetrov: Also: reworked logic a little bit.

sbuettner: You should be able to use the new @FEEL annotation support here I think.

igpetrov: > You should be able to use the new @feel annotation support here I think.

I haven't seen `@FEEL` annotation being used for service fields, only in properties records. 
@chillleader could you confirm?

igpetrov: > You should be able to use the new FEEL jackson support here I think. You could then simply accept the Function of the expression and evaluate it.

Not sure, I am following, but checking how it's done

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseMapper.java
```diff
@@ -0,0 +1,76 @@
...
+  private final FeelEngineWrapper feelEngineWrapper;
+
+  @Autowired
+  public WebhookResponseMapper(final FeelEngineWrapper feelEngineWrapper) {
+    this.feelEngineWrapper = feelEngineWrapper;
+  }
+
+  public ResponseEntity mapResponse(
+      final InboundConnectorResult correlatedResult,
+      final Map<String, Object> properties,
```
sbuettner, line 48: The extraction of the result expression should not happen here in this context. Please pass it to the mapper itself.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseMapperTest.java
```diff
@@ -0,0 +1,235 @@
...
+    when(correlatedResult.getType()).thenReturn("START_EVENT");
+    when(correlatedResult.getResponseData()).thenReturn(Optional.of(Map.of("myKey", "myVal")));
+  }
+
+  @Test
+  void produceResponse_ResponseBodyExpressionMatchesContext_ReturnMatchedResponse() {
+    // given
+    when(properties.get(RESPONSE_BODY_EXPRESSION_KEYWORD))
+        .thenReturn(
+            "=if request.body.type = \"url_verification\" then {newChallenge: request.body.challenge} else null");
```
sbuettner, line 61: This is an example where it would be way easier to handover the expression/function directly instead of passing it via the properties.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseMapper.java
```diff
@@ -0,0 +1,76 @@
...
+      final Object context) {
+    final var responseBodyExpression =
+        Optional.ofNullable(properties.get(RESPONSE_BODY_EXPRESSION_KEYWORD))
+            .map(Object::toString)
+            .orElse("");
+    if (responseBodyExpression.isBlank()) {
+      return ResponseEntity.ok(Map.of(PROCESS_DATA_KEY, correlatedResult));
+    }
```
sbuettner, line 56: Wouldnt this change the current response format?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseMapper.java
```diff
@@ -0,0 +1,76 @@
...
+      final String responseBodyExpression, final Object context) {
+    if (responseBodyExpression.trim().length() == 0) {
+      return Collections.emptyMap();
+    }
+    try {
+      return feelEngineWrapper.evaluate(responseBodyExpression, context);
+    } catch (FeelEngineWrapperException e) {
+      throw new ConnectorInputException(e);
+    }
+  }
```
sbuettner, line 75: Please expect a `Function` instead of parsing it manually here. This would also avoid the length and string checking.

sbuettner (changes requested):

PR #870, opened 2023-07-10
fix(rabbitmq): edit default properties value to start from '='

chillleader:
## Description

We force this field to be a FEEL expression, but provide the default value without `=`. Web modeler does not implicitly add an equals sign, and we can't deserialize the properties into an object because it arrives to us as a string.

## Related issues

closes #


sbuettner (approved):
Good catch!

PR #871, opened 2023-07-10
fix kafka value serialization; change group id property to include elementId and processDefinitionKey

markfarkas-camunda:
## Description

Bug was found in Kafka serialization, which had to be fixed.  Now it converts FEEL expression (which represented as a scala Map) properly to json.


sbuettner (approved):
Great! Thanks for taking care at that time.

igpetrov (approved):
LGTM, thank you

PR #873, opened 2023-07-11
Standardize jackson configuration

chillleader:
## Description

With the recent issues we encountered in several connectors due to missing modules, it makes sense to have the same object mapper config everywhere.

## Related issues

closes #


sbuettner (approved):
Thank you for taking care.

PR #882, opened 2023-07-11
fix(lambda): fixes correct JSON parsing of the payload

igpetrov:
fix(lambda): fixes correct JSON parsing of the payload

### DIFF
--- a/connectors/aws/aws-lambda/src/main/java/io/camunda/connector/awslambda/model/AwsLambdaResult.java
+++ b/connectors/aws/aws-lambda/src/main/java/io/camunda/connector/awslambda/model/AwsLambdaResult.java
@@ -8,6 +8,7 @@ package io.camunda.connector.awslambda.model;
 
 import com.amazonaws.services.lambda.model.InvokeResult;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
 
 public class AwsLambdaResult {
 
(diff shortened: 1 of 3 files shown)


connectors/aws/aws-lambda/src/main/java/io/camunda/connector/awslambda/model/AwsLambdaResult.java
```diff
@@ -18,8 +19,13 @@ public class AwsLambdaResult {
   public AwsLambdaResult(final InvokeResult invokeResult, final ObjectMapper objectMapper) {
     this.statusCode = invokeResult.getStatusCode();
     this.executedVersion = invokeResult.getExecutedVersion();
-    this.payload =
-        objectMapper.convertValue(new String(invokeResult.getPayload().array()), Object.class);
+    try {
+      this.payload =
+          objectMapper.readValue(new String(invokeResult.getPayload().array()), Object.class);
```
sbuettner, line 24: Lets use:
```
try {
  this.payload = objectMapper.readValue(invokeResult.getPayload().array(), Object.class);
} catch (IOException e) {
  throw new RuntimeException(e);
}
```

igpetrov: Done

PR #889, opened 2023-07-12
fix: filter inbound eligible elements before performing 'sanity check'

markfarkas-camunda:
## Description

Sometimes we threw a `java.lang.IllegalStateException: Sanity check failed: IntermediateCatchEvent` unnecessary which caused issues.

In the fix we filter the elements to only check those elements which actually have an inbound template applied.

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
@@ -105,7 +105,7 @@ public class ProcessDefinitionInspector {
             .filter(
                 element -> {
                   Map<String, String> zeebeProperties = getRawProperties(element);
-                  if (zeebeProperties != null && zeebeProperties.get("inbound.type") != null) {
+                  if (zeebeProperties != null && zeebeProperties.get(Keywords.INBOUND_TYPE_KEYWORD) != null) {
                     return true;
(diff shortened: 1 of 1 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
```diff
@@ -102,16 +102,25 @@ private List<InboundConnectorDefinitionImpl> inspectBpmnProcess(
     List<BaseElement> inboundEligibleElements =
         INBOUND_ELIGIBLE_TYPES.stream()
             .flatMap(type -> process.getChildElementsByType(type).stream())
+            .filter(
+                element -> {
+                  Map<String, String> zeebeProperties = getRawProperties(element);
+                  if (zeebeProperties != null && zeebeProperties.get("inbound.type") != null) {
```
chillleader, line 108: nit: you can reference `Keywords.INBOUND_TYPE_KEYWORD`

igpetrov (approved):

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
markfarkas-camunda: You are right, we should avoid these kind of magic strings, fixed it 👍

PR #895, opened 2023-07-13
fix: use ExecutorService instead of default ForkJoinPool

markfarkas-camunda:
## Description

Kafka inbound connector was only able to activate the subscription for limited amount of connector instances, due to the usage of CompletableFuture which uses ForkJoinPool, which has a limitation of how many threads it can run parallelly. [Documentation](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html) says "[...] parallelism level; by default, equal to the number of available processors"

This solution uses ExecutorService instead of ForkJoinPool to eliminate the limitations of threads.

## Related issues

https://github.com/camunda/connectors-bundle/issues/845

closes #
https://github.com/camunda/connectors-bundle/issues/845


sbuettner (approved):
Lets see how this behaves in case of a lot of kafka connectors / threads but this should unblock the user. Thanks for taking care.

PR #896, opened 2023-07-13
Jackson module for FEEL expressions

chillleader:
## Description

Supports deserializing FEEL expressions into Java functional types (`Function` and `Supplier`).
 
## Related issues

closes https://github.com/camunda/connectors-bundle/issues/851


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/feel/jackson/FeelFunctionDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+    this.outputType = outputType;
+  }
+
+  private final FeelEngineWrapper feelEngineWrapper = new FeelEngineWrapper();
+
+  @Override
+  protected Function<IN, OUT> doDeserialize(String expression) {
+    return (input) -> {
+      Object result = feelEngineWrapper.evaluate(expression, input);
+      return feelEngineWrapper.getObjectMapper().convertValue(result, outputType);
```
sbuettner, line 24: Curious why this is necessary. Cant we just return the result casted to the type?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/feel/jackson/FeelSupplierDeserializer.java
```diff
@@ -0,0 +1,36 @@
...
+  protected FeelSupplierDeserializer(Class<OUT> outputType, FeelEngineWrapper feelEngineWrapper) {
+    super(feelEngineWrapper);
+    this.outputType = outputType;
+  }
+
+  @Override
+  Supplier<OUT> doDeserialize(String expression) {
+    return () -> {
+      Object result = feelEngineWrapper.evaluate(expression, Map.of());
+      return feelEngineWrapper.getObjectMapper().convertValue(result, outputType);
```
sbuettner, line 22: same as above

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/feel/jackson/FeelFunctionDeserializer.java
chillleader: As discussed synchronously, we need that to do the final type conversion into the `Function` output type.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/feel/jackson/FeelSupplierDeserializer.java
```diff
@@ -0,0 +1,51 @@
...
+  protected FeelSupplierDeserializer(Class<OUT> outputType, FeelEngineWrapper feelEngineWrapper) {
+    super(feelEngineWrapper);
+    this.outputType = outputType;
+  }
+
+  @Override
+  Supplier<OUT> doDeserialize(String expression) {
+    // evaluate eagerly to fail fast
+    OUT result = feelEngineWrapper.evaluate(expression, Map.of(), outputType);
+    return () -> result;
```
sbuettner, line 38: Shouldnt we only evaluate expressions eagerly if the target type is neither Function nor Supplier?
I would therefore see three case for the different target types:
* Supplier -> eval lazy
* Function -> eval during application
* Everything else -> eval eagerly (this can be used to directly bind a FEEL expression to a List<String> for example.

chillleader: How would we implement a custom deserializer for the third case? As mentioned, I didn't find a way to contextually deserialize a String into generic type if that string is not guaranteed to be a FEEL expression (if it's just a string, we would need to somehow delegate to another deserializer which we don't know about).

Re. the eager evaluation for supplier, I think it makes sense to validate it and fail fast before running the connector, as it's just an expression without arguments

sbuettner: @chillleader We would only eagerly evaluate if we tested whether the string is a FEEL expression (as we do already) and then evaluate and bind the result right away.

chillleader: I implemented the deserializer you mentioned (`FeelDeserializer`) but couldn't find a way to register it for any property due to Jackson's deserializer selection precedence. So I suggest that we manually enable this deserializer for properties where eager FEEL expression evaluation is required, like this: `@JsonDeserialize(using = FeelDeserializer.class)`.

I think this approach would be better than to wrap the values that need to parsed into some other wrapper type or annotating them with a custom annotation. And if the future we find a way to enable this resolution automatically, the old way with the standard `@JsonDeserialize` annotation will still work and will be easy to migrate from. WDYT?

sbuettner: I think thats a good comprosmise for what we initially planned. Great effort and thanks for further investigating the topic.

sbuettner (approved):

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/feel/jackson/FeelDeserializer.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package io.camunda.connector.runtime.core.feel.jackson;
+
+import com.fasterxml.jackson.databind.BeanProperty;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import io.camunda.connector.runtime.core.feel.FeelEngineWrapper;
+import java.util.Map;
+
+public class FeelDeserializer extends AbstractFeelDeserializer<Object> {
```
sbuettner, line 25: The only issue I see is that connectors would need to depend on the runtime to make use of it. Maybe move it to the SDK?

chillleader: Right, we will then have to move a few dependencies like feel engine and jackson module scala to the SDK.

chillleader: https://github.com/camunda/connector-sdk/pull/525

PR #920, opened 2023-07-21
feat(webhook): rename jwtRoleExpression -> permissionsExpression

Oleksiivanov:
## Description

rename `jwtRoleExpression` to `permissionsExpression` in webhook

### DIFF
--- a/connectors/webhook/element-templates/webhook-connector-intermediate.json
+++ b/connectors/webhook/element-templates/webhook-connector-intermediate.json
@@ -2,7 +2,7 @@
   "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
   "name": "Webhook connector",
   "id": "io.camunda.connectors.webhook.WebhookConnectorIntermediate.v1",
-  "version": 4,
+  "version": 3,
   "description": "Configure webhook to receive callbacks",
   "documentationRef": "https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/http-webhook/",
   "category": {
@@ -265,7 +265,7 @@
       "optional": true,
       "binding": {
         "type": "zeebe:property",
-        "name": "inbound.auth.jwt.jwtRoleExpression"
+        "name": "inbound.auth.jwt.permissionsExpression"
       },
       "condition": {
         "property": "authorizationType",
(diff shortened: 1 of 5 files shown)


connectors/webhook/element-templates/webhook-connector-intermediate.json
```diff
@@ -265,7 +265,7 @@
       "optional": true,
       "binding": {
         "type": "zeebe:property",
-        "name": "inbound.auth.jwt.jwtRoleExpression"
+        "name": "inbound.auth.jwt.permissionsExpression"
```
igpetrov, line 268: It hasn't been updated in web-modeler yet, right? Otherwise we would need to bump versions.

igpetrov (approved):
LGTM. Please wait an approval from @markfarkas-camunda who has more context.

connectors/webhook/element-templates/webhook-connector-intermediate.json
sbuettner: @igpetrov We dont need to bump the version or did we release JWT yet?

sbuettner (approved):

connectors/webhook/element-templates/webhook-connector-intermediate.json
igpetrov: @sbuettner I just checked, we didn't do it.

I however noticed, that we still have `version: 2` in [webmodeler](https://github.com/camunda/web-modeler/blob/main/webapp/src/App/Pages/Diagram/BpmnJSExtensions/connectorsExtension/.camunda/element-templates/webhook-connector-intermediate.json) but `version: 4` in [bundle](https://github.com/camunda/connectors-bundle/blob/main/connectors/webhook/element-templates/webhook-connector-intermediate.json). I don't think it's critical but still don't know what happens if we jump over one version. cc: @Oleksiivanov

markfarkas-camunda: @igpetrov @sbuettner because we haven't released it yet, there is still an open [PR about the element-template](https://github.com/camunda/web-modeler/pull/5422). In this PR you [can see](https://github.com/camunda/web-modeler/pull/5422/files#diff-99809af484b1351c03d41b4753bebd14717b05d5266b9a4f11cc22a95c0b41bd) that we have added version 3. 
@Oleksiivanov I will add this change to that PR. 
Is it clearly defined when should we change version in the element-template?

sbuettner: @markfarkas-camunda Always when we need to change something that brings new functionality. We dont necessarily need to do it for simply description changes.

igpetrov (approved):
LGTM

markfarkas-camunda (approved):
All good, thank you! 👍

PR #931, opened 2023-07-24
fix(webhook): ignore subtracting sign in activation condition

Oleksiivanov:
## Description

allow and ignore subtracting sign in activation condition
with this fix we can set next activation condition : 
Screenshot 2023-07-25 at 01 41 47
 

closes #
https://github.com/camunda/connectors-bundle/issues/930


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/correlation/InboundCorrelationHandler.java
```diff
@@ -146,13 +149,41 @@ protected boolean isActivationConditionMet(
       return true;
     }
     try {
+      // quick fix cases when activation condition with '-' like
+      // '=(request.body.detail-type="MyEvent")'
```
igpetrov, line 153: I think, we should not provide fixes on such cases. It seems to be FEEL engine specifics.

Customers could use `=(get value(request.body, "detail-type") = "MyEvent")` instead.

PR #934, opened 2023-07-26
Add SDK to the monorepo

chillleader:
## Description

Adds Connector SDK to the bundle monorepo.

Notable changes:
- Project was yet again restructured
  - `connector-core` previously contained the `impl` package with runtime-specific classes. They have been moved to the `connector-runtime-core` package
  - FEEL Jackson module was extracted into a separate artifact; now it can be consumed by the runtime, by `connector-test`, or by individual connector implementations, if needed. This module now also defines the common `ObjectMapper` that will be used everywhere instead of creating new instances.
- Adjusted the `@FEEL` annotation (it previously depended on `FeelDeserializer`, making the whole `runtime-core`
 dependant on the Jackson stuff). Now it uses `JacksonAnnotationIntrospector` to assign the deserializer in a decoupled way.
- New root `README.md`

## ❓How to review this PR?

Most of the changes are just code moved from the SDK that doesn't require a re-review.

Suggested review approach:
- Check out the branch locally and get acquainted with the proposed project structure
- Review the following commits in detail:
  - [License policies](https://github.com/camunda/connectors-bundle/pull/934/commits/456f6474d7400d191f4f9937b6300c75e3ba2de8) (Check if all the licenses make sense to you and that we don't suddenly make something open source when it should be source available)
  - [README update](https://github.com/camunda/connectors-bundle/pull/934/commits/ee831248bee638a9339cd23086fc7edefd351605)
  - [API / Implementation split for the Core module + Jackson module extraction](https://github.com/camunda/connectors-bundle/pull/934/commits/2e8d9fa5c91ce9aa40cf156924e660f556aaec15)

## Related issues

closes https://github.com/camunda/team-connectors/issues/454


sbuettner:
@chillleader I would prefer to have a `connector-sdk` project folder on the rootpath and that the other sdk projects like `test`, `validation` etc live under it. WDYT?

chillleader:
@sbuettner I decided to go for a flat structure because IMO the borderline between SDK and non-SDK is somewhat fuzzy, and we considered certain things parts of the SDK just because they were easier to place in the SDK repository. For example, the gcp secret provider implementation probably doesn't belong to the SDK. Parent pom with plugin configs is also probably a top-level module.

If we still like the hierarchical structure more, then I would only place 3 modules there:
```
|/connector-sdk
|--/core
|--/validation
|--/test
|/parent
|/feel-jackson
|/secret-provider
```

sbuettner:
Regarding our feel module, we could name it like: 

https://github.com/zalando/jackson-datatype-money
https://github.com/FasterXML/jackson-datatypes-misc
https://github.com/FasterXML/jackson-datatype-jsr310

and go with:  `jackson-datatype-feel`?

I think its main use-case is the `connector-sdk` for now. Hence we could put it there as well for now as long as we dont publish it as a separate module.

chillleader:
@sbuettner moved the modules according to your suggestion

sbuettner (approved):

PR #940, opened 2023-07-27
chore: made controller tests more explicit

igpetrov:
chore: made controller tests more explicit

### DIFF
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
@@ -22,6 +22,7 @@ import static io.camunda.zeebe.spring.test.ZeebeTestThreadSupport.waitForProcess
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertInstanceOf;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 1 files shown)


connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
```diff
@@ -205,6 +212,44 @@ public void testSuccessfulProcessingWithFailedActivation() throws Exception {
...
+                            "myPath",
+                            new HashMap<>(),
+                            "{}".getBytes(),
+                            new HashMap<>(),
+                            new MockHttpServletRequest());
+
+    assertEquals(200, responseEntity.getStatusCode().value());
+
+    var result = responseEntity.getBody();
+    assertNull(result);
```
sbuettner, line 252: Would it make sense to json assert the response?

igpetrov: it's null :)

as per contract - if no response body expression given, we just return `200 OK` and no body. Json assert won't be different here.

chillleader (approved):

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
sbuettner: Right, I meant to add a case where its not null.

igpetrov: I see your point. For that case we do parsing already, like:

```
ResponseEntity<InboundConnectorResult<?>> responseEntity =
        (ResponseEntity<InboundConnectorResult<?>>)
            controller.inbound(
                "myPath",
                new HashMap<>(),
                "{}".getBytes(),
                new HashMap<>(),
                new MockHttpServletRequest());
```

And I added those use-cases as asserts by actual object properties.

sbuettner: I meant a real test against the actual rendered json response and not the `ResponseEntity`.

igpetrov: But we don't have these now. Or at least I am not aware of. If so, I would suggest to have a separate task that I can take after I am back. Today is the last day before I leave for FTO, and there are more urgent tasks left - regression testing and documentation.

sbuettner: Alright.

sbuettner (approved):

PR #943, opened 2023-07-27
chore: remove whatsapp inbound connectors

igpetrov:
chore: remove whatsapp inbound connectors


sbuettner (approved):
Thanks for taking care!

PR #952, opened 2023-07-28
fix: runs image as non-root user

jessesimpson36:
## Description

Hello Connectors Team,

I wrote this patch as part of  [Enhanced security principles for docker images](https://github.com/camunda/product-hub/issues/717), which is related to some customer issues that my team (Distribution Team) has had to deal with, and is also required to another issue I'm part of [Getting Self-Managed helm chart on AWS Marketplace](https://github.com/camunda/product-hub/issues/1275) 

Some customers have requirements from their IT teams that their containers must run as unprivileged users, and occasionally to run on read-only filesystems. In some cases, this can be implemented in the helm chart with runAsUser: <insertNumber>. However, many clients are unable to do that because the images are scanned for this configuration rather than scanning for the user id in runtime.

It's also in general, a security best practice to run as unprivileged users in case we have any hackers breaking into the container and then escaping onto the host machine.

This change will create a user inside the connectors-bundle docker image, and set the current user to the created one.

I see there are some other Dockerfiles,
```
> find . -name "Dockerfile*"
./bundle/camunda-saas-bundle/Dockerfile
./bundle/default-bundle/Dockerfile
./bundle/default-bundle/Dockerfile-focal
./connector-runtime/connector-runtime-application/Dockerfile
./connector-runtime/connector-runtime-application/example/Dockerfile
```
Please let me know if you think this patch should go inside the other dockerfiles as well.

I've tested the new docker image by running the self-managed helm chart and verifying that the application comes up as healthy.

## Related issues

Related:
* https://github.com/camunda/product-hub/issues/717
* https://github.com/camunda/product-hub/issues/1275

Other PRs for similar changes in other components:
* https://github.com/camunda/operate/pull/5020
* https://github.com/camunda-cloud/identity/issues/1967
* https://github.com/camunda/zeebe/pull/13418


sbuettner:
Hi @jessesimpson36 thanks for the context and the patch. It would be great if you could add your changes to the other images as well. Thank you!

jessesimpson36:
Sounds good. I checked our camunda-operator repo (from Controller team) and have verified that they do not have any persistent volume claims that might get messed up by improper permissions.  I think this change will be safe for them.

PR #981, opened 2023-08-01
feat(spring): Support secret lookup from spring config sources

### LINKED ISSUE #947: Use Spring Environment facility instead of System.getEnv

Spring users expect the out-of-the-box secret provider to include all information that is available in Spring's [Environment](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html) context to make use of application properties (and other configures configuration providers) when resolving secrets in addition to the existing support of environment variables based on `System.getEnv`.

Inject and use the [Environment](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html) when resolving secrets and register it as the default secret provider for the self-managed bundle.

sbuettner:
## Related issues

closes #947


connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java
```diff
@@ -39,6 +45,9 @@
 @EnableConfigurationProperties(ConnectorProperties.class)
 public class OutboundConnectorsAutoConfiguration {
 
+  @Value("${camunda.connector.secret-provider.discovery.enabled:true}")
+  Boolean secretProviderLookupEnabled;
```
chillleader, line 49: I would add this one and `camunda.connector.secret-provider.environment.enabled` to [ConnectorProperties](https://github.com/camunda/connectors/blob/68a47f72e82c3b284ba838c6ca7da46bf44038cb/connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/ConnectorProperties.java) to enable autocompletion in properties files and keep all our properties in the same place.

I think the `additional-spring-configuration-metadata` file won't be needed then right?

chillleader (approved):
LGTM, just a minor comment. I like the approach, specifically how the lookup is now extracted from `SecretProviderAggregator`. The fact that we now support using both Spring-defined and SPI secret providers at the same time is probably also a good thing.

connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java
sbuettner: Thanks for the hint. Fixed!

PR #989, opened 2023-08-02
feat(spring): Support spring bean based connectors

### LINKED ISSUE #809: Support Connectors configured with Spring

Although the Connector Runtime is based on Spring, so far we have offered limited support of what Spring has to offer. We have defined 2 problems related to our Spring support.
- If you are using the default Connector Runtime, _(outbound)_ Connectors can be discovered if they are defined as Spring beans (see `OutboundConnectorAnnotationProcessor`), but we don't use the Spring container to actually get the Connector bean instances. Instead, we parse the configuration and create the Connector instances ourselves. This is confusing, as the two different approaches are mixed. It would be great to be able to configure Connectors as Spring beans and rely on Connector Runtime to actually use the Spring container to create the instances of Connectors.
- While Connectors are often developed in a Spring environment, it is still impossible to use Spring dependency injection to initialize Connector objects. This creates additional complexity for Spring-based projects, especially when Connectors are integrated into existing projects where pre-configured beans are already available.

We would like to improve the Spring support in Connectors, so that:
- Connectors can be defined and consumed by the Runtime as Spring beans
- Developers can rely on Spring DI while creating custom Connectors.

Here is what a Connector may look like if these requirements are met:

```java
@Component
@OutboundConnector(
    name = "MYAPPCONNECTOR",
    inputVariables = {"myProperty", "authentication"},
    type = "io.camunda:my-app-connector:1")
public class MyAppConnector implements OutboundConnectorFunction {

  private MyAppClient client;

  @Autowired
  public MyAppConnector(MyAppClient client) { this.client = client; }

  @Override
  public Object execute(OutboundConnectorContext context) throws Exception {
    // connector code
  }
}
```

Relying on Spring DI makes Connectors written for Spring-based Connector Runtime potentially incompatible with other Connector Runtimes. Because of this, we can consider implementing the same functionality based on factory bean pattern (e.g. user defines a Runtime-specific factory bean that populates Connector objects, while the Connector code remains unaware of the upstream Spring dependency).

This would probably be less convenient for developers, but this way we can guarantee that Connector code remains runtime-agnostic.

sbuettner:
## Related issues

closes #809

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java
@@ -19,13 +19,12 @@ package io.camunda.connector.runtime.core.outbound;
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 import io.camunda.connector.runtime.core.ConnectorHelper;
 import io.camunda.connector.runtime.core.config.OutboundConnectorConfiguration;
-import io.camunda.connector.runtime.core.discovery.EnvVarsConnectorDiscovery;
-import io.camunda.connector.runtime.core.discovery.SPIConnectorDiscovery;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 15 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java
```diff
@@ -36,7 +35,7 @@ public class DefaultOutboundConnectorFactory implements OutboundConnectorFactory
   private static final Logger LOG = LoggerFactory.getLogger(DefaultOutboundConnectorFactory.class);
 
   // stores all pre-initialized outbound connectors
-  private Map<OutboundConnectorConfiguration, OutboundConnectorFunction> functionMap;
+  private Map<String, OutboundConnectorConfiguration> functionMap;
```
chillleader, line 38: nit: naming and comment can be updated

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java
```diff
@@ -49,32 +48,32 @@ public DefaultOutboundConnectorFactory() {
...
-        functionMap.keySet().stream()
-            .filter(config -> config.type().equals(type))
-            .findFirst()
-            .orElseThrow(
-                () -> new NoSuchElementException("Connector " + type + " is not registered"));
-    return functionMap.get(configuration);
+    return Optional.ofNullable(functionMap.get(type))
+        .map(this::createInstance)
+        .orElseThrow(() -> new NoSuchElementException("Connector " + type + " is not registered"));
+  }
```
chillleader, line 59: Why not keep the previous mechanism of instance caching? Or is it to support a case where someone creates a stateful `OutboundConnectorFunction` and wants to use it with prototype scope in Spring?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorAnnotationProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+        .getAnnotation(InboundConnector.class)
+        .map(inboundConnector -> registerInboundConnector(inboundConnector, beanInfo));
+  }
+
+  private InboundConnectorConfiguration registerInboundConnector(
+      InboundConnector inboundConnector, BeanInfo beanInfo) {
+    var scope = configurableBeanFactory.getMergedBeanDefinition(beanInfo.getBeanName()).getScope();
+    if (!SCOPE_PROTOTYPE.equals(scope)) {
+      throw new IllegalStateException(
+          "Unsupported inbound connector scope: \"" + scope + "\" for bean: " + beanInfo.getBean());
```
chillleader, line 69: nit: we tell here that the scope is unsupported but don't give a hint what scope is expected

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/app/TestSpringBasedInboundConnector.java
```diff
@@ -0,0 +1,53 @@
...
+import io.camunda.connector.api.annotation.InboundConnector;
+import io.camunda.connector.api.inbound.InboundConnectorContext;
+import io.camunda.connector.api.inbound.InboundConnectorExecutable;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.context.annotation.Scope;
+import org.springframework.core.env.Environment;
+
+@InboundConnector(name = "TEST_INBOUND_SPRING", type = "io.camunda:test-inbound-spring:1")
+@Scope(BeanDefinition.SCOPE_PROTOTYPE)
+public class TestSpringBasedInboundConnector implements InboundConnectorExecutable {
```
chillleader, line 28: Curious, how come that we need a `@Component` annotation on the `TestSpringBasedOutboundConnector` but not here?

chillleader (approved):
Nice!

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/app/TestSpringBasedInboundConnector.java
sbuettner: Good catch. Actually we dont as we include the connector into the test classes reference for the context component lookup. Removed it.

PR #1001, opened 2023-08-04
feat(webhook): improve HMAC scopes 

Oleksiivanov:
## Description

Refactor HMAC scopes, added ability to generate HMAC from body when HTTP method get

Documentation:  https://github.com/camunda/camunda-platform-docs/pull/2395 - PR, and for view - https://github.com/camunda/camunda-platform-docs/blob/add_hmac_scopes_in_webhook/docs/components/connectors/protocol/http-webhook.md

## Related issues
https://github.com/camunda/team-connectors/issues/476

### DIFF
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/HMACEncodingStrategy.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/HMACEncodingStrategy.java
@@ -1,0 +1,15 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.inbound.signature.strategy;
+
+import io.camunda.connector.api.inbound.webhook.WebhookProcessingPayload;
+import java.io.IOException;
+
+public sealed interface HMACEncodingStrategy
+    permits BodyEncodingStrategy, URLAndBodyEncodingStrategy, URLAndParametersEncodingStrategy {
+  byte[] getBytesToSign(WebhookProcessingPayload payload) throws IOException;
+}
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
@@ -25,6 +25,8 @@ import io.camunda.connector.inbound.model.WebhookConnectorProperties.WebhookConn
 import io.camunda.connector.inbound.model.WebhookProcessingResultImpl;
 import io.camunda.connector.inbound.signature.HMACAlgoCustomerChoice;
 import io.camunda.connector.inbound.signature.HMACSignatureValidator;
+import io.camunda.connector.inbound.signature.strategy.HMACEncodingStrategy;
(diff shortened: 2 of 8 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -111,7 +111,7 @@ private ResponseEntity<?> processWebhook(
       } else if (e instanceof WebhookConnectorException webhookConnectorException) {
         connectorResponse = handleWebhookConnectorException(webhookConnectorException);
       } else {
-        connectorResponse = ResponseEntity.internalServerError().build();
+        connectorResponse = ResponseEntity.internalServerError().body(e.getMessage());
```
chillleader, line 114: I'm not sure if this is always a good idea to return the arbitrary exception in body. I think there is a chance that we may leak sensitive information this way or expose something unwanted, as this is an endpoint open to the internet. WDYT?

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/HMACEncodingStrategy.java
```diff
@@ -0,0 +1,14 @@
...
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.inbound.signature.strategy;
+
+import io.camunda.connector.api.inbound.webhook.WebhookProcessingPayload;
+import java.io.IOException;
+
+public interface HMACEncodingStrategy {
```
chillleader, line 12: Suggestion: we can implement a sealed hierarchy here like with [`WebhookAuthorizationHandler`](https://github.com/camunda/connectors/blob/2589015a805272da8688610b4d829fcb8cb28e3b/connectors/webhook/src/main/java/io/camunda/connector/inbound/authorization/WebhookAuthorizationHandler.java#L22) and its subclasses.

```suggestion
public sealed interface HMACEncodingStrategy {
```

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/StrategyContext.java
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  public boolean containsAllScopes(HMACScope... scopes) {
+    return hmacScopes.length == scopes.length
+        && Arrays.stream(hmacScopes).allMatch(s -> Arrays.asList(scopes).contains(s));
+  }
+
+  public boolean isGetMethod() {
+    return HttpMethods.get.name().equalsIgnoreCase(method);
+  }
```
chillleader, line 29: Suggestion: we could place this functionality into static methods in `HMACEncodingStrategyFactory` and avoid creating objects of this class

chillleader (commented):
Great work! Added a few suggestions.

I think we can also improve the element template - currently there's just a text field where user enters an array of HMAC scopes, but it would be much nicer to have a dropdown like this:
Screenshot 2023-08-04 at 16 31 49

And then we could also add a conditional and only allow URL+PARAMS scope for GET requests. What do you think about this?

chillleader:
> I am not sure that I need to change current template, but if you think that this is really required - I will correct template

Right, makes sense. I'm fine with keeping the current approach 👍

chillleader (approved):

PR #1009, opened 2023-08-08
Operate client fix

chillleader:
## Description

Turns out `OperateClientProdAutoConfiguration` is actually not an auto configuration in Spring Zeebe (despite the name), so the ordering annotations didn't work properly.


markfarkas-camunda (approved):
After fixing the formatting issues, it will be okay from my side 👍

PR #1021, opened 2023-08-10
fix: sanitize scala map output

chillleader:
## Description

Avoid exposing scala maps outside of `FeelEngineWrapper`.

### DIFF
--- a/connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/FeelEngineWrapperExpressionEvaluationTest.java
+++ b/connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/FeelEngineWrapperExpressionEvaluationTest.java
@@ -185,6 +185,21 @@ class FeelEngineWrapperExpressionEvaluationTest {
         .hasMessageContaining("no function found with name 'camel case'");
   }
 
+  @Test
+  void shouldSanitizeScalaMapOutput() {
+    // given
+    final var expression = "={\"processedOutput\": response.callStatus }";
+    final var variables = Map.of("callStatus", "200 OK");
+
+    // when
+    final var result = objectUnderTest.evaluate(expression, variables, Object.class);
+
+    // then
+    // result is not a scala map
+    assertThat(result).isNotInstanceOf(scala.collection.Map.class);
+    assertThat(result).isEqualTo(Map.of("processedOutput", "200 OK"));
+  }
+
   class TestPojo {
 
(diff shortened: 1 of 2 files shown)


igpetrov (approved):

connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/FeelEngineWrapperExpressionEvaluationTest.java
```diff
@@ -185,6 +185,21 @@ void evaluateToJson_ShouldFail_WhenFeelEngineRaisesException() {
         .hasMessageContaining("no function found with name 'camel case'");
   }
 
+  @Test
+  void shouldSanitizeScalaMapOutput() {
+    // given
+    final var expression = "={\"processedOutput\": response.callStatus }";
+    final var variables = Map.of("callStatus", "200 OK");
```
markfarkas-camunda, line 192: Wouldn't it be better to create a scala.collection.Map instead of java.util.Map to test this?

markfarkas-camunda (commented):
Maybe we could write better tests but the solution seems all right!

connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/FeelEngineWrapperExpressionEvaluationTest.java
chillleader: I don't think it makes sense: scala maps appear during the type conversion, which is the last step in the `FeelEngineWrapper::evaluate` call. 

They are produced [here](https://github.com/camunda/connectors/blob/259d8cead5faca0332162ad9977bc268407f6eef/connector-sdk/jackson-datatype-feel/src/main/java/io/camunda/connector/feel/FeelEngineWrapper.java#L121) by `ObjectMapper` when the caller didn't provide a specific type (i.e. provided `Object.class`) - usually ObjectMapper would convert it to Java's `LinkedHashMap`, but apparently when Scala module is registered, it converts it to a scala map.

Besides we never actually provide scala collections as inputs.

markfarkas-camunda: Makes sense, thank you for the detailed explanation.

markfarkas-camunda (approved):

PR #1028, opened 2023-08-10
fix: AWS scala maps serialization

igpetrov:
fix: AWS scala maps serialization

### DIFF
--- a/connectors/aws/aws-base/src/main/java/io/camunda/connector/aws/ObjectMapperSupplier.java
+++ b/connectors/aws/aws-base/src/main/java/io/camunda/connector/aws/ObjectMapperSupplier.java
@@ -6,22 +6,12 @@
  */
 package io.camunda.connector.aws;
 
-import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
-import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
-import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
-import com.fasterxml.jackson.module.scala.DefaultScalaModule$;
+import io.camunda.connector.feel.ConnectorsObjectMapperSupplier;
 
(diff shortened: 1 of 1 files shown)


connectors/aws/aws-sqs/src/main/java/io/camunda/connector/inbound/MessageMapper.java
```diff
@@ -45,7 +45,7 @@ public static SqsInboundMessage toSqsInboundMessage(final Message message) {
 
   private static Object toObjectIfPossible(final String body) {
     try {
-      return ObjectMapperSupplier.getMapperInstance().readValue(body, Object.class);
+      return ConnectorsObjectMapperSupplier.getCopy().readValue(body, Object.class);
```
chillleader, line 48: We can move it to a variable to avoid creating a copy for each static method call

connectors/aws/aws-dynamodb/src/main/java/io/camunda/connector/aws/dynamodb/operation/item/AddItemOperation.java
```diff
@@ -25,7 +25,7 @@ public AddItemOperation(final AddItem addItemModel) {
 
   public PutItemOutcome invoke(final DynamoDB dynamoDB) throws JsonProcessingException {
     String itemStr =
-        ObjectMapperSupplier.getMapperInstance().writeValueAsString(addItemModel.getItem());
+        ConnectorsObjectMapperSupplier.getCopy().writeValueAsString(addItemModel.getItem());
```
chillleader, line 28: Same here, a variable would be nicer, or to use the `DEFAULT_INSTANCE` static instance

connectors/aws/aws-sqs/src/main/java/io/camunda/connector/inbound/MessageMapper.java
igpetrov: Actually, good point

chillleader (approved):
Thanks! 🙌

PR #1035, opened 2023-08-14
chore(http-polling): introduce template

Oleksiivanov:
## Description

Added HTTP polling template
Screenshot 2023-09-14 at 12 03 40

Screenshot 2023-09-18 at 09 03 52

Screenshot 2023-09-14 at 12 02 28

## Related issues

### DIFF
--- a/connectors/http/polling/element-templates/http-polling-connector.json
+++ b/connectors/http/polling/element-templates/http-polling-connector.json
@@ -1,0 +1,466 @@
+{
+  "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
+  "name": "HTTP Polling Connector",
+  "id": "io.camunda.connectors.http.Polling",
+  "version": 1,
+  "description": "Polls endpoint at regular intervals",
+  "icon": {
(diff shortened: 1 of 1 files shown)


connectors/http/polling/element-templates/http-polling-connector.json
```diff
@@ -0,0 +1,491 @@
...
+      "feel": "required",
+      "optional": true,
+      "binding": {
+        "type": "zeebe:property",
+        "name": "activationCondition"
+      },
+      "description": "Condition under which the connector triggers. Leave empty to catch all events. <a href='https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/http-webhook/#make-your-http-webhook-connector-for-receiving-messages-executable' target='_blank'>See documentation</a>"
+    },
+    {
+      "label": "Timer",
```
chillleader, line 407: nit: It wasn't immediately clear to me why we need Timer and HTTP request delay at the same time. I understand it now, but maybe we can call this property "Time to live", or "Expiration time"...

chillleader: Perhaps also a note saying that connector will run indefinitely if you leave this blank

connectors/http/polling/element-templates/http-polling-connector.json
```diff
@@ -0,0 +1,491 @@
...
+        "name": "inbound.timer.message"
+      },
+      "optional": true,
+      "feel": "optional",
+      "constraints": {
+        "notEmpty": false
+      }
+    },
+    {
+      "label": "Connection timeout",
```
chillleader, line 437: I would prefer to move this field into the `endpoint` group (to make it visually close to the HTTP request so users can easily understand the purpose of this field)

chillleader (commented):
Looks good 🙌 
Just a few suggestions for properties layout

igpetrov (commented):
Overall, looks good to me. I think, you should present the CX to @bastiankoerber sooner than later to make sure we're moving in the right direction.

Just a side note, do we need to add "new features" of the HTTP connector introduced by `0.22.0`, like API keys parsing, response body expression, etc?

bastiankoerber (approved):
Looks good for me, just was wondering about the use case of http delete/put/patch method during polling.

but for a super generic polling connector it's fine.

connectors/http/polling/element-templates/http-polling-connector.json
```diff
@@ -0,0 +1,460 @@
...
+      "type": "Text",
+      "feel": "required",
+      "binding": {
+        "type": "zeebe:property",
+        "name": "headers"
+      },
+      "optional": true
+    },
+    {
+      "label": "HTTP request delay",
```
sbuettner, line 181: Lets call it `interval` as its state in the description of the connector.

PR #1043, opened 2023-08-16
Migrate Connectors to Jackson

igpetrov:
Regression testing:
- [x] Slack
- [x] HTTP REST
- - [x] Error expression
- - [x] Different auth types
- - [x] Response body expression 
- [x] GraphQL
- [ ] RabbitMQ
- [ ] GCP Secrets Provider (?)


connectors/http/rest/src/main/java/io/camunda/connector/http/rest/HttpRequestMapper.java
```diff
@@ -94,9 +94,7 @@ public static HttpRequest toHttpRequest(
       content = new UrlEncodedContent(request.getBody());
     } else {
       content =
-          request.hasBody()
-              ? new JsonHttpContent(GsonComponentSupplier.gsonFactoryInstance(), request.getBody())
-              : null;
+          request.hasBody() ? new JsonHttpContent(new GsonFactory(), request.getBody()) : null;
```
sbuettner, line 97: I guess we should remove the `GsonFactory`?

connectors/http/rest/src/main/java/io/camunda/connector/http/rest/components/HttpTransportComponentSupplier.java
```diff
@@ -28,8 +29,7 @@ private HttpTransportComponentSupplier() {}
   private static final HttpTransport HTTP_TRANSPORT = new ApacheHttpTransport();
   private static final HttpRequestFactory REQUEST_FACTORY =
       HTTP_TRANSPORT.createRequestFactory(
-          request ->
-              request.setParser(new JsonObjectParser(GsonComponentSupplier.gsonFactoryInstance())));
+          request -> request.setParser(new JsonObjectParser(new GsonFactory())));
```
sbuettner, line 32: s.a.

sbuettner (approved):
Thanks for taking care. Some minor comments.

connectors/http/rest/src/main/java/io/camunda/connector/http/rest/HttpRequestMapper.java
igpetrov: I don't think we can do it an easy way. The HTTP REST connector uses Google HTTP client, which builds its entity with Gson factory. If we were to decide to get rig of the google http client, it should be handled in a separate task.

WDYT?

sbuettner: @igpetrov Looks like it supports jackson: https://googleapis.github.io/google-http-java-client/json.html

sbuettner: I see, thanks for pointing it out. I think we should replace this client anyway with a more modern stack. Would be great if you could create a task for that.

PR #1052, opened 2023-08-18
feat(http-polling): implement connector, first iteration

Oleksiivanov:
## Description

HTTP-Polling connector 

## Related issues
https://github.com/camunda/connectors/pull/1058


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
```diff
@@ -167,7 +167,19 @@ private void activateConnector(InboundConnectorDefinitionImpl newConnector) {
...
+                    .fetchProcessInstancesByDefinitionKey(processDefinitionKey)
+                    .stream()
+                    .map(
+                        instance ->
+                            new ProcessInstance(
+                                instance.getKey(),
+                                () ->
+                                    processDefinitionSearch.fetchVariablesByProcessInstanceKey(
+                                        instance.getKey())))
+                    .collect(Collectors.toList());
```
sbuettner, line 181: We should `diff` against finished instances and new/running instances first without fetching the variables. Only fetch the variables when you need them (creating the polling request).

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
```diff
@@ -167,7 +167,19 @@ private void activateConnector(InboundConnectorDefinitionImpl newConnector) {
...
+
         executable.activate(
             new InboundIntermediateConnectorContextImpl(inboundContext, processInstancesSupplier));
```
sbuettner, line 184: The `interface` hierarchy should be sealed and the `PollingConnectorExecutable` should receive the actual interface: `InboundIntermediateConnectorContext` without a need to cast it. Please move the polling parts into a subpackage `polling` similar to the `webhook` connector.

sbuettner: The `PollingConnectorExecutable` and the `WebhookConnector` executable should be sealed interfaces of the `InboundConnectorExecutable`.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/PollingConnector.java
```diff
@@ -10,17 +10,50 @@
...
+import io.camunda.connector.http.base.components.GsonComponentSupplier;
+import io.camunda.connector.http.base.components.HttpTransportComponentSupplier;
+import io.camunda.connector.http.base.services.HttpService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-@InboundConnector(name = "HTTP_POLLING", type = "io.camunda:http:polling:1")
+@InboundConnector(name = "HTTP_POLLING", type = "io.camunda:http-polling:1")
 public class PollingConnector implements PollingConnectorExecutable {
+  private static final Logger LOGGER = LoggerFactory.getLogger(PollingConnector.class);
```
sbuettner, line 21: Maybe call the class `HttpPollingConnector` as well.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/operate/ProcessInstance.java
```diff
@@ -27,7 +28,8 @@
...
  *
  * @param key The unique identifier of the process instance.
- * @param variables A map containing the variables associated with this process instance. The keys
- *     represent variable names, and the values are the variable values.
+ * @param variablesSupplier A supplier that provides a map containing the variables associated with
+ *     this process instance. The keys represent variable names, and the values are the variable
+ *     values.
  */
-public record ProcessInstance(Long key, Map<String, String> variables) {}
+public record ProcessInstance(Long key, Supplier<Map<String, String>> variablesSupplier) {}
```
sbuettner, line 35: Arent the variables a `Map<String,Object>`?

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/PollingScheduler.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  private static class Holder {
+    private static PollingScheduler INSTANCE = new PollingScheduler();
+  }
+
+  // todo maybe delete. it's broke singleton, but it's useful for test
+  public static void resetInstance() {
+    Holder.INSTANCE = new PollingScheduler();
+  }
```
sbuettner, line 32: Please remove this.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/model/PollingRequest.java
```diff
@@ -0,0 +1,77 @@
...
+package io.camunda.connector.http.polling.model;
+
+import io.camunda.connector.api.annotation.FEEL;
+import io.camunda.connector.http.base.model.CommonRequest;
+import java.util.Objects;
+
+public class PollingRequest extends CommonRequest {
+  private Timer timer;
+  @FEEL private String pollingInterval;
+  @FEEL private String initialDelay;
```
sbuettner, line 16: What is the idea behind the `initialDelay`?

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/model/PollingRequest.java
```diff
@@ -0,0 +1,77 @@
...
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.http.polling.model;
+
+import io.camunda.connector.api.annotation.FEEL;
+import io.camunda.connector.http.base.model.CommonRequest;
+import java.util.Objects;
+
+public class PollingRequest extends CommonRequest {
+  private Timer timer;
```
sbuettner, line 14: What is a `timer` in this context?

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/model/PollingRequest.java
```diff
@@ -0,0 +1,77 @@
...
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.http.polling.model;
+
+import io.camunda.connector.api.annotation.FEEL;
+import io.camunda.connector.http.base.model.CommonRequest;
+import java.util.Objects;
+
+public class PollingRequest extends CommonRequest {
```
sbuettner, line 13: Why not re-use the existing http request here and separate the polling aspect?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/operate/ProcessInstance.java
sbuettner: Please refactor this part and migrate this to an interface `ProcessContext`.  This process context can then provide the same functionality like an `OutboundConnectorContext` via `<T> T bindVariables(Class<T> cls);`.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundIntermediateConnectorContext.java
```diff
@@ -23,18 +23,20 @@
...
+   * Retrieves a list of process instances associated with this context, including a supplier to
+   * obtain their latest variables. The variables can be dynamically retrieved every time the
+   * supplier is invoked, allowing the connector to have access to the most recent state of the
+   * process instances.
    *
-   * @return a list of {@link ProcessInstance} with updated variables
+   * @return a list of {@link ProcessInstance} objects, each with a supplier for its updated
+   *     variables
    */
   List<ProcessInstance> getProcessInstances();
```
sbuettner, line 41: List<ProcessInstanceContext> getProcessInstanceContexts();

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/model/Timer.java
```diff
@@ -0,0 +1,52 @@
...
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.http.polling.model;
+
+import io.camunda.connector.api.annotation.FEEL;
+import java.util.Objects;
+
+public class Timer {
```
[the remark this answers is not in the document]
sbuettner: Is this a requirement in the epic? If not, lets skip it for now.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/SchedulingService.java
```diff
@@ -0,0 +1,79 @@
...
+        pollingScheduler.scheduleTask(
+            new RequestPollingTask(
+                context,
+                newProcessInstance,
+                httpService,
+                getCancelTaskFunction(),
+                new ConcurrentHashMap<>()),
+            0,
+            15,
+            TimeUnit.SECONDS);
```
sbuettner, line 67: Why do we use the same polling scheduler here? 

I recommend to have two components:
- ProcessInstanceManager (this component is polling all the process definitions and their active process instances for all the inbound polling connectors)
- PollingJobScheduler (this scheduler only manages the actual next polling requests) as this is a separate concern.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundIntermediateConnectorContextImpl.java
```diff
@@ -21,26 +21,28 @@
...
 
+/**
+ * Implementation of {@link InboundIntermediateConnectorContext} that delegates operations to an
+ * underlying {@link InboundConnectorContext} and {@link ProcessInstanceContext}.
+ */
 public class InboundIntermediateConnectorContextImpl
     implements InboundIntermediateConnectorContext {
   private final InboundConnectorContext inboundContext;
-  private final Supplier<List<ProcessInstance>> processInstancesSupplier;
+  private final ProcessInstanceContext processInstanceContext;
```
sbuettner, line 34: The idea was that a `ProcessInstanceContext` represents a processs instance context and not all process instances. A process instance context is the combination of a processs instance and the variables of the process instance + the connector inbound context.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionSearch.java
```diff
@@ -115,42 +113,37 @@ public List<ProcessInstance> fetchProcessInstancesWithVariables(final Long proce
...
     List<Object> variablePaginationIndex = null;
     SearchResult<io.camunda.operate.dto.Variable> searchResult;
     Map<String, String> processVariables = new HashMap<>();
     do {
       try {
         VariableFilter variableFilter =
-            new VariableFilter.Builder().processInstanceKey(processInstanceKey).build();
+            new VariableFilter.Builder()
+                .scopeKey(processInstanceKey)
+                .processInstanceKey(processInstanceKey)
```
sbuettner, line 140: Scope key and process instance key are the same. Is this on purpose?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
```diff
@@ -190,6 +193,33 @@ private void activateConnector(InboundConnectorDefinitionImpl newConnector) {
...
+      InboundConnectorDefinitionImpl newConnector,
+      InboundConnectorContextImpl inboundContext) {
+    Function<Long, Map<String, String>> fetchVariablesFunction =
+        processDefinitionSearch::fetchVariablesByProcessInstanceKey;
+    Supplier<List<ProcessInstance>> processInstancesSupplier =
+        createProcessInstancesSupplier(fetchVariablesFunction, newConnector.processDefinitionKey());
+    ProcessInstanceContext processInstanceContext =
+        new ProcessInstanceContextImpl(processInstancesSupplier, new FeelEngineWrapper());
+    InboundIntermediateConnectorContextImpl intermediateContext =
+        new InboundIntermediateConnectorContextImpl(inboundContext, processInstanceContext);
```
sbuettner, line 207: Please move this into a ContextFactory that creates the `InboundIntermediateConnectorContextImpl` instances.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
```diff
@@ -190,6 +193,33 @@ private void activateConnector(InboundConnectorDefinitionImpl newConnector) {
...
+  private Supplier<List<ProcessInstance>> createProcessInstancesSupplier(
+      Function<Long, Map<String, String>> fetchVariablesFunction, Long processDefinitionKey) {
+    return () ->
+        processDefinitionSearch.fetchProcessInstancesByDefinitionKey(processDefinitionKey).stream()
+            .map(
+                instance ->
+                    new ProcessInstance(
+                        instance.getKey(), fetchVariablesFunction.apply(instance.getKey())))
+            .collect(Collectors.toList());
+  }
```
sbuettner, line 221: This should live in the implementation of the `InboundIntermediateConnectorContextImpl` itself.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/operate/ProcessInstanceContext.java
```diff
@@ -0,0 +1,50 @@
...
+   *
+   * <ul>
+   *   <li>Any FEEL expression among the properties fails to be evaluated.
+   * </ul>
+   *
+   * @param context The inbound connector context whose properties are to be updated and evaluated.
+   * @return true if the properties from active process instances were successfully updated and all
+   *     FEEL expressions evaluated, false otherwise.
+   */
+  boolean updateProperties(InboundConnectorContext context);
```
sbuettner, line 49: Its somewhat the exact way it was discussed. The Ideas was that this returns all the process intances that are waiting for the message wrapped in a context object that is able to bind properties.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/HttpPollingConnector.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public void poll(InboundIntermediateConnectorContext context) {
+    LOGGER.debug("Polling task initiated");
+
+    PollingTask pollingTask = new PollingTask(context, httpService, this::deactivate);
+    long pollingInterval = DelayParser.getPollingIntervalInMillis(context);
+    scheduledExecutorService.scheduleWithFixedDelay(
+        pollingTask, 0, pollingInterval, TimeUnit.MILLISECONDS);
+  }
```
sbuettner, line 60: Why do we need this poll method? Cants this be done in the activation?

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingTask.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Override
+  public void run() {
+    try {
+      if (isReadyForBinding()) {
+        Optional.ofNullable(bindContextToRequest())
+            .map(this::executeHttpRequest)
+            .ifPresent(this::correlateHttpResponse);
+      }
```
sbuettner, line 44: We only should have a poll task for instances that are ready as discussed before.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/operate/ProcessInstance.java
sbuettner: Binding should happen in the context of a process instance to create the http request. I dont see why you would want to bind against the context of the inbound connector here?

We should never update the inbound context itself. It should be seen as immutable.

igpetrov (changes requested):

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultInboundConnectorContextFactory.java
```diff
@@ -0,0 +1,87 @@
...
+        Type[] actualTypeArguments = parameterizedType.getActualTypeArguments();
+        if (actualTypeArguments.length > 0
+            && actualTypeArguments[0].equals(InboundIntermediateConnectorContext.class)) {
+          inboundContext =
+              new InboundIntermediateConnectorContextImpl(
+                  inboundContext, operateClientAdapter, dynamicPropertyManager, objectMapper);
+          break;
+        }
+      }
+    }
```
igpetrov, line 83: Can we get this covered by unit tests? It feels it can be simplified, but without test I cannot experiment.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -16,45 +16,69 @@
...
+    this.properties = new HashMap<>(properties);
   }
 
-  public void setOrUpdateProperties(Long id, Map<String, Object> properties) {
-    propertiesById.putIfAbsent(id, new HashMap<>());
-    propertiesById.get(id).putAll(properties);
+  @Override
+  public void update(final Map<String, String> variables) {
+    properties.putAll(variables);
   }
```
sbuettner: Why do you need this? You receive all properties via the intermediate context as its available right from the start when creating this instance?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -16,45 +16,69 @@
...
           Object evaluatedValue = feelEngineWrapper.evaluate(value, properties);
+          if (evaluatedValue == null) {
+            return false;
+          }
           entrySet.setValue(evaluatedValue);
         } catch (FeelEngineWrapperException e) {
-          LOGGER.debug("Failed to evaluate FEEL expression", e);
           return false;
         }
       }
```
sbuettner: Why is that check needed?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundProcessInstanceContextManager.java
```diff
@@ -0,0 +1,37 @@
...
+   *
+   * @param processDefinitionKey The unique identifier associated with a process definition.
+   * @param elementId The FlowNode key in the BPMN process which identifies a specific element.
+   * @param properties Contextual properties derived from the inbound context. These properties
+   *     might influence how the process instance contexts are retrieved or managed.
+   * @return A list of active {@link ProcessInstanceContext}s that match the provided criteria.
+   */
+  List<ProcessInstanceContext> getActiveProcessInstanceContexts(
+      Long processDefinitionKey, String elementId, Map<String, Object> properties);
+}
```
sbuettner, line 37: I dont see a need for this, its part of the intermediate context no?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessInstanceContext.java
```diff
@@ -18,9 +18,43 @@
...
+   */
+  void update(Map<String, String> variables);
+
+  /**
+   * Determines if the current context is ready to bind properties. This can be used to check
+   * prerequisites or conditions before attempting to bind properties or perform further actions.
+   *
+   * @return {@code true} if the context is ready for binding, otherwise {@code false}.
+   */
+  boolean isReadyToBind();
```
sbuettner: Why is this necessary? A process instance context is only available for process instances that are ready.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessInstanceContext.java
```diff
@@ -18,9 +18,43 @@
...
 
-  Map<String, Object> getVariables();
+  /**
+   * Updates the context with new or modified variables. This method provides a way to make changes
+   * to the state of the process instance.
+   *
+   * @param variables A map containing variable names as keys and their corresponding values.
+   *     Existing variables with the same name will be overwritten.
+   */
+  void update(Map<String, String> variables);
```
sbuettner, line 41: Why do you need to update any variables? The prociess instance context encapsules all the variable handling itself.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessInstanceContext.java
```diff
@@ -18,9 +18,43 @@
...
+ * interact with, modify, and retrieve information related to a particular instance of a process.
+ */
 public interface ProcessInstanceContext {
 
+  /**
+   * Retrieves the unique key associated with the process instance.
+   *
+   * @return A unique identifier representing the process instance.
+   */
   Long getProcessInstanceKey();
```
sbuettner, line 32: Why is this necessary at all?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessInstanceContext.java
```diff
@@ -18,9 +18,43 @@
...
+
+  /**
+   * Binds properties into a specified class type. This method allows for deserialization and
+   * conversion of properties into a given data structure or object.
+   *
+   * @param <T> The type into which properties should be deserialized.
+   * @param cls The class type that provides a blueprint for the deserialization.
+   * @return An instance of type {@code T} containing deserialized and bound properties.
+   */
+  <T> T bindProperties(Class<T> cls);
```
sbuettner, line 59: Should be called just bind as it combines the process variables and the properties.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
sbuettner: I dont see a need for this. You bind agains properties + process variables. We dont pre-validate other bind processes so why do you want to do it here?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundProcessInstanceContextManager.java
sbuettner: All the information ```Long processDefinitionKey, String elementId, Map<String, Object> properties``` are part of the context itself.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/HttpPollingConnector.java
```diff
@@ -0,0 +1,75 @@
...
+import org.slf4j.LoggerFactory;
+
+@InboundConnector(name = "HTTP_POLLING", type = "io.camunda:http-polling:1")
+public class HttpPollingConnector
+    implements InboundConnectorExecutable<InboundIntermediateConnectorContext> {
+  private static final Logger LOGGER = LoggerFactory.getLogger(HttpPollingConnector.class);
+
+  private final HttpService httpService;
+  private ScheduledExecutorService scheduledExecutorService;
+  private final ExecutorService executorService;
```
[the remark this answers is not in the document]
sbuettner: @Oleksiivanov If you have lets say two inbound polling connectors. Both will use the `ScheduledExecutorService` that only has one Thread? I would recommend to just use one pool that is shared between all inbound polling instances.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultInboundConnectorContextFactory.java
sbuettner: Please refactor the implementation. You always create an instance of `InboundConnectorContextImpl` even if you throw it away at a later point. Please extract the lookup into a separate method as only create the right context when needed.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextFactory.java
```diff
@@ -0,0 +1,52 @@
...
+   *
+   * @param definition The specific inbound connector definition which gives details about the
+   *     connector and its related properties.
+   * @param cancellationCallback Callback that gets invoked during connector execution errors or
+   *     cancellations.
+   * @param executableClass Class representation of the executable connector in use.
+   * @return A newly created {@link InboundConnectorContext} instance, tailored to the provided
+   *     parameters.
+   */
+  <T extends InboundConnectorExecutable<?>> InboundConnectorContext createContext(
```
sbuettner, line 48: Shouldnt the returned `InboundConnectorContext` be generic as well as you return different context types here?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/operate/OperateClientAdapterImpl.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class OperateClientAdapterImpl implements OperateClientAdapter {
+
+  private final ProcessDefinitionSearch processDefinitionSearch;
+
+  public OperateClientAdapterImpl(final ProcessDefinitionSearch processDefinitionSearch) {
+    this.processDefinitionSearch = processDefinitionSearch;
+  }
```
sbuettner, line 32: Are you aware that the process definition search is stateful? I would recommend to not use it in this context.

sbuettner: Please just use the operate client here.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/DynamicPropertyManager.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Manages dynamic properties associated with a given property ID. This interface provides methods
+ * to set, update, retrieve, and check the readiness of properties for binding, as well as clearing
+ * the properties.
+ */
+public interface DynamicPropertyManager {
+
+  /**
+   * Sets or updates the properties associated with the specified property ID.
```
sbuettner: I dont see a need for this. Could be removed as the client only deals with `ProcessInstanceContext`.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultInboundConnectorContextFactory.java
```diff
@@ -0,0 +1,91 @@
...
+              new InboundIntermediateConnectorContextImpl(
+                  inboundContext,
+                  operateClientAdapter,
+                  validationProvider,
+                  feelEngineWrapper,
+                  objectMapper);
+          break;
+        }
+      }
+    }
```
sbuettner, line 87: Please extract into separate method and have a conditional return statement: isIntermediateContext()

Dont create the other context if not needed.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -0,0 +1,105 @@
...
+          return false;
+        }
+      }
+    }
+    return true;
+  }
+
+  private boolean isFeelExpression(String value) {
+    return value != null && value.trim().startsWith("=");
+  }
```
sbuettner, line 99: We dont need this check.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/HttpPollingConnector.java
```diff
@@ -0,0 +1,64 @@
...
+  @Override
+  public void activate(final InboundIntermediateConnectorContext context) {
+    long pollingInterval = DelayParser.getPollingIntervalInMillis(context);
+    scheduleTask(context, pollingInterval);
+  }
+
+  private void scheduleTask(InboundIntermediateConnectorContext context, long pollingInterval) {
+    Runnable task = new HttpPollingTask(context, httpService, executorService);
+    executorService.scheduleWithFixedDelay(task, 0, pollingInterval, TimeUnit.MILLISECONDS);
+  }
```
sbuettner, line 57: Can be in the same method.

sbuettner: Dont you have to schedule each individual process instance request?

sbuettner: Here you schedule only the job that polls process instances no?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
sbuettner: @Oleksiivanov Why would `headerValue` be null "for now" if we always request processes that have all variables available? The variables should all be there when we execute the bind call.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
sbuettner: But we can bind values and evaluate properties already for inbound contexts. Can we re-use:
```
  public <T> T bindProperties(Class<T> cls) {
    var mappedObject = objectMapper.convertValue(getPropertiesWithSecrets(), cls);
    getValidationProvider().validate(mappedObject);
    return mappedObject;
  }
```

?

sbuettner: Hi @Oleksiivanov I though we only bind against the inbound context variables and then (if needed) add the. process variables for the evaluation. For example:
```
@FEEL Function<Map<String,Object>, String> url;
```
and if you want to use the `url` you call url.apply(processContext.variables()). Maybe @chillleader or @igpetrov have other ideas but this kind of "pre-evaluation" mixes both concepts.

chillleader: Yes, I like the approach @sbuettner suggested, it would be nice to reuse the delayed evaluation concept. The only concern is that this might cause confusion if customers try to use `contextProperties` (i.e. defined directly in the element template of the polling connector) as variables in FEEL expression (as opposed to only using the process variables we fetched from Operate). On the other hand, I'm not sure if there is really a valid use case for this.

And, apparently, another issue with delayed evaluation: we can't reuse the HTTP request data models directly (the `HttpCommonRequest`) when binding properties, as we would have to wrap all fields into `@FEEL`-annotated `Function`.

chillleader: One more idea: what if we customized the effect of the `FEEL` annotation on the object mapper based on the connector type?
We could have 2 implementations of the `FeelDeserializer`, the first one would just be the current implementation that only uses process definition context to resolve variables, and the second would also include the process instance variables.

The first implementation could stay active by default, and it can be manually overridden in the `ProcessInstanceContext` by creating a local copy of the object mapper and enabling a modified jackson module on it. `ProcessInstanceContext` can pre-configure the custom jackson module by providing the variables from Operate.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/operate/OperateClientConfiguration.java
```diff
@@ -14,12 +14,18 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class OperateClientConfiguration {
+  @Bean
+  public OperateClientAdapter springOperateClientAdapter(
+      CamundaOperateClient camundaOperateClient) {
+    return new OperateClientAdapterImpl(camundaOperateClient);
+  }
+}
```
chillleader, line 31: I might be wrong, but it seems that this bean method can fail if inbound connectors are disabled by setting `camunda.connector.polling.enabled=false`, and Operate client bean is not created.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
sbuettner: @chillleader We could then maybe start with a simpler approach:
* Remove the `bind` method on the `ProcessInstanceContext` for now
* Replace it with a `getVariables()` method
* Just use the existing infrastructure and bind against the inbound context plus add the variables when access needed?

WDYT?

sbuettner: > Not necessarily, I think in theory we could use [deserialization context attributes](https://stackoverflow.com/a/46893967) to pass the process instance ID or even already resolved variables, thus allowing to have a common mapper instance for all ProcessInstanceContext objects.

That could work although it seems to create a new object reader instance for every process instance.

> So in the last suggestion, do I get it correctly that we would then add an overload of the bindProperties method to enrich the context with instance variables? Or how would you add them?
I would just add a `getVariables()` method to the `ProcessInstanceContext` and leave the existing `bindProperties` method as it is for now.

If we go this route we could discuss your idea in our architecture session tomorrow and plan it as an improvement. WDYT?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/operate/OperateClientConfiguration.java
chillleader: Sorry, false alarm - as discussed with @Oleksiivanov this is actually safe, because `OperateClientConfiguration` is imported in the `InboundConnectorRuntimeConfiguration`, which has the `ConditionalOnProperty` annotation.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/HttpPollingConnector.java
```diff
@@ -0,0 +1,72 @@
...
+    executorService.scheduleWithFixedDelay(
+        httpRequestTask, 0, intervals.getHttpRequestInterval().toMillis(), TimeUnit.MILLISECONDS);
+
+    PollingOperateTask pollingOperateTask = new PollingOperateTask(context, httpRequestTask);
+    executorService.scheduleWithFixedDelay(
+        pollingOperateTask,
+        0,
+        intervals.getOperatePollingInterval().toMillis(),
+        TimeUnit.MILLISECONDS);
+  }
```
sbuettner, line 65: Why do you schedule two tasks here?

Shouldnt we only schedule the polling operate task and if the process instance is not already scheduled for polling create the HttpPolling task and schedule that one?

sbuettner: Sure, its understood but this will for example run each individual http request sequentially. Please implement the following:
1. Request process instances from Operate
2. If they are not scheduled yet, request variables and schedule them
3. Each individuall http request will get executed with the delay configures as the request interval (not the operate polling interval)
4. Remove process instances from scheduling that were not fateched from operate (they finished already or were cancelled because of a boundary event)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultInboundConnectorContextFactory.java
```diff
@@ -0,0 +1,102 @@
...
+    for (Type genericInterface : genericInterfaces) {
+      // Check if the current type is a parameterized type (i.e., has generic parameters).
+      if (genericInterface instanceof ParameterizedType parameterizedType) {
+        // Retrieve actual type arguments of the parameterized type.
+        Type[] actualTypeArguments = parameterizedType.getActualTypeArguments();
+
+        // If the first generic type argument matches InboundIntermediateConnectorContext,
+        // then this class represents an intermediate context.
+        if (actualTypeArguments.length > 0
+            && actualTypeArguments[0].equals(InboundIntermediateConnectorContext.class)) {
```
igpetrov, line 95: To be on a safe side, suggesting the following:
```suggestion
            && InboundIntermediateConnectorContext.class.equals(actualTypeArguments[0])) {
```

connectors/http/rest/src/test/java/io/camunda/connector/http/rest/HttpJsonFunctionTest.java
```diff
@@ -189,6 +192,14 @@ void execute_shouldReturnNullFieldWhenResponseWithContainNullField()
     assertThat(asJsonObject.get("unknown").isNull()).isTrue();
   }
 
+  @Test
+  public void test() {
+    Map<String, String> map = new HashMap<>();
+    map.put("test2", "t3");
+    String s = Optional.ofNullable(map.get("test")).orElseGet(() -> map.get("test2"));
+    System.out.println(s);
```
igpetrov, line 200: Don't forget to remove sysout

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultInboundConnectorContextFactory.java
```diff
@@ -0,0 +1,102 @@
...
+
+    if (isIntermediateContext(executableClass)) {
+      inboundContext =
+          new InboundIntermediateConnectorContextImpl(
+              inboundContext,
+              operateClientAdapter,
+              validationProvider,
+              feelEngineWrapper,
+              objectMapper);
+    }
```
sbuettner, line 77:
```
if (isIntermediateContext(executableClass)) {
...
else {
inboundContext =
        new InboundConnectorContextImpl(
            secretProviderAggregator,
            validationProvider,
            newConnector,
            correlationHandler,
            cancellationCallback,
            objectMapper);
}
```

Lets try to avoid object creation if not needed.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/HttpPollingConnector.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public HttpPollingConnector(final String proxyFunctionUrl) {
+    this(
+        new HttpService(
+            ConnectorsObjectMapperSupplier.getCopy(),
+            HttpTransportComponentSupplier.httpRequestFactoryInstance(),
+            proxyFunctionUrl),
+        HttpPollingSharedExecutorService.getInstance());
+  }
```
sbuettner, line 41: `proxyFunctionUrl` will always be null this way.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingOperateTask.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * Schedules or refreshes HttpRequestTasks based on the given processInstanceContexts. Each
+   * ProcessInstanceContext and flow node together uniquely define a single HttpRequestTask.
+   */
+  private void scheduleOrRefreshTasks(List<ProcessInstanceContext> contexts) {
+    contexts.forEach(
+        processInstanceContext -> {
+          String taskKey = context.getDefinition().elementId() + processInstanceContext.getKey();
```
sbuettner, line 76: Please extract this and:
```
.map(ProcessInstanceContext::getKey)
            .map(key -> context.getDefinition().elementId() + key)
            
```
into a simple and separate function `String getRequestTaskKey(processInstaceContext);`

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingOperateTask.java
```diff
@@ -0,0 +1,99 @@
...
+          // Check if an HttpRequestTask already exists for this unique task key.
+          if (executorService.containsTask(taskKey)) {
```
sbuettner, line 78: Why are there two places? Lets just have one place where tasks are tracked. Can just be this class.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -0,0 +1,116 @@
...
+    this.operatePropertiesSupplier = operateVariables;
+  }
+
+  @Override
+  public Long getKey() {
+    return processInstanceKey;
+  }
+
+  @Override
+  public <T> T bind(final Class<T> cls) {
```
sbuettner, line 65: As discussed, please replace the `bind` method with `getVariables()` and add a simple check whether the variables have been fetched already.

sbuettner: Not really as its about creating a `bind` method that works using jackson. We wanted to use `getVariables()` now to avoid this custom FEEL handling code we have here.

sbuettner: @Oleksiivanov You dont as you dont need validation for process instance variable binding.

sbuettner: You can create a record class that has a Function<Map<String,Object>> and bind it to the url. Then call it with the `getVariables()` result to use it. You dont have to use the Http Request class 1:1. Just pass/map the values.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultInboundConnectorContextFactory.java
sbuettner: Ah, sorry. Didnt see that. Works for me.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -0,0 +1,116 @@
...
+            processInstanceKey);
+        return null;
+      }
+    } catch (Exception e) {
+      LOGGER.warn(
+          "Error occurred while binding properties for processInstanceKey {}: {}",
+          processInstanceKey,
+          e.getMessage());
+      return null;
+    }
```
sbuettner, line 87: You can remove the try and the `null` return value as this should just lead to an error.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -0,0 +1,116 @@
...
+            return false;
+          }
+          entry.setValue(evaluatedValue);
+        } catch (FeelEngineWrapperException e) {
+          return false;
+        }
+      }
+    }
+    return true;
+  }
```
sbuettner, line 106: No need to return a boolean. Just throw a runtime exception here.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingOperateTask.java
```diff
@@ -0,0 +1,92 @@
...
+    contexts.forEach(
+        processInstanceContext -> {
+          String taskKey = getRequestTaskKey(processInstanceContext);
+
+          if (!executorService.containsTask(taskKey)) {
+            HttpRequestTask newTask =
+                new HttpRequestTask(taskKey, httpService, this.context::correlate);
+            newTask.setProcessInstanceContext(processInstanceContext);
+
+            runningHttpRequestTaskIds.add(taskKey);
```
sbuettner, line 81: This is not needed as discussed yesterday :)

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/service/HttpPollingSharedExecutorService.java
```diff
@@ -0,0 +1,77 @@
...
+  public static HttpPollingSharedExecutorService getInstance() {
+    return INSTANCE;
+  }
+
+  public void scheduleWithFixedDelay(
+      AbstractScheduledTask task, long initialDelay, long delay, TimeUnit unit) {
+    ScheduledFuture<?> future =
+        executorService.scheduleWithFixedDelay(task, initialDelay, delay, unit);
+    tasks.put(task.getId(), future);
+  }
```
sbuettner, line 46: Please use `computeIfAbsent` and schedule only if not present. Please rewrite the code to make it thread-safe:
```
scheduleWithFixedDelay(taskId,  long initialDelay, long delay, TimeUnit unit, Function<String, AbstractScheduledTask> taskCreator) {
tasks.computeIfAbsent("taskId", () -> taskCreator) 
}
```

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingOperateTask.java
sbuettner: This should be tracked by the executor service. Just add a method: `executorService.getTaskIds()`;

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingOperateTask.java
sbuettner: Pleae look at my other comment below. We agreed on that its provided by the executor service.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
sbuettner: As discussed yesterday we keep the bind method and refactor later.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/PollingOperateTask.java
sbuettner: Just schedule the operate polling task with the same executorservice but without this task abstraction.

sbuettner: Just use: `this.executorService.schedule...`

sbuettner (approved):
@Oleksiivanov You made it 🎉Great work. 

Great learnings for all of us regarding the whole interaction with operate and zeeb. 

Lets get this into QA and start testing.

PR #1058, opened 2023-08-21
feat(http): move common classes to base

Oleksiivanov:
## Description

moved common classes to HTTP base, Preparing to HTTP-Polling connector, this PR need for doing PR with Polling connector more readable and understandble (https://github.com/camunda/connectors/pull/1052)


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/HttpInteractionService.java
```diff
@@ -42,13 +42,13 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class HTTPService {
+public class HttpInteractionService {
```
sbuettner, line 45: Why not `HttpService`?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/CommonRequest.java
```diff
@@ -25,22 +26,23 @@
 
 public class CommonRequest {
```
sbuettner, line 27: Can we maybe call it `HttpRequest`?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/CommonRequest.java
```diff
@@ -25,22 +26,23 @@
...
+  @Pattern(regexp = "^(=|http://|https://|secrets|\\{\\{).*$")
   private String url;
 
-  @NotBlank private String method;
+  @FEEL @NotBlank private String method;
 
   @Valid private Authentication authentication;
 
   @Pattern(regexp = "^([0-9]+|secrets\\..+)$")
   private String connectionTimeoutInSeconds;
```
sbuettner, line 39: We can map it to an Integer/Long now as its supported. We can also remove the pattern as its outdated.

sbuettner: @Oleksiivanov We should support it as we use jackson to bind the values which allows to map from `"10"` to `10`.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/HttpInteractionService.java
sbuettner: Ok, as a reader the whole code is a little bit tough to follow and not easy to understand why there is an `HttpService` and an `HttpInteractionService` to begin with. Its ok to move forward now but there is a need for refactoring as we create new instances of these services for each call like this:
```
 private HttpCommonResult executeRequestDirectly(HttpCommonRequest request)
      throws IOException, InstantiationException, IllegalAccessException {
    ...
    HttpInteractionService httpInteractionService = new HttpInteractionService(gson);
    AuthenticationService authService = new AuthenticationService(gson, requestFactory);
...
}
```

which indicates these are rather utilities and not services.

sbuettner (approved):
We need some more refactoring in the feature as mentioned before but we can move on for now.

Issue #842, opened 2023-07-06
Add mTLS Support


upgradingdave:
__Is your feature request related to a problem? Please describe.__

Customers may want to configure the HTTP REST Connector to communicate over mTLS

__Describe the solution you'd like__

I'm not an expert of mLS, but here are some ideas (Feel free to update!)

The server side certificate validation is already provided via the normal tls infrastructure used all over the web. But, we'd need to provide a way to validate client certificates. 

In SaaS, a customer could configure a Secret representing the client's public certificate. Then, the http rest connector backend could use that to validate client certs.

In SM, the client public cert could be added to JKS. Then, the http rest connector backend could use that to validate client certs.

__Describe alternatives you've considered__

It's currently possible for customers to write their own http rest connector to support mTLS

__Additional context__

- Slack Discussion: https://camunda.slack.com/archives/C02JLRNQQ05/p1688656536293809
- mTLS is mentioned in issue #108
- Nice overview of mTLS: https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/

sbuettner:
We will cover this topic in the future again but its currently not easily doable with the current proxy model.

sbuettner:
Additional evidence: https://camunda.slack.com/archives/C05FVD0AZ2R/p1761833077419299

crobbins215:
See @falko [note](https://camunda.slack.com/archives/C05FVD0AZ2R/p1765554440689649?thread_ts=1761833077.419299&cid=C05FVD0AZ2R) Customer ask for SM only.

Explored implementation [#5742 Enable mTLS support for HTTP REST Connector via system SSL context](https://github.com/camunda/connectors/pull/5742)

sbuettner:
@johnBgood I was looking into this and was wondering: To me it seems like that we will have to create a custom http client instance as the `SSLSocketFactory`  is not a "per-request" configuration. If so, we would need to cache the actual http client object to not create a new one for each request. 

As an alternative we could leverage this library which supports managing multiple identities on a URL basis: https://github.com/Hakky54/ayza#routing-identity-material-to-specific-host

WDYT?

johnBgood:
@sbuettner 
Currently, we create a client per-request anyway, so we could implement the mTLS support without much effort.

As follow-up task, if needed, we could cache our clients; this should be trivial. This is how we do it currently:
```java
    try (var client =
        new ProxyAwareHttpClient(
            new ProxyAwareHttpClient.TimeoutConfiguration(
[...]

PR #1067, opened 2023-08-23
feat: add mTLS support

markfarkas-camunda:
## Description

STILL IN PROGRESSS! (early feedbacks are welcomed ofc)

Add mTLS support for REST connector. This so far uses proper certificate files (for localhost), but the idea is to use base64 encoded certs as an element template input.
Integration test already added.

Currently this solution fails with the message: ``` java.net.SocketException: An established connection was aborted by the software in your host machine ```

## Related issues

https://github.com/camunda/connectors/issues/842 

closes #


sbuettner:
@markfarkas-camunda Can we close this? We can come back if its requested.

Issue #518, opened 2023-05-02
Connector remapping configuration


chillleader:
## Description

We would like to introduce the new configuration properties support in the connector runtime to target the following use-cases.

### 1. Disable connectors

I have a process application that bundles some Camunda connectors. When the application starts, all connectors start as well. In specific cases, I want to disable one or all connectors in the configuration. Currently, it is not possible to disable a connector. 

For other job workers, I can disable them via config:

```
zeebe.client.worker.override.foo.enabled=false
```

Or, by defining a bean that implements `ZeebeWorkerValueCustomizer`.

But both methods don't work for connectors.

### 2. Connector type remapping

It would be useful to allow overriding/remapping connector types. This issue occurs e.g. when running a self-hosted connector runtime that connects to the Camunda 8 SaaS cluster. As the two connector runtimes are unaware of each other's presence, it is impossible to configure which runtime should receive a job for a certain connector. If there are Kafka connectors running in both runtimes, both will occasionally receive jobs from Zeebe, but sometimes we want to use the self-managed runtime only e.g. to access the Kafka cluster on-premises.

It would be nice to allow to redefine the types registered for each connector, e.g. change the default one from `io.camunda:connector-kafka:1` to `com.mycompany:connector-kafka:1`, or even allow both.

## Proposal

Introduce a connector configuration. It should cover both remapping and disabling and may look like this:
```
camunda.connector:
  outbound:
    KAFKA:
      enanled: false
    RABBITMQ:
      enabled: true
      types:
        - io.camunda:connector-rabbitmq:1
        - com.mycompany:connector-rabbitmq:1
```

bastiankoerber:
let's discuss this in the architecture sync

chillleader:
Adjusted the issue description based on our decision.

bastiankoerber:
linked to https://github.com/camunda/product-hub/issues/1363

sbuettner:
We agreed on the following environment format for remapping the connector type information: 
```CONNECTOR_AWSEventBridge_TYPE=my-aws-event-bridge```

We should do this in the static methods:
* ConnectorUtil.getRequiredOutboundConnectorConfiguration(cls)
* ConnectorUtil.getRequiredInboundConnectorConfiguration(cls)

Note from @igpetrov: We should revisit our current connector names and lookup names like `AWS DynamoDB` as `CONNECTOR_AWS_DYNAMODB_TYPE`.

PR #1068, opened 2023-08-23
feat: adds connector type override mechanism

igpetrov:
## Description

feat: adds connector type override mechanism

- Despite PR being huge, the real change is at `connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java` and related test.
- Connector runtime name changes are cosmetic.
- Element template only bump version, and make task definition type visible.

## Related issues

PH: https://github.com/camunda/product-hub/issues/1363


connectors/asana/element-templates/asana-connector.json
```diff
@@ -42,10 +46,15 @@
   ],
   "properties": [
     {
-      "type": "Hidden",
+      "type": "String",
+      "group": "taskDefinitionType",
+      "description": "Task definition type should correspond to connector's runtime. Details in the <a href=\"https://docs.camunda.io/docs/components/connectors/use-connectors/\" target=\"_blank\">documentation</a>",
```
chillleader, line 51: I'm a bit concerned that we want to expose this property for every connector template. It doesn't make sense in SaaS to change the template type and may confuse SaaS users. Self-managed users, on the other hand, should be capable of editing the hidden field, given they have applied a custom properties configuration to the connector runtime.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
chillleader, line 1: Perhaps it also makes sense to support disabling connectors as part of this task - as described in the issue [here](https://github.com/camunda/connectors/issues/518). WDYT?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
```diff
@@ -33,11 +34,27 @@ private ConnectorUtil() {}
...
   public static Optional<OutboundConnectorConfiguration> getOutboundConnectorConfiguration(
       Class<? extends OutboundConnectorFunction> cls) {
-    return Optional.ofNullable(cls.getAnnotation(OutboundConnector.class))
-        .map(
-            annotation ->
+    Map<String, String> env = System.getenv();
+    var annotation = Optional.ofNullable(cls.getAnnotation(OutboundConnector.class));
+    if (annotation.isPresent()) {
+      final var normalizedConnectorName =
+          "CONNECTOR_" + toNormalizedConnectorName(annotation.get().name()) + "_TYPE";
```
sbuettner, line 41: Please extract into a static method and use it for inbound and outbound

connectors/asana/element-templates/asana-connector.json
igpetrov: This is what, in my understanding, we agreed last time during discussion with @bastiankoerber. But I am happy to discuss this during architecture session today.

---

> It doesn't make sense in SaaS to change the template type and may confuse SaaS users.

I can say that we have misunderstanding in this statement. I did it specifically for SaaS. Say, if you have a standard cluster: Zeebe + Connectors. You also would like to have an instance of Connectors with some super-critical configuration inside your private network.

If you would like to run your BPMN process through SaaS Connector instance - you do nothing - Keep `io.camunda.myconnector:1`. If you would like to run a BPMN process through your local, then you have to change the `taskDefinitionType` as `io.camunda.myconnector:2` in you diagram.

That's what I showed on demo on Friday.

But again, happy to discuss, maybe I missed some important use-case.

---

Regarding the confusing part. It is a bit from the CX part but we requested a feature to make that group collapsable by default. See this one: https://github.com/camunda/camunda-modeler/issues/3807.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
```diff
@@ -33,11 +34,27 @@ private ConnectorUtil() {}
...
+      if (env.containsKey(normalizedConnectorName)) {
+        return annotation.map(
+            ann ->
                 new OutboundConnectorConfiguration(
-                    annotation.name(), annotation.inputVariables(), annotation.type(), cls));
+                    toNormalizedConnectorName(ann.name()),
```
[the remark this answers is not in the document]
sbuettner: Why not just remap the id instead of creation the whole new configuration like:
var type = Optional,ofNullable(env.containsKey(normalizedConnectorName)).orElse(annotation.type())

And this could be done in a separate method.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
igpetrov: Last time we decided not to implement disabling functionality but we can reconsider today during architecture session.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
igpetrov: I didn't want to have 2 constants, that would be ugly. On the other hand, since this code is executured during startup, I can use `String.format`.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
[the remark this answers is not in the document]
igpetrov: Just trying to make it look nicer.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
sbuettner: Not sure what you mean. I meant that his code exists in both methods for inbound and outbound and can be reused.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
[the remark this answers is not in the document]
igpetrov: > Why not just remap the id instead of creation the whole new configuration like:

I personally see this as a matter of style. If you like that one more, I can do it.

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/ConnectorUtilTest.java
```diff
@@ -60,6 +68,164 @@ public void shouldHandleMissingConnectorConfiguration() {
...
+        @Override
+        public Object execute(OutboundConnectorContext context) throws Exception {
+          return null;
+        }
+      }
+
+      // when
+      EnvironmentVariables environmentVariables =
+          new EnvironmentVariables(
+              "CONNECTOR_NONNORMALIZED_NAME_TYPE", "io.camunda:connector:XXXXXXX");
```
sbuettner, line 113: Nice

connectors/asana/element-templates/asana-connector.json
chillleader: Got it - thanks for the clarification, that makes sense. Let's discuss further in the architecture session.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
sbuettner: The code is fine its just duplicated.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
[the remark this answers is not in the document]
igpetrov: btw, annotation here is `Optional` too, so might look a bit ugly

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/ConnectorUtilTest.java
igpetrov: thanks!

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
igpetrov: As per architecture review - will have a separate task

connectors/asana/element-templates/asana-connector.json
igpetrov: As per architecture session, will take it offline with @bastiankoerber. For now I removed all element template changes and if needed, I'll make a separate PR.

PR #1077, opened 2023-08-27
fix: collect all process definitions before extracting inbound connectors

markfarkas-camunda:
## Description

Issue:
We query operate in the chunks of 20, and only process these 20 process definitions. Processing means activating any inbound connector in it. If you have more then 20 versions of a certain process definition, the application might activate an inbound connector which is not part of the latest version but a previous one. Even if we will override this (deactivate the old one and activate the new one) later, there is still a brief moment when we have the old version as an activated inbound connector, which may lead to unexpected behavior.

## Solution

We first collect all the process definitions (not just 20), and only start extracting inbound connectors afterwards. This will result in only processing the latest process definitions as expected.

## Related issues

There is no related issue.

closes #


chillleader (approved):
Nice one!

PR #1094, opened 2023-08-31
feat: generated REST connector template

chillleader:
## Description

- Add necessary annotations to enable automatic generation of the REST Connector template.

- Support Hybrid templates generation with the element template generator

## Related issues

related to https://github.com/camunda/team-connectors/issues/481


chillleader:
After testing I can conclude that the only important difference is that now method names are upper-case, and in the manual element template they were lower-case. We could override it using `@TemplateProperty` but it doesn't look very nice, so I decided to just bump the element template version - this way we should be able to exclude any issues.

sbuettner (approved):
Great and nice addition with the hybrid support. Lets discuss later whether we want so always allow this to be set instead of managing two templates.

igpetrov (approved):

PR #1105, opened 2023-09-04
feat: Add multi-tenancy support for inbound connectors

sbuettner:
## Related issues

closes [#440](https://github.com/camunda/team-connectors/issues/440)


parent/pom.xml
```diff
@@ -65,10 +65,10 @@ limitations under the License.</license.inlineheader>
     <nexus.sonatype.url>https://s01.oss.sonatype.org</nexus.sonatype.url>
 
     <!-- Camunda internal libraries -->
-    <version.zeebe>8.2.12</version.zeebe>
-    <version.spring-zeebe>8.2.4</version.spring-zeebe>
+    <version.zeebe>8.3.0-alpha4</version.zeebe>
+    <version.spring-zeebe>8.3.0-alpha5</version.spring-zeebe>
     <version.feel-engine>1.16.1</version.feel-engine>
-    <version.operate-client>8.2.0</version.operate-client>
+    <version.operate-client>8.3.0-alpha5</version.operate-client>
```
chillleader, line 71: Are `zeebe` and `operate-client` version bumps still necessary after `spring-zeebe` version bump? It should include both of them right?

chillleader (approved):

igpetrov (approved):

parent/pom.xml
sbuettner: True but it might be confusing to let them point to older versions. Maybe we should just remove them completely as they are transitive dependencies of `spring-zeebe` anyway?

chillleader: Yes, it would be ok to remove them from my perspective.

PR #1110, opened 2023-09-06
feat(generator): support validation annotations

chillleader:
## Description

Support generating validation constraints based on Jakarta Bean Validation API.

Also some new documentation has been added, clarifying how the generator works and what rules are applied by default.

## Related issues

closes https://github.com/camunda/team-connectors/issues/490


sbuettner (approved):
Great, thanks for providing documentation as well.

PR #1118, opened 2023-09-07
feat(generator): support icons

chillleader:
## Description

Support icons in the element template generator.

The icon can be defined in the `@ElementTemplate` annotation. It must reference an image placed in the resources directory. Maven plugin manipulates the resources of each individual target so they are always available in the class loader.

## Related issues

closes https://github.com/camunda/team-connectors/issues/491


igpetrov (approved):
LGTM

sbuettner (approved):
Nice!

PR #1128, opened 2023-09-11
fix(webhook): mark inbound.activationCondition as deprecated

Oleksiivanov:
marked i`nbound.activationCondition` as deprecated

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDefinitionImpl.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDefinitionImpl.java
@@ -51,7 +51,8 @@ public record InboundConnectorDefinitionImpl(
   }
 
   public String activationCondition() {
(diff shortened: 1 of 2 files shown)


igpetrov (approved):
LGTM. Do we need to reflect the same changes as from https://github.com/camunda/web-modeler/pull/6070/files in Connectors too?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDefinitionImpl.java
```diff
@@ -51,7 +51,8 @@ public String resultVariable() {
   }
 
   public String activationCondition() {
-    return rawProperties.get(Keywords.ACTIVATION_CONDITION_KEYWORD);
+    return Optional.ofNullable(rawProperties.get(Keywords.ACTIVATION_CONDITION_KEYWORD))
+        .orElseGet(() -> rawProperties.get(Keywords.DEPRECATED_ACTIVATION_CONDITION_KEYWORD));
```
sbuettner, line 55: Can this throw if its `null`?

sbuettner (approved):

PR #1136, opened 2023-09-12
feat(connectors): added subprocess support

igpetrov:
feat(connectors): added subprocess support

### Webhook

Screenshot 2023-09-13 at 15 48 45

### Kafka

Screenshot 2023-09-13 at 15 48 30

### Two webhooks (start and intermediate)

Screenshot 2023-09-13 at 15 59 03

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
@@ -16,11 +16,13 @@
  */
 package io.camunda.connector.runtime.inbound.importer;
 
+import static io.camunda.connector.runtime.core.Keywords.CORRELATION_KEY_EXPRESSION_KEYWORD;
+import static io.camunda.connector.runtime.core.Keywords.INBOUND_TYPE_KEYWORD;
+
 import io.camunda.connector.api.inbound.InboundConnectorDefinition;
 import io.camunda.connector.api.inbound.correlation.MessageCorrelationPoint;
 import io.camunda.connector.api.inbound.correlation.ProcessCorrelationPoint;
 import io.camunda.connector.api.inbound.correlation.StartEventCorrelationPoint;
-import io.camunda.connector.runtime.core.Keywords;
 import io.camunda.connector.runtime.core.inbound.InboundConnectorDefinitionImpl;
 import io.camunda.operate.CamundaOperateClient;
 import io.camunda.operate.dto.ProcessDefinition;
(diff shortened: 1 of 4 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
```diff
@@ -98,20 +105,21 @@ public List<InboundConnectorDefinitionImpl> findInboundConnectors(
...
+        if (clz.isInstance(element)) {
+          // this FlowElement type currently does not support
+          // inbound connectors implementations
+          continue;
+        }
+        final var zeebeProperties = getRawProperties(element);
+        if (zeebeProperties != null && zeebeProperties.get(INBOUND_TYPE_KEYWORD) != null) {
+          inboundEligibleElements.add(element);
+        }
+      }
```
sbuettner, line 121: Had to think so some seconds about what is happening here, could we simplify it a little bit?

`INBOUND_ELIGIBLE_TYPES` can be a Set<Class>

```
if(INBOUND_ELIGIBLE_TYPES.contains(element.getClass()) && ...) {
  ...
}
```

Lets also move this method into `retrieveEligibleElementsFromProcess` and filter there. I would assume that a method called `retrieveEligibleElementsFromProcess` only returns eligibile lements.

sbuettner (commented):
Looks good, just a smaller nit regarding readability.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspector.java
igpetrov: Sure, can rework. Please note, that `INBOUND_ELIGIBLE_TYPES.contains(element.getClass())` won't work because, for example, for start events, the `INBOUND_ELIGIBLE_TYPES` contain `StartEvent.class` interface, while the actual result of `element.getClass()` will be `class io.camunda.zeebe.model.bpmn.impl.instance.StartEventImpl` thus will be always `false`.

igpetrov: Done

sbuettner: @igpetrov Thank you! Maybe we should rely on `ModelElementType.getInstanceType()` or `ModelElementType.getTypeName()` for a more robust and simpler approach in the future.

sbuettner (approved):
Great improvement!

PR #1138, opened 2023-09-13
feat(bundle): add user-managed folder for mounting

igpetrov:
feat(bundle): add user-managed folder for mounting

Related to https://github.com/camunda/team-connectors/issues/496.

#### Connector volume defined

Pay attention at the argument `-v $PWD/connector-template-0.1.0-SNAPSHOT-with-dependencies.jar:/opt/additionalLibs/connector.jar`

```
 ~/W/t/custom-connector $ tree
.
└── connector-template-0.1.0-SNAPSHOT-with-dependencies.jar

1 directory, 1 file
 ~/W/t/custom-connector $ docker run --rm --name=HybridConnector -v $PWD/connector-template-0.1.0-SNAPSHOT-with-dependencies.jar:/opt/additionalLibs/connector.jar -e ZEEBE_CLIENT_SECURITY_PLAINTEXT=false -e ZEEBE_CLIENT_CLOUD_CLUSTER-ID='[REDACTED]' -e ZEEBE_CLIENT_CLOUD_CLIENT-ID='[REDACTED]' -e ZEEBE_CLIENT_CLOUD_CLIENT-SECRET='[REDACTED]' -e ZEEBE_CLIENT_CLOUD_REGION='bru-2' -e CAMUNDA_OPERATE_CLIENT_URL='[REDACTED]' -e CONNECTOR_HTTP_REST_TYPE='io.camunda:http-json:2' camunda/connectors-bundle:999.999.1
```

Provides the following output:

```
...
2023-09-13T13:32:53.309Z  INFO 1 --- [           main] i.c.z.s.c.jobhandling.JobWorkerManager   : . Starting Zeebe worker: ZeebeWorkerValue{type='io.camunda:template:1', name='MYCONNECTOR', timeout=null, maxJobsActive=null, requestTimeout=null, pollInterval=null, autoComplete=true, fetchVariables=[authentication, message], enabled=null, methodInfo=null}
...
```

#### Connector volume NOT defined (regression testing)

Connector starts normally without `-v` flag.

```
docker run --rm --name=HybridConnector -e ZEEBE_CLIENT_SECURITY_PLAINTEXT=false -e ZEEBE_CLIENT_CLOUD_CLUSTER-ID='[REDACTED]' -e ZEEBE_CLIENT_CLOUD_CLIENT-ID='[REDACTED]' -e ZEEBE_CLIENT_CLOUD_CLIENT-SECRET='[REDACTED]' -e ZEEBE_CLIENT_CLOUD_REGION='bru-2' -e CAMUNDA_OPERATE_CLIENT_URL='[REDACTED]' -e CONNECTOR_HTTP_REST_TYPE='io.camunda:http-json:2' camunda/connectors-bundle:999.999.1
```

### DIFF
--- a/bundle/camunda-saas-bundle/Dockerfile
+++ b/bundle/camunda-saas-bundle/Dockerfile
@@ -1,6 +1,7 @@
 FROM eclipse-temurin:17.0.8_7-jre
 
 RUN mkdir /opt/app
+
 # Download connectors from maven central
 COPY target/*-with-dependencies.jar /opt/app/
 
--- a/bundle/default-bundle/Dockerfile-focal
+++ b/bundle/default-bundle/Dockerfile-focal
@@ -1,6 +1,9 @@
 FROM eclipse-temurin:17.0.8_7-jre-focal
 
-RUN mkdir /opt/app
+# The /opt/app is used for the Connectors runtime, and out-of-the-box connectors
+# Use the /opt/additionalLibs to mount your own connectors, secret providers, or include other jars into the classpath
+RUN mkdir /opt/app && mkdir /opt/custom
+
 COPY target/*-with-dependencies.jar /opt/app/
 
 # Using the start script from the base connector runtime image
--- a/bundle/default-bundle/start.sh
+++ b/bundle/default-bundle/start.sh
@@ -16,4 +16,4 @@ if [[ -n ${DEBUG_JVM_PRINT_JAVA_OPTS} ]]; then
   echo "Applied JVM options: $JAVA_OPTS"
 fi
 
-exec java ${JAVA_OPTS} -cp '/opt/app/*' "io.camunda.connector.runtime.app.ConnectorRuntimeApplication"
+exec java ${JAVA_OPTS} -cp "/opt/app/*:/opt/custom/*" "io.camunda.connector.runtime.app.ConnectorRuntimeApplication"
--- a/bundle/default-bundle/Dockerfile
+++ b/bundle/default-bundle/Dockerfile
@@ -1,6 +1,9 @@
 FROM eclipse-temurin:17.0.8_7-jre
 
-RUN mkdir /opt/app
+# The /opt/app is used for the Connectors runtime, and out-of-the-box connectors
+# Use the /opt/additionalLibs to mount your own connectors, secret providers, or include other jars into the classpath
+RUN mkdir /opt/app && mkdir /opt/custom
+
 COPY target/*-with-dependencies.jar /opt/app/
 
 # Using the start script from the base connector runtime image


bundle/camunda-saas-bundle/Dockerfile
```diff
@@ -1,6 +1,12 @@
 FROM eclipse-temurin:17.0.8_7-jre
 
-RUN mkdir /opt/app
+# The /opt/app is used for the Connectors runtime, and out-of-the-box connectors
+# Use the /opt/additionalLibs to mount your own connectors, secret providers, or include other jars into the classpath
+RUN mkdir /opt/app && mkdir /opt/additionalLibs
```
sbuettner, line 5: Can we use something simpler like `/opt/custom`?

sbuettner: Just want to avoid any issues with snakeCaseUserProblems

bundle/camunda-saas-bundle/Dockerfile
```diff
@@ -9,4 +15,4 @@ RUN addgroup --gid 1003 connectors && useradd -g connectors -u 1003 -M connector
 USER connectors:connectors
 
 # Using entry point to allow downstream images to add JVM arguments using CMD
-ENTRYPOINT ["java", "-cp", "/opt/app/*", "io.camunda.connector.runtime.saas.SaaSConnectorRuntimeApplication"]
+ENTRYPOINT ["java", "-cp", "/opt/app/*:/opt/additionalLibs/*", "io.camunda.connector.runtime.saas.SaaSConnectorRuntimeApplication"]
```
sbuettner, line 18:
```suggestion
ENTRYPOINT ["java", "-cp", "/opt/app/*;/opt/additionalLibs/*", "io.camunda.connector.runtime.saas.SaaSConnectorRuntimeApplication"]
```

bundle/camunda-saas-bundle/Dockerfile
igpetrov: @sbuettner how about `/opt/extras/`?

sbuettner, line 5: I dont think we need this feature in saas. We can keep the simpler approach here.

bundle/default-bundle/Dockerfile-focal
```diff
@@ -1,6 +1,9 @@
 FROM eclipse-temurin:17.0.8_7-jre-focal
 
-RUN mkdir /opt/app
+# The /opt/app is used for the Connectors runtime, and out-of-the-box connectors
+# Use the /opt/additionalLibs to mount your own connectors, secret providers, or include other jars into the classpath
+RUN mkdir /opt/app && mkdir /opt/additionalLibs
```
sbuettner, line 5: What if we use something simpler like `/opt/custom`? This would avoid user problems as snakeCaseOftenLeadsToProblems 😄

bundle/default-bundle/start.sh
```diff
@@ -16,4 +16,4 @@ if [[ -n ${DEBUG_JVM_PRINT_JAVA_OPTS} ]]; then
   echo "Applied JVM options: $JAVA_OPTS"
 fi
 
-exec java ${JAVA_OPTS} -cp '/opt/app/*' "io.camunda.connector.runtime.app.ConnectorRuntimeApplication"
+exec java ${JAVA_OPTS} -cp '/opt/app/*:/opt/additionalLibs/*' "io.camunda.connector.runtime.app.ConnectorRuntimeApplication"
```
sbuettner, line 19:
```suggestion
exec java ${JAVA_OPTS} -cp '/opt/app/*;/opt/additionalLibs/*' "io.camunda.connector.runtime.app.ConnectorRuntimeApplication"
```

AFAIK you have to use a semicolon here: https://stackoverflow.com/questions/58507668/set-java-classpath-to-multiple-folders-that-have-jar-files-and-run-from-command

bundle/camunda-saas-bundle/Dockerfile
igpetrov: Sounds good

bundle/default-bundle/Dockerfile-focal
sbuettner: `extras` works for me as well. Thought about custom because we call it custom connector and custom secret provider.

sbuettner:
@igpetrov I added another requirement. It would be great if we would allow customers to override the name of the folder via an environment variable but the default should be our folder name. This would give customers more flexibility in terms of naming.

igpetrov:
> @igpetrov I added another requirement to the original ticket as I wasnt aware you already implemented it. It would be great if we would allow customers to override the name of the folder via an environment variable but the default should be our folder name. This would give customers more flexibility in terms of naming.

That would require re-building Docker image. Is that what you expect, because I thought this change for those users who would like to use image from scratch? Otherwise, creating folders is possible even without this change with standard tools: Docker Compose, K8S.

sbuettner:
@igpetrov The idea was that a customer could simply define the variable that will be used in the classpath search. Why should a rebuild be necessary? If you mount and override one woouldnt need to create a new folder.

The environment variable `CUSTOM_CLASSPATH` for example would by default point to `/opt/custom/*`. If a customer overrides it, it would point to whatever they like to mount.

WDYT?

bundle/camunda-saas-bundle/Dockerfile
```diff
@@ -1,6 +1,10 @@
 FROM eclipse-temurin:17.0.8_7-jre
 
 RUN mkdir /opt/app
+
+# Use this folder to add/mount your own libraries that you wish
+# to load into classpath, such as custom connectors, or secret providers
+
```
sbuettner, line 7: Can ew remove this for the saas bundle?

bundle/default-bundle/start.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 JAVA_OPTS="${JAVA_OPTS}"
+CUSTOM_CLASSPATH_FOLDER="${CUSTOM_CLASSPATH_FOLDER:-/opt/custom}"
```
sbuettner, line 4: I thought we support `CUSTOM_CLASSPATH` and just use that variable and default it to: `/opt/custom/*` in case customers want to mount a single file instead of a directory.

igpetrov: Removed env variable

sbuettner (approved):

sbuettner:
Thanks for the quick implementation!

PR #1140, opened 2023-09-14
feat: support process definition deletions

### LINKED ISSUE #1117: Support process definition deletions for inbound connectors
__What should we do?__

The runtime should support deletions of process definition. This requires a change on how the runtime resolves process definitions from Operate. The current approach assumes that no process definitions get deleted. This needs to be changes and might require the runtime to fetch all process definition data from operate instead of gradually fetchting the latest process definitions as its currently implemented.

__Why should we do it?__

Deletion of process defintions are currently not supported but will be part of Operate as part of the 8.3 release. The connectors runtime should support the deletion of process defintions and individual process definition versions.

https://github.com/camunda/product-hub/issues/615

__How we do it?__

Operate currently doesnt provide an endpoint to receive only the latest process definition versions. We therefore have to fetch/paginate through all versions of all definitions and have to detect deleted versions and definitions and update the inbound connectors accordingly.

chillleader:
## Description

Adds support for process definition deletions. New algorithm:
- Fetch all process definitions every 5 seconds (by chunks of 50 due to how Operate API works, but combine all pages afterwards)
- Keep only latest versions of every process to simplify further comparisons
- Compare to saved versions and determine which process definitions are completely new, which ones are updated and which ones have been deleted.

## Related issues

closes #1117


sbuettner (approved):
Great!

PR #1160, opened 2023-09-20
feat(job): Provide access to job information

### LINKED ISSUE #1153: Add possibility to retrieve job-related general purpose information from the context in a connector

I want to add certain general purpose information from a running process instance calling my connector into my logic, e.g. to distinguish a local file path for different process instances using my outbound connector I would like to receive the _processInstanceKey_ from the context or from a separate object.

For _@JobWorker_ methods I can get the _ActivatedJob_ parameter in my method to get these kind of information. Something similar for a connector would be great. Either from the context or from a similar object as the ActivatedJob object in JobWorkers.

Alternatively, I could think of having these kind of general purpose variables available through FEEL, so I can at least set them from the outside.

__Additional context__

None

sbuettner:
feat(job): Provide access to job information via JobContext in Outbou…ndConnectorContext
chore(secret): Remove deprecated @Secret annotation.

closes #1153


connector-sdk/core/src/main/java/io/camunda/connector/api/outbound/JobContext.java
chillleader, line 1: Should we also add `tenantId` to `JobContext`? Or wait until it is supported normally?

sbuettner: Good point @chillleader. Added it. Please check again.

chillleader (approved):

itsmestefanjay:
Wow that was fast. Thanks!

PR #1180, opened 2023-09-25
build: revisit release workflows

chillleader:
## Description

New workflows proposal:

1. In the beginning of every release cycle (i.e. immediately after the previous successful minor release), we will run the `CREATE_RELEASE_BRANCH` workflow. It will create a branch for the next release and set the corresponding version in the release branch, as well as in `main`.
2. When we do the monthly alpha release, we simply do it from the main branch with the current snapshot version.
3. After the monthly alpha release, the release branch is rebased into main to include the changes that were introduced that month.
4. When it is time for the minor release (e.g. 8.4, 8.5 etc.), we create a release tag from the release branch and draft a new release. This triggers the `RELEASE` workflow that publishes artifacts and updates the drafted release with the release details.

Feedback is welcome in relation to both workflows code and the general approach 🙂 

The changes are still in test, missing parts:
- [ ] Support alpha releases from main branch

## Related issues

related to https://github.com/camunda/team-connectors/issues/453


igpetrov (approved):
LGTM

sbuettner:
> When we do the monthly alpha release, we simply do it from the main branch with the current snapshot version.

What about that we rebase the release branch to main and release the alphas from there to be able to have code freezes for alphas. Otherwise we run into the same issues we have right now where unstable PRs have to wait as we want to prepare and test a relase.

igpetrov (approved):

PR #1185, opened 2023-09-27
fix(element-template): Adjust Description for Inbound Operations

ManuelDittmar:
## Description

Element Template uses same descriptions as Outbound Connector, even though another operation is performed.


igpetrov (approved):
LGTM. Updated also newly created message start event element template

PR #1202, opened 2023-10-03
fix(http): Payload not supported in DELETE method

Oleksiivanov:
## Description
Payload not supported in DELETE method

Exception : Apache HTTP client does not support DELETE requests with content.


sbuettner (approved):

sbuettner (changes requested):
The element templates for the http json connector are generated. Please change the respective annotations.

sbuettner (approved):

PR #1224, opened 2023-10-05
Simplify the inbound context correlation API

sbuettner:


chillleader (approved):

markfarkas-camunda (approved):
LGTM!

igpetrov (approved):
LGTM, just a question to make sure we don't have a regression.
So, if we don't return a `CorrelationResult` now, can we still use `=correlation` in the return body expression.

sbuettner:
@igpetrov I added a compatibility layer just now but as we upgraded to the latest FEEL engine an expression that uses it and tries to access a `null` value would not lead to any errors.

igpetrov:
> @igpetrov I added a compatibility layer just now but as we upgraded to the latest FEEL engine an expression that uses it and tries to access a `null` value would not lead to any errors.

Sorry, I meant something different. There were numerous requests from customers that wanted to have `=correlation` result, that's why we initially introduced that. Just wanted to confirm whether we still support or want to support that.

chillleader (approved):

sbuettner:
@igpetrov no, we dont want to support that for now.

parent/pom.xml
```diff
@@ -44,6 +44,7 @@
     <version.java>17</version.java>
```
chillleader, line 44: Shouldn't it be `java.version`?

chillleader: Nevermind, it is consumed as `version.java` in camunda-release-parent

igpetrov (approved):
LGTM

PR #1225, opened 2023-10-05
Fix release workflows

chillleader:
## Description

Release workflow debugging, possibly not the last PR

### DIFF
--- a/.github/workflows/RELEASE.yaml
+++ b/.github/workflows/RELEASE.yaml
@@ -3,7 +3,7 @@ run-name: Release a new version ${{ inputs.version }}
 
 on:
   release:
-    types: [ published ]
+    types: [ created ]
 
 jobs:
   build-and-push:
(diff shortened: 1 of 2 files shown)


.github/workflows/RELEASE.yaml
```diff
@@ -104,7 +104,7 @@ jobs:
           RELEASE_VERSION: ${{ github.event.release.tag_name }}
 
       - name: Deploy artifacts to Artifactory and Maven Central
-        run: mvn -B compile generate-sources source:jar javadoc:jar deploy -PcheckFormat -Psonatype-oss-release
+        run: mvn -B compile generate-sources source:jar javadoc:jar deploy -PcheckFormat -Psonatype-oss-release -DskipTests
```
markfarkas-camunda, line 107: Do we skip tests because at this point we already ran them in a previous workflow? Just wanna understand.

markfarkas-camunda (approved):
Just one question out of curiosity, but looks good👍

igpetrov (approved):

.github/workflows/RELEASE.yaml
chillleader: Yes, because main branch is always tested, as well as the feature branches before merging.

PR #1280, opened 2023-10-13
feat(jackson): support dynamic FEEL context resolution in datatype-feel

### LINKED ISSUE #1161: FEEL deserializer should support process instance variables for inbound connectors

We use the `@FEEL` annotation and the corresponding Jackson deserializer to evaluate FEEL expressions during deserialization, which is useful for inbound connectors where input properties don't pass through Zeebe. We don't provide any context for FEEL evaluation in such cases.

Now we have a new use-case that involves fetching the process variables and including them into FEEL expression evaluation (HTTP Polling connector). The [current implementation](https://github.com/camunda/connectors/pull/1052) of this functionality is imperative. To reduce complexity, improve maintainability and developer experience, it would be great to transition to a more declarative programming model where `@FEEL` annotation handler supports process instance variables too.

* `FeelDeserializer` is adjusted to enrich the FEEL evaluation context with process instance variables:
  * Either process instance key is provided and the deserializer fetches process instance variables from Operate lazily
  * Or variables are pre-fetched and passed to the deserializer
* Jackson's [deserialization context](https://stackoverflow.com/questions/37756939/set-jackson-deserializationcontext-attribute-in-per-request-thread-safe-manner/46893967#46893967) is used to pass the parameters.
* `ProcessInstanceContext` passes the process instance key (or prefetched variables) when invoking the deserialization via `bind` method
* `InboundConnectorContext` doesn't pass the process instance key / variables when invoking the `bindProperties` method (current behavior is retained).

Keep the current imperative solution (simple, but might be more difficult to understand for connector developers)

__Additional context__

See [this thread](https://github.com/camunda/connectors/pull/1052#discussion_r1328323880) for more context.

chillleader:
## Description

* Dynamic FEEL context resolution support for FEEL deserializers
* FEEL module restructuring: new module for `connector-feel-wrapper`, so `jackson-datatype-feel` has a cleaner structure now

The dynamic context supplier can be provided using deserialization context attributes. An API is provided to create an ObjectReader that already contains a context supplier - `FeelContextAwareObjectReader`.

## Related issues

closes #1161


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/DefaultProcessInstanceContext.java
```diff
@@ -65,34 +65,18 @@ public Long getKey() {
...
-      }
+    try {
+      T mappedObject =
+          FeelContextAwareObjectReader.of(objectMapper)
+              .withContextSupplier(operatePropertiesSupplier)
+              .readValue(processDefinitionProperties, cls);
+      validationProvider.validate(mappedObject);
+      return mappedObject;
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
sbuettner, line 76: Nice 😄

connector-sdk/feel-wrapper/src/test/java/io/camunda/connector/feel/FeelEngineWrapperExpressionEvaluationTest.java
```diff
@@ -202,16 +202,5 @@ void shouldSanitizeScalaMapOutput() {
...
-
-    public TestPojo(String value) {
-      this.value = value;
-    }
-
-    public String getValue() {
-      return value;
-    }
-  }
+  record TestPojo(String value) {}
```
sbuettner, line 205: 👍

PR #1293, opened 2023-10-17
Allow configurable prefix for environment secret provider

### LINKED ISSUE #1290: Allow Prefix for Spring Environment Secret Provider

The secret provider can be modified, but this requires actual coding.

It would be handy to have a configurable way of creating a scoped secret provider.

The Spring Environment Secret Provider should be adjustable with a prefix for the secret name that is applied.

This will allow for a dedicated secret namespace so that client credentials will not be exposed.

Implementing a separate secret provider and making interchangeable with different profiles

__Additional context__

https://jira.camunda.com/browse/SEC-689

jonathanlukas:
## Description

The Spring Environment Secret Provider has been adjusted to accept a prefix. This allows for easily narrowing down the environment that is accessible through the secret provider.
## Related issues

closes #1290

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/env/SpringEnvironmentSecretProviderTest.java
+++ b/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/env/SpringEnvironmentSecretProviderTest.java
@@ -16,22 +16,20 @@
  */
 package io.camunda.connector.runtime.env;
 
-import io.camunda.connector.api.secret.SecretProvider;
-import org.springframework.core.env.Environment;
+import static org.assertj.core.api.Assertions.*;
 
-public class SpringEnvironmentSecretProvider implements SecretProvider {
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.env.MockEnvironment;
 
-  private final Environment environment;
-  private final String prefix;
+public class SpringEnvironmentSecretProviderTest {
 
-  public SpringEnvironmentSecretProvider(Environment environment, String prefix) {
-    this.environment = environment;
-    this.prefix = prefix;
-  }
-
-  @Override
-  public String getSecret(String name) {
-    String prefixedName = prefix == null ? name : prefix + name;
-    return environment.getProperty(prefixedName);
+  @Test
+  void shouldApplyPrefix() {
+    MockEnvironment env = new MockEnvironment();
+    env.setProperty("secrets.MY_TOTAL_SECRET", "beebop");
+    SpringEnvironmentSecretProvider secretProvider =
+        new SpringEnvironmentSecretProvider(env, "secrets.");
+    String myTotalSecret = secretProvider.getSecret("MY_TOTAL_SECRET");
+    assertThat(myTotalSecret).isEqualTo("beebop");
   }
 }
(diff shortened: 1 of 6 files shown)


connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/secret/SpringEnvironmentSecretProviderTest.java
```diff
@@ -27,12 +27,15 @@
 
 @SpringBootTest(
     classes = {TestConnectorRuntimeApplication.class},
-    properties = {"camunda.connector.secretprovider.discovery.enabled=false"})
+    properties = {
+      "camunda.connector.secretprovider.discovery.enabled=false",
+      "camunda.connector.secret-provider.environment.prefix=secrets."
```
sbuettner, line 32:
```suggestion
      "camunda.connector.secretprovider.discovery.enabled=false",
      "camunda.connector.secretprovider.environment.prefix=secrets."
```

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/env/SpringEnvironmentSecretProviderTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.springframework.mock.env.MockEnvironment;
+
+public class SpringEnvironmentSecretProviderTest {
+
+  @Test
+  void shouldApplyPrefix() {
+    MockEnvironment env = new MockEnvironment();
+    env.setProperty("secrets.MY_TOTAL_SECRET", "beebop");
```
sbuettner, line 29: Idea for better readability: What about using something that looks more like a prefix and not like our own secrets prefix like to not confuse the reader: `my_secret_prefix_` or similar.

connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java
```diff
@@ -49,7 +49,7 @@ public class OutboundConnectorsAutoConfiguration {
   @Value("${camunda.connector.secretprovider.discovery.enabled:true}")
   Boolean secretProviderLookupEnabled;
 
-  @Value("${camunda.connector.secret-provider.environment.prefix:}")
+  @Value("${camunda.connector.secretprovider.environment.prefix:secrets.}")
```
sbuettner, line 52: Lets not have a default here to be backwards compatible without any suprises.

sbuettner (changes requested):
Please remove the default prefix.

jonathanlukas:
@sbuettner I added a PostConstruct method on the SpringEnvironmentSecretProvider that will log WARN if no prefix is configured.

Should we add another log like this before each invocation of `getSecret`?

sbuettner:
> I added a PostConstruct method on the SpringEnvironmentSecretProvider that will log WARN if no prefix is configured.
I dont think this is necessary as the prefix is completely optional. You could add an info log if you like. Otherwise it could be confusing to users.

jonathanlukas:
Alright, I can adjust this. As example, I used keycloak that logs WARN if being run in a dev setup (which is considered unsecure), but the threat is much less with connectors... I will adjust the log level ;)

jonathanlukas:
@sbuettner warn log was replaced by info log now

sbuettner (approved):

PR #1298, opened 2023-10-18
Enable Stackdriver logging

### LINKED ISSUE #1297: [runtime/docker]: Enable Stackdriver JSON logging configuration

Currently, the configuration for using Stackdriver JSON logging in the connectors runtime is not trivial, as it requires:
* an additional depedency
* a custom-logback-spring.xml
* An environment variable to bind the application to this xml: `LOGGING_CONFIG=file:/opt/app/logging/custom-logback-spring.xml`

The used dependency is the [gcp stackdriver logging](https://cloud.spring.io/spring-cloud-static/Greenwich.RC1/multi/multi__stackdriver_logging.html)

There is an [example](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/tree/main/spring-cloud-gcp-logging) available.

An simple environment variable to alter the log format (just like in [zeebe](https://docs.camunda.io/docs/self-managed/zeebe-deployment/configuration/logging/#google-stackdriver-json-logging))

`CONNECTORS_LOG_APPENDER=Stackdriver`

which would enable Stackdriver-compatible logging.

Documenting the first approach clearly. This should be carried out in the meantime.

__Additional context__

https://camunda.slack.com/archives/C02JLRNQQ05/p1697629871877529

jonathanlukas:
## Description

To enable gcp stackdriver json logging, there were dependencies added to:
* provide an appender for the format
* evaluate expressions inside the `logback-spring.xml`

Now, you can switch on stackdriver logging by using `CONNECTORS_LOG_APPENDER=Stackdriver`

## Related issues

closes #1297


connector-runtime/connector-runtime-application/src/main/resources/logback-spring.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <if condition='property("CONNECTORS_LOG_APPENDER").equalsIgnoreCase("stackdriver")'>
```
sbuettner, line 3: What about using https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.logging.logback-extensions.environment-properties here? This allows an `application.properties` based config.

jonathanlukas:
> Hey @jonathanlukas, great contribution as always. What is the goal of this change here? Are we talking about a SaaS or SM usecase? Asking as we should be careful introducing additional libraries to SM that are cloud provider specific. If you look at the SaaS bundle you can already find a json config that enables json logging for GCP.

A customer asked for json logging. As all other components use the stackdriver format, it would be an alignment with the others.

Background is probably that we run on GCP, but a FR for other JSON logging formats could cover all components.

sbuettner (approved):

sbuettner:
@jonathanlukas Thank you for taking care. Would be great to see this reflected in our docs.

PR #1301, opened 2023-10-19
[WIP] feat(generator): HTTP element template SDK

chillleader:
## Description

First iteration of the new template generator SDK layer for building HTTP connector templates.

## Related issues

### DIFF
--- a/connector-sdk/element-template-generator-http/pom.xml
+++ b/connector-sdk/element-template-generator-http/pom.xml
@@ -24,5 +24,26 @@
       <artifactId>connector-http-base</artifactId>
       <version>${project.version}</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.skyscreamer</groupId>
+      <artifactId>jsonassert</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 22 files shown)


chillleader:
A small note about templating path / body & properties:

First attempt was to implement `HttpOperation` and its builder in a way so they accept FEEL expressions and don't do any parsing (or minimal). Like this:

`pathFeelExpression="/examples/" + exampleId`

But this proved to be quite inconvenient, as in real life we need to add more quotes, escape the inner double quotes or use the multi-line preformatted strings. Plus, no syntax highlighting or validation.

Instead, I'd like to try a different approach with templating syntax, that can be used directly in a format in which it is provided in the OpenAPI spec (which is a bonus):

`/examples/{exampleId}`

Under the hood this will be parsed and replaced with the first example that can be used in the element template. This should make the DSL more user-friendly.

sbuettner:
> But this proved to be quite inconvenient, as in real life we need to add more quotes, escape the inner double quotes or use the multi-line preformatted strings. Plus, no syntax highlighting or validation.

Hey @chillleader is it really inconvenient as using FEEL based URLs wont/shouldnt be exposed to end users for generated templates right? It will be a hidden property like we already have for protocol connectors right?

chillleader:
@sbuettner It's convenient to consume but inconvenient to provide 😄 I'm talking about the developer experience and how the code of the protocol parser module will look like. You can see the example of how FEEL approach looks in the only test that I provided so far. I would like to achieve something more structured and easy to use.

sbuettner:
Are you referring to https://github.com/camunda/connectors/pull/1301/files#diff-bf55573348be3134e4dd046705de1738ddf50fb7bb5c72e5b36b253ac49a13f0R71 ?

It would be nice of course to be able to transform URL and parse it. Maybe we can leverage something like this: https://github.com/mikaelhg/urlbuilder as an intermediate format.

sbuettner (approved):

Issue #1307, opened 2023-10-20
README links to the old archived connector-runtime-docker 


rob2universe:
__Describe the Bug__

The Docker section:

> Refer to the [Connector Runtime Docker image documentation](https://github.com/camunda/connector-runtime-docker) for further details.

links to the old archived connector-runtime-docker

__Expected Behavior__

It should link to the new Dockerfile location:
https://github.com/camunda/connectors/tree/main/connector-runtime/connector-runtime-application/example

PR #1308, opened 2023-10-20
Make README.md point to new Dockerfile example

rob2universe:
## Description

point to new Dockerfile example instead of archived repository

## Related issues

closes # 
https://github.com/camunda/connectors/issues/1307


chillleader (approved):
Thank you! ✨

PR #1326, opened 2023-10-25
feat(generator): OpenAPI element template generator

chillleader:
### OpenAPI parsing module

OpenAPI parsing module. Includes the `OpenApiOutboundTemplateGenerator` class and various adjustments to the `http-dsl` module discovered during implementation.

The PR only includes a mock test for OpenAPI generator itself, but I would rather add tests later because the generator behavior is still unstable and will be tweaked as we test and develop it further.

### CLI tool

Introduces the element template generator CLI tool `congen`

Features:
- Can work with any template generator that implements `CliCompatibleTemplateGenerator` interface (basically an extended version of `Element Template Generator`)
- Generators can be bundled with the tool via Java ServiceLoader
- Able to can the generation source to dry run the generation and list the supported operations, detailed reason is provided for each non-supported operation
- Configurable template name, id, version, hybrid support

Example usage:
1. `congen openapi-outbound scan https://petstore3.swagger.io/api/v3/openapi.json`
  Here, `openapi-outbound` is the generator name, followed by a source link
2. `congen openapi-outbound generate https://petstore3.swagger.io/api/v3/openapi.json createOrder`
  This will generate the actual template and include only the `createOrder` operation
3. `congen -h -i MyTemplate openapi-outbound generate https://petstore3.swagger.io/api/v3/openapi.json`
  This generates a template with a custom ID `MyTemplate` and a configurable `type` property (hybrid mode).

## Related issues


chillleader:
Added the [CLI PR](https://github.com/camunda/connectors/pull/1346) into this branch for convenience as changes are somewhat interconnected.

element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpOutboundElementTemplateBuilder.java
```diff
@@ -0,0 +1,133 @@
...
+      List<HttpAuthentication> authentication) {
+    this.authentication = authentication;
+    return this;
+  }
+
+  public OutboundElementTemplate build() {
+    if (operations == null || operations.isEmpty()) {
+      throw new IllegalStateException("Could not find any supported operations");
+    }
+    builder.elementType("bpmn:ServiceTask").appliesTo(Set.of("bpmn:Task"));
```
sbuettner, line 120: Would an enum make this more type safe?

element-template-generator/openapi-parser/src/test/java/io/camunda/connector/generator/openapi/ExampleTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class ExampleTest {
+
+  private final ObjectMapper mapper = new ObjectMapper();
+
+  @Test
+  void generate() throws JsonProcessingException {
+    // given
+    var parser = new OpenAPIV3Parser();
+    var openApi = parser.read("https://modeler.cloud.camunda.io/v3/api-docs/api-v1");
```
sbuettner, line 33: This and the other test below might fail in case the modeler is not available. What about testing against a local open api spec file that is located in the test resources?

sbuettner (approved):
LGTM, minor comments.

element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpServerData.java
```diff
@@ -14,7 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.camunda.connector.generator.http;
+package io.camunda.connector.generator.dsl.http;
 
 /** Data for the HTTP base URL selector property of the generated element template. */
 public record HttpServerData(String baseUrl, String label) {}
```
sbuettner, line 20: Nice idea to have make this available in the DSL. Lets see whether it can be used for other generators as well.

element-template-generator/congen-cli/src/main/java/io/camunda/connector/generator/cli/command/Generate.java
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  public Integer call() {
+    CliCompatibleTemplateGenerator<Object, ?> generator =
+        (CliCompatibleTemplateGenerator<Object, ?>) loadGenerator(connectorGen.generatorName);
+    Object input;
+    try {
+      input = generator.prepareInput(params);
+    } catch (Exception e) {
+      System.err.println("Error while preparing input data: " + e.getMessage());
+      return -1;
```
igpetrov, line 54: Thought (maybe not for this PR, but future consideration): error codes should be unique. This is not critical here but if at some point, somebody decides to write bash script on top of CLI, they might have issues with identifying exact errors.

igpetrov (dismissed):
LGTM. There was one optional/non-urgent suggestion. Another non-urgent is `verificationExpression` I introduced a day ago (see latest Slack inbound connectors). Do we need to do anything here?

element-template-generator/congen-cli/src/main/java/io/camunda/connector/generator/cli/command/Generate.java
chillleader: I set the following exit codes for now: 1 for execution exception, 2 for invalid input. These are defaults offered by the picocli framework.

chillleader:
`verificationExpression` seems unrelated as we are only covering outbound connectors here

igpetrov (approved):
Thanks for addressing my comments. LGTM

PR #1332, opened 2023-10-26
feat(sdk): adding verification expression

igpetrov:
## Description

feat(sdk): adding verification expression

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
@@ -24,6 +24,7 @@ import static org.springframework.web.bind.annotation.RequestMethod.PUT;
 
 import io.camunda.connector.api.error.ConnectorException;
 import io.camunda.connector.api.inbound.webhook.MappedHttpRequest;
+import io.camunda.connector.api.inbound.webhook.VerifiableWebhook;
 import io.camunda.connector.api.inbound.webhook.WebhookConnectorException;
 import io.camunda.connector.api.inbound.webhook.WebhookConnectorException.WebhookSecurityException;
 import io.camunda.connector.api.inbound.webhook.WebhookConnectorExecutable;
@@ -41,6 +42,7 @@ import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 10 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound.webhook;
+
+import java.util.Map;
+import java.util.function.Function;
+
+public interface VerifiableWebhook {
```
sbuettner, line 22: The idea was to add a single method to the `WebhookExecutable` interface.

connectors/slack/src/main/java/io/camunda/connector/slack/inbound/SlackInboundWebhookExecutable.java
```diff
@@ -89,7 +82,21 @@ public void activate(InboundConnectorContext context) throws Exception {
...
+
+  @Override
+  public WebhookHttpVerificationResult verify(WebhookProcessingPayload payload) throws Exception {
+    verifySlackRequestAuthentic(payload);
+    return props
+        .verificationExpression()
+        .apply(
+            Map.of(
+                "body",
+                bodyAsMap(payload.headers(), payload.rawBody()),
```
sbuettner, line 97: Why do we add the headers to the body? What about query params?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
```diff
@@ -0,0 +1,48 @@
...
+  default WebhookHttpVerificationResult verify(WebhookProcessingPayload payload) throws Exception {
+    return null;
+  }
+
+  default Function<Map<String, Object>, WebhookHttpVerificationResult> verificationExpression() {
+    return response -> null;
+  }
```
sbuettner, line 29: Is this method necessary?

connectors/slack/src/main/java/io/camunda/connector/slack/inbound/SlackInboundWebhookExecutable.java
igpetrov: We don't. The `bodyAsMap` requires headers to determine content type, because it can be either JSON or form data.
Good point on query params. Keep in mind, the PR is draft and still WIP (work in progress).

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
sbuettner: Why not just execute `verify` and check the result?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -89,8 +91,22 @@ private ResponseEntity<?> processWebhook(
...
-      var webhookResult =
-          ((WebhookConnectorExecutable) connector.executable()).triggerWebhook(payload);
+      WebhookConnectorExecutable connectorHook =
+          (WebhookConnectorExecutable) connector.executable();
+
+      if (connectorHook.verificationExpression() != null) {
+        var verificationResult = connectorHook.verify(payload);
+        if (verificationResult != null) {
+          HttpHeaders headers = new HttpHeaders();
+          Optional.of(verificationResult)
```
sbuettner, line 101: `verificationResult` should never be null here.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -89,8 +91,22 @@ private ResponseEntity<?> processWebhook(
...
+              .map(VerifiableWebhook.WebhookHttpVerificationResult::headers)
+              .ifPresent(map -> map.forEach(headers::add));
+          return new ResponseEntity<>(
+              verificationResult.body(), headers, verificationResult.statusCode());
```
sbuettner, line 105: Lets use single return statements and I would recommend to introduce a dedicated method here.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: Why not? If it's `null` (validation failed), then return immediately - fail fast strategy. Otherwise, we need to introduce some sort of flag whether vaildation result was "success" or "failure". WDYT?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: This is fail-fast exit, in case validation failed. In that case, there is no sense to follow down the method. If we want single return statement, it adds up branching in code, which is already too much in this class.

sbuettner: > it adds up branching in code, which is already too much in this class.

A good indicator to split up the method 😄

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
sbuettner: Because you already checked it with: ```verificationResult != null```

sbuettner: Wondering why the use of Optional here.

igpetrov: aaahhh, I see. Sorry, got confused myself :)
Yes, some of those will be cleaned up. As a part of our commitment, we have to show progress every two days, so the code you seeing is still in progress.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
igpetrov: Removed

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: `Optional` still makes sense here, since I am using `map` and `ifPresent`-`forEach` structure. Without `Optional` I would need to write a `for`-loop.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: Splitted it up. Please take a look.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -89,44 +91,79 @@ private ResponseEntity<?> processWebhook(
...
+  }
+
+  protected ResponseEntity verify(
+      WebhookConnectorExecutable connectorHook, WebhookProcessingPayload payload) throws Exception {
+    var verificationResult = connectorHook.verify(payload);
+    if (verificationResult != null) {
+      HttpHeaders headers = new HttpHeaders();
+      Optional.of(verificationResult)
+          .map(VerifiableWebhook.WebhookHttpVerificationResult::headers)
+          .ifPresent(map -> map.forEach(headers::add));
```
sbuettner, line 132: Shouldnt it rather be `Optional.of(verificationResult.headers())`?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -89,44 +91,79 @@ private ResponseEntity<?> processWebhook(
...
+      return new ResponseEntity<>(
+          verificationResult.body(), headers, verificationResult.statusCode());
+    }
+    return null;
+  }
+
+  protected ResponseEntity buildBodyExpressionResponseOrOk(
+      WebhookResult webhookResult, WebhookResultContext processVariablesContext) {
+    if (webhookResult.responseBodyExpression() != null) {
+      var httpResponseData = webhookResult.responseBodyExpression().apply(processVariablesContext);
```
sbuettner, line 142: Couldnt this throw an exception?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -89,44 +91,79 @@ private ResponseEntity<?> processWebhook(
...
+      return ResponseEntity.ok(httpResponseData);
+    } else {
+      return ResponseEntity.ok().build();
```
sbuettner, line 145: Please avoid multiple return statements.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -89,44 +91,79 @@ private ResponseEntity<?> processWebhook(
...
+        return handleWebhookConnectorException(webhookConnectorException);
       } else {
-        connectorResponse = ResponseEntity.internalServerError().build();
+        return ResponseEntity.unprocessableEntity()
+            .body(
+                new ErrorResponse(
+                    connectorException.getErrorCode(), connectorException.getMessage()));
       }
+    } else {
+      return ResponseEntity.internalServerError().build();
```
sbuettner, line 165: Please avoid multiple return statements similar to the previous implementation.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: Very much depends on `apply()` implementation. The method call is covered behind `try-catch`.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: Reworked this and other methods. I am generally okay with multiple `ret`s. I am wondering, why don't you like multiple `return` statements? We agreed on Google Java Code Style but it says nothing about returns.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
igpetrov: @sbuettner do you have any specific issue in mind?

connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/jackson/FeelFunctionDeserializerTest.java
```diff
@@ -195,4 +213,6 @@ private record TargetTypeInteger(Function<InputContextInteger, Integer> function
   private record TargetTypeList(Function<InputContextInteger, List<Long>> function) {}
 
   private record TargetTypeMap(Function<InputContextInteger, Map<String, Long>> function) {}
+
+  private record TargetTypeFoldedMap(Function<InputContextInteger, Map<String, Object>> function) {}
```
chillleader, line 217: nit:

The original issue actually involved an `Object` return type that contained a map as one of its fields, right? I think it might be more representative to test the similar use case here:

```
private record FoldedMapRecord(Map<String, Object> folded) {}

private record TargetTypeFoldedMap(Function<InputContextInteger, FoldedMapRecord> function) {}
```

The reason this caught my eye, I think the current test case could have been handled by the previous implementation as well (by [this code](https://github.com/camunda/connectors/blob/ab6a44e56660c39ce7c1bdd59f32e0df85345ed3/connector-sdk/feel-wrapper/src/main/java/io/camunda/connector/feel/FeelEngineWrapper.java#L101), as the root type is already a map). But I might be wrong here.

chillleader (approved):
LGTM, just a minor comment to the test case

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
sbuettner: Not really but usually we catch exceptions where they happen.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
sbuettner: I think its usually (not always) easier to follow but we can also discuss with the team.

sbuettner (approved):

connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/jackson/FeelFunctionDeserializerTest.java
igpetrov: Spent some time trying to rewrite and didn't manage to make it working 😕 
Let me merge it as-is now, and I'll follow up on this in [this task](https://github.com/camunda/connectors/issues/1354).

PR #1343, opened 2023-10-30
feat(test): Add e2e tests for http rest connector

sbuettner:
## Related issues

closes https://github.com/camunda/team-connectors/issues/532


connectors-e2e-test/connector-http-json-test/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+  <parent>
+    <groupId>io.camunda.connector</groupId>
+    <artifactId>connector-parent</artifactId>
+    <relativePath>../../parent/pom.xml</relativePath>
+    <version>8.4.0-SNAPSHOT</version>
+  </parent>
+
+  <description>Tests</description>
+  <artifactId>connector-http-json-test</artifactId>
+  <packaging>jar</packaging>
```
chillleader, line 14: Will we need this module packaged as a jar for anything apart from our CI/CD pipelines? If not, then I would suggest disabling `maven-shade-plugin` for it (originally defined in the parent) to save some build time.

connectors-e2e-test/connector-http-json-test/src/test/java/io/camunda/connector/e2e/rest/HttpJsonConnectorTest.java
```diff
@@ -0,0 +1,226 @@
...
+    classes = {TestConnectorRuntimeApplication.class},
+    properties = {
+      "spring.main.allow-bean-definition-overriding=true",
+      "camunda.connector.webhook.enabled=false",
+      "camunda.connector.polling.enabled=false"
+    })
+@ZeebeSpringTest
+@ExtendWith(MockitoExtension.class)
+// TODO discuss whether we want to enable it by default
+// @EnabledIfEnvironmentVariable(named = "CONNECTORS_E2E_TEST", matches = "true")
```
chillleader, line 60: Yes, that would be great 👍  We could then add a new pipeline for running e2e tests or add a checkbox to optionally run it during snapshot / release builds.

chillleader (commented):
This looks great!

connectors-e2e-test/connector-http-json-test/pom.xml
sbuettner: The `maven-shade-plugin` is configured for the connectors. The tests project is not inheriting from the connectors pom and therefore the plugin is not enabled.

PR #1350, opened 2023-11-01
chore: added banners and /tmp volume

igpetrov:
## Description

- Added banner: addresses an issue to show current running version of connectors
- Added a volume, which is related to https://github.com/camunda/team-connectors/issues/525

## Banner

Screenshot 2023-11-01 at 14 46 24

## Related issues

- https://github.com/camunda/team-connectors/issues/525

### DIFF
--- a/bundle/README.md
+++ b/bundle/README.md
@@ -25,7 +25,7 @@ docker run --rm --name=connectors -d \
            -e MY_SECRET=secret \              # Set a secret with value
            -e SECRET_FROM_SHELL \             # Set a secret from the environment
            --env-file secrets.txt \           # Set secrets from a file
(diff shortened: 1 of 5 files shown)


markfarkas-camunda (approved):
Awesome 👍

chillleader (approved):
Nice 🙌 

Should we place the banner in `connector-runtime/connector-runtime-application`? I think it would be applied to both bundles too, then.

igpetrov:
> Nice 🙌
> 
> Should we place the banner in `connector-runtime/connector-runtime-application`? I think it would be applied to both bundles too, then.

I tired that one too but never worked for me.

Issue #1364, opened 2023-11-03
Cannot fetch inbound connectors if any process contains a polling connector


markfarkas-camunda:
__Describe the Bug__

If any of the active processes contain a polling connector one can not fetch the list of inbound connectors calling the `http://localhost:8080/inbound` endpoint. Calling this results the following exception on the backend:
```
java.lang.ClassCastException: class io.camunda.connector.runtime.core.inbound.InboundIntermediateConnectorContextImpl cannot be cast to class io.camunda.connector.runtime.core.inbound.InboundConnectorContextImpl (io.camunda.connector.runtime.core.inbound.InboundIntermediateConnectorContextImpl and io.camunda.connector.runtime.core.inbound.InboundConnectorContextImpl are in unnamed module of loader 'app')
	at io.camunda.connector.runtime.inbound.lifecycle.InboundConnectorRestController.mapToResponse(InboundConnectorRestController.java:56)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.AbstractList$RandomAccessSpliterator.forEachRemaining(AbstractList.java:722)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at io.camunda.connector.runtime.inbound.lifecycle.InboundConnectorRestController.getActiveInboundConnectors(InboundConnectorRestController.java:51)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:205)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:150)
[...]

PR #1371, opened 2023-11-06
fix(inbound): fix health report on /inbound endpoint

markfarkas-camunda:
## Description

The /inbound returned 500 error, because of class cast exception. This fix eliminates this error by using the more generic InboundConnectorExecutable instead of the InboundConnectorContextImpl.

## Related issues

closes #
https://github.com/camunda/connectors/issues/1364

### DIFF
--- a/connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
+++ b/connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
@@ -100,5 +100,13 @@ public interface InboundConnectorContext {
    */
   void reportHealth(Health health);
 
+  /**
+   * Provides a Health object to get information about the current status of the Connector with
+   * optional details.
+   *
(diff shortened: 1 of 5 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -99,4 +99,6 @@ public interface InboundConnectorContext {
    * implementation requires it.
    */
   void reportHealth(Health health);
+
+  Health getHealth();
```
sbuettner, line 103: Maybe add a simple javadoc.

sbuettner (approved):
LGTM

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
markfarkas-camunda: Good point, thank you!

sbuettner (approved):

PR #1392, opened 2023-11-08
feat(webhook): add verification expression support

igpetrov:
feat(webhook): add verification expression support

### DIFF
--- a/connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
+++ b/connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
@@ -19,25 +19,12 @@ package io.camunda.connector.api.inbound.webhook;
 import java.util.Map;
 
 public interface VerifiableWebhook {
-  default WebhookHttpVerificationResult verify(WebhookProcessingPayload payload) throws Exception {
-    return null;
-  }
+  WebhookHttpVerificationResult verify(WebhookProcessingPayload payload);
 
-  record WebhookHttpVerificationResult(Object body, Map<String, String> headers, int statusCode) {
-    public WebhookHttpVerificationResult() {
-      this(Map.of(), null, 200);
-    }
-
-    public WebhookHttpVerificationResult(Object body) {
-      this(body, null, 200);
-    }
-
-    public WebhookHttpVerificationResult(Object body, int statusCode) {
-      this(body, null, statusCode);
-    }
-
-    public WebhookHttpVerificationResult(Object body, Map<String, String> headers) {
-      this(body, headers, 200);
+  record WebhookHttpVerificationResult(
+      Object body, Map<String, String> headers, Integer statusCode) {
+    public WebhookHttpVerificationResult {
+      statusCode = statusCode == null || statusCode < 100 ? 200 : statusCode;
     }
   }
 }
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
@@ -125,7 +125,12 @@ public class InboundWebhookRestController {
   protected ResponseEntity verify(
       WebhookConnectorExecutable connectorHook, WebhookProcessingPayload payload) throws Exception {
     ResponseEntity response = null;
-    var verificationResult = connectorHook.verify(payload);
+
+    VerifiableWebhook.WebhookHttpVerificationResult verificationResult = null;
+    if (connectorHook instanceof VerifiableWebhook verifiableWebhook) {
+      verificationResult = verifiableWebhook.verify(payload);
+    }
+
     if (verificationResult != null) {
       HttpHeaders headers = new HttpHeaders();
       Optional.of(verificationResult)
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
@@ -313,19 +313,23 @@ class WebhookControllerTestZeebeTests {
     assertEquals("expression", responseEntity.getBody().expression());
   }
 
+  interface MyVerifiableWebhook extends WebhookConnectorExecutable, VerifiableWebhook {}
+
   @Test
   @SuppressWarnings("unchecked")
   public void testSuccessfulProcessingWithVerification() throws Exception {
-    WebhookConnectorExecutable webhookConnectorExecutable = mock(WebhookConnectorExecutable.class);
+    var webhookConnectorExecutable = mock(MyVerifiableWebhook.class);
     WebhookResult webhookResult = mock(WebhookResult.class);
     when(webhookResult.request()).thenReturn(new MappedHttpRequest(Map.of(), Map.of(), Map.of()));
     when(webhookResult.response())
         .thenReturn(new WebhookHttpResponse(Map.of("keyResponse", "valueResponse"), null));
     when(webhookConnectorExecutable.triggerWebhook(any(WebhookProcessingPayload.class)))
         .thenReturn(webhookResult);
+
     when(webhookConnectorExecutable.verify(any(WebhookProcessingPayload.class)))
         .thenReturn(
-            new VerifiableWebhook.WebhookHttpVerificationResult(Map.of("result", "GOOD"), 201));
+            new VerifiableWebhook.WebhookHttpVerificationResult(
+                Map.of("result", "GOOD"), Map.of(), 201));
 
     var webhookDef = webhookDefinition("processA", 1, "myPath");
     var webhookContext =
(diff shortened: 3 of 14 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
```diff
@@ -39,5 +39,10 @@ public WebhookHttpVerificationResult(Object body, int statusCode) {
     public WebhookHttpVerificationResult(Object body, Map<String, String> headers) {
       this(body, headers, 200);
     }
+
+    @Override
+    public int statusCode() {
```
sbuettner, line 44: Lets return Integer here instead.

igpetrov: Done

igpetrov:
@sbuettner do you consider to put it into `8.3.1`? If so, I will go ahead, and create a QA ticket.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
```diff
@@ -32,12 +33,17 @@ public WebhookHttpVerificationResult(Object body) {
       this(body, null, 200);
     }
 
-    public WebhookHttpVerificationResult(Object body, int statusCode) {
+    public WebhookHttpVerificationResult(Object body, Integer statusCode) {
```
sbuettner, line 36: Why not create a helper method like 
```
static WebhookHttpVerificationResult.ok(Object body) {
   ...
}```

Comment (igpetrov) [reply connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java:36]: A matter of style

**Review by sbuettner — COMMENTED** [review]
I am curious by the way why we went for a dedicated interface `public interface VerifiableWebhook` and extend it automatically in the `WebhookConnectorExecutable`?

If it would be optional to the implementor we could remove the default implementation etc.

File: connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java (hunk 2 above)
Comment (sbuettner) [reply connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java:36]: You would not need the extra code in the `statusCode`

**Comment by igpetrov** [comment]
> I am curious by the way why we went for a dedicated interface `public interface VerifiableWebhook` and extend it automatically in the `WebhookConnectorExecutable`?
> 
> If it would be optional to the implementor we could remove the default implementation etc.

I prefer to work with smaller interfaces with explicitly defined purpose. In that case, verification expression is covered by a dedicated interface.

File: connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java (hunk 2 above)
Comment (igpetrov) [reply connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java:36]: > You would not need the extra code in the `statusCode`

We still would need that tho. if the verification expression does not  specify `statusCode: xxx`, it will still evaluate to `null`. Constructors are only used as quality of life improvement mostly for tests, and in this case I think it's a matter of taste.

The only option to remove lines in the `statusCode()` is to add `if-else` checks upstream somewhere, but I would like to avoid that and encapsulate it here.

**Comment by sbuettner** [comment]
> explicitly defined purpose
Agreed, but why extend it in the default `WebhookConnectorExecutable`? You could get rid of the whole default behaviour if its optional for users to implement.

File: connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java (hunk 2 above)
Comment (sbuettner) [reply connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java:36]: I think we are mixing the custom implementation with the expression here.
When evaluating the expression you could analyze the data and create a record accordingly no?

**Comment by igpetrov** [comment]
>  explicitly defined purpose
> > Agreed, but why extend it in the default `WebhookConnectorExecutable`? You could get rid of the whole default behaviour if its optional for users to implement.

Sorry, maybe I am not getting your concern right. `WebhookConnectorExecutable` is a core interface, which is used in many places, like `InboundConnectorManager` and `InboundConnectorRestController`. By extending `WebhookConnectorExecutable` with `VerifiableWebhook` I am implicitly defining relation `WebhookConnectorExecutable-[is-a]-VerifiableWebhook` which several `instanceof` calls. Or maybe I misunderstand your concern 🤔 .

**Comment by sbuettner** [comment]
@igpetrov Usually one would extends the `VerifiableWebhook` with the `WebhookConnectorExecutable`. Meaning that its an extra capability that one can implement optionally. If you implement a `VerifiableWebhook` it verifies but its also executable. You implemented in a way that every `WebhookConnectorExecutable` is verifiable. I hope that makes it clearer.

File: connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java [hunk 3]
```diff
@@ -32,12 +33,17 @@ public WebhookHttpVerificationResult(Object body) {
...
       this(body, null, statusCode);
     }
 
     public WebhookHttpVerificationResult(Object body, Map<String, String> headers) {
       this(body, headers, 200);
     }
+
+    @Override
+    public Integer statusCode() {
+      return statusCode == null || statusCode < 100 ? 200 : statusCode;
```
sbuettner, line 46: @igpetrov Records supports custom constructors for these use-cased: https://www.baeldung.com/java-records-custom-constructor#how-to-make-a-custom-constructor-for-a-java-record

sbuettner: There you can put 200 as the default if you like

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
igpetrov: It's evaluated automatically. If the expression is `= {body:{"key":"value"}}`, then it will create an object with `null` as status code and headers. This is beyond of my control. Custom constructor is an option but also a custom getter. Unfortunatelly, we cannot use kotlin style default init like `record Data(body, statusCode=200)`.

igpetrov:
> @igpetrov Usually one would extends the `VerifiableWebhook` with the `WebhookConnectorExecutable`. Meaning that its an extra capability that one can implement optionally. If you implement a `VerifiableWebhook` it verifies but its also executable. You implemented in a way that every `WebhookConnectorExecutable` is verifiable. I hope that makes it clearer.

I guess, I understand what you mean and it makes sense. What I am trying to achieve with it are two things: (1) `VerifiableWebhook` is more like an extention to make code clearer, I think we agreed on that; (2) we currently heavily rely on `WebhookConnectorExecutable` interface in runtime. If I make `VerifiableWebhook` extending `WebhookConnectorExecutable`, then I may need to perform a bit of refactoring.

So extending `WebhookConnectorExecutable` with `VerifiableWebhook` is just more practical. In addition to that, `VerifiableWebhook` has just a `verify` method. Potentially, we could be able to extend with it other interfaces, such as `HttpPollingConnector`. 
Maybe we should consider the following: (1) give it better naming (`Verifiable`, `VerifiableInboundConnector`?), and (2) rework it in such a way, that it fits not only webhooks? WDYT?

sbuettner:
Not sure if there is a use-case for a `VerifiableInboundConnector` but having a simple interface could allow this. Although I would be careful as we return http related information (status) with the verification here which doesnt apply for other connectors like kafka for example.

> we currently heavily rely on WebhookConnectorExecutable interface in runtime.
Exactly and its totally fine. The nice thing with having a simple `Verifiable` or `VerifiableWebhook` is that you can just check whether a webhook implements that interface and execute the verification if its the case. We could remove the whole default behaviour this way and keep the interfaces clean, simple and combinable.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
sbuettner: @igpetrov You could map to to a `Map<K,V>` and evaluate and construct the result manually no?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/webhook/VerifiableWebhook.java
sbuettner: The benefit is that you construct a valid object that has the right values. Similar to kotlin data classes.

sbuettner:
@igpetrov Please take a look at this proposal to make it more clear: https://github.com/camunda/connectors/tree/webhook-verification-expression-idea

igpetrov:
> @igpetrov Please take a look at this proposal to make it more clear: https://github.com/camunda/connectors/tree/webhook-verification-expression-idea

That works for me. I thought it to be more cohesive but that works for me as well. Let me think about it a bit first.

sbuettner:
@igpetrov Of course, no need to rush here and I could live with both approaches.

Also chiming in @chillleader for a second pair of eyes.

igpetrov:
> @igpetrov Of course, no need to rush here and I could live with both approaches.
> 
> Also chiming in @chillleader for a second pair of eyes.

So, after thinking about the [proposal](https://github.com/camunda/connectors/commit/ccaeb96eac407e25b6b2361502d2ecf4051ad223), it seems to me alright, my only concern now is that previously customer had to implement an interface `WebhookConnectorExecutable` which was essentially a single one they need to know about when talking about webhook implementation. Now they need to learn a new one `VerifiableWebhook`. This may theoretically cause some confusion.

But overall, I like the approach. Happy to discuss it with the team. WDYT @sbuettner @chillleader @markfarkas-camunda @Oleksiivanov?

sbuettner:
@igpetrov 
> it seems to me alright, my only concern now is that previously customer had to implement an interface WebhookConnectorExecutable which was essentially a single one

Valid point, lets continue as it is right now

sbuettner (approved):

PR #1418, opened 2023-11-15
Made zeebe job timeout configurable via annotation and env

### LINKED ISSUE #1417: Allow custom Zeebe Job Timeout on connector

Some connectors might require a custom zeebe job timeout. Currently, it is only possible to set the job timeout for the zeebe client (which will affect all connectors of a runtime)

I would like to be able to control the job timeout per connector in a bundled runtime.

Build a separate runtime for the connector that requires a custom timeout and use it as hybrid connector

__Additional context__

Feature Request: https://jira.camunda.com/browse/SUPPORT-19266

jonathanlukas:
## Description

A connector might require a custom zeebe job timeout. This can be configured on the connector now

## Related issues

closes #1417

### DIFF
--- a/connector-runtime/connector-runtime-core/README.md
+++ b/connector-runtime/connector-runtime-core/README.md
@@ -58,6 +58,7 @@ Outbound Connector configuration example:
 CONNECTOR_SLACK_FUNCTION=io.camunda.connector.runtime.util.outbound.SlackFunction
 CONNECTOR_SLACK_TYPE=io.camunda.connector:SLACK
 CONNECTOR_SLACK_INPUT_VARIABLES=foo,bar
+CONNECTOR_SLACK_TIMEOUT=10000 # optional
 ```
 
 Inbound Connector configuration example:
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/AnnotatedFunction.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/AnnotatedFunction.java
@@ -23,7 +23,8 @@ import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 @OutboundConnector(
     name = "ANNOTATED",
     inputVariables = {"a", "b"},
-    type = "io.camunda:annotated")
+    type = "io.camunda:annotated",
+    timeout = 20000)
 public class AnnotatedFunction implements OutboundConnectorFunction {
 
   @Override
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/AnnotatedFunction.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/AnnotatedFunction.java
@@ -23,8 +23,7 @@ import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 @OutboundConnector(
     name = "ANNOTATED",
     inputVariables = {"a", "b"},
-    type = "io.camunda:annotated",
-    timeout = 20000)
+    type = "io.camunda:annotated")
 public class AnnotatedFunction implements OutboundConnectorFunction {
 
   @Override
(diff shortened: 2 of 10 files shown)


connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/AnnotatedFunction.java
```diff
@@ -23,7 +23,8 @@
 @OutboundConnector(
     name = "ANNOTATED",
     inputVariables = {"a", "b"},
-    type = "io.camunda:annotated")
+    type = "io.camunda:annotated",
+    timeout = 20000)
```
sbuettner, line 27: Can you incorporate this change without changing the annotation? I would like to avoid runtime configuration in the code and rather make use the application config (environment).

jonathanlukas: I could. However, I was thinking whether it could make sense to have the timeout configurable somewhere on the connector side.

Maybe by rephrasing it from timeout to duration or invocationDuration?

sbuettner: I think we should separate the runtime behavior form the code and therefore not introduce another field in the annotation.

jonathanlukas: Alright. In this case, the only way to configure the timeout will be via env variable

jonathanlukas:
As follow up task, the documentation should be extended as well

connector-runtime/connector-runtime-core/README.md
chillleader, line 1: Maybe it would also be nice to allow overriding the timeout for connectors defined using SPI discovery. It seems that currently if I want to define the timeout, I have to register the whole connector via env vars. We could also allow overriding it for SPI-registered connectors in a similar way to how we can override the connector type here:
https://github.com/camunda/connectors/blob/main/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java#L40-L46

This is an optional suggestion, feel free to disregard or submit a follow-up task. I'm good with the current solution too 👍

igpetrov (approved):
LGTM. Just wondering, whats the reasoning behind making that connector-level config, and not a runtime? Was there a specific customer request?

connector-runtime/connector-runtime-core/README.md
jonathanlukas: When trying out hybrid connector runtimes, it seemed that overriding only one attribute will make all others default to the values from the annotation. Example:

```
CONNECTOR_HTTP_REST_TYPE='my-crazy-job-type'
```

would only change the job type, everything else would default to the values from the HTTP REST connector.

So, configuring 

```
CONNECTOR_HTTP_REST_TIMEOUT='60000'
```

would register the default HTTP REST connector with the custom timeout.
Is that the expected behviour?

jonathanlukas:
The reasoning is creating a workaround for running a hybrid connector that is motivated in a higher zeebe job timeout (see linked FR in issue).

connector-runtime/connector-runtime-core/README.md
chillleader: Yes, it makes sense as this was designed specifically for overriding the default connector type. Maybe @igpetrov can confirm.

igpetrov: @jonathanlukas  I don't think, we currently support overriding custom properties except `CONNECTOR_*_TYPE`. Were you able to confirm that `CONNECTOR_HTTP_REST_TIMEOUT=XXX` works?

jonathanlukas: There is a test in place that asserts that the timeout is applied on a pure env discovery. If you could point me to the place where the `CONNECTOR_*_TYPE` is applied to the SPI discovery, that would be helpful.

sbuettner: @jonathanlukas https://github.com/camunda/connectors/blob/d213c8944e980f7ce47973e6fcf89ee9aad6d7a6/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java#L39

jonathanlukas: Done 👍 I have been there at an earlier stage, but I removed the code again. Thank you!

sbuettner (approved):
As always, great contribution. Thanks for considering the feedback.

Issue #1411, opened 2023-11-13
Enhance inbound connector health report


markfarkas-camunda:
__Is your feature request related to a problem? Please describe.__

Part of https://github.com/camunda/product-hub/issues/1827

Enhance the inbound connector health report to show more details about the inbound connector state in the web-modeler.

__Describe the solution you'd like__

Improve the `/inbound` endpoint to return with a list of key value pairs in the **details** in addition to the status of the connector.

Add an **additional endpoint** which will provide the **last activities** happened in connection with the connector. No persistent storage is necessary, we can store these information in memory. A minimum of datetime, title, description is necessary.

The creator of the connector will be responsible to provide these data, by calling the specific methods. 

__Additional context__

https://github.com/camunda/product-hub/issues/1827

__PR__

https://github.com/camunda/connectors/pull/1420

PR #1420, opened 2023-11-15
feat(health report): extend health report with error object

markfarkas-camunda:
## Description

Context: https://github.com/camunda/product-hub/issues/1827 

This PR adds a new property called `error` to the Health object to make more detailed error reports possible. It also changes how we handle state changes: with this PR we won't always create a new Health object but merge with the existing one.

These is also a new endpoint `/inbound/*/*/*/logs` which lets you fetch the activity logs.

You can add to the activity log queue(size limited) any time by using the example code below:
```
    inboundConnectorContext.log(
        ActivityLog.level(Severity.INFO)
            .tag("Your tag")
            .message("Your message "));
```

## Related issues
https://github.com/camunda/connectors/issues/1411

closes #
https://github.com/camunda/connectors/issues/1411


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -94,14 +99,28 @@ public InboundConnectorDefinition getDefinition() {
 
   @Override
   public void reportHealth(Health health) {
-    this.health = health;
+    if (this.health == null) {
+      this.health = health;
+      return;
+    }
+    this.health = this.health.merge(health);
```
sbuettner, line 106: Can we use and if/else statement here instead of an if combined with an early return?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -94,14 +99,28 @@ public InboundConnectorDefinition getDefinition() {
...
 
+  @Override
+  public void log(ActivityLog log) {
+    this.logs.add(log);
+  }
+
+  @Override
+  public Queue<ActivityLog> getLogs() {
+    return this.logs;
+  }
```
sbuettner, line 122: Same with `getHealth()` should we really expose it to the clients?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
```diff
@@ -47,34 +45,44 @@ public List<ActiveInboundConnectorResponse> getActiveInboundConnectors(
...
+  BiPredicate<ActiveInboundConnector, String> tenantIdMatch =
+      (connector, tenantId) -> {
+        var definition = connector.context().getDefinition();
+        return tenantId != null && tenantId.equals(definition.tenantId());
+      };
+
+  private Queue<ActivityLog> mapToInboundLogsResponse(ActiveInboundConnector connector) {
+    var logs = connector.context().getLogs();
+    return logs;
+  }
```
sbuettner, line 73: Not sure whether this requires a dedicated method

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
```diff
@@ -39,12 +59,18 @@ public Map<String, Object> getDetails() {
     return details;
   }
 
+  public HealthError getError() {
+    return error;
+  }
+
```
sbuettner, line 65: I thought we return the error as part of the details?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/HealthError.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+public record HealthError(Severity severity, String text, String log, String title) {}
```
sbuettner, line 19: Why should we separate between `text` and `log`? Do we need a `title`?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Severity.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+public enum Severity {
+  DEBUG,
+  INFO,
+  WARNING,
+  ERROR,
+  FATAL
```
sbuettner, line 24: What is there difference between these too?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
markfarkas-camunda: @sbuettner Sure, but I think the early return can reduce complexity. In general in this kind of case, if in the future we'll add a new condition inside the else branch that would result in nested if-else statements which I would avoid. If we implement this now with early return, we make it easier to eliminate nested if-else statements in the future. I'll gladly rewrite this, but please elaborate on why you think that's better.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/HealthError.java
markfarkas-camunda: My initial thought was having  a short version and longer version of information. In text we could add something like: NullPointerException. In the logs we would add something like the stacktrace. In this way, it would make sense to rename these to title, and stacktrace, but I would keep it as it is, since on the FE it would be beneficial to show the short version while having the possibility to show the longer version if requested. We might not need the title. @sbuettner Wdyt?

sbuettner: @markfarkas-camunda I thought we only need one string and support expanding it on the client side.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
```diff
@@ -102,20 +133,51 @@ public static Health down(Throwable ex) {
...
+  public Health merge(Health newHealth) {
+    Map<String, Object> mergedDetails =
+        this.getDetails() != null ? this.getDetails() : new HashMap<>();
+    if (newHealth.getDetails() != null && !newHealth.getDetails().isEmpty()) {
+      mergedDetails.putAll(newHealth.getDetails());
+    }
+    return Health.status(newHealth.getStatus()).details(mergedDetails).error(newHealth.getError());
+  }
```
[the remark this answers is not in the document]
markfarkas-camunda: Right now we have only 1 single Heath instance per inbound connector instance. If we would like to add a new details for example, than we would like to keep the original details(e.g.: we already had the path, but we also want to add the authentication type just to show it there), and only append the new one instead of overriding it completely. Hence the name of the method `merge`.
Better example is when you wanna change the status attribute of the Health object. We always have to show the details(e.g. path for webhook), so we cannot override that, so merging the information is our only option.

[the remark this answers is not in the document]
sbuettner: @markfarkas-camunda So should we maybe rather have a merge for properties only?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
markfarkas-camunda: You are right, will remove it.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
markfarkas-camunda: I agree with your thinking, let's keep the interface as slim as possible. But how can we expose this to ourselves without creating unnecessary complexity?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
```diff
@@ -47,34 +45,44 @@ public List<ActiveInboundConnectorResponse> getActiveInboundConnectors(
...
   }
 
-  private ActiveInboundConnectorResponse mapToResponse(ActiveInboundConnector connector) {
+  @GetMapping("/inbound/{bpmnProcessId}/{elementId}/{tenantId}/logs")
+  public List<Queue<ActivityLog>> getActiveInboundConnectorLogs(
+      @PathVariable String bpmnProcessId,
+      @PathVariable String elementId,
+      @PathVariable String tenantId) {
+    var result =
+        inboundManager.query(new ActiveInboundConnectorQuery(bpmnProcessId, elementId, null));
```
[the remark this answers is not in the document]
markfarkas-camunda: I do not have anything against that, my initial thought was to avoid adding more parameters to the ActiveInboundConnectorQuery contructor. Will add the `tenantId`.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
sbuettner: We could add a `InboundConnectorReportingContext` interface to get access to this data points but keep it out of the SDK for example.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/HealthError.java
markfarkas-camunda: @sbuettner I thought we need at least 2 strings, since in the design I can see that there is a title written in bold, so we definitely have to differentiate between that and the longer text below it. If we only have 1 text, we have no chance to write anything in bold. So I would say we need 2 strings, rn I cannot see any other way. WDYT?

image

markfarkas-camunda: Or is this title always just: webhook inactive? If so than I misunderstood and you are right, one text is enough 🙂

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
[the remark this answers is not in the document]
sbuettner: @markfarkas-camunda Sorry, I meant then maybe just allow users to pass details instead of a whole health object.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/HealthError.java
sbuettner: I think we should keep it simple for the first iteration and only have a message

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Severity.java
sbuettner: Yeah, lets keep it simple for now.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
sbuettner: I thought we wanted to check the details for a property named `error`. If its there the modeler renders it as an error.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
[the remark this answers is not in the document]
sbuettner: But we query later on anyhow so why not add it there.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
markfarkas-camunda: @sbuettner based on our previous conversations my assumption was that we agreed on adding an error construct.
Architecture session:
image
Slack:
image
WDYT?

markfarkas-camunda: But maybe I misunderstood it, and it is enough to add this to the details. My issue with that is we then need to be able to effectively remove it, but if we accept that the "error" is a reserved keyword, than we can certainly do that. This way it makes sense, I will add it to the details with the key "error". Sorry for the confusion.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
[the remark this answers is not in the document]
markfarkas-camunda: @sbuettner but somehow they need to change the status, ~and error as well~(this is irrelevant if we will use the details), and they might want to change those at the same time. Do you mean we should have different methods for changing different attributes? So one method for merging details, one method for changing status and maybe one for creating health object? But in this case how could we prevent them to accidentally override the Health object, and with that deleting the details? Since we should handle the "error" detail specially we would also have to have at least one method for changing that.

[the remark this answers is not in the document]
sbuettner: What kind of details should we keep and we cant easily created again?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ActivityLog.java
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.inbound;
+
+import java.time.OffsetDateTime;
+
+public record ActivityLog(Severity severity, String tag, OffsetDateTime timestamp, String message) {
```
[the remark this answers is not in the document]
sbuettner: Good point, lets go with `Activity`.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/Health.java
```diff
@@ -23,8 +23,12 @@
 public class Health {
 
   private final Status status;
+  private final Message message;
+
   private final Map<String, Object> details;
 
+  public record Message(Severity severity, String message) {}
```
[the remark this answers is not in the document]
sbuettner: Good point. So would you just go with a single dedicated property `message` and derive the severity from the `status`?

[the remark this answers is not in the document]
sbuettner: Or would you rather have an optional `error` property without any severity? Could be similar to what we had before:
```
public record Error(code, message)
``` 
and we only make us of that in case of an error. No info or any other severity?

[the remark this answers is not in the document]
chillleader: Yeah, an optional `error` looks reasonable - and since it's a separate class, we will be able to extend it if needed in the future without breaking the API 👍

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -46,6 +49,8 @@ public class InboundConnectorContextImpl extends AbstractConnectorContext
 
   private Health health = Health.unknown();
 
+  private EvictingQueue<Activity> logs = EvictingQueue.create(10);
```
chillleader, line 52: We could make the queue size configurable via a property and allow the client to decide how many log entries to request, to avoid changing the code when we decide to show more or less records than 10.

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/security/SecurityConfiguration.java
```diff
@@ -57,7 +57,10 @@ public WebSecurityCustomizer webSecurityCustomizer() {
   public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
     http.csrf(csrf -> csrf.ignoringRequestMatchers("/inbound/**"))
         .authorizeHttpRequests(
-            auth -> auth.requestMatchers("/inbound").hasAuthority("SCOPE_inbound:read"))
+            auth -> {
+              auth.requestMatchers("/inbound").hasAuthority("SCOPE_inbound:read");
+              auth.requestMatchers("/tenants/**").hasAuthority("SCOPE_inbound:read");
```
chillleader, line 62: This can be a single line: `auth.requestMatchers("/inbound", "/tenants/**").hasAuthority("SCOPE_inbound:read");`

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/HealthError.java
sbuettner: As discussed in meeting with product, we keep a `code` property.

markfarkas-camunda:
After discussing with @YanaSegal , we decided that having a 3rd type of information in the Error record is necessary in order to have all the FE functionalities. We shall call this `log`. We should also rename the `Error` record because its name is too broad. We should rename the `code` attribute to `title` in the Error record to reflect its purpose more clearly.
Adding a limitation for the number of characters for the `tag` attribute in the Activity class may make sense.

markfarkas-camunda:
FYT: @YanaSegal Had a discussion with @bastiankoerber , who came up with a great idea: let's only have `code` and `message` in the `HeathError` object. `code` will be a String, but we will define an enum to have the actual codes for the expected errors (Kafka topic is not found, connection refused etc.). In case of these expected errors we will have a message which will provide some guidance about what a user can do to mitigate the error. In case of an unexpected error (we will have a code for that as well) the message will not contain any guidance for the user, but the actual stack trace. This way we can keep the SDK slim and still provide useful information for all cases.

sbuettner (approved):
Great improvement for our inbound connector management.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundIntermediateConnectorContextImpl.java
```diff
@@ -120,6 +121,16 @@ public void reportHealth(final Health health) {
 
   @Override
   public Health getHealth() {
-    return inboundContext.getHealth();
+    return ((InboundConnectorReportingContext) inboundContext).getHealth();
```
sbuettner, line 124: Could we get rid of the casting if we apply https://github.com/camunda/connectors/pull/1420/files#r1411746859 ?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundIntermediateConnectorContextImpl.java
```diff
@@ -120,6 +121,16 @@ public void reportHealth(final Health health) {
...
+  }
+
+  @Override
+  public void log(Activity log) {
+    inboundContext.log(log);
+  }
+
+  @Override
+  public Queue<Activity> getLogs() {
+    return ((InboundConnectorReportingContext) inboundContext).getLogs();
```
sbuettner, line 134: Same as above

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -106,12 +106,11 @@ public interface InboundConnectorContext {
...
+   * data can be used to report data about processed requests or failures.
    *
-   * <p>Use the {@link #reportHealth(Health)} method to set this information
-   *
-   * @return Health object
+   * <p>This method can be called as often as needed and the internal state of the inbound Connector
+   * implementation requires it.
    */
-  Health getHealth();
+  void log(Activity log);
```
sbuettner, line 115:
```suggestion
  void log(Activity activity);
```

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/task/HttpRequestTask.java
```diff
@@ -33,6 +41,16 @@ public void run() {
...
           HttpCommonResult httpResponse = httpService.executeConnectorRequest(httpRequest);
           processInstanceContext.correlate(httpResponse);
+          this.context.log(
+              Activity.level(Severity.INFO)
+                  .tag(httpRequest.getMethod().toString())
+                  .message(
+                      "Polled url: "
+                          + httpRequest.getUrl()
+                          + ", body: "
+                          + httpRequest.getBody()
```
sbuettner, line 51: This could leak sensitive information. Lets skip it for now.

sbuettner (approved):
Thanks for incorporating the feedback. LGTM!

PR #1426, opened 2023-11-16
feat(inbound): re-introduce correlation result

### LINKED ISSUE #1337: Broken API of `  InboundConnectorContext#correlate(Object variables)  `
`  InboundConnectorContext#correlate(Object variables) 
`  does not return `CorrelationResult` even though the `InboundCorrelationHandler` still does. Due to this code change between 0.22 and 8.3.0 it is not possible to know whether the correlation was successful or whether the activation condition has not been met and therefore it is not possible to nack the message so it is available for processing somewhere else. 

I'd expect the Correlation result still to be used as otherwise the class does not need to be there at all.

chillleader:
## Description

This PR adds a new method, `correlateWIthResult` (naming to be discussed but the current name is a mirror of a [similar interface](https://docs.camunda.org/javadoc/camunda-bpm-platform/7.6/org/camunda/bpm/engine/runtime/MessageCorrelationBuilder.html#correlateWithResult()) in C7.

Unfortunately, we can't simply change the return type of the existing method: this would require to recompile all existing inbond connectors, including external developers.

## Related issues

closes #1337

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
@@ -22,7 +22,9 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.connector.api.error.ConnectorException;
 import io.camunda.connector.api.error.ConnectorInputException;
 import io.camunda.connector.api.inbound.CorrelationResult;
-import io.camunda.connector.api.inbound.CorrelationResult.ErrorCode;
+import io.camunda.connector.api.inbound.CorrelationResult.Failure.ActivationConditionNotMet;
+import io.camunda.connector.api.inbound.CorrelationResult.Failure.Other;
+import io.camunda.connector.api.inbound.CorrelationResult.Success;
 import io.camunda.connector.api.inbound.Health;
 import io.camunda.connector.api.inbound.InboundConnectorContext;
 import io.camunda.connector.api.inbound.InboundConnectorDefinition;
@@ -68,25 +70,19 @@ public class InboundConnectorContextImpl extends AbstractConnectorContext
   @Override
   public void correlate(Object variables) {
     var result = correlationHandler.correlate(definition, variables);
-    if (result instanceof CorrelationResult.Failure failure
-        && ErrorCode.ACTIVATION_CONDITION_NOT_MET != failure.code()) {
-
-      if (failure.error() == null) {
-        throw new ConnectorException(
-            String.format(
-                "Failed to correlate inbound event: %s, error code: %s",
-                failure.code(), failure.message()));
-      }
-      if (failure.error() instanceof ConnectorException
-          || failure.error() instanceof ConnectorInputException) {
-        throw (RuntimeException) failure.error();
-      }
-      throw new ConnectorException(
-          String.format(
-              "Failed to correlate inbound event: %s, error code: %s",
-              failure.code(), failure.message()),
-          failure.error());
+    if (result == null) {
+      throw new ConnectorException("Failed to correlate inbound event, result is null");
     }
+    if (result instanceof ActivationConditionNotMet || result instanceof Success) {
+      return;
+    }
+    if (result instanceof CorrelationResult.Failure.InvalidInput invalidInput) {
+      throw new ConnectorInputException(invalidInput.message(), invalidInput.error());
+    }
+    if (result instanceof Other exception) {
+      throw new ConnectorException(exception.error());
+    }
+    throw new ConnectorException("Failed to correlate inbound event, details: " + result);
   }
 
   @Override
@@ -95,7 +91,7 @@ public class InboundConnectorContextImpl extends AbstractConnectorContext
       return correlationHandler.correlate(definition, variables);
     } catch (Exception e) {
       LOG.error("Failed to correlate inbound event", e);
-      return new CorrelationResult.Failure(ErrorCode.UNKNOWN, e.getMessage(), e);
+      return new CorrelationResult.Failure.Other(e);
     }
   }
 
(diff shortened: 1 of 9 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -66,6 +67,11 @@ public void correlate(Object variables) {
     correlationHandler.correlate(definition, variables);
   }
 
+  @Override
+  public CorrelationResult correlateWithResult(Object variables) {
+    return correlationHandler.correlate(definition, variables);
+  }
```
sbuettner, line 73: I think we can return a `CorrelationResult` for the existing code without breaking any contracts right?

chillleader: On the source code level - yeah, but the resulting byte code on the caller side will become incompatible.

I tested it with a simpler example by compiling the java class, compiling another class where it was used, then replacing the first class without recompiling the second one - it throws a `NoSuchMethodException` even though I only changed the return type.

So this will make all connectors that are already compiled incompatible with the new runtime.

sbuettner: Did you change the return type from `void` to a type or did you change the type lets say from `int` to `String`?

chillleader: From void to a type

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationResult.java
```diff
@@ -0,0 +1,35 @@
...
+    private final boolean isCorrelated;
+
+    public boolean isCorrelated() {
+      return isCorrelated;
+    }
+
+    CorrelationResultCode(boolean isCorrelated) {
+      this.isCorrelated = isCorrelated;
+    }
+  }
```
sbuettner, line 34: What about a sealed `CorrelationResult` class and `CorrelationResult.Success` and `CorrelationResult.Failed(reason)` subclasses?

chillleader: Would this also mean returning e.g. Zeebe connection errors in the response? (in other words would this method then never throw exceptions?)

sbuettner: Yes, that was the idea. As I think it would be cumbersome to deal with the result object in addition to exceptions. WDYT?

chillleader: Agreed, done

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -27,23 +27,28 @@ public interface InboundConnectorContext {
   /**
    * Correlates the inbound event to the matching process definition
    *
+   * @deprecated since 8.4. Use {@link #correlateWithResult(Object)} instead.
+   */
+  @Deprecated(since = "8.4")
+  void correlate(Object variables);
```
sbuettner, line 33: So sad there is no easy way to replace it 😄

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationResult.java
```diff
@@ -0,0 +1,45 @@
...
+    public static final Success INSTANCE = new Success();
+  }
+
+  record Failure(ErrorCode code, String message, Throwable error) implements CorrelationResult {}
+
+  enum ErrorCode {
+    INVALID_INPUT(false),
+    ACTIVATION_CONDITION_NOT_MET(false),
+    FAULT_ZEEBE_CLIENT_STATUS(true),
+    FAULT_IDEMPOTENCY_KEY(false),
```
sbuettner, line 32: Should `FAULT_IDEMPOTENCY_KEY` indicate an already existing message?

chillleader: No, I think it's for when we can't extract the idempotency key from the payload (i.e. no message id / message id expression)

sbuettner: Shouldnt this rather be an INVALID_INPUT error then?

sbuettner: And another one: We return a success in case of an already existing message right?

chillleader: No, as far as I know this would raise a ClientStatusException as the message would be rejected ([src](https://docs.camunda.io/docs/components/concepts/messages/#message-uniqueness)).

I'm ok with removing this error code - will do. I just copied it as it was initially.

sbuettner: Yes, we have to. What about a dedicated class `Correlation.Failure.Zeebe` that allows to inspect the actual zeebe correlation error?

sbuettner: We might also want to be able to provide access to the exception for `Unknown` WDYT?

sbuettner: We could offer `Correlation.Failure.Exception(e)`

chillleader: I would name it `Correlation.Failure.Other` - Exception is probably not the best name as it conflicts with the standard Java Exception class.

chillleader: Reworked the sealed hierarchy as discussed

sbuettner (approved):

PR #1427, opened 2023-11-16
fix(generator): various OpenAPI generator bugfixes

chillleader:
## Description

This PR fixes several issues discovered while testing the OpenAPI generator.

| Issue description | Fix description |
|---|---|
| Request body could not be parsed when it was provided as a reference to the `Components` object  | Fetch body from components if ref is provided, see `OperationUtil::extractBodyExample`  |
| Inconsistent operation naming (sometimes method + path, sometimes description) | Use only method + path naming schema, as descriptions are inconsistent among different API providers |
| Examples for string parameters are falsely enclosed in quotes  | Fixed in `OperationUtil::extractBodyExample` |
| If path param contains a dash, it is treated by the FEEL engine as a minus sign  | Replace all dashes in path parameters with underscores, see `OperationUtil::extractPath` |
| Enum param values are not displayed in the property example value | If property is an enum array, we include all enum values as examples. If it is a single enum value, it is included as a dropdown (`generateFakeDataStringFromSchema`) |
| Parameters named like this: `operationId_path` conflict with our internal properties that have the same name  | Replaced the internal property name `operationId_path` with `operationId_$path` to make the chance of such conflicts lower |
| Huge YAML files can't be deserialized | Added a system prop override, see https://github.com/swagger-api/swagger-parser/issues/1857 | 

Additionally, with these changes I moved most of properties shading / processing logic one level up, to the `PropertyUtil` in HTTP DSL. This makes the HTTP DSL cleaner and avoid some data redundancies.

## Related issues

### DIFF
--- a/element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpOperation.java
+++ b/element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpOperation.java
@@ -16,7 +16,6 @@
  */
 package io.camunda.connector.generator.dsl.http;
 
-import io.camunda.connector.generator.dsl.Property;
 import io.camunda.connector.http.base.model.HttpMethod;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


igpetrov (approved):
Awesome, thank you! Eager to try this out.

sbuettner (approved):
Great improvements.

PR #1442, opened 2023-11-21
fix(rest): support lowercase http method names in element template

chillleader:
## Description

After replacing the REST connector template with an auto generated one, the backwards compatibility was broken by changing the method name case. This lead to condition issues for users upgrading their already applied old versions of the template.

## Related issues

related to https://github.com/camunda/team-connectors/issues/552


igpetrov (approved):

chillleader:
Closing this as we discovered a different root cause of the observed issues..

PR #1445, opened 2023-11-22
Create incident from error expression

### LINKED ISSUE #1383: Support a `technicalError` FEEL function in error expressions

The Connectors runtime supports a `bpmnError()` FEEL function that can be used in error expressions to create a `newThrowErrorCommand` to raise a business error. 

There is the need to raise technical errors based on valid responses as some APIs return technical errors as part of a valid API response.

Example: A JSON API that returns a HTTP 200 (Ok) response which contains technical error information that the request was invalid instead of a 400 (Bad Request). The error is inside the body and needs to be extracted and raised manually via an error expression.

An additional FEEL function `technicalError()` that leads to a [`FailJob`](https://docs.camunda.io/docs/apis-tools/grpc/#failjob-rpc) RPC call would solve allow users to raise technical errors for valid connector responses.

* Support `FailJob` parameters: 
  * `errorMessage` (Optional)
  * `retries` (the retries that are left)
  * `retryBackoff` (retryBackOff in ms)
  * `variables`(Optional) passed as a FEEL `context` object

__Additional context__

This feature is relevant in the context of [technical error handling](https://docs.camunda.io/docs/components/best-practices/development/dealing-with-problems-and-exceptions/) in Camunda 8.

### LINKED ISSUE #1444: Create Incident from connector via error handling expression

Currently, an incident is created if the connector throws a technical error (for example http status code >= 400)

Some connectors or their implementation might require the creation of incidents as technical errors might be part of a response and not be generic.

In the error handling expression, make a function `incident()` available that allow to provide an incident description.
The function should only accept an error message and the retries will be 0 on first invocation.

Each connector that has alternative incident capabilities would require its own implementation of an incident condition.

__Additional context__

Discussion with connectors team and the customer

jonathanlukas:
## Description

Currently, the error expression can only handle the creation of `bpmnError`. This PR enables creation of `incident`.
## Related issues

closes #1444 
closes #1383

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
@@ -26,6 +26,6 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
 @JsonTypeInfo(use = Id.NAME, property = ERROR_TYPE_PROPERTY)
 @JsonSubTypes({
   @Type(value = BpmnError.class, name = BPMN_ERROR_TYPE_VALUE),
-  @Type(value = Incident.class, name = INCIDENT_TYPE_VALUE)
+  @Type(value = FailJob.class, name = FAIL_JOB_TYPE_VALUE)
 })
-public sealed interface ConnectorError permits BpmnError, Incident {}
+public sealed interface ConnectorError permits BpmnError, FailJob {}
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
@@ -26,6 +26,6 @@ import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
 @JsonTypeInfo(use = Id.NAME, property = ERROR_TYPE_PROPERTY)
 @JsonSubTypes({
   @Type(value = BpmnError.class, name = BPMN_ERROR_TYPE_VALUE),
-  @Type(value = FailJob.class, name = FAIL_JOB_TYPE_VALUE)
+  @Type(value = JobError.class, name = JOB_ERROR_TYPE_VALUE)
 })
-public sealed interface ConnectorError permits BpmnError, FailJob {}
+public sealed interface ConnectorError permits BpmnError, JobError {}
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
@@ -14,25 +14,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package io.camunda.connector.runtime.core.error;
 
-import java.util.Collections;
-import java.util.Map;
+import static io.camunda.connector.feel.FeelConnectorFunctionProvider.*;
 
-/**
- * Container record for BPMN error data. This is used to indicate when a BPMN error should be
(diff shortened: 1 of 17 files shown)


connector-sdk/feel-wrapper/src/main/java/io/camunda/connector/feel/FeelConnectorFunctionProvider.java
```diff
@@ -55,32 +61,39 @@ public class FeelConnectorFunctionProvider extends JavaFunctionProvider {
...
-  public List<JavaFunction> resolveFunctions(String functionName) {
-    return functions.getOrDefault(functionName, Collections.emptyList());
-  }
+  // Incident
+  private static final String INCIDENT_FUNCTION_NAME = "incident";
+  private static final List<String> INCIDENT_FUNCTION_ARGUMENTS = List.of("message");
+  private static final JavaFunction INCIDENT_FUNCTION =
+      new JavaFunction(
+          INCIDENT_FUNCTION_ARGUMENTS,
+          args ->
```
sbuettner, line 81: We wanted to go with `technicalError` or `jobError` as it will not immediately lead to an incident. Only after failing the retries. We also wanted to support variables as described in: https://github.com/camunda/connectors/issues/1383

sbuettner: or just `failJob` wiith optional arguments.

jonathanlukas: The function is now called `failJob`

jonathanlukas:
@sbuettner One challenge is open: To provide a reasonable `retries` value, one needs to know the current retries. I saw that the `ConnectorResult.responseValue()` is used as context for the error handling expression, but I not sure where to inject the retries (or maybe the whole job context).

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/ConnectorError.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+
+@JsonTypeInfo(use = Id.NAME, property = ERROR_TYPE_PROPERTY)
+@JsonSubTypes({
+  @Type(value = BpmnError.class, name = BPMN_ERROR_TYPE_VALUE),
+  @Type(value = FailJob.class, name = FAIL_JOB_TYPE_VALUE)
```
sbuettner, line 29: Thinking about this: `JobError` would maybe a better matching name as `BpmnError` is a type of error and `FailJob` is more like a command.

sbuettner:
@jonathanlukas Great question. I just checked the source of the `FeelEngineWrapper`. It internally transforms the `final Object variables` to a `scala.collection.immutable.Map<String, Object>`  during the evaluation:
```
var variablesAsMap = ensureVariablesMap(variables);
var variablesAsMapAsScalaMap = toScalaMap(variablesAsMap);
```

We could allow an optional and additional context that gets merged into `Map` before handing it over to the FEEL engine like: 
```
public String evaluateToJson(final String expression, final Object variables, final Map<String,Object> context)
```

Or we refactor the `FeelEngineWrapper` to accept only a `final Map<String,Object> context` so you could add it to the context before the evaluation.

jonathanlukas:
@sbuettner I made the context in the feelEngineWrapper varArgs of objects that are merged to a single context.

Pro: I can add as many contexts as I want to a single expression

Con: The contexts can override each other, the last one wins

wdyt?

sbuettner (approved):
Great work @jonathanlukas! Lets make it part of 8.4 😄

jonathanlukas:
@sbuettner is there anything I need to do for this?

PR #1455, opened 2023-11-27
feat: use identity to authenticate to operate

dlavrenuek:
## Description

This work is part of the https://github.com/camunda/product-hub/issues/1849 epic.

- Adds `identity-spring-boot-starter` for auto-configuration of `Identity` bean (Identity SDK), that supports generic OpenIDC and Azure authentication
- Injects Identity authorization handling into Operate client with minimal destructivity 💥 (via proxy on operate properties class)
- Only uses Identity authorization if it is configured properly. If run with the same env variables, the current behavior should not change

Please feel free to add tests based on your test strategy
 
## Related issues

Based on https://github.com/camunda-cloud/identity/pull/2379 (the PR in Identity has to be merged first, in order for the new helper method to be available)
Closes https://github.com/camunda-cloud/identity/issues/2341

### DIFF
--- a/connector-runtime/spring-boot-starter-camunda-connectors/pom.xml
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/pom.xml
@@ -47,6 +47,10 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>identity-spring-boot-starter</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/InboundConnectorsAutoConfiguration.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/InboundConnectorsAutoConfiguration.java
@@ -69,28 +69,28 @@ public class InboundConnectorsAutoConfiguration {
   }
 
   private OperateClientConfigurationProperties operatePropertiesProxy(
-      OperateClientConfigurationProperties properties,
-      Identity identity) {
-    return (OperateClientConfigurationProperties) Proxy.newProxyInstance(
-        OperateClientConfigurationProperties.class.getClassLoader(),
-        new Class[] {OperateClientConfigurationProperties.class},
-        (proxy, method, args) -> {
-          try {
-            if (method.getReturnType().equals(AuthInterface.class)
-                && identity.authentication().isAvailable()) {
-              return new IdentityAuth(identity, operateAudience);
-            }
+      OperateClientConfigurationProperties properties, Identity identity) {
+    return (OperateClientConfigurationProperties)
+        Proxy.newProxyInstance(
+            OperateClientConfigurationProperties.class.getClassLoader(),
+            new Class[] {OperateClientConfigurationProperties.class},
+            (proxy, method, args) -> {
+              try {
+                if (method.getReturnType().equals(AuthInterface.class)
+                    && identity.authentication().isAvailable()) {
+                  return new IdentityAuth(identity, operateAudience);
+                }
 
-            return method.invoke(properties, args);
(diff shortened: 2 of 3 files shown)


sbuettner:
Thank you @dlavrenuek for the support. I assume we continue with https://github.com/camunda/connectors/pull/1471 for now right?

1nb0und:
@dlavrenuek 
1. Can we hold off for this change? Reason is the `operate` dependency being reference on this PR will be deprecated.

The newer `operate` dependency has now been absorbed and part of `main` in [spring-zeebe](https://github.com/camunda-community-hub/spring-zeebe/tree/main/camunda-sdk-java/java-client-operate)

So even if we merge this PR, it will be deprecated and needs rework soon. M

2. Should the identity-sdk be used in `connectors` ? or in `spring-zeebe` ? 
3. In spring-zeebe, i introduced [support](https://github.com/camunda-community-hub/spring-zeebe/pull/533) for `scope`. As soon as there is a snapshot or 8.4.0 rc available, we can test those changes.
4. Spring-zeebe can incorporate the identity SDK at 8.4 / January release.

dlavrenuek:
@sbuettner to be honest, I can't see how the changes from these two PRs overlap (wich my limited knowledge about Connector's code base) but I have the feeling that my Operate authentication hack could be too "hacky" considering the PR for using the new Operate client. Should be ignore my changes for alpha2 and review what needs to be done for final 8.4.0?

dlavrenuek:
@1nb0und The goal is to use authentication logic from `identity-sdk` (or rather `identity-spring-boot-starter`) when connecting to Operate (via the Operate client). In my opinion it would be best to include the `identity-spring-boot-starter` in spring-zeebe and use directly there. The code changes should be pretty simple, a new Auth class ([here I have used `IdentityAuth`](https://github.com/camunda/connectors/pull/1455/files#diff-fb20c048f64a19f7f9d2d7b64f7d59dbe0789d1a3ba4c0e15354c32ff319d3c1R91)) and [a condition based on Identity](https://github.com/camunda/connectors/pull/1455/files#diff-fb20c048f64a19f7f9d2d7b64f7d59dbe0789d1a3ba4c0e15354c32ff319d3c1R80) instead of own variables:

```java
if  (identity.authentication().isAvailable()) {
    return new IdentityAuth(identity, operateAudience);
}
```

One missing piece (not sure if is is already included in the operate client) is the `opereateAudience`. I also see it in the client itself rather than in connectors or Identity.

1nb0und:
@dlavrenuek 

Question on this PR. I'm still grasping things but I see there could be an issue with integration whether thru here in connectors or in Spring-zeebe

Here in connectors, we defined Operate credential properties in two ways

Option 1
```
zeebe.cloud.client.client-id=xxx
zeebe.cloud.client.client-secret=xxx
```

Option 2
```
camunda.operate.client.client-id=xxx
camunda.operate.client.client-secret=xxx
```

And I was reading the Identity SDK and it seems we need to require credentials in a different way

```
camunda.identity.client-id=xxx
camunda.identity.client-secret=xxx
```

## Question 1

Do we expect connectors or `Spring-zeebe` to use your `camunda.identity.*` scheme? If yes, this could be a breaking change and we might not have enough time since we need to update the readmes here, and the consequent docs and any other references. 

## Question 2

`Spring-zeebe` uses multiple clients (zeebe, operate, optimize, tasklist, modeler, etc). AFAIK, Identity SDK is just limited to using a single credential. My concern would be what if the customer decides to use different credentials across different modules. How would Identity SDK adopt to that? I would solve it maybe not in Identity code but on `Spring-zeebe` by assigning different `Identity` instances per client and go from there (not sure how on the top of my head but it sounds possible).

Happy to hear your thoughts and please correct my understanding if I'm mistaken.

sbuettner:
@1nb0und Can we use `camunda.identity.client-id` (and secret) as a fallback value for all clients in case client specific config is not provided?

sbuettner:
and BTW @1nb0und: ```zeebe.cloud.client.client-id=xxx``` were always also used for the Operate Client in case the
Operate Client credentials were not set.

sbuettner:
@dlavrenuek @1nb0und As we decided to incorporate the changes into the `spring-zeebe` project as part of the Camunda SDK Java initiative I will close the PR here.

PR #1471, opened 2023-11-30
feat(operate): try to integrate the new operate

1nb0und:
## Description

Spring-Zeebe has absorbed Operate as it moves to Camunda Spring SDK. Trying to integrate the new changes here

- use new Operate submodule in spring-zeebe
- refactor operate client properties from `camunda.operate.client.x` to `operate.client.x`

## Related issues

closes # No issue


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/operate/OperateClientAdapterImpl.java
```diff
@@ -98,7 +99,7 @@ public List<FlowNodeInstance> fetchActiveProcessInstanceKeyByDefinitionKeyAndEle
       return result.stream()
           .map(
               node ->
-                  new FlowNodeInstance(
+                  new io.camunda.connector.runtime.core.inbound.FlowNodeInstance(
```
1nb0und, line 102: Can we use directly the new class [FlowNodeInstance](https://github.com/camunda-community-hub/spring-zeebe/blob/main/camunda-sdk-java/java-client-operate/src/main/java/io/camunda/operate/model/FlowNodeInstance.java) here? (so that we can just use 1 class)

It has the same name and same fields now. Thoughts?

sbuettner: Yes, of course.

chillleader (approved):
LGTM, thanks for taking care of this @1nb0und!

chillleader:
⚠️ To support the new property structure, we will need to make changes in our SaaS configuration and update the documentation

1nb0und:
@chillleader @sbuettner 
I made some adjustments to this PR:
- bump to spring-zeebe [8.3.4.1](https://github.com/camunda-community-hub/spring-zeebe/releases/tag/8.3.4.1) which should support both old and new Operator properties

So this means, no need to reach out to controller team and other teams to update the documentation, as Spring SDK will support the old scheme. But I have marked it as deprecated. Moving forward we should plan to adopt to the new scheme on coming releases

sbuettner (approved):

PR #1485, opened 2023-12-05
refactor: change dockerfile uid and gid

aabouzaid:
## Description

Using a non-numeric user and group ID in Dockerfile makes it harder to configure in K8s since it's not possible to detect if it's a root or non-root.

So we need to make that change to remove explicit config related to the user and group ID. 

**Please note that:**
- This change should be part of 8.4 and not part of 8.3.x release.
- I've also changed the GID to be the same as the UID (no particular need to have them different).

## Related issues

Related to: https://github.com/camunda/distribution/issues/122


chillleader (approved):
Thank you @aabouzaid!

PR #1491, opened 2023-12-05
feat: openapi generator improvements

chillleader:
## Description

- Adds `list` CLI command to output the available generators and their usage information (because it is not visible in the help page)
- Fix body example corner cases that caused the body example to remain undetected
- New body parsing mode that breaks down the body into individual fields (enabled by default for simple body objects or primitives, can be disabled with a CLI option)
- Support schemas without operation IDs (we generate our own ID in that case)

## Related issues

closes https://github.com/camunda/team-connectors/issues/549


igpetrov (approved):
LGTM

PR #1561, opened 2023-12-11
feat(generator): Enhanced nested properties support

### LINKED ISSUE #1497: Element template generator: support nested discriminator properties

We should support nested discriminator properties. Currently, the conditions created for nested discriminators are incorrect.

Support structures like this one: https://github.com/camunda-consulting/consulting-connectors/blob/feature/soap-connector/keystore-usage/soap-connector/src/main/java/com/camunda/consulting/connector/soap/SoapConnectorInput.java

__Additional context__

Reported by @jonathanlukas here: https://camunda.slack.com/archives/C02JLRNQQ05/p1701424054254949

chillleader:
## Description

This PR introduces enhanced nested properties features for the annotation-based element template generator.

### 1. Support nested sealed hierarchies

Previously it was impossible to wrap sealed hierarchies into sealed hierarchy subtypes. The new solution changes the way discriminator property IDs are treated by the generator: 

### 2. `@NestedProperties` annotation with support for bulk condition.

`@TemplateProperty` annotation was previously used to configure the `addNestedPath` feature (deciding whether path would be prefixed for nested properties). This was semantically incorrect: nested property containers are technically not properties, besides it was confusing that all the other fields of `@TemplateProperty`, like `id` or `label`, didn't cause any effects on the generated template.

The new approach utilizes a special annotation called `NestedProperties` that allows to configure only those options that can be applied to nested hierarchies. To accomodate the request from @jonathanlukas, bulk condition support has also been added to this annotation.

## Related issues

closes #1497

### DIFF
--- a/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
+++ b/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
@@ -28,13 +28,13 @@ import io.camunda.connector.generator.dsl.Property.FeelMode;
 import io.camunda.connector.generator.dsl.PropertyBinding;
 import io.camunda.connector.generator.dsl.PropertyBinding.ZeebeInput;
 import io.camunda.connector.generator.dsl.PropertyBuilder;
-import io.camunda.connector.generator.dsl.PropertyCondition;
+import io.camunda.connector.generator.dsl.PropertyCondition.AllMatch;
 import io.camunda.connector.generator.dsl.PropertyCondition.Equals;
-import io.camunda.connector.generator.dsl.PropertyCondition.OneOf;
 import io.camunda.connector.generator.dsl.PropertyGroup;
 import io.camunda.connector.generator.dsl.PropertyGroup.PropertyGroupBuilder;
 import io.camunda.connector.generator.dsl.StringProperty;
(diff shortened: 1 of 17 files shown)


element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
```diff
@@ -277,11 +278,29 @@ private static List<PropertyBuilder> handleSealedType(Class<?> type) {
...
+                        property.condition(
+                            new AllMatch(
+                                List.of(
+                                    property.getCondition(),
+                                    new Equals(
+                                        discriminatorIdAndName.getKey(),
+                                        subTypeIdAndName.getKey()))));
+                      }
+                    }
+                  })
```
sbuettner, line 303: This and the whole method is quite complex. Maybe we can split it up in the future

sbuettner (approved):

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
chillleader: Added a follow-up: https://github.com/camunda/connectors/issues/1565

I will go ahead and merge this now to unblock the SOAP connector activities.

PR #1591, opened 2023-12-14
fix: remove length constraints on Google Sheets element template fields

### LINKED ISSUE #1049: Increase Maximum Length Value in Google Sheet Connectors

Today, the Google Sheet connector does not allow for more than 100 characters to be parsed. 

Screenshot 2023-08-17 at 11 31 30

When using FEEL expressions together with variable names, the number of characters can easily go over this limit; which in turn limit the number of use cases. 

Optimally, we should not have a limit of the number of characters we can insert in this field, or have it much larger (ie. 5000)

The only work around today is not limit the amount of variables sent or change the names of those variables to be shorter. Both of those solutions degrade the user experience and the maount of use cases this connector provides. 

__Additional context__

See insight: https://camunda.productboard.com/insights/notes/view/1016-all-notes/notes/34959700

ev-codes:
## Description

Remove all length constraints on Google Sheets connector template fields.

The previous limits of 200 and 100 characters on `Value` and `Enter values` fields were preventing some otherwise valid FEEL queries from being used with this template.

The initial length constraints were added early in the Element Templates development process. Now, the team feels that (in general) templates should only apply length constraints when the API itself requires it. This is because:
- We can't always determine limits imposed by 3rd party systems
- 3rd party systems could change limits at any time
- We don't want to unnecessarily limit our customers where those limits aren't strictly needed

Although Camunda has a [4MB process instance payload size limit](https://docs.camunda.io/docs/components/concepts/variables/#variable-size-limitation), we can't easily determine how much each field "should" be allowed to take up. 

I quickly checked whether the [Google Sheets API](https://developers.google.com/sheets/api/reference/rest) has any documented length constraints for the fields this PR edits, but I couldn't find any.

### Testing

- [x] Validation in Desktop Modeler works as expected
    - For each field, check if we can input a value longer than the previous length requirement
 - [x] Search for "maxValues" inside `element-templates/google-sheets-connector.json` to ensure no length constraint left behind

## Related issues

closes #1049 

## Links

- [Camunda process instance payload size limit](https://docs.camunda.io/docs/components/concepts/variables/#variable-size-limitation)
- [Google Sheets REST API](https://developers.google.com/sheets/api/reference/rest)

### DIFF
--- a/connectors/google/google-sheets/element-templates/google-sheets-connector.json
+++ b/connectors/google/google-sheets/element-templates/google-sheets-connector.json
@@ -316,8 +316,7 @@
         "name": "operation.worksheetName"
       },
       "constraints": {
-        "notEmpty": false,
-        "maxLength": 100
+        "notEmpty": false
       },
       "condition": {
         "property": "operationType",
--- a/connectors/google/google-sheets/element-templates/google-sheets-connector.json
+++ b/connectors/google/google-sheets/element-templates/google-sheets-connector.json
@@ -516,7 +516,6 @@
       },
       "constraints": {
         "notEmpty": false,
-        "maxLength": 100,
         "pattern": {
           "value": "^([0-9]*$)",
           "message": "Must be a number."
@@ -540,7 +539,6 @@
       },
       "constraints": {
         "notEmpty": false,
-        "maxLength": 100,
         "pattern": {
           "value": "^([0-9]*$)",
           "message": "Must be a number."
--- a/connectors/google/google-sheets/element-templates/google-sheets-connector.json
+++ b/connectors/google/google-sheets/element-templates/google-sheets-connector.json
@@ -245,8 +245,7 @@
         "name": "operation.spreadsheetName"
       },
       "constraints": {
-        "notEmpty": true,
-        "maxLength": 255
+        "notEmpty": true
       },
       "condition": {
         "property": "operationType",
--- a/connectors/google/google-sheets/element-templates/google-sheets-connector.json
+++ b/connectors/google/google-sheets/element-templates/google-sheets-connector.json
@@ -561,8 +561,7 @@
         "name": "operation.cellId"
       },
       "constraints": {
-        "notEmpty": true,
-        "maxLength": 100
+        "notEmpty": true
       },
       "condition": {
         "property": "operationType",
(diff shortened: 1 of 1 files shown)


markfarkas-camunda:
Nice, detailed, clear description explaining the reasons, thank you @ev-codes . Do you think it would also make sense to modify or even remove the constraints for other properties e.g.:  Worksheet name ?

ev-codes:
@markfarkas-camunda thank you!

I think it makes sense to consider modifying or removing constraints on the other 25 fields.

However, I think that work falls outside the scope of the ticket (#1049) and this PR. I propose instead we create a new ticket to check the rest of the fields (and perhaps fields in other connectors as well).

What do you think about that?

chillleader:
Great start @ev-codes! Nice work exploring the topic and documenting the effort.

- Speaking about technical limitations, the only one I'm aware of is that the maximum variable size should not exceed [4 MB](https://docs.camunda.io/docs/components/concepts/variables/#variable-size-limitation).
- Generalizing, I don't think it makes sense to have these kinds of constraints unless we know that **1)** the field is always a plain string input (not a FEEL expression), and **2)** the underlying system's API has a strict limitation.
  - FEEL expressions can be evaluated to a string of arbitrary size and we shouldn't forget that this constraint would apply to the length of a FEEL expression and not its evaluation result.
  - Finding out about such limitations in third-party systems we integrate with is usually not trivial (there's sometimes no documentation, we have to test on our own, etc.). I don't think it's a good investment of our time, as the added value is usually minimal.

Given this, I would be ok with just removing this length constraint and other similar constraints. Curious what others think.

ev-codes:
Thanks @chillleader! I think that's a fair point.

Because removing constraints (or increasing each constraint to around 3 million characters) will be a lot less effort than researching all 25 of them, I'd be open to doing that for all (or most) the Google Sheets fields in this PR, if the team agrees.

I'll raise our approach to constraints in Monday's sync, as I think we should agree on an approach as a team.

Also: I now realize that for the Create Row "Enter values" field, Google Sheets would ultimately accept inputs that are much longer. This is because the field accepts an array. For example, [50k char string, 50k char string, 50k char string] = 150k+ characters. So regardless, I think we should remove this specific Create Row "Enter values" limit.

igpetrov (approved):
LGTM, awesome! Please don't forget to rebase to 1 commit before merge.

ev-codes:
Thanks for the rebase reminder! Just rebased and pushed up.

PR #1593, opened 2023-12-15
fix(operate): deal with nulls from process definition search

chillleader:
## Description

Handle null values when there are no processes found (previously it returned an empty list and we didn't have this issue)

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionSearch.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionSearch.java
@@ -77,7 +77,7 @@ public class ProcessDefinitionSearch {
       }
 
     } while (processDefinitionResult.getItems() != null
-        && processDefinitionResult.getItems().size() > 0);
+        && !processDefinitionResult.getItems().isEmpty());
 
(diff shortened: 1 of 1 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionSearch.java
```diff
@@ -71,9 +71,13 @@ public void query(Consumer<List<ProcessDefinition>> resultHandler) {
...
 
-      processDefinitions.addAll(processDefinitionResult.getItems());
+      var items = processDefinitionResult.getItems();
+      if (items != null) {
+        processDefinitions.addAll(items);
+      }
 
-    } while (processDefinitionResult.getItems().size() > 0);
+    } while (processDefinitionResult.getItems() != null
+        && processDefinitionResult.getItems().size() > 0);
```
markfarkas-camunda, line 80: 🙃 Just a small note: maybe instead of `processDefinitionResult.getItems().size() > 0` we could also use `!processDefinitionResult.getItems().isEmpty()` if you would like to.

igpetrov (dismissed):

markfarkas-camunda (dismissed):
Just a minor thing, all good 👍

markfarkas-camunda (approved):
LGTM! 👍

Issue #1238, opened 2023-10-06
Webhook ID / path should be validated to only contain characters allowed in URLs


chillleader:
__Describe the Bug__

The current element template (and the connector runtime) doesn't validate the webhook ID (path). The user can deploy a webhook where the webhook ID contains characters that need to be URL-encoded.

__Steps to Reproduce__

1. Link a webhook connector to any start or intermediate event
2. Enter a URL that contains characters that are typically URL-encoded (e.g. whitespace)
3. Ensure no error is displayed

__Expected Behavior__

Characters that must be URL-encoded in order to be part of the URL should be forbidden to prevent unclarity and undefined behavior in the runtime.

The slash symbol is typically used to delimit path parts and it should also be discussed whether we allows to use it.

__Environment__

C8 SaaS 8.3

sbuettner:
Good finding @chillleader. It should also be validated on the server side when extracting the path.

ev-codes:
@sbuettner Thanks for the suggestion!

Where in the code were you thinking the server-side validation should happen?

I'm trying to avoid grouping too many concerns in one place (like in `WebhookConnectorRegistry`), and I also don't want to throw any unrecoverable exceptions for existing connectors that don't pass this validation.

ev-codes:
@sbuettner and I decided I should pick up a more complex ticket that involves more parts of the codebase I haven't worked with yet.

(As discussed in today's Weekly)

I'll close the associated PR and move this ticket back to Ready.

sbuettner:
@ev-codes The `WebhookConnectorRegistry` is a good place for this general validation. We should also reflect it in our element templates. 

For this task we can limit the scope to the element templates of `webhook` connector. A follow-up task should be created to cover other `webhook` based connectors as well.

Lets support the characters mentioned in this blog post: https://blog.sucuri.net/2023/01/bad-paths-the-importance-of-using-valid-url-characters.html but limit it even further:

```
Alphanumeric characters: A-Z, a-z, and 0-9
Hyphen: –
Underscore: _
```

Valid paths should always start with an alphanumeric character and end with one.

Valid:
`hello-world`
`123-456_789`

Invalid:
`-my-path`
`my_path_`

Please create one regular expression that covers this requirement and use it in the runtime as well as in the element templates.

bastiankoerber:
> @ev-codes The `WebhookConnectorRegistry` is a good place for this general validation. We should also reflect it in our element templates.
> 
[...]

PR #1629, opened 2024-01-03
fix: add URI validation to webhook connectors

ev-codes:
## Description

Check that webhook ID contains only valid URI characters as outlined in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) for the following:
- Webhook Start Event Connector
- Webhook Intermediate Event Connector

Before:
Screenshot 2024-01-03 at 16 04 44

After:
Screenshot 2024-01-03 at 16 05 56

Webhook ID's form part of a URL. As such, the ID needs to be a valid URI in order to prevent obviously malformed URL's. See #1238 for more details.

### Allowing "/"

There was discussion in #1238 to disallow the forward slash "/" as this could be misconstrued as a path delimiter.

I personally feel we should still allow "/" because it is still syntactically correct. I don't want to enforce semantics, as it's possible there are some use cases where a Webhook ID could have a "/".

### Test Cases

Checked the following for all connectors in Desktop Modeller:

Should show error message:
- "Start event"
- "Start event€"
- " Start event "

Should not show error message:
- "12345"
- "abcde"
- "\_start_event\_"
- "%20start%20event%20"

### Next Steps

- Other generic webhook connectors (e.g., `webhook-connector-boundary`)
- 3rd party service webhook connectors (e.g., Twilio)

I wanted to wait to get feedback on my first implementation before expanding this approach to many other connectors.

## Related issues

https://github.com/camunda/connectors/issues/1238

### DIFF
--- a/connectors/webhook/element-templates/webhook-connector-intermediate.json
+++ b/connectors/webhook/element-templates/webhook-connector-intermediate.json
@@ -2,7 +2,7 @@
   "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
   "name": "Webhook Intermediate Catch Event Connector",
   "id": "io.camunda.connectors.webhook.WebhookConnectorIntermediate.v1",
-  "version": 5,
+  "version": 6,
   "description": "Configure webhook to receive callbacks",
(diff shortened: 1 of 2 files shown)


ev-codes:
Closing for now while we pause work on #1238. See issue for more details.

PR #1632, opened 2024-01-04
fix(deps) update spring-zeebe to 8.3.4.4

Oleksiivanov:
## Description

Update spring-zeebe version to 8.3.4.4


sbuettner:
@Oleksiivanov I think we should not upgrade it to the spring-zeebe Operate client.

PR #1679, opened 2024-01-12
fix(generator): OpenAPI generator issues with Petstore example

chillleader:
## Description

1. Added a "target" prefix to all property IDs so they look like `operationId_body_propertyName` or `operationId_headers_propertyName`. Sometimes properties in body (derived from the JSON schema) can have the same name as parameters.
2. Fixed an issue with `scan` command where unsupported operations were never displayed
3. Changed the `generate` command output so it returns a single JSON object when only one template was generated, instead of an array of element template JSON objects.
4. Adjusted the README: removed the invalid example, added a note about unsupported operations.

## Related issues

See https://camunda.slack.com/archives/C0350959WG1/p1705053321728369


sbuettner (approved):
Great improvements.

PR #1772, opened 2024-01-22
fix(generator): ConGen - improve help, fix multi-template bugs

chillleader:
## Description

- Improved help command output for configuration parameters
- Fixed a couple newly discovered bugs found while experimenting with generating multiple templates at once for different element types:
  - non-idempotent `FeelContextBuilder.build()`
  - unformatted output when only one element template is returned
  - extra call to `appliesTo()` with the hard-coded value`bpmn:Task`, not needed anymore since we set this value in a different place now.


igpetrov (approved):
LGTM, thank you

PR #1773, opened 2024-01-22
feat(webhook): restrict webhook path/ID on element template, add warn in runtime

ev-codes:
## Description

Two main changes:
1. **Element template:** Introduce new versions of the Webhook element templates that allow only webhook paths/IDs that meet the following requirements:
- Must only contain alphanumeric characters, hyphens, and underscores
- Must start and end with an alphanumeric character

2. **Connectors runtime:** Warn in `WebhookConnectorRegistry` if a webhook path/ID fails the above criteria. We want to allow existing webhook paths/IDs that violate the criteria so we don't break existing connectors.

We're making this change to avoid unexpected issues with running this code on different platforms and with different URL parsing logic. [See related issue](https://github.com/camunda/connectors/issues/1238) for more discussion.

### Element Template Error Message Screenshot

Screenshot 2024-01-24 at 3 42 42 PM

## Related issues

https://github.com/camunda/connectors/issues/1238


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -43,7 +52,14 @@ public boolean isRegistered(ActiveInboundConnector connector) {
   public void register(ActiveInboundConnector connector) {
     var properties = connector.context().bindProperties(CommonWebhookProperties.class);
     var context = properties.getContext();
+
+    logIfWebhookPathDeprecated(connector, context);
+
     var existingEndpoint = activeEndpointsByContext.putIfAbsent(context, connector);
+    logIfEndpointExists(existingEndpoint, context);
```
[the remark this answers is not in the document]
ev-codes: Smart idea! I changed to `checkIfEndpointExists`.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -16,20 +16,29 @@
...
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class WebhookConnectorRegistry {
 
+  private static final String WEBHOOK_CONNECTOR_REGISTRY = "WebhookConnectorRegistry";
+  public static final String DEPRECATED_WEBHOOK_MESSAGE_PREFIX = "Deprecated webhook path: ";
```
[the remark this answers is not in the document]
ev-codes: Thanks for this suggestion!

Who would be the audience of this log message? 

Currently, this PR's code only logs using `InboundConnectorRestController.log()` which is accessible via the `/inbound` REST API - perhaps we should log elsewhere or in an additional place.

Even if we keep the coding logic the same, I agree that adding more context would be helpful :)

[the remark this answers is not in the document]
chillleader: These logs will be visible in the web modeler, under "Last activities" 🙂 

Screenshot 2024-01-24 at 16 06 42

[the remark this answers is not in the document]
ev-codes: Thanks for this context!

Since that Webhook panel also contains more technical debugging info like the HTTP method, full URL's, and unformatted dates, I'm assuming the panel is intended for a more technical persona.

If my assumption is right, then I'm ok with putting a regex pattern in the log message.

It also doesn't seem like we need to log anywhere else.

Here's what the new log entry looks like from the inbound logs endpoint:
```
[[{"severity":"WARNING","tag":"WebhookConnectorRegistry","timestamp":1706114818.142828000,"message":"Deprecated webhook path: €. This may lead to unexpected behavior. Consider adjusting the path to match the pattern: ^[a-zA-Z0-9]([a-zA-Z0-9_-]*[a-zA-Z0-9])?$"}]]
```

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -54,6 +74,18 @@ public void register(ActiveInboundConnector connector) {
     }
   }
 
+  private static void logIfWebhookPathDeprecated(ActiveInboundConnector connector, String webhook) {
+
+    if (!CURRENT_WEBHOOK_PATH_PATTERN.matcher(webhook).matches()) {
+      String message =
+          DEPRECATED_WEBHOOK_MESSAGE_PREFIX + webhook + DEPRECATED_WEBHOOK_MESSAGE_SUFFIX;
+      Activity activity =
+          Activity.level(Severity.WARNING).tag(WEBHOOK_CONNECTOR_REGISTRY).message(message);
```
sbuettner, line 83: As this string will be shown in the webmodeler as a tag I am not sure how useful the information `WebhookConnectorRegistry` would be for a user. I would go with `Warning` or `Validation` or something. Maybe @YanaSegal has a good idea. 

Bildschirmfoto 2024-01-25 um 09 44 54

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -43,7 +56,14 @@ public boolean isRegistered(ActiveInboundConnector connector) {
   public void register(ActiveInboundConnector connector) {
     var properties = connector.context().bindProperties(CommonWebhookProperties.class);
     var context = properties.getContext();
+
+    logIfWebhookPathDeprecated(connector, context);
```
sbuettner, line 60: I would recommend to log this message after the check for an existing endpoint.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -16,16 +16,29 @@
...
+  private static final String WEBHOOK_CONNECTOR_REGISTRY = "WebhookConnectorRegistry";
+
+  // Reflect changes to this pattern in webhook element templates
+  private static final Pattern CURRENT_WEBHOOK_PATH_PATTERN =
+      Pattern.compile("^[a-zA-Z0-9]([a-zA-Z0-9_-]*[a-zA-Z0-9])?$");
+  private static final String DEPRECATED_WEBHOOK_MESSAGE_PREFIX = "Deprecated webhook path: ";
+  private static final String DEPRECATED_WEBHOOK_MESSAGE_SUFFIX =
+      ". This may lead to unexpected behavior. Consider adjusting the path to match the pattern: "
+          + CURRENT_WEBHOOK_PATH_PATTERN;
+
```
sbuettner, line 41: We could move this code and the static method in some WebhookValidation utility class to reduce the noise in this class but its rather a cosmetic change and not a blocker.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -16,16 +16,29 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class WebhookConnectorRegistry {
 
+  private static final String WEBHOOK_CONNECTOR_REGISTRY = "WebhookConnectorRegistry";
+
+  // Reflect changes to this pattern in webhook element templates
+  private static final Pattern CURRENT_WEBHOOK_PATH_PATTERN =
+      Pattern.compile("^[a-zA-Z0-9]([a-zA-Z0-9_-]*[a-zA-Z0-9])?$");
```
sbuettner, line 36: This allows paths like this: ```123__-23131```. Would be great to only allow hyphens and underscores between alphanumeric sections like this: `123_456-789`

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
ev-codes: Good idea!

Given this context, even "Warning" would be better than "WebhookConnectorRegistry", so I changed the tag to "Warning" for now. Will update based on Yana's feedback :)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
ev-codes: Sure!

Let me challenge some of my other assumptions about the desired pattern:

Should each "block" between hyphens and underscores should be exclusively numbers or exclusively letters? e.g.,
Valid: 123456-abcdef
Invalid: 1a2b3c-4d5e6f

Should we only allow lower case letters? e.g.,
Valid: abcdef
Invalid: ABCdef, ABCDEF

These might seem like silly questions, but I'm trying to build up my intuition on what's required here 😄 I'll update the pattern once I feel I understand the requirements.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
ev-codes: Fair point - I'll circle back to this one once I get clarity on the other remaining comments.

That way, I can avoid auto-resolving (or changing references to) any existing comments on these static fields.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
sbuettner: Its really up to us. We should support upper + lower case from my perspective.

ev-codes: Ok! In my opinion, the only new restriction we should add is:

> only allow hyphens and underscores between alphanumeric sections like this: 123_456-789

I believe adding this restriction this would result in the pattern:
`^[a-zA-Z0-9]([a-zA-Z0-9]*[a-zA-Z0-9_-])?([a-zA-Z0-9_-][a-zA-Z0-9]*)*`

I'll update the code and unit tests and push in a few minutes.

ev-codes: I pushed the updated code, unit tests, and element templates.

@sbuettner let me know what you think

sbuettner: Looks good! Thanks for taking care.

chillleader (dismissed):

sbuettner:
Looks good @ev-codes. Maybe you can do the validation extraction and we are good.

sbuettner (approved):
Thanks for taking care @ev-codes!

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
ev-codes: @sbuettner I extracted the static constants and method into a new class `WebhookConnectorValidationUtil`.

What do you think?

PR #1774, opened 2024-01-22
fix(webhook): inconsistent webhook de-registration

chillleader:
## Description

There was a bug in the webhook connector registration code: we attempted to deregister a webhook connector that was not registered due to a conflicting webhook endpoint path.

Consider the scenario:
- A process (`P1`) is deployed containing a webhook connector with path `myPath`. It is successfully registered with `WebhookConnectorRegistry`.
- We attempt to deploy another process (`P2`) that contains a webhook connector with the same path, `myPath`.
- This results in an error during registration, because this path is already used. This is checked in `webhookConnectorRegistry::register`.
- We make a change to `P2` and try to deploy a newer version
  - Process definition for the previous version of `P2` is registered
  - We know it contained a webhook connector, so we try to deactivate it
  - In `WebhookConnectorRegistry::deactivate`, the code doesn't check whether we are actually trying to deactivate the same connector that we tried to activate before. It only checks if paths are the same - which can also happen for different processes, like in our case.
  - The code removes `P1`'s webhook instead of the `P2`'s webhook, as only path is checked.

**Proposed solution**

1. Check the connector during the deregistration attempt and throw an exception if it is not the original connector (to be fail-safe)
2. In `InboundConnectorManager`, make a preliminary call to check whether the connector we are trying to deactivate, had actually been activated before (to avoid bumping into an exception)

## Related issues

closes https://github.com/camunda/team-connectors/issues/586

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorManager.java
@@ -60,7 +60,7 @@ public class InboundConnectorManager {
   //  setup
   private final Map<Long, Set<ActiveInboundConnector>> activeConnectorsByProcDefKey =
       new HashMap<>();
-  private Set<Long> registeredProcessDefinitions = new HashSet<>();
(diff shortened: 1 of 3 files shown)


igpetrov (approved):
LGTM, thank you for the improvememnt

PR #1783, opened 2024-01-23
feat(generator): support group definition for nested properties and binding name override

chillleader:
## Description

2 improvements to the java-based element template generator:
1. Support defining groups for nested properties, which comes in handy when the whole nested input should be part of the same group.
2. Support overriding property binidng name. By default, field name is used, but we allow overriding the name. Implemented as a nested annotation in case we decide to allow overriding the binding type too at some point (but it doesn't make sense to allow this as of now).


sbuettner (approved):
Awesome!

Issue #1799, opened 2024-01-24
Health reporting and activity logs is not working properly in some cases 


markfarkas-camunda:
__Describe the Bug__

In some cases health reporting is not working properly, and shows connector status as unknown despite it is up and running. 

Error reporting also needs to be refined leveraging the activity logs. If an FEEL parsing error happens we should report it using the activity logs.

__Steps to Reproduce__

Wrong connector state:
1. Create a bpmn process definition with a Slack inbound connector in it
2. Deploy or run the process definition
3. In the web-modeler open the Webhook tab. The status will be unknown even though the connector is activated properly

Not showing the error:
1. Create a bpmn process definition with a Webhook connector in it with faulty result expression
2. Deploy or run the process definition
3. Trigger the Webhook. An exception will be thrown in the BE. 
4. In the web-modeler open the Webhook tab. Observe that the error is not shown anywhere.

__Expected Behavior__

In the first case, we should show the UP state.

In the second case we should have the error in the activity logs.

__Environment__

- OS: Windows 11
- Library version: N/A
- Camunda version: 8.4.0

__PR__
https://github.com/camunda/connectors/pull/1800

PR #1800, opened 2024-01-24
fix(health report): add general health report and feel exception logging for webhooks

markfarkas-camunda:
## Description

Added a general health report logic, which reports the webhook connector health as UP, if it is still in unknown state. This happens after the activate method call, so we can be sure that at this point the connector must be up unless the status is set to DOWN.

Also added an error logging for Feel exceptions to show those on the FE. This way users have a feedback about the error happened.

## Related issues

closes #
https://github.com/camunda/connectors/issues/1799


igpetrov (approved):
LGTM, thank you

PR #1812, opened 2024-01-25
Add RPC support to the RabbitMQ Connector

### LINKED ISSUE #1735: Use the RabbitMQ connector to make RPC calls

I want to use the RabbitMQ Connector to publish an RPC message.

This is actually suggested in a Camunda blog post a while ago: https://camunda.com/blog/2021/03/communication-between-loosely-coupled-microservices/

There is a nice RPC pattern that RabbitMQ suggests here: https://www.rabbitmq.com/tutorials/tutorial-six-java.html

To make an RPC call to another application, you first publish a message, but fill in the `replyTo` and `correlationId` headers. These values can be autogenerated GUID's.

Next, you listen to the queue you have specified in the `replyTo` header (you should declare the queue before publishing, to avoid a race condition).

When a reply arrives at the queue, the correlationId should probably already match (since the queue name was also auto generated). This reply then gets send to the zeebe engine to complete the activity.

An timeout value can also be passed as a parameter to indicate when we should stop waiting for a RPC reply.

See here: https://forum.camunda.io/t/using-the-rabbitmq-connectors-to-make-rpc-calls/49799

__Additional context__

* https://www.rabbitmq.com/tutorials/tutorial-six-java.html
* https://camunda.com/blog/2021/03/communication-between-loosely-coupled-microservices/

michaelarnauts:
## Description

I've extended the RabbitMQ Connector to add RPC support based on the example here: https://www.rabbitmq.com/tutorials/tutorial-six-java.html

This is still a draft to get feedback about this. I'll also need to rebase this to `main` later.

Some code should probably be moved to a common model, but I'll keep it seperate for now to minimize code changes in unrelated files.

## Related issues

closes #1735

### DIFF
--- a/connectors/rabbitmq/element-templates/rabbitmq-connector-rpc.json
+++ b/connectors/rabbitmq/element-templates/rabbitmq-connector-rpc.json
@@ -1,0 +1,263 @@
+{
+  "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
+  "name": "RabbitMQ Outbound RPC Connector",
+  "id": "io.camunda.connectors.RabbitMQ.RPC.v1",
+  "description": "Send RPC to RabbitMQ",
+  "documentationRef": "https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/rabbitmq/?rabbitmq=outbound",
+  "version": 3,
+  "category": {
+    "id": "connectors",
+    "name": "Connectors"
+  },
+  "appliesTo": [
+    "bpmn:Task"
+  ],
+  "elementType": {
+    "value": "bpmn:ServiceTask"
+  },
+  "groups": [
+    {
+      "id": "authentication",
+      "label": "Authentication"
+    },
+    {
+      "id": "routing",
+      "label": "Routing"
+    },
+    {
+      "id": "message",
+      "label": "Message"
+    },
+    {
+      "id": "output",
+      "label": "Output mapping"
+    },
+    {
+      "id": "errors",
+      "label": "Error handling"
+    }
+  ],
+  "properties": [
+    {
+      "type": "Hidden",
+      "value": "io.camunda:connector-rabbitmq-rpc:1",
+      "binding": {
+        "type": "zeebe:taskDefinition:type"
+      }
+    },
+    {
+      "id": "connectionType",
+      "label": "Connection type",
+      "group": "authentication",
+      "type": "Dropdown",
+      "value": "uri",
+      "binding": {
+        "type": "zeebe:input",
+        "name": "authentication.authType"
+      },
+      "choices": [
+        {
+          "name": "URI",
+          "value": "uri"
+        },
+        {
+          "name": "Username/Password",
+          "value": "credentials"
+        }
+      ]
+    },
+    {
+      "label": "URI",
+      "description": "URI should contain username, password, host name, port number, and virtual host",
+      "group": "authentication",
+      "type": "String",
+      "binding": {
+        "type": "zeebe:input",
+        "name": "authentication.uri"
+      },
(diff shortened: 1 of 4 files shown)


chillleader (commented):
Thank you for raising this issue and taking the initiative here!

Looking at the implementation, I'm a bit concerned about job timeouts. If the RPC response timeout happens to be bigger than the job timeout, the connector invocation could be retried, creating a potential idempotency issue. I understand though that this approach may still work perfectly for you given the preconditions (hosting connectors on-premises => you are in control of the job timeout).

I discussed this with the @camunda/connectors team and we reached a consensus that we'd rather try to solve this on the modelling level, bootstrapping the existing outbound & inbound RabbitMQ connectors to create a reusable BPMN subprocess (similar to what I suggested [on the forum](https://forum.camunda.io/t/using-the-rabbitmq-connectors-to-make-rpc-calls/49799/2?u=chillleader)). We'd like to avoid maintaining a separate connector function tailored specifically for RMQ RPC.

Here's what we could do:
- If you prefer to proceed with the current solution, would you maybe consider publishing in our [community hub](https://github.com/camunda-community-hub)?
- Alternatively, we could try to solve the issues you've experienced with the modeling approach. For instance, if generating a UUID is something you're missing, it can also be implemented as a custom [FEEL function](https://camunda.github.io/feel-scala/docs/reference/developer-guide/function-provider-spi).

WDYT?

michaelarnauts:
@chillleader thanks for reviewing.

I've first tried to do it with in the BPMN as suggested on the forums, but I ran into a few issues.
* There is no `UUID()` function to generate the callback queue and correlation ID. Not sure if it will be accepted, since it's not in the DMN, but I've opened an issue here: https://github.com/camunda/feel-scala/issues/796
* It's just a lot of work to model this in BPMN (probably in every process where i want to do a RPC call) while I feel that this should be a basic feature of the connector.

Regarding the timeout, doesn't this also occur with the other connectors?

I think we don't have another choice then to maintain the changes ourselves. I'm not sure about the legal possibilities of publishing my changes in the Connector HUB, since it's based on the current RabbitMQ Producer code, what doesn't have an open license. :(

michaelarnauts:
Hmm, now thinking about it. Since the inbound connector requires Operate to register the callback (and therefore, requires a license), is it even possible to generate a UUID at runtime and have it register the queue? I thought that all inbound connectors were registered when the process is deployed.

sbuettner:
> I thought that all inbound connectors were registered when the process is deployed.

Exactly. The deployed process definitions are fetched from Operator and scanned for inbound Connector definitions. There is one inbound Connector instance per process definition. They are not instantiated on a process level but on a process definition level.

> is it even possible to generate a UUID at runtime and have it register the queue

Outbound Connectors can be seen as singletons. You would therefore do everything inside the function. You could generate any identifier there and potentially cache it based on some key if that is needed.

chillleader:
> is it even possible to generate a UUID at runtime and have it register the queue

What @sbuettner said above is true for most inbound connectors, but we also have an example where an inbound connector accesses process instance variables - it's the [HTTP Polling connector](https://github.com/camunda/connectors/blob/main/connectors/http/polling/src/main/java/io/camunda/connector/http/polling/HttpPollingConnector.java). Notice how it uses `InboundIntermediateConnectorContext` which is a modification of `InboundConnectorContext` that also exposes individual process instances. However this is much more complicated than a regular inbound or outbound connector.

> Regarding the timeout, doesn't this also occur with the other connectors?

True - we'll need to address this issue for other connectors too. There's a new Zeebe API that allows updating the timeout for activated jobs, we're currently considering exposing it in the connector context.

> it's based on the current RabbitMQ Producer code, what doesn't have an open license

Sadly, have to agree with you here. I guess the only viable option would be to eliminate the dependency on the OOTB connector.

PR #1830, opened 2024-01-26
fix(webhook): update tag from Warn to Warning

ev-codes:
## Description

Update webhook deprecated path warning tag to "Warning" instead of "Warn".

After discussion with @YanaSegal , we felt this would be the better tag name. Further UX suggestions we discussed would require `web-modeler` changes and a separate issue.

## Related issues

[[Bug](https://github.com/camunda/connectors/issues/1238)


igpetrov (approved):
LGTM, thank you

markfarkas-camunda (approved):
LGTM 👍

sbuettner (approved):

PR #1942, opened 2024-02-09
FeelEngineWrapper: only wrap the actual response to the response context

jonathanlukas:
## Description

Right now, the job context is wrapped to the response context, leading to the jobs' retries being available as `response.job.retries` which is not valid.

This PR fixes this (documented and tested) so that only the FIRST element of the varArgs for an evaluation is considered to be the response and wrapped into a response context.

## Related issues

nothing to close

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorHelper.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorHelper.java
@@ -16,6 +16,8 @@
  */
 package io.camunda.connector.runtime.core;
 
+import static io.camunda.connector.feel.FeelEngineWrapperUtil.*;
+
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
@@ -58,7 +60,10 @@ public class ConnectorHelper {
 
     Optional.ofNullable(resultExpression)
         .filter(s -> !s.isBlank())
-        .map(expression -> FEEL_ENGINE_WRAPPER.evaluateToJson(expression, responseContent))
+        .map(
+            expression ->
+                FEEL_ENGINE_WRAPPER.evaluateToJson(
+                    expression, responseContent, wrapResponse(responseContent)))
         .map(json -> parseJsonVarsAsTypeOrThrow(json, Map.class, resultExpression))
         .ifPresent(outputVariables::putAll);
 
@@ -74,7 +79,8 @@ public class ConnectorHelper {
         .filter(s -> !s.isBlank())
         .map(
             expression ->
-                FEEL_ENGINE_WRAPPER.evaluateToJson(expression, responseContent, jobContext))
+                FEEL_ENGINE_WRAPPER.evaluateToJson(
+                    expression, responseContent, wrapResponse(responseContent), jobContext))
         .filter(json -> !json.equals("null"))
         .filter(json -> !parseJsonVarsAsTypeOrThrow(json, Map.class, errorExpression).isEmpty())
         .map(json -> parseJsonVarsAsTypeOrThrow(json, ConnectorError.class, errorExpression))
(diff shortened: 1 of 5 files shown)


sbuettner:
It feels wrong to handle this on that level. WDYT? What about moving the `response` wraping into the client code and simplify the FeelEngineWrapper in that regard.

jonathanlukas:
I had these thoughts as well. However, this seemed to be breaking as the response wrapping will then happen in the connector runtime.

But in general, I like the idea :)

sbuettner:
I dont see an issue with that. We can just refactor the code. I like the change but I would prefer to have it on a different level so that its easier to reason about.

jonathanlukas:
@sbuettner I have refactored this now. The `FeelEngineWrapper` is not aware of any `response` context.

The creation of this context has been moved to a util which is used by tests and the `ConnectorHelper` now.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorHelper.java
```diff
@@ -58,7 +60,10 @@ public static Map<String, Object> createOutputVariables(
 
     Optional.ofNullable(resultExpression)
         .filter(s -> !s.isBlank())
-        .map(expression -> FEEL_ENGINE_WRAPPER.evaluateToJson(expression, responseContent))
+        .map(
+            expression ->
+                FEEL_ENGINE_WRAPPER.evaluateToJson(
+                    expression, responseContent, wrapResponse(responseContent)))
```
sbuettner, line 66: Why do we add the `responseContent` twice?

sbuettner: In the case of a successful Connector result you would have access to the result of the evaluation of the result variable and result expression: https://github.com/camunda/connectors/blob/1003797ccc93f4526073e6451599957d03ec8b38/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java#L176

jonathanlukas: Yes, but in the current implementation, you can access for example the body of a rest invocation in the result expression via `response.boy` OR `body`. Passing the context twice (wrapped AND unwrapped) will ensure that this is still possible.

sbuettner: Right, good point. We just need to verify that this was the only path where it was supported. Maybe it was similar for the result expressions. 

Do you know by chance @chillleader ?

connector-sdk/feel-wrapper/src/main/java/io/camunda/connector/feel/FeelEngineWrapper.java
```diff
@@ -132,7 +130,8 @@ private <T> T sanitizeScalaOutput(T output) {
    * Evaluates an expression with the FEEL engine with the given variables.
    *
    * @param expression the expression to evaluate
-   * @param variables the variables to use in evaluation
+   * @param variables the variables to use in evaluation, the first context provided will be
+   *     available wrapped as "response"
```
chillleader, line 134: Seems like that's not true anymore, same for other methods in this class

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorHelper.java
chillleader: I only ever saw this in relation to result expressions and error expressions, looks like both use cases are covered

connector-sdk/feel-wrapper/src/main/java/io/camunda/connector/feel/FeelEngineWrapper.java
```diff
@@ -90,7 +88,6 @@ private static String trimExpression(final String expression) {
   private static scala.collection.immutable.Map<String, Object> toScalaMap(
       final Map<String, Object> responseMap) {
     final HashMap<String, Object> context = new HashMap<>(responseMap);
-    context.put(RESPONSE_MAP_KEY, responseMap);
```
jonathanlukas, line 93: @sbuettner this line adds the whole `responseMap` under `response`. In the line above, the `responseMap` becomes the context itself first

sbuettner (approved):
Thanks @jonathanlukas for taking care!

PR #1943, opened 2024-02-09
Refactor Slack to use element template generator

igpetrov:
Refactor Slack to use element template generator


element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
```diff
@@ -109,7 +110,8 @@ public static List<PropertyBuilder> extractTemplatePropertiesFromType(Class<?> t
                 .peek(
                     builder -> {
                       if (nestedPropertiesAnnotation != null
-                          && nestedPropertiesAnnotation.condition() != null) {
+                          && StringUtils.isNotBlank(
+                              nestedPropertiesAnnotation.condition().property())) {
```
igpetrov, line 114: Hi @chillleader 

The reason for this change is the following:

1. `nestedPropertiesAnnotation.condition()` cannot be `null` as annotation requires constant.
2. Previous change generated the following empty condition.

```
"condition" : {
      "property" : "",
      "oneOf" : [ ],
      "type" : "simple"
    },
```

I wasn't sure that this wasn't expected so it would be nice if you take a look at this snippet too.

Thank you.

chillleader: Makes sense!

connectors/slack/element-templates/slack-outbound-connector-hybrid.json
```diff
@@ -0,0 +1,324 @@
...
+      "value" : "conversations.invite"
+    } ]
+  }, {
+    "id" : "data.messageType",
+    "label" : "Message type",
+    "optional" : false,
+    "value" : "plainText",
+    "group" : "message",
+    "binding" : {
+      "name" : "data.messageType",
```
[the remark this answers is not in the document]
igpetrov: Good catch! However, `messageType` is a new change not included in current release. Even in web-modeler, it's still open: https://github.com/camunda/web-modeler/pull/7942/files. I will update web-modeler with current change later

PR #1952, opened 2024-02-15
fix(generator): property type override via annotation

### LINKED ISSUE #1948: ETG: `propertyType` inside `@TemplateProperty` is not considered for non-primitive types
__Describe the Bug__

Consider an example:

```
public record ConnectorInput(
    @TemplateProperty(propertyType = PropertyType.String) Date prop
) {}
```

The element template generator never considers the property type specified in the annotation and instead tries to interpret the type as a container, which may result in different sorts of errors including Stack Overflow Exception (no meaningful error message provided in that case).

__Expected Behavior__

The generator should not try to handle the type as a container if manual override is provided in the annotation.

Additionally, we can create a rule to map `LocalDate` to string properties by default and improve error reporting.

chillleader:
## Description

Property type override was not applied to composite/container types. This lead to classes such as `Date` being always interpreted as container types no matter which property type is specified by the user in `@TemplateProperty(type=...)`.

Also:
- `LocalDate` and other Java 8 date/time types are implicitly mapped to `String` properties
- Improved handling of StackOverflowError for misconfigured container types
- Format changes for better readability

## Related issues

closes #1948

### DIFF
--- a/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
+++ b/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
@@ -61,9 +61,7 @@ import java.util.stream.Collectors;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.StringUtils;
 
-/**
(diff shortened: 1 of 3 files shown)


sbuettner (dismissed):

igpetrov (approved):
LGTM, thank you

Issue #1955, opened 2024-02-15
Exception Handling for Missing Event Definitions in  Inbound Connector Elements


Oleksiivanov:
__What should we do?__

Develop and integrate a new exception handling class named InvalidInboundConnectorDefinitionException into the Camunda Connectors error handling process. Adjust the error logging level in the getCorrelationPointForElement method to better categorize errors and prevent the overflow of non-critical error logs.

__Why should we do it?__

This task is crucial to maintain a clean and efficient logging system, ensuring that only genuine critical issues raise alerts. It will help in quicker identification of actual problems, reducing noise in the logs, and allowing the team to focus on significant errors that impact system stability and performance. 

link to slack chat : https://camunda.slack.com/archives/C05Q7C4HJ5Q/p1708007276714739

PR #1956, opened 2024-02-15
feat(logging): Introduce InvalidInboundConnectorException and adjust log level for missing definitions

Oleksiivanov:
## Description

Added `InvalidInboundConnectorException` class. This exception is specifically thrown in cases where the sanity check for connector definitions fails, such as when the required event definitions are missing.

Modified the error handling strategy within the getCorrelationPointForElement method to catch InvalidInboundConnectorException and log it as a warning. This adjustment is made to ensure consistency in how we handle and log similar errors across different connector processing methods, such as getCorrelationPointForStartEvent, where similar issues could arise.

## Related issues

closes #
https://github.com/camunda/connectors/issues/1955

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/InvalidInboundConnectorDefinitionException.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/InvalidInboundConnectorDefinitionException.java
@@ -19,35 +19,111 @@ package io.camunda.connector.runtime.core.error;
 import java.io.Serial;
 
 /**
- * Represents an unchecked exception indicating an invalid definition for an inbound connector. This
- * exception is thrown when the connector's definition does not meet the expected format or
- * standards.
+ * Represents an unchecked exception indicating an invalid definition for an inbound connector.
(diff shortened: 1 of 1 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/error/InvalidInboundConnectorDefinitionException.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package io.camunda.connector.api.error;
+
+import java.io.Serial;
+
+/**
+ * Unchecked exception indicating an invalid definition for an inbound connector. This exception is
+ * thrown when a connector definition does not meet the required criteria.
+ */
+public class InvalidInboundConnectorDefinitionException extends RuntimeException {
```
sbuettner, line 25: What about adding some more information like the process definition and the version?

sbuettner: Cant we please move this exception to the runtime-core?

sbuettner (approved):
Awesome, thanks @Oleksiivanov

PR #1958, opened 2024-02-16
feat(generator): inbound element template generation

chillleader:
## Description

This PR replaces the `OutboundClassBasedTemplateGenerator` with the `ClassBasedTemplateGenerator` capable of generating both outbound and inbound element templates.

### Highlights

Apart from the main feature itself (inbound template generation), please note the following:
- `@ElementTemplate` annotation now allows defining multiple element types and configuring template ID and name overrides for each of them
- Configuration of the Template Generator Maven plugin now has a different format and supports file name overrides for each element type and hybrid option for each connector inside the project individually
  - This change is applied to all existing connectors
- `OutboundClassBasedTemplateGenerator` is now `ClassBasedTemplateGenerator`. The domain model of the generators is simplified and now there is no separation of interfaces between inbound and outbound.
- Webhook Connector is now using the template generator

## Readiness checklist

- [x] Core implementation
- [x] Tests for inbound generation
- [x] Apply the new generator to one of the connectors

## Related issues

related to https://github.com/camunda/team-connectors/issues/547


connectors/aws/aws-dynamodb/element-templates/aws-dynamodb-outbound-connector.json
```diff
@@ -943,11 +943,11 @@
     },
     "type" : "Dropdown",
     "choices" : [ {
-      "name" : "DELETE",
```
igpetrov, line 946: out of curiosity, why do we remove it?

connectors/soap/src/main/java/io/camunda/connectors/soap/SoapConnector.java
```diff
@@ -29,7 +29,7 @@
     type = SoapConnector.SOAP_CONNECTOR_TYPE,
     inputVariables = {})
 @ElementTemplate(
-    id = "io.camunda:soap",
+    id = "io.camunda.consulting:soap",
```
igpetrov, line 32: I believe, it shouldn't be consulting anymore.

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/JWTProperties.java
```diff
@@ -7,14 +7,33 @@
...
 import java.util.Objects;
 import java.util.function.Function;
 
 public record JWTProperties(
-    @FEEL List<String> requiredPermissions,
-    Function<Object, List<String>> permissionsExpression,
-    @FEEL String jwkUrl) {
+    @TemplateProperty(
+            label = "JWK url",
+            description = "Well-known url of JWKs",
```
igpetrov, line 19:
```suggestion
            label = "JWK URL",
            description = "Well-known URL of JWKs",
```

connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
```diff
@@ -43,6 +46,43 @@
...
+      @PropertyGroup(id = "authentication", label = "Authentication"),
+      @PropertyGroup(id = "authorization", label = "Authorization"),
+      @PropertyGroup(id = "webhookResponse", label = "Webhook response")
+    },
+    elementTypes = {
+      @ConnectorElementType(
+          appliesTo = BpmnType.START_EVENT,
+          elementType = BpmnType.START_EVENT,
+          templateIdOverride = "io.camunda.connectors.webhook.WebhookConnector.v1",
+          templateNameOverride = "Webhook Start Event Connector"),
```
igpetrov, line 69: I am not 100% sure, but feels like we wanted to keep the first word and the `Connector` capitalized, in this case smth like: 

```suggestion
          templateNameOverride = "Webhook start event Connector"),
```

Maybe @YanaSegal can confirm or disproof?

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookAuthorization.java
```diff
@@ -22,14 +26,50 @@
...
+  @TemplateSubType(id = "APIKEY", label = "API Key")
+  record ApiKeyAuth(
+      @TemplateProperty(
+              label = "API Key",
+              description = "Expected API key",
+              group = "authorization")
+          @FEEL
+          String apiKey,
+      @TemplateProperty(
+              label = "API Key locator",
```
igpetrov, line 65:
```suggestion
  @TemplateSubType(id = "APIKEY", label = "API key")
  record ApiKeyAuth(
      @TemplateProperty(
              label = "API key",
              description = "Expected API key",
              group = "authorization")
          @FEEL
          String apiKey,
      @TemplateProperty(
              label = "API key locator",
```

igpetrov (approved):
LGTM, thank you. Some minor suggestions.

connectors/aws/aws-dynamodb/element-templates/aws-dynamodb-outbound-connector.json
chillleader: It was just moved a few lines down, the reason is that I made a change in the generator that preserves the order of dropdown options from the annotation.

sbuettner (approved):

sbuettner (approved):

connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
chillleader: Looking at other connectors, we are largely using capital case so I'd rather leave it as is and tackle possible renaming separately if we agree on that

igpetrov: Sure

igpetrov (approved):

PR #1967, opened 2024-02-20
chore(jackson): disable time as array serialization

igpetrov:
chore(jackson): disable time as array serialization


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorHelper.java
```diff
@@ -39,7 +40,10 @@ public class ConnectorHelper {
   public static FeelEngineWrapper FEEL_ENGINE_WRAPPER = new FeelEngineWrapper();
   // TODO: Check if this is properly configured
   public static ObjectMapper OBJECT_MAPPER =
-      new ObjectMapper().registerModule(new JavaTimeModule());
+      new ObjectMapper()
```
sbuettner, line 43: Can we re-use the object mapper from [ConnectorsObjectMapperSupplier.java](https://github.com/camunda/connectors/pull/1967/files#diff-da754f27ed015a23aa4c6f76ac24826962d8afd12b58d325d47dad5d64cea865)?

sbuettner: And remove the comment: `/** The ConnectorHelper provide utility functions used to build connector runtimes. */`

igpetrov: I did it in a first place but didn't find the reason it was introduced. Maybe it has a purpose? 🤔 
Regardless, tests did pass, so assuming it's safe to swap.

PR #1971, opened 2024-02-21
feat(generator): support constraints in @TemplateProperty

chillleader:
## Description

Adds a way to define property constraints in `@TemplateProperty` instead of using Java bean validation API annotations.

Context: https://camunda.slack.com/archives/C0350959WG1/p1708515841961049


igpetrov (approved):
Awesome, thanks a lot!

PR #2038, opened 2024-02-27
fix(importer): prevent OOM by filtering out older process versions early on

chillleader:
## Description

We need to filter out older process definitions versions at an earlier stage to prevent storing all process definitions in a list which may lead to an OOM error when there are too many processes deployed.

## Related issues

closes https://github.com/camunda/team-connectors/issues/647


sbuettner (approved):

igpetrov (approved):
LGTM

PR #2053, opened 2024-02-28
feat(openai): Add a temperature input, Add the possibility to use a custom OpenAI model

### LINKED ISSUE #1959: OpenAI Connector models are fixed and point to outdated versions
## Issue

The OpenAI connector allow only the choice of _GPT-3.5 Turbo_ or _GPT-4_. That leads to two issues:

1. Impossible to pin model versions for reliable behaviour. These two are aliases and OpenAI changes what model they point to over time.
2. Multiple alternative and often newer and more performant models are unavailable to the connector user.

## Solution

Change Model field in `connectors/connectors/openai/element-templates/openai-connector.json` from a dropdown to a text field to allow any model input.

johnBgood:
## Description

This PR updates our **openai** connector:
- New `temperature` field
  - with a validation, the value must be between 0 and 2 (floats are ok) according to their [documentation](https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature)
- New **Custom** entry in the models dropdown.
  - When **Custom** is selected a new TextField appears, any value can be typed there

https://github.com/camunda/connectors/assets/1797846/b6cd2db8-c22b-47f2-a33c-c9b30652ed5b

## Related issues

closes #1959

### DIFF
--- a/connectors/openai/element-templates/openai-connector.json
+++ b/connectors/openai/element-templates/openai-connector.json
@@ -183,6 +183,10 @@
         {
           "name": "GPT-4",
           "value": "gpt-4"
+        },
+        {
+          "name": "GPT-4-turbo-preview",
+          "value": "gpt-4-turbo-preview"
         }
       ],
       "binding": {
--- a/connectors/openai/element-templates/openai-connector.json
+++ b/connectors/openai/element-templates/openai-connector.json
@@ -2,7 +2,7 @@
   "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
   "name": "OpenAI Outbound Connector",
   "id": "io.camunda.connectors.OpenAI.v1",
-  "version": 3,
+  "version": 4,
   "description": "Interact with ChatGPT or moderation API",
   "icon": {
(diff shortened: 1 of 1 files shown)


igpetrov (approved):
LGTM!

Few suggestions:
- In the `Assignee` field put yourself
- In the project, where you selected `Connectors` you also need to change status to `In review`

connectors/openai/element-templates/openai-connector.json
```diff
@@ -170,11 +166,16 @@
...
       "type": "Dropdown",
       "value": "gpt-3.5-turbo",
       "choices": [
+        {
+          "name": "Custom",
+          "value": "custom"
+        },
         {
           "name": "GPT-3.5-turbo",
           "value": "gpt-3.5-turbo"
```
sbuettner, line 181: Would be great if we could add the other known supported ones as well https://platform.openai.com/docs/models/continuous-model-upgrades namely: `gpt-4` and `gpt-4-turbo-preview`

sbuettner: Another reference: https://platform.openai.com/docs/models/model-endpoint-compatibility

johnBgood: Done ✔️

sbuettner (approved):
LGTM and congrats to your first PR! 🥳

PR #2055, opened 2024-02-29
feat(kafka): revert to correlate without result

Oleksiivanov:
## Description

Previous commits introduced correlation with the result in the Kafka consumer. In the weekly meeting, it was decided to revert to the previous implementation for now; 
thus, we will correlate without the result


igpetrov (approved):
Thank you

PR #2056, opened 2024-02-29
feat(kafka inbound): topic and consumer group not support FEEL

Oleksiivanov:
## Description

As discussed with the team, the topic and consumer group ID will not support FEEL expressions. Templates have been updated; template versions have been incremented.

next step : 
- https://github.com/camunda/web-modeler/pull/8355


igpetrov (approved):
LGTM

Issue #1970, opened 2024-02-21
Remove SPEC_PASSWORD_EMPTY_PATTERN from BasicAuthentication 


markfarkas-camunda:
__What should we do?__

We should remove the SPEC_PASSWORD_EMPTY_PATTERN constant from BasicAuthentication class and from easy post connector template. Instead we should just use empty string to indicate that the password was not provided.

__Why should we do it?__

Snyk code scanning finds this code suspicious, and this also adds unnecessary complexity to the code.

ev-codes:
The connectors runtime changes are merged and ready for release!

Adding the new template to the web-modeler [is in review](https://github.com/camunda/web-modeler/pull/8422). Since the required web-modeler reviewer is [busy with code freeze](https://github.com/camunda/web-modeler/pull/8422#issuecomment-1980287272), I'll:
- follow up the reviewer next week
- keep [this issue](https://github.com/camunda/connectors/issues/1970) as "In Review" until the web-modeler PR is merged.

ev-codes:
We discussed today whether we can introduce a new EasyPost template without a password field.

The problem is a new template without a password field will not work with <8.5.0 connectors runtime.

We discussed a few options:
- introducing a new template with the same ID (won't work)
- intorudcing a new template with a new iD and new template type (overly complicated and painful for customers
- add dropdown with "old" and "new" authentication type (overly confusing for customers)
- start deprecation process for "old" template (too painful for customers)

More to come.

ev-codes:
We won't backport this to 8.4 (as discussed among Connectors team) as this is not strictly a bug fix.

PR #2061, opened 2024-03-01
Remove use of `SPEC_PASSWORD_EMPTY_PATTERN`

ev-codes:
## Description

Allow passing in an empty String for a password.

Previously, we needed this field [because Zeebe would not accept](https://github.com/camunda/connectors/pull/386) non-empty password fields.

Zeebe does not give any errors for this when I deploy to my local Zeebe setup with the changes in this PR.

## Related issues

#1970 (I will close after I add the new EasyPost template to `web-modeler`)

### DIFF
--- a/connectors/easy-post/element-templates/easy-post-connector.json
+++ b/connectors/easy-post/element-templates/easy-post-connector.json
@@ -114,20 +114,6 @@
       }
     },
     {
-      "id": "authentication.password",
-      "description": "This field will replace to empty string. EasyPost API use only username field for put API key",
-      "group": "authentication",
-      "type": "Hidden",
-      "optional": true,
-      "constraints": {
-        "notEmpty": false
-      },
-      "binding": {
-        "type": "zeebe:input",
-        "name": "authentication.password"
-      }
-    },
-    {
       "description": "post method",
       "type": "Hidden",
       "optional": false,
--- a/connectors/easy-post/element-templates/easy-post-connector.json
+++ b/connectors/easy-post/element-templates/easy-post-connector.json
@@ -2,7 +2,7 @@
   "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
   "name": "Easy Post Outbound Connector",
   "id": "io.camunda.connectors.EasyPost.v1",
-  "version": 4,
+  "version": 5,
   "description": "Allows you to create addresses, parcels, and shipments, as well as purchase and verify shipments",
   "documentationRef": "https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/easy-post/",
   "category": {
--- a/connectors/easy-post/element-templates/easy-post-connector.json
+++ b/connectors/easy-post/element-templates/easy-post-connector.json
@@ -119,7 +119,7 @@
       "group": "authentication",
       "type": "Hidden",
       "optional": true,
-      "value": "SPEC_PASSWORD_EMPTY_PATTERN",
+      "value": "",
       "constraints": {
         "notEmpty": false
       },
--- a/connectors/easy-post/element-templates/easy-post-connector.json
+++ b/connectors/easy-post/element-templates/easy-post-connector.json
@@ -119,7 +119,6 @@
       "group": "authentication",
       "type": "Hidden",
       "optional": true,
-      "value": "",
       "constraints": {
         "notEmpty": false
       },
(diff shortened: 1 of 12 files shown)


johnBgood (dismissed):
LGTM 👍

connectors/easy-post/element-templates/easy-post-connector.json
```diff
@@ -119,7 +119,7 @@
       "group": "authentication",
       "type": "Hidden",
       "optional": true,
-      "value": "SPEC_PASSWORD_EMPTY_PATTERN",
+      "value": "",
```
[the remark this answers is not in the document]
ev-codes: Thanks for the tip! I made the needed change in `BasicAuthentication`.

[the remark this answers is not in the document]
sbuettner: @ev-codes why not remove the property?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/auth/BasicAuthentication.java
```diff
@@ -22,30 +22,23 @@
...
 
   @FEEL
   @NotEmpty
   @TemplateProperty(group = "authentication")
   private String username;
 
   @FEEL
-  @NotEmpty
   @TemplateProperty(group = "authentication")
   private String password;
```
sbuettner, line 36: Please add the condition via an element-template generator constraint that this should not be empty. This will render the property as required in the element templates but wont require a `@NotEmpty` constraint on the backend.

sbuettner: And please re-generate the element templates for the http base connector based connectors (REST & GraphQL)

ev-codes: Thanks, @sbuettner!

Have you seen [Oleksii's comment](https://github.com/camunda/connectors/pull/2061#pullrequestreview-1911664972)?

I'll wait to address your comments here until we can discuss Oleksii's comment.

johnBgood: It's probably somewhere, but if not should we add a section (the command basically) in element-template-generator/maven-plugin/README.md describing how to generate the element templates? 
Might be useful if we hire people without prior Maven experience?

sbuettner: @ev-codes Yes, I saw it and my proposal would lead to keeping the constraint in the template but removing it from the backend.

sbuettner: @johnBgood Good idea, but should be tackled in a separate issue. Please create one, so we dont forget it.

ev-codes: Summary:
- @Oleksiivanov agreed with @sbuettner 's approach
- I implemented @sbuettner 's suggested change
- @johnBgood [created the issue here](https://github.com/orgs/camunda/projects/23/views/1?filterQuery=doc&pane=issue&itemId=55227336)

connectors/easy-post/element-templates/easy-post-connector.json
```diff
@@ -119,7 +119,6 @@
       "group": "authentication",
       "type": "Hidden",
       "optional": true,
-      "value": "SPEC_PASSWORD_EMPTY_PATTERN",
       "constraints": {
         "notEmpty": false
```
sbuettner, line 123: Shouldnt we remove the property?

sbuettner:
@ev-codes I can see that you re-generate the webhook element templates but I expected to see changes in the Rest and GraphQL outbound connector element templates.

connectors/easy-post/element-templates/easy-post-connector.json
ev-codes: Sure! Done

sbuettner:
@ev-codes You are right, the outcome should be same regarding the element-template. I was irritated by the webhook element template changes as they should not be affected by this. I would therefore recommend to remove them from this PR.

ev-codes:
@sbuettner I understand your irritation, I also don't like this.

However, the re-ordering in the webhook element templates will happen any time we re-generate the templates with `mvn clean verify` going forward. (If I understand correctly).

So I'd rather include these changes once in this PR when the impact is clear. Otherwise, whoever runs `mvn clean verify` next will be even more confused, since they don't have the context.

What do you think? Perhaps we can generate the constraints in a way that doesn't inadvertently touch the webhook connectors?

sbuettner:
Thats the point, your changes should not affect the webhook element templates at all. Thats why the changes dont belong in this PR. AFAIK they also use a different data class. 

Is that related to the inbound element template change @chillleader ?

ev-codes:
@sbuettner : @chillleader let me know that perhaps I have an outdated version of the template generator plugin on my local machine.

While I determine what's wrong on my local machine, I removed the webhook connector changes from this PR.

PR #2063, opened 2024-03-04
chore(spring-zeebe): update to relevant dep

igpetrov:
chore(spring-zeebe): update to relevant dep


igpetrov:
The failed build seems to be the reason why backport has failed

PR #2075, opened 2024-03-05
fix: incorrect context binding in wh element template

chillleader:
## Description

Fixes an error with the incorrect binding name (was `inbound.inbound.context` due to nested path resolution).

### DIFF
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
@@ -43,7 +43,7 @@ public record WebhookConnectorProperties(
             label = "Webhook ID",
             group = "endpoint",
             description = "The webhook ID is a part of the URL",
-            binding = @PropertyBinding(name = "inbound.context"))
+            binding = @PropertyBinding(name = "context"))
         @NotBlank
         @Pattern(
             regexp = "^[a-zA-Z0-9]+([-_][a-zA-Z0-9]+)*$",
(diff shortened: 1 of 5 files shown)


1petr0v (approved):
LGTM

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
```diff
@@ -43,7 +43,7 @@ public record WebhookConnectorProperties(
             label = "Webhook ID",
             group = "endpoint",
             description = "The webhook ID is a part of the URL",
-            binding = @PropertyBinding(name = "inbound.context"))
+            binding = @PropertyBinding(name = "context"))
```
johnBgood, line 46: dumb question but why don't we need `inbound` here, but we need it when using 
` @PropertyCondition(property = "inbound.shouldValidateHmac", equals = "enabled"))` ?

johnBgood (approved):
👍

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
johnBgood: another thing, the `version` properties are not bumped for any of the json files

chillleader: `WebhookConnectorProperties` are wrapped into `WebhookConnectorPropertiesWrapper` that contains a single nested field `inbound`. The generator takes wrapper as an input instead of the real properties, and then it is able to add the `inbound` prefix automatically the same way it resolves other nested props (we add this prefix by default but this is configurable).

chillleader: > another thing, the version properties are not bumped for any of the json files

I'd say that in this case we don't need the version bump because we bumped the versions in the [previous PR](https://github.com/camunda/connectors/pull/1958) and that version never made it to prod because of this bug.

Issue #1236, opened 2023-10-06
[Support] [Bug] Escaped characters incorrectly deserialize from Kafka topic


igpetrov:
__Describe the Bug__

It was identified by the `SUPPORT-18661` that escaped special characters may lead to broken JSON parsing. See ticket for more information.

__Steps to Reproduce__

1. Create any Kafka inbound connector assigned to a topic
2. Publish into the topic the following message with escaped character, e.g. `\"`

```
{
  "key": "test",
  "value": {
    "payload": {
      "foo": "a\"string"
    }
  }
}
```

__Expected Behavior__

Message parses correctly.

__Environment__

- Any

__Related To__
- [SUPPORT-20896](https://jira.camunda.com/browse/SUPPORT-20896)

igpetrov:
@sbuettner @bastiankoerber needs to be groomed

Vadimnick:
@markfarkas-camunda Hi, can you please tell me when the fix is. We look forward to being corrected.   🙏🙏🙏

PR #2091, opened 2024-03-05
fix(kafka): fix deserialization of escaped characters

markfarkas-camunda:
## Description

Allow parsing properly unescaped control characters, to prevent parsing failures.

## Related issues

closes https://github.com/camunda/connectors/issues/1236


johnBgood (dismissed):
Do we have some UTs already covering this ticket's use case (line breaks)? If not would be nice to add some

sbuettner (dismissed):

johnBgood (approved):
Great thanks!

PR #2102, opened 2024-03-05
fix(webhook): missing default and constraint for the api key locator field

chillleader:
## Description

Default value and constraint were missing after migrating to the auto-generated template.

## Related issues

Found during QA: https://github.com/camunda/team-connectors/issues/655


igpetrov (approved):
LGTM

PR #2104, opened 2024-03-05
chore(kafka): fix broken kafka test after backport

markfarkas-camunda:
## Description

After backporting https://github.com/camunda/connectors/pull/2091 a test was failing. This commit fixes that test.


igpetrov (approved):
LGTM, thank you

PR #2164, opened 2024-03-14
fix: revert to v4 EasyPost template

ev-codes:
## Description

v5 was never released to customers.

v5 will not work with runtimes before 8.5.0.

We don't yet have a mechanism to inform customers which runtimes a template is supported by.

Because this template was only cleaning up technical debt and not introducing new features, we believe the best option is to not release a new version for now.

[See the discussion here](https://github.com/camunda/connectors/issues/1970#issuecomment-1997022685) for more details.

## Related issues

https://github.com/camunda/connectors/issues/1970


ev-codes:
@Oleksiivanov 

> Can we add a backport label for 8.5, or would it be better to manually revert in the 8.5 branch?

I'm not sure, what do you think?

ev-codes:
@Oleksiivanov ok! I added `backport/8.5`.

Because this PR doesn't touch the runtime code, and this template was never released to customers, I won't take additional actions to release a new patch.

PR #2183, opened 2024-03-18
chore(soap): update SOAP connector to the latest generator

igpetrov:
chore(soap): update SOAP connector to the latest generator

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-soap/src/test/java/io/camunda/connector/e2e/soap/SoapConnectorBaseTest.java
+++ b/connectors-e2e-test/connectors-e2e-test-soap/src/test/java/io/camunda/connector/e2e/soap/SoapConnectorBaseTest.java
@@ -40,7 +40,7 @@ import org.springframework.boot.test.web.server.LocalServerPort;
 public abstract class SoapConnectorBaseTest {
 
   protected static final String ELEMENT_TEMPLATE_PATH =
-      "../../connectors/soap/element-templates/soap-connector.json";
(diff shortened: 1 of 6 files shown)


johnBgood:
Not part of your issue but [SoapConnector](https://github.com/camunda/connectors/blob/2b4997bfc2219e5f3f7f6e18ef1bb48b602e482e/connectors/soap/src/main/java/io/camunda/connectors/soap/SoapConnector.java#L30) is the only one (I guess) with an empty `inputVariables` param. Is it on purpose?

igpetrov:
> Not part of your issue but [SoapConnector](https://github.com/camunda/connectors/blob/2b4997bfc2219e5f3f7f6e18ef1bb48b602e482e/connectors/soap/src/main/java/io/camunda/connectors/soap/SoapConnector.java#L30) is the only one (I guess) with an empty `inputVariables` param. Is it on purpose?

Fair point for the normal circumstances, but the idea for this connector was to move it "as-is" from consulting repo. Refactoring planned as the next step.

sbuettner:
@igpetrov I think its a valid concern and considered best practice. Would be great if you could quickly incorporate these changes.

igpetrov:
> @igpetrov I think its a valid concern and considered best practice. Would be great if you could quickly incorporate these changes.

okay, shouldn't take long

igpetrov:
Hi @Oleksiivanov,

Sorry for back-and-forth, could you please give one more review/approval?

Thank you!

PR #2185, opened 2024-03-19
fix(sendgrid): fix sendgrid template

Oleksiivanov:
## Description

 - fix sendgrid template properties;
 - added return value;
 - added hybrid template


sbuettner (approved):
Thanks and please also update the template in the web-modeler

PR #2217, opened 2024-03-20
fix(feel): fixed timestamp serialization

igpetrov:
fix(feel): fixed timestamp serialization


jonathanlukas:
There is no test in place to assert the correct serialization. 

@igpetrov do you plan to add a test?

sbuettner (approved):

igpetrov:
> There is no test in place to assert the correct serialization.
> 
> @igpetrov do you plan to add a test?

Added a test

PR #2219, opened 2024-03-20
chore(rest): added validation to OAuth token URL

igpetrov:
chore(rest): added validation to OAuth token URL

### DIFF
--- a/connectors/http/graphql/element-templates/graphql-outbound-connector.json
+++ b/connectors/http/graphql/element-templates/graphql-outbound-connector.json
@@ -4,7 +4,7 @@
   "id" : "io.camunda.connectors.GraphQL.v1",
   "description" : "Execute GraphQL query",
   "documentationRef" : "https://docs.camunda.io/docs/components/connectors/protocol/graphql/",
-  "version" : 5,
+  "version" : 6,
   "category" : {
     "id" : "connectors",
     "name" : "Connectors"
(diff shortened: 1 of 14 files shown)


connectors/http/graphql/element-templates/graphql-outbound-connector.json
```diff
@@ -193,11 +193,15 @@
...
     "id" : "authentication.oauthTokenEndpoint",
-    "label" : "Oauth token endpoint",
+    "label" : "OAuth 2.0 token endpoint",
     "description" : "The OAuth token endpoint",
     "optional" : false,
     "constraints" : {
-      "notEmpty" : true
+      "notEmpty" : true,
+      "pattern" : {
+        "value" : "^(=|http://|https://|secrets|\\{\\{).*$",
```
johnBgood, line 202: I could use `secretsssss` for instance, or `http://///dldl`, not sure we want a full validation though so feel free to ignore my comment.

connectors/http/graphql/element-templates/graphql-outbound-connector.json
```diff
@@ -193,11 +193,15 @@
...
+        "message" : "Must be a http(s) URL"
```
johnBgood, line 203: could be http URL, secret, expression etc

johnBgood (approved):
Minor comments only, great that you refactored the code 👍

connectors/http/graphql/element-templates/graphql-outbound-connector.json
igpetrov: I guess, we're fine with that. We discussed the same issue with generic URL validation but decided not to go over-strict

igpetrov: I will create a ticket to revisit the URL pattern

igpetrov: There it is https://github.com/camunda/connectors/issues/2236

Issue #2239, opened 2024-03-22
[Support] [Bug] Kafka inbound connector deserialization fails if json string contains "\"


markfarkas-camunda:
__What should we do?__

Issue was raised in SUPPORT-21234. Kafka intermediate catch event fails if json string contains "\". We introduced a fix for this in https://github.com/camunda/connectors/issues/1236 but it does not work in every case. 
We should fix the json deserialization in the kafka inbound connector.

__Why should we do it?__

It is a critical bug.

PR #2240, opened 2024-03-22
fix(kafka): fix kafka deserialization

markfarkas-camunda:
## Description

- Removed the unescaping logic
- Improved logging
- Added new test case

## Related issues

https://github.com/camunda/connectors/issues/2239

closes https://github.com/camunda/connectors/issues/2239

### DIFF
--- a/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaPropertyTransformer.java
+++ b/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaPropertyTransformer.java
@@ -132,8 +132,7 @@ public class KafkaPropertyTransformer {
         kafkaInboundMessage.setValue(objectReader.readTree(value));
       }
     } catch (Exception e) {
-      LOG.error(e.getMessage());
-      LOG.error("Cannot parse value to json object -> use the raw value");
+      LOG.error("Cannot parse value to json object -> use the raw value", e);
       kafkaInboundMessage.setValue(kafkaInboundMessage.getRawValue());
     }
   }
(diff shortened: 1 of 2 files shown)


igpetrov (approved):
LGTM. thank you

connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaPropertyTransformer.java
```diff
@@ -129,11 +129,11 @@ private static void setValue(
...
         String value = (String) consumerRecord.value();
         kafkaInboundMessage.setRawValue(value);
-        var json = StringEscapeUtils.unescapeJson(value);
-        kafkaInboundMessage.setValue(objectReader.readTree(json));
+        kafkaInboundMessage.setValue(objectReader.readTree(value));
       }
     } catch (Exception e) {
-      LOG.debug("Cannot parse value to json object -> use the raw value");
+      LOG.error(e.getMessage());
+      LOG.error("Cannot parse value to json object -> use the raw value");
```
johnBgood, line 136: Is it an error? Something we want to act on?

johnBgood (approved):
🥇

connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaPropertyTransformer.java
markfarkas-camunda: Good question! The reason behind this is to be able to handle the case when we process a malformed message. The raw value (string value) will still be available, even though we cannot deserialize the message.

johnBgood: Ok, so you changed the log to an error one so that we can spot these kinds of issues in a GCP log for instance 👌 

I don't know if there's a reason for the `LOG.error(e.getMessage());` but I think you can use:

`LOG.error("Cannot parse value to json object -> use the raw value", e);`

markfarkas-camunda: Good idea, adjusted it.

igpetrov (approved):

johnBgood (approved):

PR #2248, opened 2024-03-26
fix(webhook): adjust property IDs in the element templates

chillleader:
## Description

Changes the property IDs of the webhook element templates that deviated from the original values after migrating to the template generator. This does not change the runtime behavior since binding names don't change, i.e. IDs are simply brought in accordance with bindings for clarity.

### DIFF
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
@@ -13,7 +13,6 @@ import io.camunda.connector.feel.annotation.FEEL;
 import io.camunda.connector.generator.dsl.Property.FeelMode;
 import io.camunda.connector.generator.java.annotation.TemplateProperty;
 import io.camunda.connector.generator.java.annotation.TemplateProperty.DropdownPropertyChoice;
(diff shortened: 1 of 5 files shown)


connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
```diff
@@ -39,11 +38,11 @@ public record WebhookConnectorProperties(
             defaultValue = "any")
         String method,
     @TemplateProperty(
-            id = "webhookId",
+            id = "context",
             label = "Webhook ID",
             group = "endpoint",
             description = "The webhook ID is a part of the URL",
-            binding = @PropertyBinding(name = "context"))
```
chillleader, line 46: Bindings are optional here as they are equal to property IDs.

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
```diff
@@ -39,11 +38,11 @@ public record WebhookConnectorProperties(
...
+            feel = FeelMode.disabled)
```
chillleader, line 45: FEEL was disabled in previous versions before the generator

johnBgood (approved):

sbuettner (approved):

Issue #2251, opened 2024-03-27
Set Default Value to 'false' for All Inbound Connectors' FEEL Expressions in the Template Generation


Oleksiivanov:
__What should we do?__

In the template generator, the default value for all inbound connectors should be 'false', while remaining optional for all outbound connectors

related : 
 - https://github.com/camunda/connectors/pull/2056

PR #2264, opened 2024-03-27
feat(template generator): set default value feel = false for all inbound connectors

Oleksiivanov:
## Description

In the template generator, the default value for all inbound connectors should be 'false', while remaining optional for all outbound connectors;

In the template generator, the default value for all inbound connectors is set to 'false', with this setting remaining optional for all outbound connectors. 

Additionally, a system_default value for FeelMode has been introduced: if it's inbound, the default value is 'disabled'; otherwise, it's 'optional'. 

Furthermore, feel = FeelMode.optional has been added to all webhook templates to ensure this setting remains unchanged.

## Related issues

closes #
 - https://github.com/camunda/connectors/issues/2251

### DIFF
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/model/JWTProperties.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/model/JWTProperties.java
@@ -17,6 +17,7 @@ public record JWTProperties(
     @TemplateProperty(
             label = "JWK URL",
             description = "Well-known URL of JWKs",
+            feel = FeelMode.optional,
             group = "authorization")
         @FEEL
         String jwkUrl,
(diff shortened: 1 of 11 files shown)


connectors/webhook/element-templates/webhook-connector-boundary.json
```diff
@@ -199,14 +199,14 @@
     },
     "type" : "Dropdown",
     "choices" : [ {
-      "name" : "None",
-      "value" : "NONE"
-    }, {
```
chillleader, line 204: Looks like you have an outdated version of the template generator plugin installed locally (template diff is not related to the changes in the PR). Please run `mvn clean install` for the whole project locally and then regenerate the templates again

chillleader: Yeah, the dropdown ordering has been changed a while ago and the current template before your diff seems correct (NONE should be the first, as declared in the annotation). I still suspect that somehow you're using an outdated plugin. Could you try purging the related directory in your local Maven repository? `~/m2/repository/io/camunda/connector`

chillleader:
FYI: I fixed a related issue reported by Igor, please rebase before merging https://github.com/camunda/connectors/pull/2286

chillleader (approved):

PR #2279, opened 2024-03-29
chore(slack): migrate slack inbound to generator

igpetrov:
chore(slack): migrate slack inbound to generator

### DIFF
--- a/connectors/slack/src/main/java/io/camunda/connector/slack/inbound/model/SlackWebhookProperties.java
+++ b/connectors/slack/src/main/java/io/camunda/connector/slack/inbound/model/SlackWebhookProperties.java
@@ -8,14 +8,41 @@ package io.camunda.connector.slack.inbound.model;
 
 import com.slack.api.app_backend.SlackSignature;
 import io.camunda.connector.api.inbound.webhook.VerifiableWebhook;
+import io.camunda.connector.generator.dsl.Property;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import io.camunda.connector.generator.java.annotation.TemplateProperty.PropertyType;
+import jakarta.validation.constraints.NotBlank;
 import java.util.Map;
 import java.util.function.Function;
 
 public record SlackWebhookProperties(
-    String context,
-    String slackSigningSecret,
-    Function<Map<String, Object>, VerifiableWebhook.WebhookHttpVerificationResult>
-        verificationExpression) {
+    @TemplateProperty(
+            id = "context",
+            label = "Webhook ID",
+            group = "endpoint",
+            description = "The webhook ID is a part of the URL endpoint",
+            feel = Property.FeelMode.disabled)
+        @NotBlank
+        String context,
+    @TemplateProperty(
+            id = "slackSigningSecret",
+            label = "Slack signing secret",
+            group = "endpoint",
+            description =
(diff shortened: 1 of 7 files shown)


connectors/slack/src/main/java/io/camunda/connector/slack/inbound/model/SlackWebhookProperties.java
```diff
@@ -8,14 +8,39 @@
...
+            group = "endpoint",
+            description =
+                "Slack signing secret. <a href='https://api.slack.com/authentication/verifying-requests-from-slack' target='_blank'>See documentation</a> regarding the Slack signing secret")
+        @NotBlank
+        String slackSigningSecret,
+    @TemplateProperty(
+            id = "verificationExpression",
+            group = "endpoint",
+            type = PropertyType.Hidden,
+            feel = Property.FeelMode.disabled,
```
chillleader, line 38: Shouldn't it be `required`? Looks like we're expecting a FEEL expression here

connectors/slack/src/main/java/io/camunda/connector/slack/inbound/model/SlackWebhookProperties.java
```diff
@@ -8,14 +8,39 @@
...
 public record SlackWebhookProperties(
-    String context,
-    String slackSigningSecret,
-    Function<Map<String, Object>, VerifiableWebhook.WebhookHttpVerificationResult>
-        verificationExpression) {
+    @TemplateProperty(
+            id = "context",
+            label = "Webhook ID",
+            group = "endpoint",
+            description = "The webhook ID is a part of the URL endpoint")
```
chillleader, line 23: I would also add `feel = FeelMode.disabled` to this property and `slackSigningSecret`. We can change the default behavior for inbound connectors to set `disabled` (As we discussed last week with @Oleksiivanov)

chillleader: I created this issue, let's rather change the default value: https://github.com/camunda/connectors/issues/2282

I will tackle it later this week (currently focused on a different task), or feel free to take the lead if you feel like it :)

connectors/slack/src/main/java/io/camunda/connector/slack/inbound/model/SlackWebhookProperties.java
igpetrov: The field is hidden, so it doesn't really matter, and if we remove disabled, it sets `feel: optional` which is a bit confusing. In general, Slack has very specific verification expression, so we sort of preventing user to mess with it.

connectors/slack/src/main/java/io/camunda/connector/slack/inbound/model/SlackWebhookProperties.java
igpetrov: Good point, I'll change that to `disabled`

igpetrov: @chillleader it seems like `disabled` is not working for inbound connectors. Setting `feel = Property.FeelMode.disabled` had no effect. Except the field is `hidden`, then it works. 🤔

chillleader: > Setting feel = Property.FeelMode.disabled had no effect.

Good catch, I could confirm the issue - here is the proposed fix: https://github.com/camunda/connectors/pull/2286

PR #2288, opened 2024-04-02
fix(generator): standardize line endings in generated element templates

markfarkas-camunda:
## Description

On windows machines the generator used CRLF line endings instead of LF, so the generated files were all different. In this PR we specify the line endings explicitly, so it will be the same on Windows as well.


chillleader (approved):
Thanks!

PR #2291, opened 2024-04-02
Support POSTing application-x-www-form-urlencoded Content-Type

upgradingdave:
## Description

Its currently not possible make a POST request and send data in the body with `Content-Type` of `application/x-www-form-urlencoded`. 

The unqork api requires this and I think other folks can benefit from the ability as well. 

The way I implemented this is that if there's a header set like this: 

```
{
  "Content-Type": "application/x-www-form-urlencoded"
}
```

Then a body like this will be interpreted correctly: 

```
{
  "grant_type":"password",
  "username":unqorkUsername,
  "password":unqorkPassword
}
```

Screen Shot 2024-04-02 at 5 57 55 PM

## Related issues

[camunda-8-rest-connector-for-body-other-than-json](https://forum.camunda.io/t/camunda-8-rest-connector-for-body-other-than-json/43166)

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/HttpRequestMapper.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/HttpRequestMapper.java
@@ -21,7 +21,6 @@ import static io.camunda.connector.http.base.constants.Constants.APPLICATION_X_W
 import static org.apache.http.entity.ContentType.APPLICATION_FORM_URLENCODED;
 import static org.apache.http.entity.ContentType.APPLICATION_JSON;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.client.http.GenericUrl;
 import com.google.api.client.http.HttpContent;
 import com.google.api.client.http.HttpHeaders;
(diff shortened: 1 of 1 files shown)


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/HttpRequestMapper.java
```diff
@@ -92,12 +98,12 @@ public static HttpRequest toHttpRequest(
     }
 
     HttpContent content = null;
-    if (request.getMethod().supportsBody) {
-      if (APPLICATION_FORM_URLENCODED.getMimeType().equalsIgnoreCase(headers.getContentType())) {
+    if (request.getMethod().supportsBody && request.hasBody()) {
+      if (APPLICATION_FORM_URLENCODED.getMimeType().equalsIgnoreCase(headers.getContentType())
+          && request.getBody() instanceof LinkedHashMap) {
```
sbuettner, line 103:
```suggestion
          && request.getBody() instanceof Map) {
```

sbuettner:
@upgradingdave Thanks for the contribution. Would you add a test (possibly using Wiremock) that demonstrates that the approach works? You can for example create one based on our e2e test suite: https://github.com/camunda/connectors/blob/main/connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/HttpTests.java#L97

Or inside the Connector project itself of course.

sbuettner:
@johnBgood can we close this one as its covered by https://github.com/camunda/connectors/pull/2660

johnBgood:
@sbuettner sure 💯

sbuettner:
Closing this as it will be covered by the new REST Connector implementation.

PR #2296, opened 2024-04-04
chore(github): fix duplicated property IDs in template

Oleksiivanov:
## Description

fix duplicated property IDs in template of GitHub connector

## Related issues

https://github.com/camunda/team-connectors/issues/708


johnBgood (approved):
Tested locally, lists are all the same 👍

PR #2297, opened 2024-04-04
chore(gitLab): fix duplicated property IDs in template

Oleksiivanov:
## Description

 fix duplicated property IDs in template of GitLab connector

## Related issues

https://github.com/camunda/team-connectors/issues/708


igpetrov (approved):
LGTM, consider bumping version

PR #2302, opened 2024-04-05
fix(rest api): specify request param names explicitly

### LINKED ISSUE #1869: After spring-boot version update, some of the endpoints throw exception
__Describe the Bug__

Calling the `/inbound` endpoint results in the exception below:

```
Servlet.service() for servlet [dispatcherServlet] in context with path [/aaaaaaaa-aaaa-4444-8888-877e6a418b53] threw exception [Request processing failed: java.lang.IllegalArgumentException: Name for argument of type [java.lang.String] not specified, and parameter name information not available via reflection. Ensure that the compiler uses the '-parameters' flag.] with root cause
java.lang.IllegalArgumentException: Name for argument of type [java.lang.String] not specified, and parameter name information not available via reflection. Ensure that the compiler uses the '-parameters' flag.
```

__Steps to Reproduce__

1. Call the /inbound endpoint
2. Observe the error response

__Expected Behavior__

Should work exactly as it did before the spring-boot version update.

__Environment__

N/A

__Additional context__

See https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-6.x#parameter-name-retention

chillleader:
## Description

Specify the parameter names explicitly (after Spring Boot version upgrade)

## Related issues

closes #1869

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/lifecycle/InboundConnectorRestController.java
@@ -42,9 +42,9 @@ public class InboundConnectorRestController {
 
   @GetMapping("/inbound")
   public List<ActiveInboundConnectorResponse> getActiveInboundConnectors(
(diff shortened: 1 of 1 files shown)


johnBgood (approved):
🔥

PR #2306, opened 2024-04-09
feat(azure,openai): introduced Azure OpenAI prompt connector

igpetrov:
feat(azure,openai): introduced Azure OpenAI prompt connector

### DIFF
--- a/connectors/microsoft/README.md
+++ b/connectors/microsoft/README.md
@@ -3,6 +3,7 @@
 The Microsoft Connectors allow you to:
 
 * Interact with Office 365 Mail
+* Interact with Azure OpenAI models
 
 This Connector reuses the base implementation of [REST Connector](./../http/rest) by providing a compatible element template.
 
--- a/connectors/microsoft/azure-open-ai/element-templates/azure-open-ai-connector.json
+++ b/connectors/microsoft/azure-open-ai/element-templates/azure-open-ai-connector.json
@@ -1,0 +1,1136 @@
+{
+  "$schema" : "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
+  "name" : "Azure OpenAI Connector",
+  "id" : "io.camunda.connectors.AzureOpenAI.outbound.v1",
+  "description" : "Interact with Azure OpenAI from your BPMN process",
+  "version" : 1,
+  "category" : {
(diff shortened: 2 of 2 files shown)


johnBgood:
LGTM overall, but it's not exactly consistent with the other OpenAI Connector. You can check how `[internal_systemMessage](https://github.com/camunda/connectors/blob/d068effc9d553dc0e7495d8fef0b0b09ae10354d/connectors/openai/element-templates/openai-connector.json#L249)` is handled for instance.
We're not asking for the complete Json ({"role":"system".....}), we're [creating](https://github.com/camunda/connectors/blob/d068effc9d553dc0e7495d8fef0b0b09ae10354d/connectors/openai/element-templates/openai-connector.json#L324) it instead.

Not sure but I guess we should have the same behavior for all our OpenAI connectors, wdyt? 
Maybe @YanaSegal @bastiankoerber have an opinion as well?

Screenshot 2024-04-10 at 09 51 56

igpetrov:
> LGTM overall, but it's not exactly consistent with the other OpenAI Connector. You can check how [internal_systemMessage](https://github.com/camunda/connectors/blob/d068effc9d553dc0e7495d8fef0b0b09ae10354d/connectors/openai/element-templates/openai-connector.json#L249) is handled for instance. We're not asking for the complete Json ({"role":"system".....}), we're [creating](https://github.com/camunda/connectors/blob/d068effc9d553dc0e7495d8fef0b0b09ae10354d/connectors/openai/element-templates/openai-connector.json#L324) it instead.
> 
> Not sure but I guess we should have the same behavior for all our OpenAI connectors, wdyt? Maybe @YanaSegal @bastiankoerber have an opinion as well?

Good point. We already discussed that previously with @bastiankoerber.
The difference between 'classical' OpenAI / ChatGPT and current connector is that this one is as much technical as possible. Besides, AFAIK, system message adds `system` role, however, Azure OpenAI allows to pass `system`, `user`, `assistant`, `tool`, or `function` roles.

johnBgood (dismissed):
Happy to approve then :)

bastiankoerber:
can you check required fields and remove the prompt?

bastiankoerber:
looks good, i would just remove the prefilled messages / prompts

bastiankoerber (approved):

PR #2308, opened 2024-04-10
fix(rabbitmq): fix auth regex

igpetrov:
fix(rabbitmq): fix auth regex

Also related to https://github.com/camunda/web-modeler/pull/8943


johnBgood (approved):
Sorry I probably broke that during the migration :/

Issue #2060, opened 2024-02-29
[FR] Status code support in response expression


igpetrov:
## About

A Connectors user requested a status code in the response expression: https://forum.camunda.io/t/webhook-connector-response-statuscode/50684/2.

sbuettner:
As discussed in the architecture session, we will support a dedicated new property `responseExpression`:
```
{
  headers: {} 
  body: {}
  statusCode: 201
}
```

We will keep supporting the existing `responseBodyExpression` and will deprecate it at a later point in time.

abdulazizali77:
@igpetrov id like to work on this

igpetrov:
Hi @abdulazizali77 

Contributions are always welcome!

igpetrov:
Hi @abdulazizali77 

Do you have an ETA for a PR?

sbuettner:
https://github.com/camunda/team-connectors/issues/741

PR #2309, opened 2024-04-11
feat(webhook): Support response expressions for Webhooks

sbuettner:
Refactor verification and response body expression handling

## Description

Support a `responseExpression` property that allow users to define the whole response including the body, status code and headers. Its backwards compatible and supports the already existing `responseBodyExpression`.

## Related issues

closes https://github.com/camunda/connectors/issues/2060


connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
```diff
@@ -111,6 +110,15 @@ public record WebhookConnectorProperties(
         @FEEL
         HMACScope[] hmacScopes,
     WebhookAuthorization auth,
+    @TemplateProperty(
+            id = "responseExpression",
+            label = "Response expression",
+            type = PropertyType.Text,
```
johnBgood, line 116: When should we use Text and when should we use String, out of curiosity?

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
```diff
@@ -111,6 +110,15 @@ public record WebhookConnectorProperties(
...
+            group = "webhookResponse",
+            description = "Expression used to generate the HTTP response",
+            feel = FeelMode.required,
+            optional = true)
+        Function<WebhookResultContext, WebhookHttpResponse> responseExpression,
```
johnBgood, line 121: It's probably done/planned already, but we might consider updating https://docs.camunda.io/docs/components/connectors/protocol/http-webhook/#response-body-expression

connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
```diff
@@ -83,65 +86,99 @@
...
+  private boolean webhookSignatureIsValid(WebhookProcessingPayload payload) {
+    try {
+      if (shouldValidateHmac()) {
+        HMACEncodingStrategy strategy =
+            HMACEncodingStrategyFactory.getStrategy(props.hmacScopes(), payload.method());
+        byte[] bytesToSign = strategy.getBytesToSign(payload);
+        return validateHmacSignature(bytesToSign, payload);
+      }
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
johnBgood, line 181: I don't have much context, but if some error is thrown here, does that make sense to consider the signature invalid and `return false;` instead of propagating the exception? Or is it a real error case that could be solved otherwise?

johnBgood (dismissed):
Curiosity comments, LGTM 🏑

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
sbuettner: Planned, yes.

connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
sbuettner: I just moved the code and did not want to introduce additional changes in this PR.

connectors/webhook/src/main/java/io/camunda/connector/inbound/model/WebhookConnectorProperties.java
sbuettner: I think there is a difference regarding rendering but I would have to look it up.

sbuettner: https://docs.camunda.io/docs/components/modeler/desktop-modeler/element-templates/defining-templates/#types

johnBgood (approved):
💯

Issue #2133, opened 2024-03-07
Add `deduplicationId` to inbound connector templates using the template generator


chillleader:
__What should we do?__

As an outcome of #2132, a new common property called `deduplicationId` for all inbound connectors will be supported by the connector runtime.

We need to add this property to the template generator, thus propagating the change to all inbound connector templates that use the template generator.

🛑 **Was blocked by #2132**

This issue is part of the following epic: https://github.com/camunda/product-hub/issues/1996

chillleader:
We still need to update Kafka and SQS connector templates

PR #2310, opened 2024-04-12
feat(generator): support inbound deduplication; boolean props; enhanced conditions

chillleader:
## Description

This PR implements inbound deduplication ID support to the element template generator, along with some auxiliary functionality related to new features of the element template JSON schema.

Screenshot 2024-04-16 at 09 27 51

<br>
Screenshot 2024-04-16 at 09 16 43
<br>
Screenshot 2024-04-16 at 09 28 21

Key features:
- Deduplication properties are added to generated inbound connector templates
- They are hidden behind a feature flag that can be enabled in the `pom.xml`
- Boolean values (chechboxes) are now supported in the template generator
- Tooltips can be added to grop labels and properties
- New condition types are supported, including `allMatch` and `isActive` (the latter one determines whether property is visible and not hidden by other conditions).

## Related issues

closes https://github.com/camunda/connectors/issues/2133

### DIFF
--- a/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
+++ b/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
@@ -22,6 +22,7 @@ import io.camunda.connector.generator.api.ElementTemplateGenerator;
 import io.camunda.connector.generator.api.GeneratorConfiguration;
 import io.camunda.connector.generator.api.GeneratorConfiguration.ConnectorElementType;
 import io.camunda.connector.generator.api.GeneratorConfiguration.ConnectorMode;
+import io.camunda.connector.generator.api.GeneratorConfiguration.GenerationFeature;
 import io.camunda.connector.generator.dsl.BpmnType;
 import io.camunda.connector.generator.dsl.ElementTemplateBuilder;
 import io.camunda.connector.generator.dsl.ElementTemplateIcon;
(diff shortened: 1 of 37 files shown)


chillleader:
fyi @YanaSegal design vs first iteration, feel free to leave comments 🙂 

image

Screenshot 2024-04-16 at 09 16 43
<br>
Screenshot 2024-04-16 at 09 28 21

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
```diff
@@ -86,6 +87,8 @@ public List<io.camunda.connector.generator.dsl.ElementTemplate> generate(
             PropertyGroup.builder()
                 .id(group.id())
                 .label(group.label())
+                .tooltip(group.tooltip().isBlank() ? null : group.tooltip())
+                .openByDefault(group.openByDefault() == Boolean.TRUE ? null : false)
```
johnBgood, line 91: we return `null` when the property is `TRUE` ?

johnBgood (approved):
Amazing new features thank you 🥇

YanaSegal:
Looks good. 
Is it possible to add placeholder text in the ID field? If so, we can use this space to guide the user by listing the supported characters. 
image

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
chillleader: If we don't specify anything then it is assumed to be true by default, so I'd like to avoid adding this parameter to every property in every connector 🙂

chillleader:
> Is it possible to add placeholder text in the ID field?

Unfortunately no, but I can add another validation rule that will just limit the set of characters the user can input, and show an error message similar to the one in the screenshot in case the field contains some forbidden chars.

YanaSegal:
@chillleader, that sounds good. And please, let's cover this in the docs so that if users are unsure what they are doing wrong, they can find guidelines in the documentation. 
Thanks :)

johnBgood (approved):

PR #2319, opened 2024-04-15
deps(jdbc): Bump connector snapshot version

chillleader:
## Description

Version was bumped globally in #2304 but there was a time gap between PR creation and merging where the JDBC connector merge has happened, so it was not included in that PR.


johnBgood:
Is it ok if we close this one? I'll commit this change in my opened PR

chillleader:
@johnBgood I just thought to bump it separately because it messes up with my local build for some reason :)

johnBgood (approved):
Let's merge it then, thanks :)

PR #2332, opened 2024-04-22
fix(webhook): incorrect deactivation handling

chillleader:
## Description

Fix for the incorrect comparison during webhook connector deactivation that resulted in webhooks never being properly deactivated.

### DIFF
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerPlainJavaTests.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerPlainJavaTests.java
@@ -42,7 +42,6 @@ import io.camunda.connector.runtime.inbound.webhook.WebhookConnectorRegistry;
 import io.camunda.connector.validation.impl.DefaultValidationProvider;
 import java.util.List;
 import java.util.Map;
-import java.util.UUID;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerPlainJavaTests.java
```diff
@@ -163,7 +166,7 @@ public static InboundConnectorContextImpl buildContext(InboundConnectorDetails d
   public static InboundConnectorDetails webhookDefinition(
       String bpmnProcessId, int version, String path) {
     return new InboundConnectorDetails(
-        UUID.randomUUID().toString(),
+        bpmnProcessId + version + path,
```
igpetrov, line 169: Where will it be used? If it's user exposed, maybe consider concatenating with dashes or something 🤔 . Just a suggestion.

```suggestion
        bpmnProcessId + "-" + version + "-" + path,
```

igpetrov (approved):
Other than comment mentioned, rest looks good.

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerPlainJavaTests.java
chillleader: It's a mock value for tests only 🙂

PR #2333, opened 2024-04-22
feat(inbound)!: message acknowledgement config

### LINKED ISSUE #2134: Support acknowledgement configuration for inbound message queue connectors

Inbound message queue connectors such as RabbitMQ Connector currently acknowledge all messages that were consumed without errors. This behavior doesn't take into account the fact that successfully consumed events not always lead to message correlation in Zeebe. For instance, if the activation condition did not match, the message would not be correlated, but the message would be consumed and removed from the queue. Therefore, in certain cases a different approach is needed, when a connector only acknowledges messages that lead to message correlation.

Message queue connectors should allow configuring whether messages are acknowledged always or only after successful correlation.

- A new dropdown property is added to the element templates, containing 2 choices: "Acknowledge all" and "Acknowledge only correlated events" (exact naming to be reviewed).
- The following inbound connectors should support this property: Kafka, RabbitMQ, SQS.

__Additional context__

This issue is part of the following epic: https://github.com/camunda/product-hub/issues/2157

chillleader:
## Description

Adds message acknowledgement properties to the element template and the corresponding section to the connector result

## Related issues

closes #2134

### DIFF
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
@@ -28,7 +28,6 @@ import static org.mockito.Mockito.when;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.EvictingQueue;
 import io.camunda.connector.api.inbound.CorrelationResult;
-import io.camunda.connector.api.inbound.CorrelationResult.Failure;
 import io.camunda.connector.api.inbound.ProcessElement;
 import io.camunda.connector.api.inbound.webhook.MappedHttpRequest;
 import io.camunda.connector.api.inbound.webhook.WebhookConnectorExecutable;
@@ -185,9 +184,7 @@ class WebhookControllerTestZeebeTests {
 
     var correlationHandlerMock = mock(InboundCorrelationHandler.class);
     when(correlationHandlerMock.correlate(any(), any()))
-        .thenReturn(
-            new CorrelationResult.Failure.ActivationConditionNotMet(
-                Failure.HandlingStrategy.FORWARD_ERROR_TO_UPSTREAM));
+        .thenReturn(new CorrelationResult.Failure.ActivationConditionNotMet(false));
 
     var webhookDef = webhookDefinition("nonExistingProcess", 1, "myPath");
     var webhookContext =
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
@@ -28,6 +28,7 @@ import static org.mockito.Mockito.when;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.EvictingQueue;
 import io.camunda.connector.api.inbound.CorrelationResult;
+import io.camunda.connector.api.inbound.CorrelationResult.Failure;
 import io.camunda.connector.api.inbound.ProcessElement;
 import io.camunda.connector.api.inbound.webhook.MappedHttpRequest;
 import io.camunda.connector.api.inbound.webhook.WebhookConnectorExecutable;
@@ -184,7 +185,9 @@ class WebhookControllerTestZeebeTests {
 
     var correlationHandlerMock = mock(InboundCorrelationHandler.class);
     when(correlationHandlerMock.correlate(any(), any()))
-        .thenReturn(new CorrelationResult.Failure.ActivationConditionNotMet());
+        .thenReturn(
+            new CorrelationResult.Failure.ActivationConditionNotMet(
+                Failure.HandlingStrategy.FORWARD_ERROR_TO_UPSTREAM));
 
     var webhookDef = webhookDefinition("nonExistingProcess", 1, "myPath");
     var webhookContext =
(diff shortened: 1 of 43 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationResult.java
```diff
@@ -40,28 +42,54 @@ record MessageAlreadyCorrelated(@JsonIgnore ProcessElementContext activatedEleme
...
+      return ForwardErrorToUpstream.RETRYABLE;
+    }
+
+    /**
+     * @deprecated Use {@link #handlingStrategy()} instead
+     */
+    @Deprecated(forRemoval = true)
     default boolean isRetryable() {
       return false;
     }
```
chillleader, line 55: Perhaps we can remove it right away without deprecation? I'm not sure if there is someone out there implementing and using it 🙂

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationFailureHandlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+   * the upstream system, however, it is expected that the upstream system receives the error and
+   * can handle it.
+   *
+   * @param isRetryable whether the error is retryable or not. If the error is retryable, the
+   *     upstream system may retry the operation at a later point in time (e.g. by redelivering the
+   *     message).
+   */
+  record ForwardErrorToUpstream(boolean isRetryable) implements CorrelationFailureHandlingStrategy {
+
+    public static ForwardErrorToUpstream RETRYABLE = new ForwardErrorToUpstream(true);
```
johnBgood, line 46:
```suggestion
    public static final ForwardErrorToUpstream RETRYABLE = new ForwardErrorToUpstream(true);
```

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationFailureHandlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+    public static ForwardErrorToUpstream NON_RETRYABLE = new ForwardErrorToUpstream(false);
```
johnBgood, line 47:
```suggestion
    public static final ForwardErrorToUpstream NON_RETRYABLE = new ForwardErrorToUpstream(false);
```

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationResult.java
johnBgood: If they do, it means they have built their own runtime, isn't it? If so, let's just remove it 🔫

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationFailureHandlingStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+ * <ul>
+ *   <li>the type of the failure
+ *   <li>the configuration of the connector provided by the user
+ *   <li>the configuration of the Connector runtime
+ * </ul>
+ */
+public sealed interface CorrelationFailureHandlingStrategy {
+
+  /**
+   * When the connector implementation receives this strategy from the runtime, it should forward
```
johnBgood, line 35: If we don't have any docs, it would nice to add an example. Maybe the initial issue we had with the RabbitMQ connector?

johnBgood: this example should probably on the `correlate` method as this is what the user will use

johnBgood (dismissed):
Great stuff 🥇

connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaConnectorConsumer.java
```diff
@@ -165,6 +170,7 @@ private void pollAndPublish() {
       handleMessage(record);
     }
     if (!records.isEmpty()) {
+      consumer.commitSync();
```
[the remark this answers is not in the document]
chillleader: Good catch!

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationResult.java
chillleader: Not necessarily, they could have just used it in custom connectors, but I'd say the change is not very invasive.

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationFailureHandlingStrategy.java
chillleader: I added this info to the `correlate` method as well 👌

johnBgood (dismissed):
🔥

johnBgood (approved):

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/WebhookControllerTestZeebeTests.java
```diff
@@ -184,7 +184,7 @@ public void testSuccessfulProcessingWithFailedActivation() throws Exception {
 
     var correlationHandlerMock = mock(InboundCorrelationHandler.class);
     when(correlationHandlerMock.correlate(any(), any()))
-        .thenReturn(new CorrelationResult.Failure.ActivationConditionNotMet());
+        .thenReturn(new CorrelationResult.Failure.ActivationConditionNotMet(false));
 
     var webhookDef = webhookDefinition("nonExistingProcess", 1, "myPath");
     var webhookContext =
```
sbuettner, line 190: Can you also add a test case when the message acknowledgement is deactivated so the webhook should fail?

chillleader: Just to clarify, do you think we should always return an error response when activation condition is not met? Or should we also make this configurable, like for RabbitMQ / Kafka?
Maybe we could have a quick sync on Monday to discuss the desired behavior

johnBgood (approved):
👍

chillleader:
I will merge this to unblock related documentation / template repository updates, and test the webhook behavior in a separate issue: https://github.com/camunda/connectors/issues/2596

Issue #2346, opened 2024-04-25
Upgrade to Identity SDK 8.5.0


chillleader:
__What should we do?__

There is an issue in Identity SDK <8.5 that prevents us from authenticating with MS Entra.

We need to make sure that we use the correct version of Identity SDK (>8.5), either by setting the correct version in Spring Zeebe (as we get it from there as a transitive dependency) or by pinning the version in our parent pom.

__Why should we do it?__

To incorporate the fix of https://github.com/camunda-cloud/identity/issues/2418

PR #2344, opened 2024-04-25
move to spring zeebe 8.5.1

jonathanlukas:
## Related issues

closes #

Issue #2343, opened 2024-04-25
Upgrade spring-zeebe version to 8.5.1


chillleader:
__What should we do?__

We should upgrade our spring zeebe client version to 8.5.1 (as soon as it's released).

__Why should we do it?__

There is a bug fix we are interested in (job workers receive no proper name if job streaming is enabled, see [here (🔒 internal)](https://camunda.slack.com/archives/C02JLRNQQ05/p1714044262797939)

PR #2351, opened 2024-04-29
chore: migrate to spring zeebe 8.5.1

### LINKED ISSUE #2313: The connector is an important component of why it doesn't match the zeebe client version(zeebe version 8.5). This version management is terrible!!

1. connector is an important component, why is it so different from the zeebe version? This appears to users to be a serious version management disorder
2. The latest zeebe client configuration is completely different, and the address is changed to grpc-address. However, the connectors still use a mismatched version of the zeebe client. This version management is terrible!!

chillleader:
## Description

Changes to support Spring Zeebe 8.5.x, including the config property format.

Replaces the previous PR: https://github.com/camunda/connectors/pull/2344

## Related issues

closes #2313 #2343 #2346

### DIFF
--- a/bundle/default-bundle/src/test/resources/application.properties
+++ b/bundle/default-bundle/src/test/resources/application.properties
@@ -1,11 +1,4 @@
 server.port=8080
-zeebe.client.broker.gateway-address=localhost:26500
-zeebe.client.security.plaintext=true
-
-# Operate config for use with docker-compose.yml
-camunda.operate.client.url=http://localhost:8081
-#camunda.operate.client.username=demo
-#camunda.operate.client.password=demo
 
 management.context-path=/actuator
 management.endpoints.web.exposure.include=metrics,health,prometheus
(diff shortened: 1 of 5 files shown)


bundle/default-bundle/src/test/resources/application.properties
```diff
@@ -1,31 +1,26 @@
...
-zeebe.client.worker.max-jobs-active=32
-zeebe.client.worker.threads=10
+# Config for use with docker-compose.yml
+camunda.client.mode=oidc
+camunda.client.auth.client-id=connectors
+camunda.client.auth.client-secret=XALaRPl5qwTEItdwCMiPS62nVpKs7dL7
+camunda.client.auth.oidc-type=keycloak
+camunda.client.auth.issuer=http://localhost:18080/auth/realms/camunda-platform
+camunda.client.identity.audience=connectors
+camunda.client.identity.base-url=http://localhost:8084
```
chillleader, line 21: I switched to the new property config because there seems to be an issue with the old one (it refuses to start without additional explicit Zeebe client configuration via env variables, i.e. zeebe client auth is not configured by spring zeebe with the old properties format).

However, the old property format still works (although deprecated), and the `docker-compose` files also work without any adjustments (because there we inject zeebe config via env variables, which solves the problem described above).

To discuss whether we want to replace the property config in `docker-compose`, Helm charts, and SaaS config.

chillleader: We will keep the configuration in Docker compose / Helm / SaaS as is

johnBgood (approved):

PR #2352, opened 2024-04-30
fix(inbound): always return health objects with error set

chillleader:
## Description

Restores compatibility with web modeler's Problems tab by always returning an error object along with the health response. Additionally, the constructors that allow creating Health objects without an Error were removed.


ev-codes (approved):
Looks great, thanks!

Ideally, a team member with more experience with the Connectors SDK would also review this 😄

johnBgood (approved):
👍

PR #2356, opened 2024-04-30
task(connector-exception): Add a possibility to pass custom variables to the ConnectorException

johnBgood:
## Description

Add a possibility to pass custom variables when throwing a `ConnectorException`.

This is possible using the `ConnectorExceptionBuilder`.
This comes in the form of a `Map<String,Object>`.

## Related issues

closes https://github.com/camunda/team-connectors/issues/676

### DIFF
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
@@ -34,6 +34,7 @@ import static org.mockito.Mockito.when;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import io.camunda.connector.api.error.ConnectorException;
+import io.camunda.connector.api.error.ConnectorExceptionBuilder;
 import io.camunda.connector.api.error.ConnectorRetryExceptionBuilder;
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 import io.camunda.connector.runtime.core.ConnectorHelper;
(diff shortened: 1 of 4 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/error/ConnectorExceptionBuilder.java
```diff
@@ -0,0 +1,52 @@
...
+package io.camunda.connector.api.error;
+
+import java.util.Map;
+
+public class ConnectorExceptionBuilder {
+
+  protected String errorCode;
+  protected String message;
+  protected Throwable cause;
+  protected Map<String, Object> errorVariables;
```
sbuettner, line 27: They can be private right?

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
```diff
@@ -806,6 +807,33 @@ void shouldCreateBpmnError_UsingExceptionCodeAndRawContext() {
       assertThat(result.getErrorMessage()).isEqualTo("Message: exception message");
     }
 
+    @Test
+    void shouldCreateBpmnError_UsingExceptionCodeAndErrorVariables() {
+      // given
+      var errorExpression =
+          "if error.code != null then "
+              + "{ \"errorType\": \"bpmnError\", \"code\": error.code, \"message\": \"Message: \" + error.message, \"variables\": error.errorVariables} "
+              + "else {}";
```
sbuettner, line 816: Could you also add a test that verifies that the error variables are successfully passed to the command in https://github.com/camunda/connectors/blob/e277b0f002987e992d16a1384a9e167e381cbe10/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java#L289 when there is no error expression?

connector-sdk/core/src/main/java/io/camunda/connector/api/error/ConnectorExceptionBuilder.java
johnBgood: fixed

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
johnBgood: fixed

sbuettner (approved):

PR #2413, opened 2024-05-02
fix(webhook): report health UP

chillleader:
## Description

Report Health.UP on startup

### DIFF
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
@@ -11,6 +11,7 @@ import static io.camunda.connector.inbound.signature.HMACSwitchCustomerChoice.en
 
 import io.camunda.connector.api.annotation.InboundConnector;
 import io.camunda.connector.api.inbound.Activity;
+import io.camunda.connector.api.inbound.Health;
 import io.camunda.connector.api.inbound.InboundConnectorContext;
 import io.camunda.connector.api.inbound.Severity;
 import io.camunda.connector.api.inbound.webhook.MappedHttpRequest;
@@ -102,6 +103,7 @@ public class HttpWebhookExecutable implements WebhookConnectorExecutable {
     props = new WebhookConnectorProperties(wrappedProps);
     authChecker = WebhookAuthorizationHandler.getHandlerForAuth(props.auth());
     responseExpression = mapResponseExpression();
+    context.reportHealth(Health.up());
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


johnBgood (dismissed):

connectors/aws/aws-sqs/src/test/java/io/camunda/connector/inbound/SqsExecutableTest.java
```diff
@@ -108,7 +108,8 @@ public void activateTest(Map<String, Object> properties) throws InterruptedExcep
   @Test
   public void deactivateTest() {
     // Given
-    consumer = new SqsQueueConsumer(sqsClient, null, null);
+    var context = createConnectorContext(Map.of(), createDefinition());
+    consumer = new SqsQueueConsumer(sqsClient, null, context);
```
johnBgood, line 112: should we assert that Health is down ?

johnBgood (approved):
👍

ev-codes:
@chillleader Thanks for this!

I suggest we keep this PR open until I confirm with Sara whether QA is required. If QA is required, then I think we should only merge once QA is complete.

connectors/aws/aws-sns/src/main/java/io/camunda/connector/sns/inbound/SnsWebhookExecutable.java
```diff
@@ -12,6 +12,7 @@
 import com.amazonaws.services.sns.message.SnsSubscriptionConfirmation;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.connector.api.annotation.InboundConnector;
+import io.camunda.connector.api.inbound.Health;
```
ev-codes, line 15: @chillleader Thanks for these changes!

I suggest we keep this PR open until I confirm with Sara whether QA is required. If QA is required, then I think we should only merge once QA is complete.

ev-codes:
@chillleader since this wasn't causing an issue in 8.5, I think we should remove the backport release/8.5 tag.

ev-codes:
@chillleader sorry one more thing - could you please update the description to indicate why this change is needed?

(i.e., otherwise health will just be stuck on "unknown")

connectors/aws/aws-sns/src/main/java/io/camunda/connector/sns/inbound/SnsWebhookExecutable.java
ev-codes: As discussed in Slack, @slolatte is ok with us merging in to conduct QA.

We can't use Connectors preview environments here because they only work with Self-Managed, but inbound connectors are a SaaS-only feature.

ev-codes (approved):
Thanks for addressing everything, approved!

PR #2442, opened 2024-05-03
fix(webhook): incorrect type check resulting in missing return data

chillleader:
## Description

The type check was invalid and didn't account for subclasses, which lead to missing `data.path` in the `/inbound` endpoint response for webhook connectors.

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
@@ -83,7 +83,7 @@ public class InboundConnectorRestController {
 
   private Map<String, Object> getData(ActiveExecutableResponse connector) {
     Map<String, Object> data = Map.of();
-    if (WebhookConnectorExecutable.class.equals(connector.executableClass())) {
+    if (WebhookConnectorExecutable.class.isAssignableFrom(connector.executableClass())) {
       try {
         var properties = connector.elements().getFirst().rawPropertiesWithoutKeywords();
         var contextPath = properties.get("inbound.context");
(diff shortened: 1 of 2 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
```diff
@@ -83,7 +83,7 @@ private List<ActiveInboundConnectorResponse> getActiveInboundConnectors(
 
   private Map<String, Object> getData(ActiveExecutableResponse connector) {
     Map<String, Object> data = Map.of();
-    if (WebhookConnectorExecutable.class.equals(connector.executableClass())) {
+    if (WebhookConnectorExecutable.class.isAssignableFrom(connector.executableClass())) {
```
ev-codes, line 86: Could you write a quick unit test that would catch the issue you discuss in PR description?

ev-codes (commented):
Source code change itself looks good, just one comment 😄

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
chillleader: Sure

igpetrov (approved):

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/InboundEndpointTest.java
```diff
@@ -0,0 +1,76 @@
...
+  static class TestWebhookExecutable implements WebhookConnectorExecutable {
+
+    @Override
+    public WebhookResult triggerWebhook(WebhookProcessingPayload payload) throws Exception {
+      return null;
+    }
+  }
+
+  @Test
+  public void testInboundEndpointResponse() {
```
ev-codes, line 51: nitpick: consider renaming this test to reflect what the input is and what the expected result is.

e.g., `testDataReturnedForWebhookConnectorExecutableSubclass`

ev-codes (approved):

PR #2451, opened 2024-05-03
ci(release): support existing tag, clarify Maven staging

ev-codes:
### Checkout existing tag

target_commitish only exists when tag doesn't exist already when the release is created.

While this is often useful, there are times when the tag already exists.

For example, if someone edits an existing release instead of creates a new one. Or if the GitHub release needs to be re-created for some reason.

### Clarify Maven Central (Staging)

To help clarify that this will not publish the artifact directly to the public - an additional step is needed as documented internally.

### ~Run Helm Integration tests~

This is more complicated than expected. We will address this in a separate issue and PR.

## Related issues

closes #

### DIFF
--- a/.github/workflows/RELEASE.yaml
+++ b/.github/workflows/RELEASE.yaml
@@ -5,12 +5,6 @@ on:
   release:
     types: [ created ]
 
-defaults:
-  run:
-    # use bash shell by default to ensure pipefail behavior is the default
-    # see https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference
-    shell: bash
-
 jobs:
   build-and-push:
     name: Build and push Docker images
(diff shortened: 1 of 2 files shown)


.github/workflows/RELEASE.yaml
```diff
@@ -264,3 +270,15 @@ jobs:
             bundle/default-bundle/target/connectors-bundle-sbom.xml
             connectors-bundle-templates-${{ github.event.release.tag_name }}.tar.gz
             connectors-bundle-templates-${{ github.event.release.tag_name }}.zip
+
+  helm-deploy:
```
sbuettner, line 274: @ev-codes can we rather invoke/integrate the already [existing action](https://github.com/camunda/connectors/blob/main/.github/workflows/INTEGRATION_TEST.yml#L4) instead of duplicating it? 

Docs:  https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

ev-codes: @sbuettner does the change I just pushed address your concerns?

Please don't merge yet as I need to test this new configuration.

ev-codes: @sbuettner I'm going to handle the Helm integration test in another PR.

As we discussed elsewhere, the Helm integration test will be more complicated to implement to work with patch versions.

sbuettner (approved):

PR #2474, opened 2024-05-06
fix(inbound): status report not shown for unhealthy connectors

chillleader:
## Description

This edge case was not handled:
- the executable registry cannot return the executable class with the status response because the executable doesn't exist
- the executable doesn't exist because of an activation failure

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
@@ -83,7 +83,8 @@ public class InboundConnectorRestController {
 
   private Map<String, Object> getData(ActiveExecutableResponse connector) {
     Map<String, Object> data = Map.of();
-    if (WebhookConnectorExecutable.class.isAssignableFrom(connector.executableClass())) {
+    if (connector.executableClass() != null
+        && WebhookConnectorExecutable.class.isAssignableFrom(connector.executableClass())) {
       try {
         var properties = connector.elements().getFirst().rawPropertiesWithoutKeywords();
         var contextPath = properties.get("inbound.context");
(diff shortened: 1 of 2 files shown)


connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/InboundEndpointTest.java
```diff
@@ -73,4 +73,30 @@ public void testDataReturnedForWebhookConnectorExecutableSubclass() {
...
+    when(executableRegistry.query(any()))
+        .thenReturn(
+            List.of(
+                new ActiveExecutableResponse(
+                    UUID.randomUUID(),
+                    null, // executable class is null
+                    List.of(
+                        new InboundConnectorElement(
+                            Map.of("inbound.context", "myPath", "inbound.type", "webhook"),
+                            new StandaloneMessageCorrelationPoint(
```
ev-codes, line 89: FYI for the 8.6 backport PR (mentioning just in case):

You'll need to edit this to use the 3-arg constructor that's available in the `release/8.6`

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
```diff
@@ -83,7 +83,8 @@ private List<ActiveInboundConnectorResponse> getActiveInboundConnectors(
 
   private Map<String, Object> getData(ActiveExecutableResponse connector) {
     Map<String, Object> data = Map.of();
-    if (WebhookConnectorExecutable.class.isAssignableFrom(connector.executableClass())) {
+    if (connector.executableClass() != null
```
ev-codes, line 86: nitpick: for readability, consider assigning `connector.executableClass()` as a variable above line 86, then use that variable in lines 86 and 87.

If you disagree, feel free to ignore 😄

ev-codes (approved):
Awesome, thanks for fixing!

Left two non-blocking comments.

connector-runtime/spring-boot-starter-camunda-connectors/src/test/java/io/camunda/connector/runtime/inbound/InboundEndpointTest.java
chillleader: Thanks for pointing this out!

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java
chillleader: Will do

ev-codes (approved):
Thank you!

PR #2502, opened 2024-05-07
feat(jdbc): Remove MySQL ConnectorJ and use MariaDb instead

johnBgood:
## Description

We can't use GPL license, so we needed to replace the MySQL connector with the compatible MariaDb connector.

This won't change anything for our end users, and to accomplish that we needed to handle the compatibility part for them.
Thus we are adding the `permitMysqlScheme` at the end of the connection string.

Note that we still need the MySQL connector for our tests, as _testcontainers_ is looking for this driver to start a MySQL server.

### DIFF
--- a/connectors/jdbc/pom.xml
+++ b/connectors/jdbc/pom.xml
@@ -51,17 +51,24 @@
     </dependency>
 
     <dependency>
-      <groupId>com.mysql</groupId>
-      <artifactId>mysql-connector-j</artifactId>
-      <version>8.4.0</version>
-    </dependency>
-
-    <dependency>
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
       <version>42.7.3</version>
     </dependency>
 
+    <!-- Mariadb is used because of an issue with the MySQL ConnectorJ licence (GPL). We're not mentioning Mariadb in our docs -->
+    <dependency>
+      <groupId>org.mariadb.jdbc</groupId>
+      <artifactId>mariadb-java-client</artifactId>
+      <version>3.3.3</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.mysql</groupId>
+      <artifactId>mysql-connector-j</artifactId>
+      <version>8.4.0</version>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>postgresql</artifactId>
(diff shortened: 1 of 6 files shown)


connectors/jdbc/pom.xml
```diff
@@ -50,18 +50,25 @@
...
+    <!-- Mariadb is used because of an issue with the MySQL ConnectorJ licence (GPL). We're not mentioning Mariadb in our docs -->
+    <dependency>
+      <groupId>org.mariadb.jdbc</groupId>
+      <artifactId>mariadb-java-client</artifactId>
+      <version>3.3.3</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.mysql</groupId>
+      <artifactId>mysql-connector-j</artifactId>
```
ev-codes, line 68: Double-checking: is this test dependency still here intentionally?

If so, I'm curious why we still need it.

This is a non-blocking comment 😄

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/utils/ConnectionHelper.java
```diff
@@ -27,13 +29,33 @@ public static Connection openConnection(JdbcRequest request) {
       JdbcConnection connection = request.connection();
       Connection conn =
           DriverManager.getConnection(
-              connection.getConnectionString(request.database()), connection.getProperties());
+              ensureMySQLCompatibleUrl(
+                  connection.getConnectionString(request.database()), request.database()),
```
ev-codes, line 33: nitpick: for readability and easier debugging later, consider defining `request.database()` as a variable earlier.

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/utils/ConnectionParameterHelper.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public static String addQueryParameterToURL(String urlString, String paramName)
+      throws URISyntaxException {
+    return addQueryParameterToURL(urlString, paramName, null);
+  }
+
+  public static String addQueryParameterToURL(String urlString, String paramName, String paramValue)
+      throws URISyntaxException {
+    URI uri = new URI(urlString);
+    // Check if the URL already has query parameters
```
ev-codes, line 28: Nice comments here, thanks 😄

ev-codes (commented):
Thanks for this!

Looks good overall, just a few comments 😄

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/utils/ConnectionHelper.java
johnBgood: good point, done

connectors/jdbc/pom.xml
johnBgood: scope is `test` so I assume it won't be exported in our artifact.
I left a comment in the PR description you probably missed it :'(

ev-codes: Indeed I missed it in my rush 😢 

Thanks!

ev-codes (approved):
Great, thank you!

PR #2504, opened 2024-05-07
fix(inbound): invalid inbound connector definitions are not stored

chillleader:
## Description

This PR introduces an additional mechanism for capturing errors related to invalid connector configuraitons that happen before creating inbound connector executables (for example, when multiple connectors that have different properties are combined using the same deduplication ID).

Instead of discarding such errors, they are now captured in an `InvalidInboundConnectorDetails` object. The `InboundConnectorDetails` class became a sealed interface. The new sealed hierarchy is needed because invalid inbound connector details won't have the same fields as the valid ones (e.g. `properties` will be missing, as they might be different for connector elements in an invalid group).

Additionally, this PR changes the validation error message in the element templates that support deduplication. The reason behind this change is that the properties panel shows an error in the following format: `<property label> <error message>`, so the new message fits more naturally.

## Related issues

closes https://github.com/camunda/team-connectors/issues/762

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
@@ -244,7 +244,7 @@ public class InboundExecutableRegistryImpl implements InboundExecutableRegistry 
     var type =
         switch (executable) {
           case Activated activated -> activated.context().getDefinition().type();
-          case FailedToActivate failed -> failed.data().connectorElements().getFirst();
+          case FailedToActivate failed -> failed.data().connectorElements().getFirst().type();
           case ConnectorNotRegistered notRegistered -> notRegistered.data().type();
           case InvalidDefinition invalid -> invalid.data().connectorElements().getFirst().type();
         };
@@ -252,13 +252,26 @@ public class InboundExecutableRegistryImpl implements InboundExecutableRegistry 
     return elements.stream()
         .anyMatch(
             element ->
-                query.bpmnProcessId() == null
-                    || query.bpmnProcessId().equals(element.bpmnProcessId())
-                        && (query.type() == null || type == null || query.type().equals(type))
-                        && (query.tenantId() == null
-                            || query.tenantId().equals(element.tenantId())
-                                && (query.elementId() == null
-                                    || query.elementId().equals(element.elementId()))));
+                processIdMatches(element, query)
+                    && typeMatches(type, query)
+                    && tenantIdMatches(element, query)
+                    && elementIdMatches(element.elementId(), query));
+  }
+
+  private boolean processIdMatches(ProcessElement element, ActiveExecutableQuery query) {
+    return query.bpmnProcessId() == null || query.bpmnProcessId().equals(element.bpmnProcessId());
+  }
+
+  private boolean tenantIdMatches(ProcessElement element, ActiveExecutableQuery query) {
+    return query.tenantId() == null || query.tenantId().equals(element.tenantId());
+  }
+
+  private boolean typeMatches(String type, ActiveExecutableQuery query) {
+    return query.type() == null || type == null || query.type().equals(type);
+  }
+
+  private boolean elementIdMatches(String elementId, ActiveExecutableQuery query) {
+    return query.elementId() == null || query.elementId().equals(elementId);
   }
 
   private ActiveExecutableResponse mapToResponse(UUID id, RegisteredExecutable connector) {


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDetails.java
```diff
@@ -22,51 +22,88 @@
 import java.util.Map;
 
 /** Group of inbound connector elements that share the same deduplication ID. */
-public record InboundConnectorDetails(
-    String type,
-    String tenantId,
-    String deduplicationId,
-    @JsonIgnore Map<String, String> rawPropertiesWithoutKeywords,
-    List<InboundConnectorElement> connectorElements) {
+public sealed interface InboundConnectorDetails {
```
ev-codes, line 25: Non-blocking questions for my own understanding: 
- Why not use a `Factory`-type pattern here (like used elsewhere in the codebase)?
- Why use a sealed interface solely (it seems) to allow inheritors to selectively implement the interface? Couldn't the invalid implementation just implement the `properties()` as returning null, for example?
- Why mix validation logic with record constructors below instead of separating the validation concern into another class?

I could see good arguments for them, I'd just like to understand a bit better 😄

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryTest.java
```diff
@@ -0,0 +1,150 @@
...
+  @BeforeEach
+  public void prepareMocks() {
+    factory = mock(InboundConnectorFactory.class);
+    contextFactory = mock(DefaultInboundConnectorContextFactory.class);
+    batchProcessor = new BatchExecutableProcessor(factory, contextFactory, null, null);
+    registry = new InboundExecutableRegistryImpl(factory, batchProcessor);
+  }
+
+  @Test
+  public void invalidDeduplicationConfig_shouldYieldInvalidDefinition() {
```
ev-codes, line 59: Nice tests here, thank you!

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
```diff
@@ -236,20 +237,24 @@ private boolean matchesQuery(RegisteredExecutable executable, ActiveExecutableQu
...
-          case FailedToActivate failed -> failed.data().type();
+          case FailedToActivate failed -> failed.data().connectorElements().getFirst();
           case ConnectorNotRegistered notRegistered -> notRegistered.data().type();
+          case InvalidDefinition invalid -> invalid.data().connectorElements().getFirst().type();
         };
 
     return elements.stream()
         .anyMatch(
             element ->
                 query.bpmnProcessId() == null
```
ev-codes, line 255: Consider extracting this 7-line conditional into its own method.

I could imagine that for a developer used to this codebase, this conditional is very clear. But for new readers, it's a little overwhelming aha 😄

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
```diff
@@ -78,14 +81,24 @@ public Map<UUID, RegisteredExecutable> activateBatch(
...
+        data = (ValidInboundConnectorDetails) maybeValidData;
+      }
 
       final RegisteredExecutable result =
           activateSingle(data, e -> cancellationCallback.accept(e, id));
 
       switch (result) {
         case Activated activated -> alreadyActivated.put(id, activated);
         case ConnectorNotRegistered notRegistered -> alreadyActivated.put(id, notRegistered);
+        case InvalidDefinition invalid -> alreadyActivated.put(id, invalid);
```
ev-codes, line 101: I like how easily this new case fits into the existing switch, nice design work 😄

sbuettner (dismissed):

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDetails.java
chillleader: * A factory pattern would perhaps be an overkill here (at least from my perspective). Unlike in other cases where we use it, we don't have any dependencies on other classes for instance creation (unless we decide to extract validation concern into another class)
* However, I can see how in the current form this code looks like an over-use of sealed interfaces 😄 I started designing this when the inheritors didn't have any common fields and an interface was mostly a marker, but then I moved a few fields to the parent space for convenience. I will try to improve the model here (and also extract the validation concern from the constructor, at least into a separate method).

chillleader: Here's what I ended up doing:
- I extracted the creation & validation logic into a static method in a separate class, `InboundConnectorDetailsUtil`
- I kept the sealed hierarchy because I liked the data consistency benefit that we get (example: I can't create an `InvalidDetails` object without an error, or a valid object without properties). Also, since we're checking the object type in several places, it's nice that we get compile-time validation in `switch` expression that we handled all possible types without adding a `default` branch to the switch expressions.

I hope this is more balanced in terms of readability, usability, and not-overengineering-things 😄

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/inbound/InboundConnectorDetailsTest.java
```diff
@@ -20,6 +20,7 @@
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 import io.camunda.connector.api.inbound.ProcessElement;
```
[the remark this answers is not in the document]
chillleader: I'd say empty/null elements are not important for us, even if we create an `InvalidDetails` object out of them, we will be unable to use it to display the error to the user.

But I added a test for different runtime properties - good point!

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
chillleader: Lol good point, I split it into a few methods for readability

element-template-generator/core/src/main/java/io/camunda/connector/generator/dsl/CommonProperties.java
```diff
@@ -181,7 +181,7 @@ public static PropertyBuilder deduplicationId() {
                 .pattern(
                     new Pattern(
                         "^[a-zA-Z0-9_-]+$",
-                        "Only alphanumeric characters, dashes, and underscores are allowed"))
+                        "can only contain alphanumeric characters, dashes, and underscores"))
```
ev-codes, line 184: @chillleader in the PR description, may you please explain why these element template changes are added?

I think the change makes sense - I just want to make this PR easier to read for future reviewers 😄

ev-codes (approved):
Thank you!

One last little comment, then feel free to merge!

PR #2508, opened 2024-05-08
fix(jdbc): Fix invalid Javadoc comment

ev-codes:
Test file is not accessible from source code. This causes the RC build to fail.


markfarkas-camunda (approved):
LGTM 👍

PR #2511, opened 2024-05-08
fix(inbound): invalid inbound connector definitions are not stored (#…

chillleader:
…2504)

* fix(inbound): invalid definitions are not stored for user convenience

* fix polling connector test

* address review suggestions

* rewrite a condition that's impossible to read

(cherry picked from commit 2866c41afbac4846315400a616e10bcd4a96941a)

## Related issues

closes #


ev-codes (approved):
Thanks!

PR #2513, opened 2024-05-08
feat(jdbc): Don't use URI as it breaks when using the hash char

johnBgood:
## Description

Previously I used a `URI` to get some specific part of the provided database URL.
But:
- it didn't provide anything as the scheme (jdbc:xxx) is not properly detected, so we can't get the query parameters anyway
- it cause an issue when a URL parameter contains a hash (#) it ignores what is after this hash


sbuettner:
@Oleksiivanov @johnBgood Feel free to merge it so it can be tested against the RDS test databases in our AWS account.

PR #2518, opened 2024-05-13
chore: added parent POM for all connectors e2e tests

Oleksiivanov:
## Description

added parent POM for all connectors e2e tests


sbuettner (approved):
Great, thanks for doing it so quickly!

PR #2525, opened 2024-05-14
feat(jdbc): Add MariaDB support 

johnBgood:
## Description

- Add support for MariaDB
- Fix password encoding issue
- Add the jdbc connector to the bundle
- Fix and reworked integration tests to run a MariaDB container

## Related issues

closes https://github.com/camunda/team-connectors/issues/740
closes https://github.com/camunda/team-connectors/issues/773

### DIFF
--- a/bundle/default-bundle/pom.xml
+++ b/bundle/default-bundle/pom.xml
@@ -1,4 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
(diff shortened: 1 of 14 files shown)


bundle/default-bundle/pom.xml
```diff
@@ -1,4 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
ev-codes, line 1: **(nitpicky) Question:** Why does the XML namespace change here?

I wouldn't expect it to change based on the changes elsewhere in the PR.

ev-codes (commented):
Looks great! Just one small question 😄

bundle/default-bundle/pom.xml
johnBgood: It's coming from my IntelliJ's reordering config. I changed it thanks

bundle/pom.xml
```diff
@@ -1,4 +1,4 @@
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
```
ev-codes, line 1: XML namespace here too :smile:

ev-codes (approved):
One more namespace change to adjust 😄

bundle/pom.xml
johnBgood: now we're good to go I guess 💯

PR #2530, opened 2024-05-14
ci(helm-integration-test): test patches for older minor versions with correct dependencies

ev-codes:
## Description

Previously, `INTEGRATION_TEST.md` would always pass in the default values. This works well for SNAPSHOT's.

However, for patches on minor versions (especially 8.3), the Helm integration tests need to be passed in dependencies that correspond to the that minor version.

This PR will:
- determine the correct dependencies to pass in
- determine the correct Helm chart to use
- call the `integration-test-template.yml` with these inputs

## Related issues

https://github.com/camunda/team-connectors/issues/764


ev-codes:
Closing in favour of https://github.com/camunda/connectors/pull/2612

The Distribution team is changing their `test-integration-template.yaml` to not require us to pass in dependency values manually.

[See this comment on 2612 for more details.](https://github.com/camunda/connectors/pull/2612#issuecomment-2127325056)

ev-codes:
Closing in favour of https://github.com/camunda/connectors/pull/2612

The Distribution team is changing their test-integration-template.yaml to not require us to pass in dependency values manually.

[See this comment for more details.](https://github.com/camunda/connectors/pull/2612#issuecomment-2127325056)

PR #2612, opened 2024-05-22
Helm integration tests for release branches

chillleader:
## Description

Follow up on https://github.com/camunda/connectors/pull/2530.

Previously, INTEGRATION_TEST.md would always pass in the default values. This works well for SNAPSHOT's.

However, for patches on minor versions (especially 8.3), the Helm integration tests need to be passed in dependencies that correspond to the that minor version.

This PR will:

 - determine the correct connectors version to use based on the maven project version (if not provided)
 - determine the correct Helm chart to use (i.e. Helm chart repository git ref)
 - call the integration-test-template.yml with these inputs

ℹ️ Compared to the PR from @ev-codes (#2530), this approach should be simpler because it assumes that we don't need to provide the whole values file in `extraValues` when calling a custom Helm charts branch.

## Related issues

closes https://github.com/camunda/team-connectors/issues/764

### DIFF
--- a/.github/workflows/INTEGRATION_TEST.yml
+++ b/.github/workflows/INTEGRATION_TEST.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Determine current Maven project version
         id: maven-version
         run: |
-          echo "version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)" >> $GITHUB_OUTPUT
+          echo "version=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout -o)" >> $GITHUB_OUTPUT
 
       - name: Determine version of the Connectors image to use
         id: connectors-version
--- a/.github/workflows/INTEGRATION_TEST.yml
+++ b/.github/workflows/INTEGRATION_TEST.yml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
     inputs:
       connectors-version:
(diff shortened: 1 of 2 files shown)


.github/workflows/INTEGRATION_TEST.yml
```diff
@@ -2,12 +2,17 @@
...
+    branches:
+      - release/*
   workflow_dispatch:
     inputs:
-      connectors_version:
-        description: 'The version of the connectors to test'
-        required: true
-        default: 'SNAPSHOT'
+      connectors-version:
+        description: 'The version of the connectors to test. If not provided, the version is determined based on the Maven project version. On main branch, the default is SNAPSHOT'
```
igpetrov, line 11: Suggestion: could also apply some 'branding'

```suggestion
        description: 'The version of the Connectors to test. If not provided, the version is determined based on the Maven project version. On main branch, the default is SNAPSHOT'
```

igpetrov (dismissed):
LGTM

igpetrov (approved):

chillleader:
I'm moving this back to draft state because the Helm integration test GHA will undergo major refactoring and the inputs will change next week. [🔒 Source](https://camunda.slack.com/archives/C06BN4SRB6C/p1716460112152469?thread_ts=1716448671.646989&cid=C06BN4SRB6C)

chillleader:
We're merging this now to benefit from this solution already while the workflow refactoring on the Distro side is still in progress

Issue #1069, opened 2023-08-23
HTTP REST Connector: migrate from Google HTTP Connector


igpetrov:
__What should we do?__

Currently, HTTP REST and GraphQL connectors use Google HTTP Connector.

As a part of this task, we need to migrate to something more mature, such as Apache HTTP client, or other options to be considered.

__Why should we do it?__

As per [discussion](https://github.com/camunda/connectors/pull/1043#discussion_r1301448197), we agreed that it makes sense to do.

sbuettner:
Relevant articles:
https://www.wiremock.io/post/java-http-client-comparison
https://www.innoq.com/en/articles/2022/01/java-http-clients/
https://reflectoring.io/comparison-of-java-http-clients/

Issue #2659, opened 2024-05-24
Extra escape backslash added to AWS Eventbridge Secret


upgradingdave:
__Describe the Bug__

An extra backslash character (`\`) is added to secrets when the secret value also contains a `\+` character sequence. 

This is a problem because AWS will generate secrets and sometimes those secrets contain backslashes `\`. If an extra backslash is added, this breaks the authentication. 

__Steps to Reproduce__

1. Add an AWS Event Bridge Outbound Connector task to a BPM Diagram
2. Configure a value of `123\+45` into the `Secret Key` field of the Credentials Authentication Section
3. Deploy and run the process
4. Note that the `authentication` instance variable for the task has an extra backslash character added: `\`. 

```
{
	"type": "credentials",
	"accessKey": "foo",
	"secretKey": "123\\+45"
}
```

![Screen Shot 2024-05-24 at 1 00 36 PM](https://github.com/camunda/connectors/assets/939229/76b51b15-b0d0-4f31-80ad-815195124a07)

__Expected Behavior__

The secret value should be sent to AWS exactly how it's entered into Web Modeler. No extra backslashes should be added. 

__Environment__

- OS: MacOS Monterey,  Chrome Version 125.0.6422.77 (Official Build) (arm64)
- Library version: Camunda SaaS Production 8.5.1
- Camunda version: Camunda SaaS Production 8.5.1

chillleader:
@upgradingdave I wasn't able to reproduce the issue with 8.5 - although in Operate you will see the backslash with an escape character (that's a limitation in JSON that those have to be escaped), in the connector itself, the backslash is unescaped and passed to AWS in its original format. Is it possible that the authentication issue was caused by something different?

On the other hand, I found that it's not possible to use unescaped backslashes inside connector secrets. Because of how the secret replacement currently works (we serialize variables to string and replace secret syntax in that string), substituting a value that contains an unescaped backslash results in invalid json, and the value cannot be deserialized again. As a workaround, such backslash characters need to be escaped in connector secrets.

FYI @sbuettner

sbuettner:
@chillleader I think we can close this one. WDYT?

chillleader:
Yep, let's close it for now. @upgradingdave please feel free to reopen if you run into a similar issue again

upgradingdave:
[...]

PR #2660, opened 2024-05-27
1069 http rest connector migrate from google http connector

johnBgood:
## Related issues

closes https://github.com/camunda/connectors/issues/1069


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/components/apache/CustomApacheHttpClient.java
```diff
@@ -0,0 +1,136 @@
...
+      throws Exception {
+    var apacheRequest = ApacheRequestFactory.get().createHttpRequest(request);
+    try {
+      HttpCommonResult response =
+          httpClientBuilder
+              .setDefaultRequestConfig(getRequestConfig(request))
+              .build()
+              .execute(apacheRequest, new HttpCommonResultResponseHandler(remoteExecutionEnabled));
+
+      if (response.getStatus() >= 400) {
```
chillleader, line 110: Thinking out loud: if we don't follow redirects, do we also consider a 3XX status an error? Maybe we could also throw an exception in this case to be consistent. WDYT?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/ErrorResponse.java
chillleader, line 1: Can we make it immutable by converting into a record?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonResult.java
```diff
@@ -16,14 +16,16 @@
  */
 package io.camunda.connector.http.base.model;
 
-import com.google.common.base.Objects;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import java.util.Map;
+import java.util.Objects;
 
 public class HttpCommonResult {
```
chillleader, line 23: Can also potentially be converted into a record

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/request/RequestFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.camunda.connector.http.base.request;
+
+import io.camunda.connector.http.base.model.HttpCommonRequest;
+
+public interface RequestFactory<T> {
```
chillleader, line 22: Nice refactoring 👍

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/request/apache/bulider/parts/ApacheRequestBodyBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+          } else {
+            setStringEntity(builder, request);
+          }
+        }
+        case List<?> ignored -> setStringEntity(builder, request);
+        case String ignored -> setStringEntity(builder, request);
+        default ->
+            throw new IllegalStateException(
+                "Unexpected value type for the request body. Excepting Map, List, or String, got "
+                    + request.getBody().getClass().getSimpleName());
```
chillleader, line 61: To my knowledge, FEEL engine can return [other types](https://docs.camunda.io/docs/components/modeler/feel/language-guide/feel-data-types) including primitives (numbers, booleans) and dates. Although obviously not a super common use case, I can imagine that someone puts a FEEL expression into the body that evaluates to one of these types.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/request/apache/bulider/parts/ApacheRequestBodyBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+                            String.valueOf(e.getKey()), String.valueOf(e.getValue())))
+                .collect(Collectors.toList()),
+            StandardCharsets.UTF_8));
+  }
+
+  private void unescapeBody(HttpCommonRequest request) {
+    if (request.getBody() instanceof String) {
+      String unescapedBody = StringEscapeUtils.unescapeJson((String) request.getBody());
+      request.setBody(unescapedBody);
+    }
```
chillleader, line 114: Might also consider unescaping header values. Although not directly related, [this issue](https://github.com/camunda/connectors/issues/2659) shows that sometimes there might be extra escape characters there too

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/CloudFunctionService.java
```diff
@@ -0,0 +1,93 @@
...
+import io.camunda.connector.http.base.model.HttpMethod;
+import io.camunda.connector.http.base.utils.CloudFunctionHelper;
+import java.io.IOException;
+import java.util.Map;
+import org.apache.hc.core5.http.ContentType;
+import org.apache.hc.core5.http.HttpHeaders;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CloudFunctionService {
```
chillleader, line 33: We could have an abstract `ProxyService` and move the GCP-specific implementation into `connectors-bundle-saas`. A couple of things to consider here:
- We might decide to switch to a different proxy in the future - there have been discussions around that
- Customers might want to use something else as a proxy in their own runtime deployments

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/components/apache/CustomApacheHttpClient.java
johnBgood: I don't have a strong opinion, in my mind a redirect is not an error + I wanted to keep the current behavior (I think the current Google HTTP client considers HTTP codes >4xx as errors).

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonResult.java
johnBgood: Done!

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/request/apache/bulider/parts/ApacheRequestBodyBuilder.java
johnBgood: Into the body I agree, but do you think the whole `body` property could be one the types you mentioned?

Could you provide an example if that's the case?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/services/CloudFunctionService.java
johnBgood: I initially planed to do that. But the authentication here is a bit specific and requires more steps than just using a user/password. It's coming directly from the Google OAuth2 API.

My intention was:
- To keep the current implementation as it worked before, but a bit cleaner.
- Later, add full proxy support (with auth) and see how to integrate our own "proxy" if possible.

It's not a usual proxy per se, it's more that we're forwarding the whole HttpComonRequest to this function. A proxy (in the customer's mind) would be a classic proxy as we know them (haproxy, nginx, etc).

I know it's not ideal, but I couldn't come with a better solution for now at least.
Is that ok or do you still think I should change this code?

chillleader: I agree, makes sense! Let's consider refactoring this separately, if we decide to migrate to a real proxy

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/request/apache/bulider/parts/ApacheRequestBodyBuilder.java
chillleader: I checked date, boolean, and a number - luckily, dates already arrive to us as strings at this point, but things like these two below result in numbers/booleans accordingly

Screenshot 2024-05-31 at 19 00 10
Screenshot 2024-05-31 at 18 59 50

sbuettner:
I think we can remove the `HttpTransportComponentSupplier` WDYT?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/utils/JsonHelper.java
```diff
@@ -19,26 +19,33 @@
...
-              try {
-                return mapper.readTree(response);
-              } catch (JsonProcessingException e) {
-                LOGGER.error("Wasn't able to create a JSON node from string: " + strResponse);
-                throw new RuntimeException(e);
-              }
-            })
-        .orElse(null);
+  public static boolean isJsonValid(Object maybeJson) throws JsonProcessingException {
+    return getAsJsonElement(maybeJson, ConnectorsObjectMapperSupplier.DEFAULT_MAPPER) != null;
```
sbuettner, line 49: Two of the three static methods directly access the object mapper via the static `ConnectorsObjectMapperSupplier.DEFAULT_MAPPER` call while the other has a parameter for it. Can we maybe refactors this utility to be passed into the class via parameter to avoid the static object mapper calls?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/constants/Constants.java
```diff
@@ -27,7 +27,6 @@ public class Constants {
   public static final String CREDENTIALS_BODY = "credentialsBody";
   public static final String PROXY_FUNCTION_URL_ENV_NAME = "CAMUNDA_CONNECTOR_HTTP_PROXY_URL";
 
-  public static final String APPLICATION_JSON_CHARSET_UTF_8 = "application/json; charset=UTF-8";
   public static final String APPLICATION_X_WWW_FORM_URLENCODED =
       "application/x-www-form-urlencoded";
```
sbuettner, line 31: This seems to be unused. Is that correct? 

I thought we now support this: https://github.com/camunda/connectors/pull/2291

johnBgood: Deleted 👍

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/utils/JsonHelper.java
johnBgood: Fixed ✅

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/constants/Constants.java
sbuettner: Just curious, do we support sending data with this mime-type if its set via the header?

johnBgood: [This](https://github.com/camunda/connectors/blob/397d35db8441b15d0716ac071bead43632b99071/connectors/http/http-base/src/test/java/io/camunda/connector/http/base/components/apache/CustomApacheHttpClientTest.java#L245) test should show this use case, is it correct?

sbuettner: Yes, awesome. Thanks for the pointer.

sbuettner:
@johnBgood I looked at the package structure and could we maybe simplify this and make it more domain driven? 
The `model` package for example does not contain the `auth` model.  The `service` package is not really domain specific. Similar to the `constants` and the `utils` package. Would be great if we can find a way to simplify this structure a little bit with less packages and more domain based packages. The `Base64Helper` for example could live in the `auth` package for example. The `OAuthService` as well. The models in `auth` could be moved to the move to the `model` package as they are used there.

sbuettner:
We have two sub packages named `apache`. Would be great to have a single one for example. WDYT? We could get rid of the generic packages like `request` and `components` and just put them into the `base` package which is empty.

sbuettner (dismissed):
Great work!

sbuettner (approved):
👍

PR #2666, opened 2024-05-27
ci(integration): prepare integration test for use in release and deploy snapshots workflows

ev-codes:
## Description

Now that the Helm integration test is more complicated, we should re-use its logic instead of trying to replicate it.

We also want to trigger the Helm integration test each time a release artifact is ready in Docker. The integration test does not currently trigger when CI commits are made to the release branch. 

We need to merge in the `INTEGRATION_TEST`-related changes separately first. This is because when calling local workflows, GitHub only allows us to use the version on the main branch. After merging this PR, I'll be able to properly test the `DEPLOY_SNAPSHOT` changes I've made in another branch.

A separate PR will come with the `DEPLOY_SNAPSHOT` and `RELEASE` changes.

### Notable Changes

- Call `INTEGRATION_TEST` on workflow_call, not push
    - We need to specify workflow_call in order to call the test from other workflows
    - I removed push because it would cause too many cancelled or non-actionable failed tests. The `test-integration-template` file only seems to allow us to run one job at a time on the connectors repo. Renovate often pushes several commits (even to release branches) within minutes of each other, so we'd end up with many cancelled tests. In addition, the Helm integration tests are relatively flakey and may require retries.
- Using ref_name instead of ref
    - In my tests with the release workflow, it is much easier to provide the non-fully-qualified branch name (release/8.5) instead of the fully qualified one (refs/heads/release/8.5).
    - I also didn't see any reason we needed to use the fully qualified branch name
    - [See GitHub Context docs for more details](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)

## Related issues

https://github.com/camunda/team-connectors/issues/764

### DIFF
--- a/.github/workflows/INTEGRATION_TEST.yml
+++ b/.github/workflows/INTEGRATION_TEST.yml
@@ -2,9 +2,16 @@
 name: Integration test
 
 on:
-  push:
-    branches:
-      - release/*
+  workflow_call:
+    inputs:
+      connectors-version:
+        description: 'The version of the Connectors to test. If not provided, the version is determined based on the Maven project version. On main branch, the default is SNAPSHOT'
+        required: false
+        type: string
+      helm-branch:
+        description: 'The branch of the Helm chart to test against. If not provided, the branch is determined based on the ref (see helm-git-refs.json)'
+        required: false
+        type: string
   workflow_dispatch:
     inputs:
       connectors-version:
--- a/.github/workflows/INTEGRATION_TEST.yml
+++ b/.github/workflows/INTEGRATION_TEST.yml
@@ -9,7 +9,11 @@ on:
         required: false
         type: string
       helm-branch:
-        description: 'The branch of the Helm chart to test against. If not provided, the branch is determined based on the ref (see helm-git-refs.json)'
+        description: 'The branch of the Helm chart to test against. If not provided, the Helm branch is determined based on the release-branch or the ref (see helm-git-refs.json)'
+        required: false
+        type: string
+      release-branch:
+        description: 'Connectors release branch containing code to test. If not provided, the Helm branch is determined based on the ref (see helm-git-refs.json)'
         required: false
         type: string
   workflow_dispatch:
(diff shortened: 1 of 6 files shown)


chillleader (approved):
Looks good already, thank you 👍

.github/workflows/INTEGRATION_TEST.yml
```diff
@@ -55,7 +66,7 @@ jobs:
         id: helm-branch
         run: |
           if [ -z "${{ github.event.inputs.helm-branch }}" ]; then
-            helm_branch=$(jq -r ".[\"${{ github.ref }}\"]" .github/workflows/helm-git-refs.json)
+            helm_branch=$(jq -r ".[\"${{ github.event.inputs.release-branch || github.ref_name }}\"]" .github/workflows/helm-git-refs.json)
```
chillleader, line 69: Maybe a naive question, but wouldn't this part:
`${{ github.event.inputs.release-branch || github.ref_name }}`
be evaluated to a boolean?

I did a quick search through the [GitHub expressions docs](https://docs.github.com/en/actions/learn-github-actions/expressions#operators) but didn't find an example like this, but maybe I just missed it or I'm looking in the wrong place 😄

chillleader (approved):
Awesome, thanks! I only have a single question for clarification, see above

.github/workflows/INTEGRATION_TEST.yml
ev-codes: Thanks for flagging!

This is a shorthand for the ternary operator in the GitHub expressions docs - basically if the `release-branch` variable is not defined, then use `ref_name`.

You can see that expression being evaluated to a string [in my test run](https://github.com/camunda/connectors/actions/runs/9268204933/job/25496334167) under "Determine Helm chart ref to use".

[Here's another example of it being used elsewhere](https://stackoverflow.com/a/78477094).

I'll go ahead and resolve this, but feel free to ask any follow-ups 😄

chillleader: Got it, thanks for clarifying!

PR #2674, opened 2024-05-27
ci: rollback target commitish change in the release workflow

chillleader:
## Description

See https://github.com/camunda/connectors/pull/2651


johnBgood (approved):

ev-codes (approved):
This looks ok!

We should revisit this release workflow at some point to handle the edge case where:
- a release workflow fails
- we want to re-run the release workflow against an already-created tag

My solution to this edge case was not fully working anyway, but I'll create a ticket for us to discuss as a team.

PR #2675, opened 2024-05-28
Element Template generation: set notEmpty: false explicitly

jonathanlukas:
## Description

On template generation, there seems to be no way to have `constraints.notEmpty` being set to true.

However, this is currently required to achieve a field that has a regex for validation but may remain empty.

This PR enables this behaviour for template property fields and jakarta validation annotations on processing.

## Related issues

https://github.com/camunda/camunda-modeler/issues/4314#issuecomment-2134494119

closes #


sbuettner (approved):

jonathanlukas:
@sbuettner could this be regarded as a bugfix and be backported to 8.5?

sbuettner:
@jonathanlukas Yes, sure

sbuettner (approved):

PR #2678, opened 2024-05-29
fix(congen): change default element type to match enum

chillleader:
## Description

This PR will change the default value to match the enum (this was forgotten during the refactoring where this enum was introduced)


markfarkas-camunda (approved):
LGTM!

PR #2697, opened 2024-06-05
ignore major and minor on release branches

jonathanlukas:
## Description

This PR adds a rule to renovate to ignore major and minor releases on release branches as they should only contain patches by nature.

## Related issues

closes #


chillleader (approved):
Thank you, lgtm!

Issue #2648, opened 2024-05-24
Cannot set variable to null after it's set to a non-null value


Oleksiivanov:
__Describe the Bug__

original request : 
- https://jira.camunda.com/browse/SUPPORT-22056?focusedCommentId=350995&atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258Y29tbWVudA%3D%3D

The [`setSerializationInclusion(JsonInclude.Include.NON_NULL)`](https://github.com/camunda-community-hub/spring-zeebe/blob/b0147f31a7f79d69297b3bec7f8104145511fad4/camunda-sdk-java/java-common/src/main/java/io/camunda/common/json/SdkObjectMapper.java#L33) configuration in the [SdkObjectMapper.java](https://github.com/camunda-community-hub/spring-zeebe/blob/main/camunda-sdk-java/java-common/src/main/java/io/camunda/common/json/SdkObjectMapper.java) file is causing null properties to be excluded from serialization, leading to inconsistencies in behavior when creating a CompleteJobCommand in Zeebe.

The `setSerializationInclusion(JsonInclude.Include.NON_NULL)` was added in [this commit ](https://github.com/camunda-community-hub/spring-zeebe/commit/6fd4d30b23900472debaca4bdc64a723a9e6508b) and was described as a fix, so we need to be careful when solving this bug

In the attached diagram, a variable called `setMeToNull` is initialised to a `non-null` value. In the next step, it is set to `null`. Finally, it checks whether it’s actually `null`.

Users would expect it to be null, but it isn’t.

<details>
<summary>Click to reveal BPMN diagram with bug</summary>

```xml
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #2700, opened 2024-06-05
fix(deps): Bump spring-zeebe to 8.5.4

johnBgood:
## Description

Bump spring-zeebe to 8.5.4

## Related issues

https://github.com/camunda/connectors/issues/2648

### DIFF
--- a/connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/model/RabbitMqMessageProperties.java
+++ b/connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/model/RabbitMqMessageProperties.java
@@ -6,11 +6,13 @@
  */
 package io.camunda.connector.rabbitmq.inbound.model;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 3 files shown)


ev-codes (approved):
Thank you!

connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/model/RabbitMqMessageProperties.java
```diff
@@ -6,11 +6,13 @@
  */
 package io.camunda.connector.rabbitmq.inbound.model;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.rabbitmq.client.AMQP;
 import java.util.Date;
 import java.util.Map;
 
 /** Model of the RabbitMQ message properties. Mirrors the {@link AMQP.BasicProperties}. */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
johnBgood, line 15: We don't want to serialize so many null fields here as it's just noise

sbuettner (approved):

PR #2737, opened 2024-06-10
fix: update azure-identity version

markfarkas-camunda:
## Description

Fix for [CVE-2024-29025](https://www.cve.org/CVERecord?id=CVE-2024-29025)


chillleader:
There's a test failure, but it could be something non-persistent, would you mind rerunning @markfarkas-camunda?

PR #2744, opened 2024-06-12
feat: FEEL function for invoking outbound connectors

chillleader:
## Description

This PR implements:
- the `ExternalConnectorExecutor` that allows invoking outbound connectors outside of the job worker context
- the `ConnectorInvocationFeelFunctionProvider` that can be dynamically added to the `FeelEngineWrapper` (e.g. as a Spring bean)
- the `TransientDataStore` to store the binary data references away from serialization/FEEL engine 

## Related issues

closes https://github.com/camunda/team-connectors/issues/785


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/external/ExternalConnectorExecutor.java
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  public Object execute(String type, Object variables) throws Exception {
+    var connector = outboundConnectorFactory.getInstance(type);
+    String variablesJson;
+    try {
+      variablesJson = objectMapper.writeValueAsString(variables);
+    } catch (Exception e) {
+      throw new RuntimeException("Error while serializing variables", e);
+    }
```
sbuettner, line 50: If we transform the `variables` to json wont we have an issue with passing data around like a byte[] for example?

chillleader: Do you have a specific scenario in mind? As I'm thinking about it, it seems that we will be fine if we:
- only pass binary data as part of the `Document` class, and
- have a proper serialization/deserialization mechanism in place for `Document` that converts it to base64.

I might be missing something, and it is possible that we can do secret replacement more efficiently 🤔

sbuettner: @chillleader Sure, lets assume we want to execute an S3 connector in the `resultExpression` of the REST Connector to store the binary body. Passing a byte[] document wouldnt really be efficiently possible if we serialize it to json first and read it back again for larger files. Here it would be better to be able to directly invoke the connector passing the objects/references in memory to be able to use streams for example.

Yes, secret replacing would be an issue but this could theoretically be done by only replacing it when an object in the variables is of type `string` before passing it to the Connector.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/feel/ConnectorInvocationFeelFunctionProvider.java
```diff
@@ -0,0 +1,106 @@
...
+              Object unpackedVariables =
+                  FeelEngineWrapperUtil.sanitizeScalaOutput(feelValueMapper.unpackVal(variables));
+              String variablesAsString;
+              try {
+                variablesAsString = objectMapper.writeValueAsString(unpackedVariables);
+              } catch (JsonProcessingException e) {
+                throw new RuntimeException(e);
+              }
+
+              var connector = connectorFactory.getInstance(typeValue);
```
johnBgood, line 77: shouldn't we use the `ExternalConnectorExecutor`here?

johnBgood (approved):
🧨

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/external/ExternalConnectorExecutor.java
chillleader: Thanks for clarifying, makes sense. This requires refactoring how we replace secrets in all outbound connectors to avoid having two secret replacement methods (we are still getting a json string from the zeebe job). I'm not opposed to that and can also attempt this as part of this PR.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/feel/ConnectorInvocationFeelFunctionProvider.java
chillleader: Haha yes that was the idea

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/external/ExternalConnectorExecutor.java
sbuettner: Great, you can also add secret support later on to figure out whether our approach here is reasonable 😄

chillleader: Changed it here :)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/external/ExternalOutboundConnectorContext.java
```diff
@@ -28,7 +28,7 @@ public ExternalOutboundConnectorContext(
       SecretProvider secretProvider,
       ValidationProvider validationProvider,
       ObjectMapper objectMapper,
-      String jsonVariables) {
+      Object jsonVariables) {
     super(secretProvider, validationProvider, objectMapper, jsonVariables);
```
sbuettner, line 32: Can we rename `jsonVariables` to `variables`?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/AbstractOutboundConnectorContext.java
```diff
@@ -16,51 +16,91 @@
...
 public abstract class AbstractOutboundConnectorContext extends AbstractConnectorContext
     implements OutboundConnectorContext {
 
   protected final ObjectMapper objectMapper;
 
-  private final String jsonVariables;
+  private final Object variables;
+  private JsonNode variablesWithReplacedSecrets = null;
+
   private String jsonWithSecrets = null;
```
sbuettner, line 43: Shouldnt we also cache the object that contains the replaced secrets?

chillleader: We do (in `variablesWithReplacedSecrets`) - this string cache was not used any more, I removed it

chillleader:
Closing this as per our decision to move along with a different approach.

PR #2754, opened 2024-06-17
chore(rest): Bump element template version

johnBgood:
## Description

Bump the REST Connector element template version.
Removes `writeTimeoutInSeconds` as it's not supported anymore by the connector.

## Related issues

https://github.com/camunda/connectors/issues/1069

### DIFF
--- a/connectors/http/rest/src/test/resources/requests/success-test-cases.json
+++ b/connectors/http/rest/src/test/resources/requests/success-test-cases.json
@@ -8,7 +8,6 @@
     },
     "connectionTimeoutInSeconds": 20,
     "readTimeoutInSeconds": 30,
-    "writeTimeoutInSeconds": 40,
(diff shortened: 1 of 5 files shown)


connectors/http/rest/src/test/resources/requests/success-test-cases.json
```diff
@@ -108,7 +104,6 @@
     "url": "http://localhost:8085/http-endpoint",
     "connectionTimeoutInSeconds": 5,
     "readTimeoutInSeconds": 67,
-    "writeTimeoutInSeconds": 34,
```
ev-codes, line 111: May you please add an explanation to the PR doc why `writeTimeoutInSeconds` is removed in all of these templates? Thanks 😄

ev-codes (approved):
Thanks! Just one note

PR #2757, opened 2024-06-17
fix(sagemaker): Parse response.body as as json based on response cont…

### LINKED ISSUE #2752: [AWS Sagemaker Connector] Response is not parsed as json
__Describe the Bug__

As described by @bastiankoerber the `response.body` that is returned by the Sagemaker API is not parsed before returning it to the Connector Runtime. This leads to escaped json which currently cant be used or parsed inside the FEEL `resultExpression`.

__Steps to Reproduce__

Call the connector and try to access data inside the `response.body` object.

__Expected Behavior__

If Sagemaker returns json and the content type reflects that the runtime should use this information to parse the response body and return a proper unescaped json object as the value of the `response.body`.

__Environment__
- Latest snapshot on SaaS of 8.6.0-alpha2.1

sbuettner:
…ent type

## Related issues

closes #2752

### DIFF
--- a/connectors/aws/aws-sagemaker/src/main/java/io/camunda/connector/sagemaker/model/SageMakerSyncResponse.java
+++ b/connectors/aws/aws-sagemaker/src/main/java/io/camunda/connector/sagemaker/model/SageMakerSyncResponse.java
@@ -7,16 +7,37 @@
 package io.camunda.connector.sagemaker.model;
 
 import com.amazonaws.services.sagemakerruntime.model.InvokeEndpointResult;
+import io.camunda.connector.aws.ObjectMapperSupplier;
+import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 
 public record SageMakerSyncResponse(
-    String body, String contentType, String customAttributes, String invokedProductionVariant) {
+    Object body, String contentType, String customAttributes, String invokedProductionVariant) {
 
   public SageMakerSyncResponse(InvokeEndpointResult result) {
     this(
-        StandardCharsets.UTF_8.decode(result.getBody()).toString(),
+        mapResponseBody(result),
         result.getContentType(),
         result.getCustomAttributes(),
         result.getInvokedProductionVariant());
+  }
+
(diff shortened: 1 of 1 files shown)


connectors/aws/aws-sagemaker/src/main/java/io/camunda/connector/sagemaker/model/SageMakerSyncResponse.java
```diff
@@ -7,16 +7,37 @@
...
         result.getContentType(),
         result.getCustomAttributes(),
         result.getInvokedProductionVariant());
   }
+
+  /* Maps the response body to the appropriate object type
+  / https://docs.aws.amazon.com/sagemaker/latest/dg/clarify-online-explainability-invoke-endpoint.html#clarify-online-explainability-response
+  */
+  private static Object mapResponseBody(InvokeEndpointResult result) {
+    if (result.getContentType().equals("application/json")) {
```
johnBgood, line 30: I don't know the `result` var, but if it can be something like `application/json; charset=utf-8` we might want to : 
1. lower case 
2. startsWith instead of equals
wdyt?

johnBgood: also wondering if we should add that in the SDK, it looks like we're doing this comparison in a lot of different places

johnBgood (approved):
1 comment, all good green flag 🟢

connectors/aws/aws-sagemaker/src/main/java/io/camunda/connector/sagemaker/model/SageMakerSyncResponse.java
sbuettner: @johnBgood I based it on this documentation: https://docs.aws.amazon.com/sagemaker/latest/dg/clarify-online-explainability-invoke-endpoint.html#clarify-online-explainability-response
  
> also wondering if we should add that in the SDK, it looks like we're doing this comparison in a lot of different places

Potentially yes but the implementation details might still be different based on the API.

PR #2771, opened 2024-06-18
chore(readme): remove wrong env var hyphen usage

Langleu:
## Description

Shells don't support hyphens or dashes in environment variables and will fail to export them. There are ways around this, but I'd not advertise those.
Kubernetes can deal with it but depends on the application within whether it can properly load those.

My origin of the change stems from the [manual setup](https://docs.camunda.io/docs/next/self-managed/setup/deploy/local/manual/) where one is typically running some form of shell and based on that the advertised conversion from `cluster-id` to `CLUSTER-ID` doesn't work.

Additionally, Spring boot has [relaxed binding rules](https://github.com/spring-projects/spring-boot/wiki/Relaxed-Binding-2.0#environment-variables) where hyphens are also ignored. Probably stems from shells typically not supporting those out of the box.

I'm not sure whether you've more occurrences in other docs or so, just as a heads up that you shouldn't advertise it this way.
The Helm chart is also not using the hyphens and drops it when configuring connectors.

Feel free to use the PR as you see fit, even close. I'd just like to avoid advertising something that does not work in where it's originally linked from.

## Related issues

none, just thought I would provide a PR instead of going through slack.

closes #

### DIFF
--- a/connector-runtime/README.md
+++ b/connector-runtime/README.md
@@ -108,7 +108,7 @@ See also [Docker: Custom set of connectors](https://docs.camunda.io/docs/self-ma
 
 The following configuration properties can be set via `src/main/application.properties` if you run Java directly.
 
-You can also set those configuration options via environment variables (then named `ZEEBE_CLIENT_CLOUD_CLUSTER-ID` instead of `zeebe.client.cloud.cluster-id`), especially useful if you run via DOCKER.
+You can also set those configuration options via environment variables (then named `ZEEBE_CLIENT_CLOUD_CLUSTERID` instead of `zeebe.client.cloud.cluster-id`), especially useful if you run via DOCKER.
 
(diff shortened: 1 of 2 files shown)


Langleu:
Actually see in the docs as well some usage of that: 

- multiple `ZEEBE_CLIENT_BROKER_GATEWAY-ADDRESS`
- ZEEBE_CLIENT_DEFAULT-TENANT-ID
- ZEEBE_CLIENT_DEFAULT-JOB-WORKER-TENANT-IDS

https://docs.camunda.io/docs/next/self-managed/connectors-deployment/connectors-configuration/?configuration=local
Would be good to have that ironed out. I know the alternative looks ugly but even Spring Boot advertises it that way.

sbuettner (approved):

Langleu:
thanks! Added it to the merge queue.
On the docs.camunda.io topic, up to your own discretion. From what I could see from other teams, they're all avoiding `hyphens` in general in their configs. Typically `lowerUpercase` and if they have one with a hyphen, they define the env var as `BLA_BLA_BLABLABLALBLA`.

PR #2773, opened 2024-06-18
653 fr slack connector post message add option to post to an existing thread

mathias-vandaele:
## Description

Users were not able to create a new thread from an existing message or add a message to an existing thread. this solves the issues.

I added a new optional field to the slack outbound connector `thread` which takes the id of a message to start a thread from.

This id has to be gotten from the output mapping `message.ts` which contains the id of the posted message

## Related issues

There might be an issue in the future when an user wants to add messages (using camunda) inside a thread that has been create by a normal user (human), as I don't know how to retrieve the id.

closes https://github.com/camunda/team-connectors/issues/653

### DIFF
--- a/connectors/slack/element-templates/hybrid/slack-outbound-connector-hybrid.json
+++ b/connectors/slack/element-templates/hybrid/slack-outbound-connector-hybrid.json
@@ -185,6 +185,22 @@
     },
     "type" : "String"
   }, {
+    "id" : "data.thread",
+    "label" : "Thread",
+    "optional" : false,
+    "feel" : "optional",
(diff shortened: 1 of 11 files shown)


johnBgood (commented):
Left some minor comments, note that you also need to update the web-modeler templates [here](https://github.com/camunda/web-modeler/blob/main/webapp/src/App/Pages/Diagram/BpmnJSExtensions/connectorsExtension/.camunda/element-templates/slack-connector.json) is an example

johnBgood (approved):
🟢 flag, great job!

Issue #2730, opened 2024-06-07
RabbitMQ Connector does unnecessary message unsescape action.


Zapolski:
__Describe the Bug__

I use out-of-the-box RabbitMQ Connector. Message for sending contains string fields with escaped symbols, like:

```
{
   "data": {
       "company": "LTD \"Three bears\""
   }
}
```

Connector sends my message but consumer receives wrong message and fells with MessageConversionException. There’s a message which consumer receives:

```
{
   "data": {
       "company": "LTD "Three bears""
   }
}
```

All symbols `"\"` disappeared. Connector’s source code contains logic with unescaping json.

__Steps to Reproduce__

1. Prepare message object for sending by RabbitMQ connector. Object must contain field with escaped symbols. For example:
```
{
   "data": {
       "company": "LTD \"Three bears\""
   }
}
```
2. Use RabbitMQ connector for sending message.

__Expected Behavior__

I expect the same message in my RabbitMQ CONSUMER, message should be without any transformation.

__Environment__

- OS: [Linux Ubuntu]
- Library version: [RabbitMQ Connector io.camunda.connectors.RabbitMQ.v1 version 5]
- Camunda version: [e.g. 8.4. (self-managed)]

[Camunda 8 forum topic](https://forum.camunda.io/t/rabbitmq-connector-unescape-json-inside-connector/52937)

sbuettner:
As discussed in our team meeting we will remove unnecessary unescaping from Connectors in the next alpha.

mathias-vandaele:
After testing, removing `.map(StringEscapeUtils::unescapeJson)` `L65 io.camunda.connector.rabbitmq.outbound.MessageUtil` corrects the issue. 
I am not so sure about removing it as it might have been added for a valid reason, any thoughts @sbuettner  ?

johnBgood:
I'm posting the forum comment I made here for visibility:

> It is due to [Zeebe escaping characters](https://github.com/camunda/camunda/issues/9859).
> Looking at this ticket, it seems that it might be resolved. If so, we will remove the unescapes in our connectors.

The way I see it, we might remove any unescaping (not only in the RabbitMQ connector), but let's wait for @sbuettner or @chillleader opinions.

chillleader:
Right, that Zeebe issue was the reason we added it there. If the escaping issues are gone, we no longer need the workaround.

mathias-vandaele:
@chillleader Should I make an update and modify it in every connector ?

[...]

PR #2778, opened 2024-06-21
feature-2730 fix the escaping feature for rabbitMQ outbound connector

mathias-vandaele:
## Description

When passing a JSON object as body for the rabbitMQ outbound connector, it is converted either to a `Map` or a `String`. 

When a `Map` was wrote as a Json, it was then unescaped.
This PR removes this feature.

We have to be sure it does not cause any weird side effect to remove this feature

## Related issues

closes https://github.com/camunda/connectors/issues/2730


sbuettner (approved):

johnBgood (approved):
💯

PR #2792, opened 2024-06-25
Kafka connector loop retry

### LINKED ISSUE #2750: Kafka inbound connector stops processing messages on any error
__Describe the Bug__

The Kafka inbound connector stops processing messages in case of any error. To resume message processing, the user has to redeploy the process.

Because of this behavior, even if the user manually edits the offsets for the consumer group used by the connector, the connector won't be able to recover from this situation.

__Steps to Reproduce__

1. Initialize an inbound Kafka connector
2. Publish a message to the topic that contains invalid JSON, or simply doesn't match the activation condition specified in the connector
3. Check the connector runtime logs, activity logs in the web modeler, and make sure that the consumer was stopped

__Expected Behavior__

The connector should retry after a certain backoff period (e.g. 30 seconds). The external consequences may lead to offsets changing, or the activation condition starting to evaluate to true, so the connector should be able to recover.

__Environment__

- all versions <=8.6-alpha2.1

chillleader:
## Description

Adds retries to the Kafka connector consumer loop.

Previously, when the consumer threw an exception, it just stopped processing until the process is redeployed. The new default is to retry this as long as the connector is active, with a backoff of 30 seconds.

## Related issues

closes #2750

### DIFF
--- a/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaExecutable.java
+++ b/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaExecutable.java
@@ -14,6 +14,8 @@ import io.camunda.connector.api.inbound.InboundConnectorExecutable;
 import io.camunda.connector.api.inbound.Severity;
 import io.camunda.connector.generator.dsl.BpmnType;
 import io.camunda.connector.generator.java.annotation.ElementTemplate;
+import io.github.resilience4j.retry.RetryConfig;
+import java.time.Duration;
 import java.util.Properties;
 import java.util.function.Function;
 import org.apache.kafka.clients.consumer.Consumer;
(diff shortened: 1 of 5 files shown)


johnBgood (dismissed):
Really nice!

connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaExecutable.java
```diff
@@ -63,13 +65,23 @@ public class KafkaExecutable implements InboundConnectorExecutable<InboundConnec
...
   }
 
   public KafkaExecutable() {
-    this(KafkaConsumer::new);
+    this(
+        KafkaConsumer::new,
+        RetryPolicy.builder()
+            .handle(Exception.class)
+            .withDelay(Duration.ofSeconds(30))
+            .withMaxAttempts(-1)
```
johnBgood, line 83: maybe having a constant here to express the meaning: INFINITE or similar (if this is what -1 means)

johnBgood (approved):
1 minor comment only 👍

connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaExecutable.java
chillleader: Good point!

sbuettner (approved):

sbuettner:
@chillleader This one might require e2e test fixes.

chillleader:
@sbuettner this should have fixed the test

johnBgood (approved):

sbuettner (approved):

Issue #1047, opened 2023-08-17
Append Operation in Google Sheet Connector


quentinleroy:
__Is your feature request related to a problem? Please describe.__

Today, the connector allows users to create a new row based on a fixed integer. This means that to add a new row in an existing sheet, the connector needs to know in which row the last data set was inserted. This adds logic (and complexity) to the model itself, which is unnecessary.

__Describe the solution you'd like__

The Google Sheet API provides a functionality that allows the user to insert a row at the bottom of an existing sheet, without the need of providing an exact row number. 
This would make it easier for users to log information (ie. variables) outside of Camunda and make them easily consumable by business users.  

__Describe alternatives you've considered__

I've been using Optimize instead to look at data, but it is not optimised to look at text, hence variables such as descriptions are not displayed as they should. 

__Additional context__

[Feature in Productboard with insights](https://camunda.productboard.com/feature-board/6990815-overview-board/features/22109426/detail)

igpetrov:
Assigning to @bastiankoerber and @sbuettner to set proper priority

PR #2795, opened 2024-06-25
feature-1047 create new feature: make the row index optional, if not …

mathias-vandaele:
…specified will just append to the next row

## Description

Add a feature:

The row index field is now optional
  - if defined, it will add the row to the specified row index
  - if not defined, it will append to the last row

closes https://github.com/camunda/connectors/issues/1047

### DIFF
--- a/connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/operation/impl/CreateRowOperation.java
+++ b/connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/operation/impl/CreateRowOperation.java
@@ -13,6 +13,7 @@ import io.camunda.connector.gsheets.operation.GoogleSheetOperation;
 import io.camunda.google.model.Authentication;
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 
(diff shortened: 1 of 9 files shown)


connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/operation/impl/CreateRowOperation.java
```diff
@@ -25,11 +26,15 @@ public CreateRowOperation(CreateRow model) {
   @Override
   public Object execute(Authentication auth) {
     ValueRange valueRange = new ValueRange().setValues(List.of(model.values()));
-    String range = buildRange(model.worksheetName(), buildRowRange(model.rowIndex()));
 
     try {
-      this.update(auth, model.spreadsheetId(), range, valueRange);
-
+      if (Objects.isNull(model.rowIndex())) {
+        String range = buildRange(model.worksheetName(), "1:" + Integer.MAX_VALUE);
```
johnBgood, line 32: `range` is not used we can probably delete it

johnBgood (approved):
👍

PR #2801, opened 2024-06-26
Fix read timeout with cloud function

johnBgood:
## Description

- When creating the Cloud Function request we didn't set any timeout, so we defaulted to HttpCommonRequest#DEFAULT_VALUE (20). We need to have an infinite timeout when calling the Cloud Function, as this is the function that will timeout.


mathias-vandaele (commented):
Good 👍

mathias-vandaele (approved):

Issue #2774, opened 2024-06-19
Webhook connector cannot handle json array in request body


markfarkas-camunda:
__Describe the Bug__

If the body of the request is a json array, the webhook connector throws a runtime exception and returns with an http 500 response. The process doesn't start.

__Steps to Reproduce__

1. Create a process with a webhook start event
2. Deploy the process definition
3. Call the webhook url with POST request containing a json array in the body

__Expected Behavior__

The request body should be parsed and the process should start

__Environment__

- Camunda version: latest

https://jira.camunda.com/browse/SUPPORT-22337

PR #2803, opened 2024-06-26
feature-2774 webhook connector can now handle json array in request body

mathias-vandaele:
## Description

This has been re-implemented after the discussion from the architecture session

Now if the body is a map, we will then create the signature as it was before (the documentation should be updated cause it does not reflect what the code is actually doing in that case, it has been implemented specifically for twilio)

Now if the body is an Array, it actually reflects what the documentation says, the body is serialized as String and used to calculate the signature

## Related issues

closes https://github.com/camunda/connectors/issues/2774


sbuettner:
@mathias-vandaele Thanks for tackling this. I think we can solve is issue in potentially simpler way. The expected type of the `MappedHttpRequest.body` property is `Object`. We can therefore simply map the incoming request body to `Object` via Jackson without any custom serialization.

mathias-vandaele:
> @mathias-vandaele Thanks for tackling this. I think we can solve is issue in potentially simpler way. The expected type of the `MappedHttpRequest.body` property is `Object`. We can therefore simply map the incoming request body to `Object` via Jackson without any custom serialization.

Made a way simpler solution :D

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -15,30 +15,45 @@
...
       throw new NullPointerException(
           "Can't extract signature data from body, because body is null");
     }
     Map<String, String> signatureData =
-        HttpWebhookUtil.transformRawBodyToMap(
-            payload.rawBody(), HttpWebhookUtil.extractContentType(payload.headers()));
+        checkedCastToMap(
+            HttpWebhookUtil.transformRawBodyToObject(
+                payload.rawBody(), HttpWebhookUtil.extractContentType(payload.headers())),
+            String.class);
```
sbuettner, line 28: Why not pass the `payload.rawBody()` into this method?

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -15,30 +15,45 @@
...
+    if (o instanceof Map<?, ?> map) {
+      for (Map.Entry<?, ?> entry : map.entrySet()) {
+        if (!(tClass.isInstance(entry.getValue())) || !(tClass.isInstance(entry.getKey()))) {
+          return Map.of();
+        }
+      }
+      return (Map<T, T>) map;
+    }
+    return Map.of();
+  }
```
sbuettner, line 53: As we cant calculate the signature of an empty map anyway. What about skipping it when the payload is not of type map?

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -15,30 +15,45 @@
...
+
     for (String key : sortedKeys) {
       builder.append(key);
       String value = signatureData.get(key);
       builder.append(value == null ? "" : value);
     }
     return builder.toString();
   }
+
+  private static <T> Map<T, T> checkedCastToMap(Object o, Class<T> tClass) {
```
johnBgood, line 43: Looks like we don't have any other use cases than the `tClass` being a `String`. I'd suggest removing the generics then for visibility

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -15,30 +15,45 @@
...
+    if (o instanceof Map<?, ?> map) {
```
johnBgood, line 44: if this is `true` do we really have to iterate over the entrySet?

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
johnBgood: If we have a `String[]` couldn't we use the values to calculate the signature? 
I don't know where it is used, so I might be wrong

johnBgood (commented):
Just a few comments :)

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
johnBgood: Not send as an array, but we could build the signature in a similar way of what we do with Maps, basically concatenating values

mathias-vandaele: So the given list would be in this format `["key1", "value1", "key2", "value2"]` @johnBgood ?

johnBgood: Yes exactly :) Conceptually it would be more `["value1", "value2"...]` but that's the spirit yes.
Instead of adding keys/values from a Map (see below), we would simply add the array's values.

```java
for (String key : sortedKeys) {
      builder.append(key);
      String value = signatureData.get(key);
      builder.append(value == null ? "" : value);
    }
```

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
+  private static String prepareForSignature(byte[] rawBody, String contentTypeHeader) {
+    if (MediaType.FORM_DATA.toString().equalsIgnoreCase(contentTypeHeader)) {
+      String bodyAsString =
+          URLDecoder.decode(new String(rawBody, StandardCharsets.UTF_8), StandardCharsets.UTF_8);
+      Map<String, String> strMap =
+          Arrays.stream(bodyAsString.split("&"))
+              .filter(Objects::nonNull)
+              .map(param -> param.split("="))
+              .collect(
+                  Collectors.toMap(param -> param[0], param -> param.length == 1 ? "" : param[1]));
```
johnBgood, line 36: do we have a test where there's only a key without any value?

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
-    Map<String, String> signatureData =
-        HttpWebhookUtil.transformRawBodyToMap(
-            payload.rawBody(), HttpWebhookUtil.extractContentType(payload.headers()));
+    Object o = mapBytesToObject(rawBody);
+    if (o instanceof Map<?, ?> map) {
+      for (Map.Entry<?, ?> entry : map.entrySet()) {
+        if (entry.getKey() instanceof String) {
+          Map<String, String> strMap = (Map<String, String>) map;
+          return strMap.keySet().stream()
+              .sorted()
```
johnBgood, line 49: do we know now why this is sorted ?

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
+    }
+    throw new RuntimeException("Can't extract signature data from body");
+  }
 
-    StringBuilder builder = new StringBuilder();
-    List<String> sortedKeys = new ArrayList<>(signatureData.keySet());
-    Collections.sort(sortedKeys);
+  private static String mapObjectToString(Object o) {
+    try {
+      return ConnectorsObjectMapperSupplier.getCopy().writeValueAsString(o);
```
johnBgood, line 63: not sure why we are creating a copy instead of using the DEFAULT

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
+    return prepareForSignature(
+        payload.rawBody(), HttpWebhookUtil.extractContentType(payload.headers()));
   }
 
-  private static String extractSignatureData(final WebhookProcessingPayload payload)
-      throws IOException {
-    if (payload.rawBody() == null || payload.rawBody().length == 0) {
-      throw new NullPointerException(
-          "Can't extract signature data from body, because body is null");
+  private static String prepareForSignature(byte[] rawBody, String contentTypeHeader) {
```
johnBgood, line 27: we can remove `static` I guess

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
+    } else {
+      return mapObjectToString(o);
+    }
+    throw new RuntimeException("Can't extract signature data from body");
+  }
 
-    StringBuilder builder = new StringBuilder();
-    List<String> sortedKeys = new ArrayList<>(signatureData.keySet());
-    Collections.sort(sortedKeys);
+  private static String mapObjectToString(Object o) {
```
johnBgood, line 61: we can remove `static` I guess

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
+    }
+  }
 
-    for (String key : sortedKeys) {
-      builder.append(key);
-      String value = signatureData.get(key);
-      builder.append(value == null ? "" : value);
+  private static Object mapBytesToObject(byte[] rawBody) {
```
johnBgood, line 69: we can remove `static` I guess

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
mathias-vandaele: Yes That's twilio specific, we have to keep it like this

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
mathias-vandaele: Absolutely

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
mathias-vandaele: Changing this right now

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
johnBgood: Then let's add a comment, because the next person will spend some time figuring this out :D

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
```diff
@@ -6,39 +6,76 @@
...
+    if (MediaType.FORM_DATA.toString().equalsIgnoreCase(contentTypeHeader)) {
+      String bodyAsString =
+          URLDecoder.decode(new String(rawBody, StandardCharsets.UTF_8), StandardCharsets.UTF_8);
+      Map<String, String> strMap =
+          Arrays.stream(bodyAsString.split("&"))
+              .filter(Objects::nonNull)
+              .map(param -> param.split("="))
+              .collect(
+                  Collectors.toMap(param -> param[0], param -> param.length == 1 ? "" : param[1]));
+      return strMap.keySet().stream()
```
johnBgood, line 37:
```
strMap.keySet().stream()
          .sorted()
          .map(key -> key.concat(strMap.get(key)))
          .reduce(String::concat)
          .orElse("");
```

this can be extracted, it's called here and a few lines below

connectors/webhook/src/main/java/io/camunda/connector/inbound/signature/strategy/URLAndBodyEncodingStrategy.java
johnBgood: @mathias-vandaele let's just add this comment and we're good to go :)

sbuettner (approved):

PR #2804, opened 2024-06-26
Google Sheets Connector: Adding a worksheet will return the properties of the worksheet

### LINKED ISSUE #2796: Unable to obtain worksheet id using the available google sheet operations which is required for create empty row or column & delete worsheet operations
__Describe the Bug__

As a user I am unable to obtain worksheet id using the available google sheet operations which is required for create empty row or column & delete worksheet operations

__Steps to Reproduce__

1. When I "create new worksheet", the result variable doesn't populate the ID. It is also not possible for the user to set the worksheet ID while creating this new worksheet.
image

2. This ID is needed for performing other important google sheet operations like create empty column or row and delete worksheet.
image
image

3. The worksheetID cannot be retrieved or known or set using existing google sheet connectors operations. 

4. Therefore, I am unable to utilize the google sheets connector in my workflow and need to utilize the official google sheets API.
5. Documentation specifies that Start and End index can be empty. However, if empty, it is not possible to deploy the diagram.
image

__Expected Behavior__

When I create the new worksheet the response should return the sheet id. So that it can be used in other operations. This is the response received using the official google sheet API
image
OR I should be able to perform create empty column or row and delete worksheet operations using the worksheet name.

__Environment__

- Camunda version: [e.g. 8.5 (SaaS environment)]

jonathanlukas:
## Description

This PR adds a useful response to the "add worksheet" operation

This will require an adjustment to the documentation.

## Related issues

closes #2796 

PS: Can this be backported?

### DIFF
--- a/connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/operation/GoogleSheetOperation.java
+++ b/connectors/google/google-sheets/src/main/java/io/camunda/connector/gsheets/operation/GoogleSheetOperation.java
@@ -8,6 +8,7 @@ package io.camunda.connector.gsheets.operation;
 
 import com.google.api.services.sheets.v4.Sheets;
 import com.google.api.services.sheets.v4.model.BatchUpdateSpreadsheetRequest;
+import com.google.api.services.sheets.v4.model.BatchUpdateSpreadsheetResponse;
 import com.google.api.services.sheets.v4.model.ValueRange;
 import io.camunda.connector.gsheets.supplier.GoogleSheetsServiceSupplier;
 import io.camunda.google.model.Authentication;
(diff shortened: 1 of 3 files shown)


johnBgood (approved):
LGTM, remember to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) format when merging the PR so that it appears in our release note :)

johnBgood:
@jonathanlukas We can probably backport it to 8.5 and 8.4. Let's see how the PRs look like

jonathanlukas:
@johnBgood I think camunda/camunda has an action to enforce this. As I am not so used to the format of conventional commits, I would tend to be reminded of this.

PR #2808, opened 2024-06-27
fix(security) any html content response from a webhook is now escaped…

mathias-vandaele:
Any html content response from a webhook is now escaped and will not be interpreted by the browser

## Description

Another function is executed right after the execution of the FEEL function>

This function will check the response type and sanitize it if necessary.

## Related issues

closes https://github.com/camunda/team-connectors/issues/814

### DIFF
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
@@ -46,7 +46,6 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.regex.Pattern;
-
 import org.apache.commons.text.StringEscapeUtils;
 import org.jetbrains.annotations.Nullable;
 import org.slf4j.Logger;
--- a/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
+++ b/connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
@@ -45,8 +45,6 @@ import java.security.NoSuchAlgorithmException;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
-import java.util.regex.Pattern;
-import org.apache.commons.text.StringEscapeUtils;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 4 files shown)


connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
```diff
@@ -162,6 +164,20 @@ private Function<WebhookResultContext, WebhookHttpResponse> mapResponseExpressio
     return responseExpression;
   }
 
+  private WebhookHttpResponse sanitizeBody(WebhookHttpResponse webhookHttpResponse) {
+    if (webhookHttpResponse.body() instanceof String bodyStr
+        && Pattern.compile("<//?[a-z][\\s\\S]*>").matcher(bodyStr).find()) {
```
johnBgood, line 169: ` Pattern.compile("<//?[a-z][\\s\\S]*>")` might be extracted as a variable to avoid creating the Pattern every time

johnBgood: I don't remember what we decided, but this escaping be based on the pattern or the content-type?

mathias-vandaele: I would like the feedback of @sbuettner for this one

sbuettner: What about putting something like this: 
```
  protected static Object escapeValue(Object value) {
    switch (value) {
      case String s:
        return HtmlUtils.htmlEscape(s);
      default:
        return value;
    }
  }
```

into the `InboundWebhookRestController` as this is the right layer. And call it from the `toResponseEntity` method?

mathias-vandaele: I modified it, you are right, this solution is simpler and provides a solution at a higher level.

sbuettner (approved):

connectors/webhook/src/main/java/io/camunda/connector/inbound/HttpWebhookExecutable.java
sbuettner: Great, happy to support

PR #2817, opened 2024-07-04
642 fix spring security config

mathias-vandaele:
## Description

This ticket was pretty tricky cause those particular behavior are needed:

- `/inbound/*` `GET` `POST` `PUT` `DELETE` must not have any authentication system, and should not even read the authorization header
- `/inbound/*` `OPTIONS` `PATCH` `HEAD` must have an authentication system, and return 401 if the bearer is incorrect.

A new SecurityWebFilter had to be added as explained in this issue on spring-security repository https://github.com/spring-projects/spring-security/issues/14120

## Related issues

closes https://github.com/camunda/team-connectors/issues/642

### DIFF
--- a/bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/security/SecurityConfiguration.java
+++ b/bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/security/SecurityConfiguration.java
@@ -19,10 +19,11 @@ package io.camunda.connector.runtime.saas.security;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
 import org.springframework.http.HttpMethod;
+import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
-import org.springframework.security.config.annotation.web.configuration.WebSecurityCustomizer;
 import org.springframework.security.oauth2.core.DelegatingOAuth2TokenValidator;
 import org.springframework.security.oauth2.core.OAuth2TokenValidator;
 import org.springframework.security.oauth2.jwt.Jwt;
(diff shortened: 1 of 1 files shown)


bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/security/SecurityConfiguration.java
```diff
@@ -43,27 +43,32 @@ public class SecurityConfiguration {
...
+        .securityMatchers(
+            requestMatcherConfigurer ->
+                requestMatcherConfigurer
+                    .requestMatchers(HttpMethod.GET, "/inbound/*")
+                    .requestMatchers(HttpMethod.POST, "/inbound/*")
+                    .requestMatchers(HttpMethod.PUT, "/inbound/*")
+                    .requestMatchers(HttpMethod.DELETE, "/inbound/*")
+                    .requestMatchers("actuator/**"))
+        .authorizeHttpRequests(auth -> auth.anyRequest().permitAll())
+        .build();
```
sbuettner, line 69: Great @mathias-vandaele, could you please document in a comment why we need a second filter chain and why the ordering is needed. It can be deducted but its not obvious. Also please move the method with the lower order upward in the code so it reflects the priority.

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/security/SecurityConfiguration.java
```diff
@@ -43,27 +43,32 @@ public class SecurityConfiguration {
...
-  @Bean
+  @Order(1)
   public SecurityFilterChain filterChain(HttpSecurity http) throws Exception {
     http.csrf(csrf -> csrf.ignoringRequestMatchers("/inbound/**"))
         .authorizeHttpRequests(
-            auth -> {
-              auth.requestMatchers("/inbound", "/tenants/**").hasAuthority("SCOPE_inbound:read");
-            })
+            auth ->
+                auth.requestMatchers("/inbound", "/tenants/**").hasAuthority("SCOPE_inbound:read"))
```
sbuettner, line 51: Could we limit these to GET requests only?

sbuettner (approved):

Issue #2520, opened 2024-05-14
[REST connector] Access response in error expression with status code 400  


saig0:
__Is your feature request related to a problem? Please describe.__

I use the [SpaceTraders connector](https://marketplace.camunda.com/en-US/apps/420889/spacetraders-sdk) in my process and try to access the response in the error expression. The connector is based on Camunda's REST connector. 

I followed the documentation for error expressions ([ref](https://docs.camunda.io/docs/components/connectors/use-connectors/#bpmn-errors-and-failing-jobs)) but I can't access the response if the API returned a status code `400`. 

In my case, the API returns a status code `400` and sends additional data about the failure:
  
```
// status code
400 Bad Request

// URL
POST https://api.spacetraders.io/v2/my/ships/C_4979-1/navigate

// response data
{
   "error":{
      "message":"Ship action failed. Ship is not currently in orbit at X1-ZQ60-A1.",
      "code":4236,
      "data":{
         "waypointSymbol":"X1-ZQ60-A1"
      }
   }
}
```

NOTE: This is specific for a failure response (i.e. status code `4xx`). If the response returns successfully (i.e. status code `2xx`), I can access the response data in the error expression.    

__Describe the solution you'd like__

I can access the response data in the error expression, even if the response has a status code `400`.

I can use the following error expression to throw a BPMN error if the error code in the response matches:

```
if matches(error.code, "4\d\d") and response.body.error.code = 4236 
then bpmnError("shipNotInOrbitError", response.body.error.message, response.body.error.data)
else null
```

__Describe alternatives you've considered__

None. Since the response has a status code `400`, I can't handle the response in the connector or the process.

__Additional context__

How to reproduce:
1. Deploy the process [Handle BPMN error.bpmn](https://github.com/camunda/connectors/files/15305252/Handle.BPMN.error.bpmn.txt) _(with .bpmn file extension)_
2. Create a new instance of the process
3. Wait and verify that the process instance has an incident at the last task

![Screenshot from 2024-05-14 10-10-26](https://github.com/camunda/connectors/assets/4305769/0f122619-cf32-4da2-8b36-151f1aba619d)

![Screenshot from 2024-05-14 10-11-15](https://github.com/camunda/connectors/assets/4305769/32a9ad3a-5788-49a0-b254-4441c62c00dd)

johnBgood:
@saig0 Thank you for reporting this.
Something we can do short term is provide the **response** (if any) in the `error.errorVariables` object.

[...]

PR #2829, opened 2024-07-10
feat(rest): Access http response in error

johnBgood:
## Description
In case of an error we now send the whole response (body + headers) in the `error.variables`. 

It can then be used as follows in an **errorExpression**:

```
if matches(error.code, \"400\") and error.variables.response.body.temp = 36 
then bpmnError(\"Too hot\", error.variables.response.body.message, error.variables.response.body)
else bpmnError(\"Not too hot\", \"The message default\",{fake: \"fakeValue\"})
```

## Related issues

closes https://github.com/camunda/connectors/issues/2520

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/exception/ConnectorExceptionMapper.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/exception/ConnectorExceptionMapper.java
@@ -16,32 +16,27 @@
  */
 package io.camunda.connector.http.base.exception;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import io.camunda.connector.api.error.ConnectorException;
-import io.camunda.connector.api.json.ConnectorsObjectMapperSupplier;
+import io.camunda.connector.api.error.ConnectorExceptionBuilder;
 import io.camunda.connector.http.base.model.HttpCommonResult;
-import io.camunda.connector.http.base.utils.JsonHelper;
-import java.io.Serial;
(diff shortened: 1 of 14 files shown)


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/exception/ConnectorExceptionMapper.java
```diff
@@ -16,32 +16,27 @@
...
+    String reason = Optional.ofNullable(result.reason()).orElse("[no reason]");
+    Map<String, Object> headers = result.headers();
+    Object body = result.body();
+    Map<String, Object> response = new HashMap<>();
+    response.put("headers", headers);
+    response.put("body", body);
+    return new ConnectorExceptionBuilder()
+        .errorCode(status)
+        .message(reason)
+        .errorVariables(Map.of("response", response))
```
sbuettner, line 39: Curious whether this is necessary or already done in https://github.com/camunda/connectors/blob/5c1577352bf5c043dade881ecadae4d20e0e783c/connector-sdk/feel-wrapper/src/main/java/io/camunda/connector/feel/FeelEngineWrapperUtil.java#L23 ?

johnBgood: This is different. The code you posted just wraps a connector's response with this "response" key. It has nothing to do with errorVariables though

sbuettner (approved):

PR #2835, opened 2024-07-11
chore(soap): Send response body when a Fault is received

### LINKED ISSUE #2743: SOAP Connector error response
The SOAP Connector does not offer the possibility to access the response body in case of webservice fault. 
Most error handling is not possible without a body because the actual error can be found in a different place for each response object.

In the connector section of errorHandling it should be possible to access error.response.body or similar. In the connector code itself, the body is already saved in the exception in the event of an error, but is not applied later to the errorResult. 

[SoapConnectorException.java holds the response](https://github.com/camunda/connectors/blob/main/connectors/soap/src/main/java/io/camunda/connectors/soap/SoapConnectorException.java#L16)
[ConnectorJobHandler.java](https://github.com/camunda/connectors/blob/main/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java#L212)

Alternatively, all SOAP calls must be realized via ServiceWorker, which means a considerable amount of additional work.

johnBgood:
## Description
- Send the response body in case of a Fault 
- Some tests improvements

The body is accessible using this kind of `errorExpression`:
`if error.code="SOAP_FAULT_RECEIVED" then bpmnError("Code", "The code", error.variables) else null`

## Related issues

closes #2743

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-soap/src/test/java/io/camunda/connector/e2e/soap/SoapConnectorBaseTest.java
+++ b/connectors-e2e-test/connectors-e2e-test-soap/src/test/java/io/camunda/connector/e2e/soap/SoapConnectorBaseTest.java
@@ -69,7 +69,21 @@ public abstract class SoapConnectorBaseTest {
       </soap:Envelope>
       """;
 
-  @TempDir File tempDir;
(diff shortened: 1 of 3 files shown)


sbuettner (approved):

sbuettner:
🚀

PR #2836, opened 2024-07-12
fix(github-connector): Issue number can now be either a String or an Int

johnBgood:
## Description

- Wrap `issueNumber` into a string 

## Related issues

closes https://github.com/camunda/team-connectors/issues/827


mathias-vandaele (approved):
LGTM 👍

PR #2838, opened 2024-07-15
fix(inbound-message-ttl): Set the message TTL to 1 hour by default

### LINKED ISSUE #2736: [Inbound] Set the message TTL to 1 hour by default, allow overriding with env vars
__What should we do?__

As part of #2341, we added support for customizable message TTL in `8.6.0-alpha2`. The default value is currently configured in the element template and is set to 0.

This might cause issues for our users who already use inbound connectors and may rely on the TTL being set to 1 hour implicitly (the 1h default originally comes from the Zeebe Java client, see https://github.com/camunda/camunda/issues/10937). We need to set the TTL to 1h again by the `8.6.0` release.

At the same time, it would be useful to be able to define the message TTL as an application config concern, and not specify it in the BPMN.

To accomodate both requirements, we should:
- Make the TTL field optional in the element templates
- Set it to 1h by default
- Add an environment variable / spring config property to the runtime that will change the implicit default value
- However, the value specified in the BPMN should have a priority (WYSIWYG)

johnBgood:
## Description

Message Start Event TTL can now be defined as follows, with this order of precedence (low priority to high priority):
- If no env var is provided, default will be 1H.
- If `camunda.connector.inbound.message.ttl` is provided, it will be used as the Runtime default
- `messageTtl` from the connector's configuration will be used if provided

## Related issues

closes #2736

### DIFF
--- a/connectors/aws/aws-sns/element-templates/aws-sns-inbound-boundary.json
+++ b/connectors/aws/aws-sns/element-templates/aws-sns-inbound-boundary.json
@@ -141,12 +141,11 @@
     "id" : "messageTtl",
     "label" : "Message TTL",
     "description" : "Time-to-live for the message in the broker (ISO-8601 duration)",
-    "optional" : false,
-    "value" : "PT0S",
+    "optional" : true,
     "constraints" : {
-      "notEmpty" : true,
+      "notEmpty" : false,
       "pattern" : {
-        "value" : "P.*",
+        "value" : "^(PT.*|)$",
         "message" : "must be an ISO-8601 duration"
       }
     },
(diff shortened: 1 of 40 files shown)


connectors/aws/aws-sns/element-templates/aws-sns-inbound-boundary.json
```diff
@@ -141,12 +141,11 @@
     "id" : "messageTtl",
     "label" : "Message TTL",
     "description" : "Time-to-live for the message in the broker (ISO-8601 duration)",
-    "optional" : false,
-    "value" : "PT0S",
+    "optional" : true,
     "constraints" : {
-      "notEmpty" : true,
+      "notEmpty" : false,
```
johnBgood, line 146: this means `messageTtl` won't be in the XML at all, just wanted to highlight this @chillleader

sbuettner: @johnBgood is that the expected behaviour?

sbuettner (approved):

chillleader (approved):

PR #2840, opened 2024-07-16
chore(config): add slim application.properties file to default-bundle

markfarkas-camunda:
## Description

add application.properties to be able to run connectors locally with docker compose (using docker-compose.core.yaml).


johnBgood (approved):
Thank you!!

PR #2841, opened 2024-07-16
fix(process-import): handle exceptions in imported process definitions explicitly

markfarkas-camunda:
## Description

Handle exceptions in imported process definitions explicitly to avoid setting the variable `ready` to false (in case of an exception in the handleImportedDefinitions), which indicates that connectors component is unhealthy.

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImporter.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImporter.java
@@ -56,7 +56,7 @@ public class ProcessDefinitionImporter {
       try {
         handleImportedDefinitions(result);
       } catch (Exception e) {
-        LOG.error("Failed to handle imported process definitions", e);
+        LOG.warn("Failed to handle imported process definitions", e);
       }
       ready = true;
     } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImporter.java
```diff
@@ -53,7 +53,11 @@ public ProcessDefinitionImporter(
...
       var result = search.query();
-      handleImportedDefinitions(result);
+      try {
+        handleImportedDefinitions(result);
+      } catch (Exception e) {
+        LOG.error("Failed to handle imported process definitions", e);
+      }
       ready = true;
     } catch (Exception e) {
       LOG.error("Failed to import process definitions", e);
```
ev-codes, line 63: Perhaps this error message should be different now, since we now have an import process definition-specific error message on line 59

Maybe something like "Operate search query failed"?

ev-codes (commented):
Thanks @markfarkas-camunda , just one comment!

I'm also wondering if it's worth using a `WARN` instead an `ERROR` log level. Not sure how fatal a misconfigured XML file is, for example.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImporter.java
ev-codes: As discussed on Slack, the error messages already in this PR better reflect the difference between `search.query()` throwing an exception and `handleImportDefinitions()` throwing one.

ev-codes (approved):
Thanks again!

sbuettner:
@markfarkas-camunda Thanks for tackling this.

It is merged already but my concern here is that we should be careful what we log as error. Anything that happens in userland that can lead to an error due to misconfiguration shouldnt necessary be logged as an `ERROR` as this will trigger an alert on our side. What do you think?

PR #2842, opened 2024-07-16
other(rabbitMQ): add integration test for rabbitMQ to verify message integrity

mathias-vandaele:
A check was added to be sure, the unicode escaping is respected 

## Description

An integration test was added to test that unicode are escaped properly

## Related issues

#2730


johnBgood (approved):
Thank you :)

PR #2845, opened 2024-07-18
build(template-generator): change the strategy order so template properties overrides Jakarta

mathias-vandaele:
Instead over using a merge strategy, we decided with @sbuettner to change the order properties are applied, so now template property properties will always overrides Jakarta validation properties

## Description

Change the order of the Fields Processor

## Related issues

closes https://github.com/camunda/team-connectors/issues/714


johnBgood:
I'm not a fan of this solution tbh, it's hidden magic just like it was before this fix. It's just that we reverted the issue. 

I think it's still misleading for ETG users, and they (or even we) might be surprised by the resulting templates.

Or did I miss something, and maybe there's no issue at all? 

@mathias-vandaele @sbuettner

sbuettner:
@johnBgood The idea is that usually one wants to see their bean validation rules are applied automatically as one is usually adding these first. If you then want to customize the template but keep their bean validation annotations in place then this order is more natural ordering as its about the element template generation. Lets discuss if thats still unclear.

johnBgood (approved):
💯

PR #2851, opened 2024-07-19
feat(template-generator): add RestTemplateGenerator interface for rest template specific methods

### LINKED ISSUE #2680: Element template generator examples are broken
__Describe the Bug__

The examples in the element template generator README file are broken.

For example `congen openapi-outbound generate https://petstore3.swagger.io/api/v3/openapi.json placeOrder` is wrong because of at least 2 reasons. 1st, the first two arguments should be `congen generate` instead of `congen openapi-outbound`, and 2nd, adding the `placeOrder` at the end results in: `Generation failed: Cannot invoke "io.camunda.connector.generator.dsl.http.HttpOperationBuilder.getId()" because the return value of "io.camunda.connector.generator.openapi.OperationParseResult.builder()" is null`

__Steps to Reproduce__

1. Open the README.md file in congen-cli project
2. Try to run for example `congen openapi-outbound generate https://petstore3.swagger.io/api/v3/openapi.json placeOrder`
3. Observe that it fails

__Expected Behavior__

The examples should work correctly.

__Environment__

- OS: Windows 11, Ubuntu 22
- Library version: latest
- Camunda version: latest

### LINKED ISSUE #2692: Modify congen to include the tags property in case of OpenAPI, and the folders in case of Postman collection
Epic: https://github.com/camunda/product-hub/issues/2049

__What should we do?__

We should modify the congen (element template generator) to include the tags (for OpenAPI), or the folders (for Postman). 
 
Implementation suggestion: 
- add `pathItem.getPost().getTags()` in the OperationUtil.java
- in PostmanOperationUtil.java in the `extractOperation` method, use the value of the `opIdentifier` variable to extract the foldername(s)

__Why should we do it?__

We should do this to provide proper info through the scan method which makes it possible to group the methods (multi-level grouping is also possible)

markfarkas-camunda:
## Description

Added a new cli command: operations which lists all the operations including the unsupported ones. This will be used for the https://github.com/camunda/product-hub/issues/2049 epic.

## Related issues

closes #2692
closes #2680


element-template-generator/core/src/main/java/io/camunda/connector/generator/api/CliCompatibleTemplateGenerator.java
```diff
@@ -50,4 +53,52 @@ record ScanResult(
...
+   * @param tags Tags of the operation
+   * @param supported Indicate whether the operation is supported or not
+   */
+  record Operation(String id, String path, String method, List<String> tags, boolean supported) {
+    public static class Builder {
+      String id;
+      String path;
+      String method;
+      List<String> tags;
+      boolean supported;
```
sbuettner, line 72: @markfarkas-camunda would it be possible to add a `description` field?  Maybe its not used in the UI yet but it could be handy in the future. Both formats provide this field.

element-template-generator/congen-cli/README.md
```diff
@@ -53,22 +54,22 @@ followed by an optional list of operation IDs to include in the generated templa
 The following command will ask the generator to include only the `placeOrder` operation.
 
 ```shell
-congen openapi-outbound generate https://petstore3.swagger.io/api/v3/openapi.json placeOrder
+congen generate openapi-outbound https://petstore3.swagger.io/api/v3/openapi.json placeOrder
 ```
```
sbuettner, line 58: 👍

element-template-generator/core/src/main/java/io/camunda/connector/generator/api/CliCompatibleTemplateGenerator.java
```diff
@@ -32,6 +32,9 @@ public interface CliCompatibleTemplateGenerator<IN> extends ElementTemplateGener
   /** Provides a usage description for the generator. This description is used in the CLI help. */
   String getUsage();
 
+  /** Scan the source and returns with operations (both supported and not supported). */
+  List<Operation> operations(IN input);
+
```
chillleader, line 37: This interface is only intended for the CLI tool, and it is designed to be high-level and not contain anything specific to individual generator implementations. Have you considered creating a new interface for invoking the HTTP-specific template generators (such as Postman/OpenAPI) instead of extending this generic one?

element-template-generator/congen-cli/README.md
```diff
@@ -34,6 +34,7 @@ Run `congen -h` to see the usage information and a list of available parameters.
 `congen` supports the following commands:
 - `generate` invokes the selected generator implementation and prints the generated template to the standard output.
 - `scan` invokes the generator in a dry-run mode and prints the information
+- `operations` prints the list of all available operations both supported and unsupported
```
chillleader, line 37: From the standpoint of the CLI tool, can this be a modification of the existing `scan` operation? I think the idea behind those two is the same, and it's just the question of data formatting, but please correct me if I'm wrong

markfarkas-camunda: I can totally see why it would seem like that, but they print different data to the output. As it says in the docs, the scan will actually perform a dry-run, create the template and return with the template data too. The template can only contain supported operations. For the https://github.com/camunda/product-hub/issues/2049 we need to list the unsupported operations as well. Also we need additional data, that's why we have to transform the data. I can see `scan` as a useful command, and the `operations` as something for a different purpose. Please feel free to share more details about your perspective if you think that would be a better solution.

element-template-generator/core/src/main/java/io/camunda/connector/generator/api/CliCompatibleTemplateGenerator.java
markfarkas-camunda: Good idea! Added that!

sbuettner (approved):

chillleader (approved):
LGTM, thank you for making the changes 👍

element-template-generator/congen-cli/README.md
chillleader: For history: this was resolved in DM, the core reason is maintaining abstraction isolation

PR #2852, opened 2024-07-19
fix(kafka): Handle consumerGroup properly + update the offset override part

### LINKED ISSUE #2846: Kafka inbound connector creates 2 processes for 1 message
__Describe the Bug__

If there are multiple connector runtimes, one message is correlated twice, and therefore 2 process instances are created instead of 1.

Support ticket: https://jira.camunda.com/browse/SUPPORT-22763

__Steps to Reproduce__

1. Deploy a process definition with a Kafka consumer
2. Send a Kafka message to the specific topic
3. Observe that multiple process instances are created

__Expected Behavior__

Only one process instance should be created for one Kafka message.

__Environment__

Every runtime with a replica node.

johnBgood:
## Description

- Fix the way we handle consumerGroups
  - The `assign` method that we used was overriding Kafka's consumer group mechanism (silently). We now use `subscribe` instead.
- Update the way we handle manual offsets.
- Update UTs.

## Related issues

closes #2846

### DIFF
--- a/connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
+++ b/connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
@@ -13,7 +13,7 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.ArgumentMatchers.anyCollection;
 import static org.mockito.Mockito.doNothing;
 import static org.mockito.Mockito.mock;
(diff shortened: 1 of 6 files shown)


connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
```diff
@@ -52,27 +53,35 @@
...
+        Arguments.of(Arrays.asList(10L, 12L), Arrays.asList(10L, 12L)));
+  }
 
   @BeforeEach
   @SuppressWarnings("unchecked")
   public void setUp() {
-    topic = "my-topic";
-    topicPartitions =
+    String topic = "my-topic";
+    List<PartitionInfo> topicPartitions =
```
markfarkas-camunda, line 84: The CodeQL scan is right, we don't use this variable anymore.

markfarkas-camunda (commented):
I am not sure yet why, but integration tests are consistently failing for me on my local machine. Could you please check if you can run them?

sbuettner (approved):

PR #2898, opened 2024-07-22
build: fix kafka connector integration test

chillleader:
## Description

This PR will fix the Kafka Integration test that has been failing since #2792 was merged.
- The test itself was adjusted to provide a custom retry policy, also a couple small errors were fixed in other test cases
- The consumer code now rethrows the exception when the retry loop fails (this shouldn't affect normal execution in production, as the retry loop is infinite there)

The test itself is marked as disabled, to be discussed whether we want to run it always.

## Related issues

closes https://github.com/camunda/team-connectors/issues/858

### DIFF
--- a/connectors/kafka/src/test/java/io/camunda/connector/kafka/integration/KafkaIntegrationTest.java
+++ b/connectors/kafka/src/test/java/io/camunda/connector/kafka/integration/KafkaIntegrationTest.java
@@ -14,6 +14,7 @@ import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
+import dev.failsafe.RetryPolicy;
 import io.camunda.connector.api.outbound.OutboundConnectorContext;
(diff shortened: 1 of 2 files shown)


connectors/kafka/src/test/java/io/camunda/connector/kafka/integration/KafkaIntegrationTest.java
```diff
@@ -448,7 +454,7 @@ void consumeMessageWithHeaders() {
             kafkaTopic,
             null,
             null,
-            List.of(0L, 0L),
+            null,
```
markfarkas-camunda, line 457: Just to clarify: Why was this change necessary?

markfarkas-camunda (commented):
Looks good, I only have 1 question

sbuettner (approved):

johnBgood (approved):
thank you!

connectors/kafka/src/test/java/io/camunda/connector/kafka/integration/KafkaIntegrationTest.java
chillleader: If we start reading from 0 again, wrong message will be consumed as we are running multiple tests in a certain order. I believe this was added by mistake.

PR #2904, opened 2024-07-22
other(ET-generator): create a @secretPattern Annotation

mathias-vandaele:
## Description

Created a PR to add a `@SecretPattern` annotation.

Not sure of the implementation, I am waiting for the reviewers to give me a feedback.

I noticed that I had to let the Jakarta `@Pattern` annotation for many tests to pass, waiting for your feedback on this as well

## Related issues

closes https://github.com/camunda/team-connectors/issues/712

### DIFF
--- a/connectors/aws/aws-sqs/element-templates/aws-sqs-boundary-connector.json
+++ b/connectors/aws/aws-sqs/element-templates/aws-sqs-boundary-connector.json
@@ -155,7 +155,7 @@
     "constraints" : {
       "notEmpty" : true,
       "pattern" : {
-        "value" : "^(secrets|([0-9]?|1[0-9]|20)).*$",
+        "value" : "^(secrets.*|([0-9]?|1[0-9]|20))$",
         "message" : ""
       }
     },
--- a/connectors/aws/aws-sqs/element-templates/aws-sqs-boundary-connector.json
+++ b/connectors/aws/aws-sqs/element-templates/aws-sqs-boundary-connector.json
@@ -155,7 +155,8 @@
     "constraints" : {
       "notEmpty" : true,
       "pattern" : {
-        "value" : "^([0-9]?|1[0-9]|20|secrets\\..+)$"
+        "value" : "^(secrets|([0-9]?|1[0-9]|20)).*$",
+        "message" : ""
       }
     },
     "group" : "messagePollingProperties",
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
@@ -49,9 +49,9 @@ public class HttpCommonRequest {
       constraints =
           @TemplateProperty.PropertyConstraints(
               notEmpty = true,
-              pattern =
-                  @TemplateProperty.Pattern(
-                      value = "^(http://|https://|secrets|\\{\\{).*$",
+              secretPattern =
+                  @TemplateProperty.SecretPattern(
+                      value = "^(http://|https://|\\{\\{).*$",
                       message = "Must be a http(s) URL")))
   private String url;
 
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
@@ -49,7 +49,8 @@ public class HttpCommonRequest {
       constraints =
           @TemplateProperty.PropertyConstraints(
               notEmpty = true,
-              pattern = @TemplateProperty.Pattern(value = "^\\d+$", message = "Must be a number")))
+              pattern =
+                  @TemplateProperty.Pattern(value = "^\\d+$", message = "Must be a http(s) URL")))
   private String url;
 
   @Valid private Authentication authentication;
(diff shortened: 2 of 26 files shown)


element-template-generator/core/src/main/java/io/camunda/connector/generator/dsl/Property.java
```diff
@@ -40,6 +40,9 @@ public abstract sealed class Property
 
   protected final String type;
 
+  /**
+   * @Pavel
```
johnBgood, line 44: ?

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/processor/TemplatePropertyFieldProcessor.java
```diff
@@ -190,6 +191,24 @@ private PropertyConstraints buildConstraints(
...
+    }
     return builder.build();
   }
+
+  private String createSecretPatternValue(String value, Property property) {
+    if (property.getFeel() == Property.FeelMode.optional
+        || property.getFeel() == Property.FeelMode.required) {
+      return "^(=|(secrets|%s)).*$".formatted(value.replaceAll("\\^", "").replaceAll("\\$", ""));
+    } else {
+      return "^(secrets|%s).*$".formatted(value.replaceAll("\\^", "").replaceAll("\\$", ""));
```
johnBgood, line 211: do we need the `.*` at the end? Or should we just use the initial regex ( `%s`) ?

element-template-generator/core/src/main/java/io/camunda/connector/generator/dsl/Property.java
mathias-vandaele: image

FYI @sbuettner @chillleader @johnBgood 

I did not understand why Simon was asking me this (the `'s` was missing). I realize now how stupid I have been 🤣🤣

connectors/aws/aws-sqs/element-templates/aws-sqs-boundary-connector.json
```diff
@@ -155,7 +155,8 @@
     "constraints" : {
       "notEmpty" : true,
       "pattern" : {
-        "value" : "^([0-9]?|1[0-9]|20|secrets\\..+)$"
+        "value" : "^(secrets|([0-9]?|1[0-9]|20)).*$",
+        "message" : ""
```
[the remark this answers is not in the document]
mathias-vandaele: Here is the pattern for the modeler. FeelMode is disabled here, no variables can be passed and 'null' as a user value could not be inserted as it does not match the pattern

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/processor/TemplatePropertyFieldProcessor.java
mathias-vandaele: Would you prefer something like 

`"^(secrets.*|(%s).*)$".formatted(value.replaceAll("\\^", "").replaceAll("\\$", ""));`

?

connectors/aws/aws-sqs/element-templates/aws-sqs-boundary-connector.json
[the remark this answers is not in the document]
chillleader: I was more wondering about the error message. I knew there was a default message when there is nothing custom provided in the template (`<propertyId> has to match the expression <provided expression>)`. So I was curious if the same message would be displayed if we provide an empty string, or if it would just display the empty string as is. I checked it in the modeler and it looks like the behavior is identical and the default message is shown in both cases, so I think we are good 👍

mathias-vandaele:
@johnBgood @chillleader Is there any others comments for this feature ?

connectors/http/rest/element-templates/http-json-connector.json
```diff
@@ -352,7 +352,7 @@
     "constraints" : {
       "notEmpty" : true,
       "pattern" : {
-        "value" : "^(=|(http://|https://|secrets|\\{\\{).*$)",
+        "value" : "^(http://|https://|secrets|\\{\\{).*$",
```
sbuettner, line 355: Would that still work with FEEL expressions?

mathias-vandaele: Yes thank you, the HttpCommonRequest had `@Pattern` instead of `@SecretPattern`

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -43,7 +43,16 @@ public class HttpCommonRequest {
...
+  @TemplateProperty(
+      group = "endpoint",
+      label = "URL",
+      constraints =
+          @TemplateProperty.PropertyConstraints(
+              notEmpty = true,
+              secretPattern =
+                  @TemplateProperty.SecretPattern(
+                      value = "^(http://|https://|\\{\\{).*$",
+                      message = "Must be a http(s) URL")))
```
sbuettner, line 55: Now we have two annotations?

sbuettner: Asking because now the pattern is duplicated in the `@Pattern` one and the `@TemplateProperty.SecretPattern`. I thought the idea was to reduce the duplication? Just curious :)

Issue #2939, opened 2024-07-24
Connectors Pod is crash looping because of bpmn importation


mathias-vandaele:
__Describe the Bug__

A connector pod is constantly crash looping due to Bpmn importation.
I noticed this is due to the function.

```
  private Map<String, String> getRawProperties(BaseElement element)
```

in `ProcessDefinitionInspector.java` L308

The log trace is 

```
Failed to inspect process definition 2251799844334767
java.lang.IllegalStateException: Duplicate key approver (attempted merging values xxxxxxxxx and yyyyyyyyyy)
```

__Steps to Reproduce__

I have seen this issue in production, I don't know how to reproduce

__Expected Behavior__

Connectors should start properly

__Environment__

- Camunda version: SASS [Camunda 8.3+gen14]

PR #2980, opened 2024-07-29
other(bpmn-import): log warning when duplicate name is found inside process definition

mathias-vandaele:
## Description

a log (warning) is printed when a duplicate key is found inside a BPMN process, The duplicate value is ignored

## Related issues

closes https://github.com/camunda/connectors/issues/2939


sbuettner (commented):
Great, could you please add a test case?

sbuettner (approved):

PR #3015, opened 2024-08-02
fix(jdbc-tests): tests were being flaky due to deprecated method being used

mathias-vandaele:
Tests were being flaky due to deprecated method being used

## Description

Not urgent, but some tests were becoming flaky in jdbc due to usage of deprecated method

## Related issues

None


ev-codes (approved):
Thanks for this improvement!

PR #3018, opened 2024-08-02
fix(kafka): close consumer before retrying on failures

chillleader:
## Description

This PR will fix the issue with the Kafka connector retry loop. Previously, we didn't close the consumer properly before retrying, which resulted in longer waiting time before the new consumer could connect to the topic (it had to wait until the old consumer is killed by timeout).

## Related issues

related: #2750, #2792


connectors/kafka/src/test/java/io/camunda/connector/kafka/inbound/KafkaExecutableTest.java
```diff
@@ -172,11 +172,11 @@ void testActivateAndDeactivate_consumerThrows() {
         .atMost(Duration.ofSeconds(5))
         .pollInterval(Duration.ofMillis(500))
         .untilAsserted(() -> assertFalse(kafkaExecutable.kafkaConnectorConsumer.shouldLoop));
-    kafkaExecutable.deactivate();
```
chillleader, line 175: This was removed because it's not really needed and the test was flaky because of it (sometimes it was deactivated before checking, sometimes not, resulting in a varying number of close() calls)

markfarkas-camunda (approved):

PR #3051, opened 2024-08-07
feat(document): serialization & deserialization

chillleader:
## Description

This PR will add the first implementation of serialization and deserialization of document references.

Highlights:
- Jackson deserializer to handle the simple case when connector wants to get a `Document` object
- Jackson deserializer to handle more complex cases when target type is not a `Document` but the document can be converted to the target type
- Operation execution
  - Lazy and eager models are available
  - Lazy is applied by default when the connector doesn't specify its target type (e.g. just `Object`). The operation will be evaluated on serialization instead in this case.
- Jackson serializer that can execute operations
- Refined document class structure

## Related issues

related https://github.com/camunda/team-connectors/issues/873


chillleader:
Re-opened the PR due to wrong feature branch base: https://github.com/camunda/connectors/pull/3066

PR #3066, opened 2024-08-07
feat(document): document deserialization & serialization

chillleader:
## Description

This PR will add the first implementation of serialization and deserialization of document references.

Highlights:

- Jackson deserializer to handle the simple case when connector wants to get a Document object
- Jackson deserializer to handle more complex cases when target type is not a Document but the document can be converted to the target type
- Operation execution
  - Lazy and eager models are available
  - Lazy is applied by default when the connector doesn't specify its target type (e.g. just Object). The operation will be evaluated on serialization instead in this case.
- Jackson serializer that can execute operations
- Refined document class structure
- DocumentStore interface and in-memory impl

## Related issues

related https://github.com/camunda/team-connectors/issues/873


sbuettner (approved):

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -79,7 +79,6 @@ public class HttpCommonRequest {
       description = "Map of HTTP headers to add to the request")
   private Map<String, String> headers;
 
-  @FEEL
```
mathias-vandaele, line 82: Why removing it `@FEEL` ?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/document/DocumentFactoryImpl.java
```diff
@@ -0,0 +1,51 @@
...
+import io.camunda.connector.api.document.DocumentReference.ExternalDocumentReference;
+import io.camunda.connector.api.document.store.DocumentCreationRequest;
+import io.camunda.connector.api.document.store.DocumentStore;
+
+public class DocumentFactoryImpl implements DocumentFactory {
+
+  private final DocumentStore documentStore;
+
+  public DocumentFactoryImpl(DocumentStore documentStore) {
+    this.documentStore = documentStore;
```
mathias-vandaele, line 32: Will this class carry all of the future possible store ? In this case, how do you make the link between the document creation request and the corresponding store?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
```diff
@@ -47,17 +52,34 @@ public class JobHandlerContext extends AbstractConnectorContext
...
+  public JobHandlerContext(
+      final ActivatedJob job,
+      final SecretProvider secretProvider,
+      final ValidationProvider validationProvider,
+      final ObjectMapper objectMapper) {
+    this(
+        job,
+        secretProvider,
+        validationProvider,
+        new DocumentFactoryImpl(new InMemoryDocumentStore()),
```
mathias-vandaele, line 79: How will it work when new store comes?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
chillleader: Good catch, this is accidental 😄

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/document/DocumentFactoryImpl.java
chillleader: In the future we will generally use only one store here for connecting to the C8 document API. The in-memory implementation will be reserved for tests. As of now, I don't see the need to have multiple store implementations active at the same time - different stores are anyway hidden behind the C8 REST API.

One thing where a different approach is needed is external documents (such as something only available to us via an external URL). But this probably will be handled separately in the `DocumentFactory`

mathias-vandaele (approved):

connector-sdk/core/src/main/java/io/camunda/connector/api/document/DocumentFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package io.camunda.connector.api.document;
+
+import io.camunda.connector.api.document.store.DocumentCreationRequest;
+
+public interface DocumentFactory {
+
+  /** Given a document reference, create the Document object */
+  Document parse(DocumentReference reference);
```
sbuettner, line 24: Maybe `resolve` or `get` would be less technical.

connector-sdk/core/src/main/java/io/camunda/connector/api/document/DocumentReference.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.api.document;
+
+public interface DocumentReference {
```
sbuettner, line 19: Should we seal this interface?

sbuettner (approved):
Great work! Some minor comments.

connector-sdk/core/src/main/java/io/camunda/connector/api/document/DocumentReference.java
chillleader: It's being extended / implemented by classes in a different package (`DocumentReferenceImpl` where operation is added) so I couldn't seal it here

sbuettner (approved):

connector-sdk/jackson-datatype-document/src/main/java/io/camunda/connector/document/annotation/jackson/DocumentReferenceModel.java
```diff
@@ -0,0 +1,84 @@
...
+    visible = true,
+    include = As.EXISTING_PROPERTY)
+@JsonSubTypes({
+  @JsonSubTypes.Type(value = CamundaDocumentReferenceModel.class, name = "camunda"),
+  @JsonSubTypes.Type(value = ExternalDocumentReferenceModel.class, name = "external")
+})
+@JsonIgnoreProperties(ignoreUnknown = true)
+public sealed interface DocumentReferenceModel extends DocumentReference {
+
+  String DISCRIMINATOR_KEY = "$documentType";
```
sbuettner, line 47:
```suggestion
  String DISCRIMINATOR_KEY = "documentType";
```

Should we remove the $-sign as done here: https://github.com/camunda/camunda/pull/21080/files#diff-a18635d76659a781a6e8d301bd0fed7c77fd6cc5c8cfc57e11c5bbc19c78c9cdR1783 ?

PR #3080, opened 2024-08-12
fix(core) rest connector inbound was not working with query params

mathias-vandaele:
## Description

removing the value parameter in annotation.

Tested for message start and intermediate catch event successfully.

## Related issues

closes https://github.com/camunda/team-connectors/issues/876


chillleader (dismissed):

sbuettner (commented):
@mathias-vandaele Can we have a simple test to make sure this doesnt happen again?

sbuettner (approved):

sbuettner (approved):

PR #3081, opened 2024-08-12
feat: add /connector-test endpoint to test outbound connector

markfarkas-camunda:
## Description

Hackdays project to test Outbound Connectors from web-modeler. A new `/connector-test` was added to trigger the execution of a connector.


markfarkas-camunda:
Based on [the decision here](https://github.com/camunda/product-hub/issues/2453#issuecomment-2368091254), we won't have a dedicated endpoint for this purpose in the connectors repo.

PR #3107, opened 2024-08-16
fix(sqs): Minor fix to prevent user from entering a 0 polling time which would result to CPU throtling.

mathias-vandaele:
## Description

It prevents the user to enter a 0 polling time value.

That's a solution I can propose to resolve this issue https://github.com/camunda/team-connectors/issues/831

@sbuettner @chillleader What do you think?

## Related issues

closes https://github.com/camunda/team-connectors/issues/831 potentially


sbuettner (commented):
We can do this but we should keep in mind that this wont change any already configured SQS Connector. We could look into the implementation and always set it to 1 second in case its set to 0. @mathias-vandaele were there any findings regarding the SQS polling code that lead to the CPU throttling beside this?

johnBgood (approved):

Issue #3108, opened 2024-08-16
[email connector] SMTP outbound connector


mathias-vandaele:
__What should we do?__

As a first step of the emails connector, the SMTP outbound connector will be made

__Why should we do it?__

Mail connector EPIC https://github.com/camunda/product-hub/issues/2430

PR #3132, opened 2024-08-20
feat(email connector): implement new email connector, Outbound SMTP, IMAP and POP3, Inbound IMAP

mathias-vandaele:
## Description

New implementation of the email connectors.

It has not been completely polished but ready for review.

I would like to come back on the label for the ET, and add new tests

## Related issues

related https://github.com/camunda/connectors/issues/3108
related https://github.com/camunda/connectors/issues/3213
related https://github.com/camunda/connectors/issues/3214
related https://github.com/camunda/connectors/issues/3215


connectors/email/src/main/java/io/camunda/connector/email/core/jakarta/JakartaEmailListener.java
```diff
@@ -0,0 +1,107 @@
...
+      idleManager.watch(imapFolder);
+    } catch (MessagingException ex) {
+      throw new RuntimeException(ex);
+    }
+  }
+
+  private List<String> createInboxList(Object folderToListen) {
+    return switch (folderToListen) {
+      case List<?> list -> list.stream().map(Object::toString).toList();
+      case String string -> Arrays.stream(string.split(",")).toList();
```
johnBgood, line 101: List.of(string.split(",")) ?

connectors/email/src/main/java/io/camunda/connector/email/outbound/protocols/Smtp.java
```diff
@@ -0,0 +1,63 @@
...
+import io.camunda.connector.email.outbound.protocols.actions.Action;
+import io.camunda.connector.email.outbound.protocols.actions.SmtpAction;
+import io.camunda.connector.email.outbound.protocols.actions.SmtpSendEmail;
+import io.camunda.connector.generator.java.annotation.NestedProperties;
+import io.camunda.connector.generator.java.annotation.TemplateSubType;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+
+@TemplateSubType(id = "smtp", label = "SMTP")
+public final class Smtp implements Protocol {
```
johnBgood, line 22: could be a record

connectors/email/src/main/java/io/camunda/connector/email/outbound/protocols/Pop3.java
```diff
@@ -0,0 +1,66 @@
...
+import io.camunda.connector.email.config.Configuration;
+import io.camunda.connector.email.config.Pop3Config;
+import io.camunda.connector.email.outbound.protocols.actions.*;
+import io.camunda.connector.generator.java.annotation.NestedProperties;
+import io.camunda.connector.generator.java.annotation.TemplateSubType;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+
+@TemplateSubType(id = "pop3", label = "POP3")
+public final class Pop3 implements Protocol {
```
johnBgood, line 20: could be a record

connectors/email/src/main/java/io/camunda/connector/email/outbound/protocols/Imap.java
```diff
@@ -0,0 +1,67 @@
...
+import io.camunda.connector.email.config.Configuration;
+import io.camunda.connector.email.config.ImapConfig;
+import io.camunda.connector.email.outbound.protocols.actions.*;
+import io.camunda.connector.generator.java.annotation.NestedProperties;
+import io.camunda.connector.generator.java.annotation.TemplateSubType;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+
+@TemplateSubType(id = "imap", label = "IMAP")
+public final class Imap implements Protocol {
```
johnBgood, line 20: could be a record

connectors/email/src/main/java/io/camunda/connector/email/config/SmtpConfig.java
```diff
@@ -0,0 +1,81 @@
...
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.email.config;
+
+import io.camunda.connector.generator.dsl.Property;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+
+public final class SmtpConfig implements Configuration {
```
johnBgood, line 14: a lot of classes could be records, I'll let you check and decide :)

connectors/email/src/main/java/io/camunda/connector/email/core/ActionExecutor.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.email.core;
```
johnBgood, line 7: maybe this package should be named `client` or similar not `core`

connectors/email/src/main/java/io/camunda/connector/email/core/jakarta/JakartaActionExecutor.java
```diff
@@ -0,0 +1,342 @@
...
+                          email.getMessageId(),
+                          email.getFrom(),
+                          email.getSubject(),
+                          email.getSize()))
+              .limit(pop3ListEmails.getMaxToBeRead())
+              .toList();
+        }
+      }
+    } catch (MessagingException e) {
+      throw new RuntimeException(e);
```
johnBgood, line 243: consider adding some error logs (if we can show some valuable information, if not ignore my comment).
Same for other exceptions thrown

johnBgood (commented):
Great work, I left some minor comments

connectors/email/src/main/java/io/camunda/connector/email/authentication/Authentication.java
```diff
@@ -0,0 +1,36 @@
...
+    name = "type",
+    defaultValue = "simple",
+    description = "")
+public sealed interface Authentication permits NoAuthentication, SimpleAuthentication {
+
+  String getSender();
+
+  Optional<String> getUser();
+
+  Optional<String> getSecret();
```
sbuettner, line 31: What about `username´ and `password` instead of `User` and `Secret`?
Also wondering why they are wrapped in an `Optional`? 

Is `sender` really part of the authentication?

connectors/email/src/main/java/io/camunda/connector/email/authentication/Authentication.java
```diff
@@ -0,0 +1,36 @@
...
+
+  default boolean isSecuredAuth() {
+    return getUser().isPresent() && getSecret().isPresent();
+  }
```
sbuettner, line 35: We have a sealed type `NoAuthentication`, why the need to have a separate method for this?

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
```diff
@@ -0,0 +1,259 @@
...
+          sentAt,
+          receivedAt,
+          message.getSize());
+
+    } catch (MessagingException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static Email createEmail(Message message) {
```
sbuettner, line 101: I would recommend to handle this outside of the `Email` model class.

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
```diff
@@ -0,0 +1,259 @@
...
+          email.getCc(),
+          email.getSentAt(),
+          email.getReceivedAt(),
+          email.getSize());
+    } catch (IOException | MessagingException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static EmailBody processMultipart(
```
sbuettner, line 129: Same as above.

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
```diff
@@ -0,0 +1,259 @@
...
+
+  public EmailBody getBody() {
+    return body;
+  }
+
+  public Integer getSize() {
+    return size;
+  }
+
+  public static class EmailBody {
```
sbuettner, line 198: Can be a record.

connectors/email/src/main/java/io/camunda/connector/email/client/ActionExecutor.java
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.email.client;
+
+import io.camunda.connector.email.outbound.model.EmailRequest;
+
+public interface ActionExecutor {
```
sbuettner, line 11: What about `EmailActionExecutor`?

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/JakartaEmailListener.java
```diff
@@ -0,0 +1,140 @@
...
+      MessageCountEvent e,
+      InboundConnectorContext connectorContext,
+      boolean triggerAdded,
+      boolean markAsRead) {
+    IMAPFolder imapFolder = (IMAPFolder) e.getSource();
+    if (triggerAdded) {
+      Arrays.stream(e.getMessages())
+          .peek(
+              message -> {
+                if (markAsRead) this.jakartaUtils.markAsSeen(message);
```
sbuettner, line 97: Shouldnt it be marked after the correlation? If the correlation fails we have already `seen` the message and wont process it again.

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/JakartaEmailListener.java
```diff
@@ -0,0 +1,140 @@
...
+          .forEach(
+              email ->
+                  connectorContext.correlateWithResult(
+                      new ReadEmailResponse(
+                          email.getMessageId(),
+                          email.getFrom(),
+                          email.getSubject(),
+                          email.getSize(),
+                          email.getBody().getBodyAsPlainText(),
+                          email.getBody().getBodyAsHtml())));
```
sbuettner, line 109: Could be extracted into a separate method and switched to a `map(this::correlateEmail)`

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/JakartaEmailListener.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+    try {
+      idleManager.watch(imapFolder);
+    } catch (MessagingException ex) {
+      throw new RuntimeException(ex);
+    }
+  }
+
+  private List<String> createInboxList(Object folderToListen) {
```
sbuettner, line 118: Can we use a proper type like `List<Inbox>`?  or something similar without string handling?

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/JakartaEmailListener.java
```diff
@@ -0,0 +1,140 @@
...
+
+  public void stopListener() {
+    try {
+      this.idleManager.stop();
+      for (IMAPFolder folder : this.imapFolders) {
+        folder.close();
+      }
+      this.store.close();
+    } catch (MessagingException e) {
+      throw new RuntimeException(e);
```
sbuettner, line 137: Logging this `error` would help to debug it.

connectors/email/src/main/java/io/camunda/connector/email/inbound/model/EmailProperties.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package io.camunda.connector.email.inbound.model;
+
+import io.camunda.connector.email.authentication.Authentication;
+import io.camunda.connector.generator.java.annotation.NestedProperties;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+
+public class EmailProperties {
```
sbuettner, line 15: `EmailProperties` is quite generic. What about `EmailInboundConnectorProperties`.

connectors/email/src/main/java/io/camunda/connector/email/inbound/model/EmailListenerData.java
```diff
@@ -0,0 +1,98 @@
...
+package io.camunda.connector.email.inbound.model;
+
+import io.camunda.connector.email.config.ImapConfig;
+import io.camunda.connector.generator.dsl.Property;
+import io.camunda.connector.generator.java.annotation.NestedProperties;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
+
+public class EmailListenerData {
```
sbuettner, line 16: Can be a record.

connectors/email/src/main/java/io/camunda/connector/email/inbound/model/EmailListenerData.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @TemplateProperty(
+      label = "Folder to listen",
+      group = "listenerInfos",
+      id = "data.folderToListen",
+      description = "",
+      optional = true,
+      feel = Property.FeelMode.optional,
+      binding = @TemplateProperty.PropertyBinding(name = "data.folderToListen"))
+  private Object folderToListen;
```
sbuettner, line 30: Lets go with a proper type instead of `Object`.

connectors/email/src/main/java/io/camunda/connector/email/inbound/model/EmailListenerData.java
sbuettner: Name is quite generic. What about `EmailListenerConfig`

sbuettner:
Great work @mathias-vandaele. Added some comments.

connectors/email/src/main/java/io/camunda/connector/email/authentication/Authentication.java
mathias-vandaele: In the future, we might add a Oauth2 authentication as some servers does not allow user/password authentication.
But those protocols works in the same ways under the hood, so secret is more generic to retrieve either the password or the secret token.

Sender might be the wrong name, that is the email we want to connect to:

`sender` is a bit more complicated, I agree it should probably be only on the SMTP send email action. But that would bring redundancy, the user would authenticate to `example@camunda.com` and he would have to enter a `from` variable on the action.

and we can not use the getUser function because in the future with Oauth2 authentication, the user is not the email address. This might be the easier method cause as of now, `getSender` and `getUser` returns the same value, but that might not be the case in another authentication implementation. I also agree the name is not the best, something like `emailAddress` might be better.

Let me know

connectors/email/src/main/java/io/camunda/connector/email/authentication/Authentication.java
mathias-vandaele: The configuration is different depending if the connexion is authenticated or not, if we remove the `NoAuthentication` then this function is not necessary

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
```diff
@@ -0,0 +1,259 @@
...
+
+public class Email {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(Email.class);
+  private final EmailBody body;
+  private final String messageId;
+  private final String subject;
+  private final List<String> from;
+  private final List<String> to;
+  private final List<String> cc;
```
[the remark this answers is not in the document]
mathias-vandaele: I don't know if it is really necessary cause addresses are really just a list of addresses, and an address is just a String. Let me know

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
mathias-vandaele: ok 😁

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/JakartaEmailListener.java
mathias-vandaele: Yes

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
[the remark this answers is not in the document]
sbuettner: The nice thing with a value record would be that you could make use of `javax.validation.constraints.Email` for validation of emails.

connectors/email/src/main/java/io/camunda/connector/email/inbound/model/EmailListenerData.java
mathias-vandaele: The Object can really depends:

I can have 
```
{
  "field":"FROM",
  "value":"example@camunda.com"
}
```

or the example given previously, so the root object is not always the same

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/Email.java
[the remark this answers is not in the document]
mathias-vandaele: Email is intermediary, I mainly use it to transform any Message into something readable, in order to return nice response to the user. 
The validation is already done on the Jakarta mail client side

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/JakartaEmailListener.java
mathias-vandaele: This function is really just to format the object, we dont know if that is a String comma separated or already a List made by the Feel engine

mathias-vandaele:
@johnBgood Would you be ok for final approvement ?

johnBgood (approved):
Nothing blocking, just a few comments 👍

PR #3135, opened 2024-08-20
fix(rabbitmq): unnecessary unescaping in inbound connector

### LINKED ISSUE #2730: RabbitMQ Connector does unnecessary message unsescape action.

chillleader:
## Description

This PR removes unnecessary unescaping from the RabbitMQ inbound connector.

## Related issues

closes #2730 

related PR: https://github.com/camunda/connectors/pull/2778
(fixed in 8.6 but not backported)

### DIFF
--- a/connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/RabbitMqConsumer.java
+++ b/connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/inbound/RabbitMqConsumer.java
@@ -22,7 +22,6 @@ import io.camunda.connector.rabbitmq.inbound.model.RabbitMqInboundResult.RabbitM
 import io.camunda.connector.rabbitmq.supplier.ObjectMapperSupplier;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import org.apache.commons.text.StringEscapeUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


johnBgood:
Thanks! Any reason why we don't backport the whole PR in this release branch? (I'm thinking about having this issue with Kafka consumer etc)

chillleader:
We originally didn't want to backport the outbound connectors changes due to a bug in the previous versions of Zeebe: https://github.com/camunda/connectors/issues/2730#issuecomment-2185701205

Good point about Kafka - I will add it as well 👍

chillleader:
Actually I'm not sure about the Kafka change now:

```
private static Object parseKey(
      ConsumerRecord<Object, Object> consumerRecord, ObjectReader objectReader) {
    try {
      return objectReader.readTree((String) consumerRecord.key());
    } catch (Exception e) {
      LOG.debug("Cannot parse key to json object -> use as string");
      return StringEscapeUtils.unescapeJson((String) consumerRecord.key());
    }
  }
```

Looks like unescaping is fallback here and will not cause such an issue directly. I'd rather keep it as is unless we experience problems with it.

sbuettner (approved):

PR #3178, opened 2024-08-27
chore(config): Disable multiple Runtime instance as the default

johnBgood:
## Description

By default, we enabled multiple Runtime instances by setting the server port to a random value determined at Runtime.

This is not practical for different reasons:
- It's not often that we need multiple instances to test some feature
- APIs like /inbound are broken as one doesn't know the port soon enough

We will keep the previous value for this port (8080), and one can still uncomment this line if multiple instances are required.


chillleader (approved):
Thank you!

PR #3205, opened 2024-08-29
fix(document): handle SDK document reference correctly in the serializer

chillleader:
## Description

This PR fixes a bug because of which document references from the connector context were not serialized correctly.


johnBgood (approved):
TY!

PR #3249, opened 2024-09-09
feat(schema-registry): Kafka Schema Registry support

johnBgood:
## Description

- Add support for Confluent Schema Registry (Avro, Json)
- Make Kafka messages `key` optional as it should be

⚠️ **Important**
This PR will break compatibility for:
- Customers using the new templates with an old runtime (Avro schema will be ignored)
- Customers using old templates with the new runtime (same)

## Related issues

closes https://github.com/camunda/team-connectors/issues/646
closes https://github.com/camunda/team-connectors/issues/874

## Checklist


connectors/kafka/element-templates/kafka-inbound-connector-start-event.json
johnBgood, line 1: is it ok that we stop supporting the StartEvent?

connectors/kafka/src/main/java/io/camunda/connector/kafka/model/schema/AvroInlineSchemaStrategy.java
```diff
@@ -4,23 +4,31 @@
...
-            label = "Avro schema",
+            label = "Schema",
             constraints = @TemplateProperty.PropertyConstraints(notEmpty = true),
-            condition =
-                @TemplateProperty.PropertyCondition(
-                    property = "serializationType",
-                    equals = "avro"),
-            description = "Avro schema for the message value")
-        String schema) {}
+            description = "Inline schema (JSON or AVRO) for the message value")
```
chillleader, line 28:
```suggestion
            description = "Inline schema (Avro) for the message value")
```

chillleader (approved):
LGTM, also great job with the tests 🙌 

As discussed offline, one thing that bothers me a little is that users will need to provide a schema for publishing messages even if they are using the schema registry. We can improve this in iterations though after we get some feedback.

PR #3280, opened 2024-09-19
chore(deps): bump identity SDK to 8.5.5

chillleader:
Increase the version of Identity SDK that comes from spring-zeebe transitively.


ev-codes (approved):
Thanks for handling this!

Assuming the tests will pass, I approve 🚀

PR #3288, opened 2024-09-23
chore(inbound): Set the readiness to True when starting

johnBgood:
## Description

Set the `ready` property to true by default, so the health check passes when the container starts. This would avoid issues when loading definitions is super long.

## Related issues

closes https://github.com/camunda/team-connectors/issues/912

## Checklist


chillleader (approved):
Thank you!

sbuettner (approved):

PR #3467, opened 2024-10-08
fix(azure openai): Dont send null values to rest connector

sbuettner:
## Description

[#938](https://github.com/camunda/team-connectors/issues/938)

## Related issues

closes https://github.com/camunda/team-connectors/issues/938

## Checklist


sbuettner:
Did not think about the Copilot but thanks for the reminder. As its an element template only change I think we dont have to "backport" it but rather update it in the web-modeler as well. Will add it to the 8.7-alpha1 milestone.

Issue #3319, opened 2024-09-24
Support inbound Connector restarts 


sbuettner:
__Is your feature request related to a problem? Please describe.__

There is currently no simple restart mechanism for Inbound Connectors that can be used by developers to restart their Connectors. In case of any client exceptions like networking issues the inbound Connectors are stopping and users have to manually re-deploy their processes to re-activate them. 

This is not a great user experience and we can improve this situation with a solution that we already have tested in the Kafka Connector: https://github.com/camunda/connectors/blob/main/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaConnectorConsumer.java#L116

__Describe the solution you'd like__

We want to simplify the developer experience of handling restarts for inbound connectors and would therefore enhance the existing `cancel(final Throwable exception)` method implementation. 

The `cancel` method should check the exception and should look for a specific type like for example `RestartException`. If this exception is thrown the runtime should restart the Connector using the exception properties like backoff time or restart delay.

PR #3486, opened 2024-10-15
feat(restart-inbound): give the possibility to the developer to plan a restart of the connector in case of failure

mathias-vandaele:
## Description

Proposal of solution for the restart of inbound connector, not tested yet

## Related issues

closes https://github.com/camunda/connectors/issues/3319

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/RestartException.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/RestartException.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.core.error;
+
+import java.time.Duration;
+
+public class RestartException extends RuntimeException {
+
+  private final Duration backoffTime;
+
+  public RestartException(String message, Throwable cause, Duration backoffTime) {
+    super(message, cause);
+    this.backoffTime = backoffTime;
+  }
+
+  public Duration getBackoffTime() {
+    return backoffTime;
+  }
+}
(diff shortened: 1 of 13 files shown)


sbuettner:
Hey @mathias-vandaele, the ticket contains an implementation suggestion: https://github.com/camunda/connectors/issues/3319 to re-use the existing cancellation mechanism. Would be interesting to understand you see the need to introduce another method?

mathias-vandaele:
> Hey @mathias-vandaele, the ticket contains an implementation suggestion: #3319 to re-use the existing cancellation mechanism. Would be interesting to understand why you see the need to introduce another method?

@sbuettner everything still happens during the cancellation mechanism 

 ```
 @Override
  public void cancel(Throwable exception) {
    try {
      cancellationCallback.accept(exception);
    } catch (Throwable e) {
      LOG.error("Failed to deliver the cancellation signal to the runtime", e);
    }
    if (Objects.requireNonNull(exception) instanceof RestartException restartException) {
      reactivationCallback.accept(restartException.getBackoffTime());
    }
  }
```

A callback is declared in order to be executed after cancelation 👍

File: connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/RestartException.java [hunk 1]
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.core.error;
+
+import java.time.Duration;
+
+public class RestartException extends RuntimeException {
```
johnBgood, line 21: Does it make sense to reuse the `ConnectorRetryException` with the `retries` and `backoff` parameters?

sbuettner:
@mathias-vandaele Ah, I was confused as I saw: ```reactivationCallback```.  Would it be simpler to re-use the existing cancellation callback and handle it in the consumer?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/RestartException.java
sbuettner: @johnBgood Good idea although we would need a way to restart without defining the number of retries. We could add it. WDYT?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
```diff
@@ -28,37 +28,89 @@
...
 import org.slf4j.LoggerFactory;
 import org.springframework.scheduling.annotation.Scheduled;
 
 public class InboundExecutableRegistryImpl implements InboundExecutableRegistry {
 
+  private static final Logger LOG = LoggerFactory.getLogger(InboundExecutableRegistryImpl.class);
+  final Map<UUID, RegisteredExecutable> executables = new HashMap<>();
   private final BlockingQueue<InboundExecutableEvent> eventQueue;
+  private final ScheduledExecutorService reactivationScheduler =
+      Executors.newSingleThreadScheduledExecutor();
```
johnBgood, line 52: any reason to use a single thread executor here?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/error/RestartException.java
johnBgood: Yep, either that or we add the retries count to the context and when throwing the exception later we use this number, similar to how it's done for Outbound connectors.

PR #3505, opened 2024-10-16
feat:(restart-inbound-connectors): create a restart inbound mechanism after cancelling

mathias-vandaele:
## Description

Second try for implementing the restart mechanism, I have tested it and it seems to be working
image

## Related issues

closes #

## Checklist


connector-sdk/core/src/main/java/io/camunda/connector/api/error/RestartException.java
```diff
@@ -0,0 +1,44 @@
...
+  public Duration getDelay() {
+    return delay;
+  }
+
+  public Integer getRetryAttempts() {
+    return retryAttempts;
+  }
+
+  @Override
+  public Throwable getCause() {
```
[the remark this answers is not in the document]
sbuettner: @mathias-vandaele Can we re-use `ConnectorRetryException`?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/CancellationManager.java
```diff
@@ -0,0 +1,119 @@
...
+      LOG.error("Inbound connector executable not found for the given ID: {}", uuid);
+      return;
+    }
+    if (toCancel instanceof RegisteredExecutable.Activated activated) {
+      activated.context().reportHealth(Health.down(throwable));
+      try {
+        activated.executable().deactivate();
+        executables.remove(uuid);
+        executables.put(
+            uuid, new RegisteredExecutable.Cancelled(activated.executable(), activated.context()));
```
sbuettner, line 109: Could you use `Map.replace()` to switch the entry?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/CancellationManager.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CancellationManager {
```
sbuettner, line 31: Can you provide some documentation what the `ConcellationManager` is responsible for?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/CancellationManager.java
```diff
@@ -0,0 +1,119 @@
...
+
+  private void tryRestart(
+      UUID uuid, RegisteredExecutable toReactivate, Duration delay, Integer retryAttempts) {
+    if (retryAttempts < 0) return;
+    if (toReactivate instanceof RegisteredExecutable.Cancelled cancelled) {
+      try {
+        cancelled.executable().activate(cancelled.context());
+        executables.remove(uuid);
+        executables.put(
+            uuid, new RegisteredExecutable.Activated(cancelled.executable(), cancelled.context()));
```
sbuettner, line 82: Could you use Map.replace() to switch the entry?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/CancellationManager.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+    this.reactivationExecutor.schedule(
+        () -> tryRestart(uuid, toReactivate, delay, retryAttempts),
+        delay.getSeconds(),
+        TimeUnit.SECONDS);
+  }
+
+  private void tryRestart(
+      UUID uuid, RegisteredExecutable toReactivate, Duration delay, Integer retryAttempts) {
+    if (retryAttempts < 0) return;
```
sbuettner, line 76: What is the specification here? We stop restart if the the attempts are for example -1? Its not clear when this would happen

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
```diff
@@ -311,6 +288,13 @@ private ActiveExecutableResponse mapToResponse(UUID id, RegisteredExecutable con
                   new Error(
                       "Activation failure", "Invalid connector definition: " + invalid.reason())),
               List.of());
+      case Cancelled cancelled ->
+          new ActiveExecutableResponse(
+              id,
+              cancelled.executable().getClass(),
+              cancelled.context().connectorElements(),
+              cancelled.context().getHealth(),
+              cancelled.context().getLogs());
```
sbuettner, line 297: Do we report the exception somewhere that cause the cancellation? Would be enough if its part of the health object.

sbuettner:
LGTM, some minor comments and please add some tests for the different paths.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
mathias-vandaele: true 👍

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/CancellationManager.java
mathias-vandaele: Here depends of what we want to do:

- In this case, `retryAttempts` means that if the user has set 3 retries, It means it will initially try then retry 3 times
- In the other case, if the user has set 3 retries, it will try initially and then try again 2 times

sbuettner: I am not sure but `retryAttempts` sounds like the number of attempts that already happened. I would prefer a sealed interface with first of two options: Just a restart and one with a delay.
Lets tackle the number of retries later.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
```diff
@@ -34,34 +35,27 @@
...
 public class InboundExecutableRegistryImpl implements InboundExecutableRegistry {
 
+  private static final Logger LOG = LoggerFactory.getLogger(InboundExecutableRegistryImpl.class);
+  final Map<UUID, RegisteredExecutable> executables = new ConcurrentHashMap<>();
   private final BlockingQueue<InboundExecutableEvent> eventQueue;
   private final ExecutorService executorService;
-
   private final BatchExecutableProcessor batchExecutableProcessor;
-
+  private final CancellationManager cancellationManager;
```
johnBgood, line 51: We don't have any docs about the `cancel()` method. 
Do you think it would be valuable to update [this](https://docs.camunda.io/docs/components/connectors/custom-built-connectors/connector-sdk/#inbound-connector-runtime-logic) documentation to show how one can use `cancel()` and what's the behavior when using it?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/CancellationManager.java
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Handles exceptions thrown during the cancellation process of an {@link
+ * io.camunda.connector.api.inbound.InboundConnectorContext#cancel(Throwable) cancel} operation.
+ * This class determines the appropriate action to take based on the specific exception encountered.
+ */
+public class CancellationManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(CancellationManager.class);
+  private final Map<UUID, RegisteredExecutable> executables;
```
chillleader, line 39: I'm not a huge fan of having access to executables state from two different classes (`CancellationManager` and `InboundExecutableRegistryImpl`). This way it's less transparent where this state is changed.

One idea how we could tackle this:
- Instead of an external `CancellationManager`, we could introduce a proxy wrapper over the executable (e.g. `StatefulExecutable` or smth along those lines). I guess it can also implement the `InboundConnectorExecutable` interface.
- This proxy wrapper can manage the lifecycle of the executable inside of itself. It can store the real executable as a private field and create a cancellation callback; when the real executable calls the `cancel` method, a cancellation callback will trigger a method inside the proxy wrapper that will lead to the recreation of the executable.
- The proxy wrapper can also keep track of the state (thus we can probably get rid of `RegisteredExecutable`).
- The rest of the connector runtime doesn't know anything about retries and the `executables` Map in the `InboundExecutableRegistry` doesn't need to be modified anywhere outside of the registry class.

WDYT? I might have missed some important details, so happy to brainstorm together.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundIntermediateConnectorContextImpl.java
```diff
@@ -118,6 +119,9 @@ public void reportHealth(final Health health) {
     inboundContext.reportHealth(health);
   }
 
+  @Override
+  public void loadCancellationCallback(Consumer<Throwable> cancellationCallback) {}
```
johnBgood, line 123: is it expected to do nothing?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -155,6 +151,11 @@ public void reportHealth(Health health) {
     this.health = health;
   }
 
+  @Override
+  public void loadCancellationCallback(Consumer<Throwable> cancellationCallback) {
```
johnBgood, line 155: I would call it `set` not `load` as it's basically a setter

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorReportingContext.java
```diff
@@ -21,9 +21,12 @@
 import io.camunda.connector.api.inbound.InboundConnectorContext;
 import java.util.List;
 import java.util.Queue;
+import java.util.function.Consumer;
 
 public interface InboundConnectorReportingContext extends InboundConnectorContext {
 
+  void loadCancellationCallback(Consumer<Throwable> cancellationCallback);
```
johnBgood, line 28: I don't really get it. The consumer seems to always be the same (coming from `cancel` in the ActivatedExecutableManager). Did you introduce this to avoid a circular dependency?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+          () -> tryRestart(delay, retryAttempts - 1), delay.getSeconds(), TimeUnit.SECONDS);
+    }
+  }
+
+  private void handleCancellation(Throwable throwable) {
+    LOG.warn("Inbound connector executable has requested its cancellation");
+    try {
+      this.executable.deactivate();
+    } catch (Exception e) {
+      LOG.error("ojhsdfnjklndskn");
```
johnBgood, line 106: not sure about this log :D

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  public boolean isCancelled() {
+    return isCancelled;
+  }
+
+  public void cancel(Throwable throwable) {
+    switch (throwable) {
+      case ConnectorRetryException connectorRetryException -> {
+        handleCancellation(throwable);
```
johnBgood, line 63: this could be `executable.deactivate()`, not sure we need an extra method there

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+        handleRestart(connectorRetryException);
+      }
+      default -> handleCancellation(throwable);
+    }
+  }
+
+  private void handleRestart(ConnectorRetryException connectorRetryException) {
```
johnBgood, line 70: maybe could we rename it to `activateWithDelay` or similar.
I also realize that `ConnectorRetryException` doesn't allow us to provide `errorVariables` that are defined in the parent `ConnectorException`. That's on me, I missed it a few month ago. You can ignore it if you want to.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private void tryRestart(Duration delay, Integer retryAttempts) {
+    try {
+      this.executable = this.inboundConnectorFactory.getInstance(context.getDefinition().type());
+      this.executable().activate(context);
+      LOG.info(
+          "Activation successful for deduplication ID: {}",
+          context.getDefinition().deduplicationId());
+    } catch (Exception e) {
+      if (retryAttempts == 0) return;
```
johnBgood, line 91: maybe log that we used all the retries

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+      LOG.error(
+          "Activation failed for deduplication ID: {}. Retrying... {} retries left",
```
johnBgood, line 93: not sure if we could log executable information (id, etc ?)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+    LOG.warn("Inbound connector executable has requested its cancellation");
+    try {
+      this.executable.deactivate();
+    } catch (Exception e) {
+      LOG.error("ojhsdfnjklndskn");
+      throw new RuntimeException(e);
+    }
+  }
+
+  public void activate() {}
```
johnBgood, line 111: not used

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+    this.reactivationExecutor = reactivateScheduler;
+    this.isCancelled = true;
+    this.context.loadCancellationCallback(this::cancel);
+  }
+
+  public InboundConnectorReportingContext context() {
+    return context;
+  }
+
+  public boolean isCancelled() {
```
johnBgood, line 56: not used?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
```diff
@@ -49,14 +50,15 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(BatchExecutableProcessor.class);
-
-  @Value("${camunda.connector.inbound.log.size:10}")
-  private int inboundLogsSize;
-
   private final InboundConnectorFactory connectorFactory;
   private final InboundConnectorContextFactory connectorContextFactory;
   private final MetricsRecorder metricsRecorder;
   private final WebhookConnectorRegistry webhookConnectorRegistry;
+  private static final ScheduledExecutorService reactivator =
```
johnBgood, line 57: any reason to create the Executor here instead of in the ActivatedExecutableManager? 
If you want it here, we might want more threads.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundIntermediateConnectorContextImpl.java
mathias-vandaele: It is because InboundIntermediate did not have a cancellation mechanism, it wraps another context which is supposed to contain the cancel mechanism. We could speak about this one over a meeting as it is a bit complicated

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
mathias-vandaele: Correct

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ActivatedExecutableManager.java
```diff
@@ -0,0 +1,116 @@
...
+public class ActivatedExecutableManager {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ActivatedExecutableManager.class);
+  private final InboundConnectorFactory inboundConnectorFactory;
+  private final InboundConnectorReportingContext context;
+  private final ScheduledExecutorService reactivationExecutor;
+  boolean isCancelled;
+  private InboundConnectorExecutable<InboundConnectorContext> executable;
+
+  public ActivatedExecutableManager(
```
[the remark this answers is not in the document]
mathias-vandaele: No, you are right, something in fundamentally flawed here. The issue is that we have to deal with multiple things:

- We have to keep track of the status of an executable (cancelled or activated) in the `InboundExecutableRegistry` because it publishes the status of every connector, without this, we would not know the status
- cancel can only be called from the context, and to cancel properly; we need access to the executable
- The only place we could cancel was somewhere we have both access to the context AND the executable,

Those solution tries to answer this: 
- The creation of ActivatedExecutableManager that wraps the executable AND the context, being itself wrapped into registeredExecutable to be accessible from `InboundExecutableRegistry`
- This class keeps track of the status of the connector (isCancelled) and therefore is accessible from `InboundExecutableRegistry`
- We pass the Consumer to the context, so it can call it when context is called

The main complexity is that we have `InboundExecutableRegistryImpl` which contains a map, which contains an RegiteredExecutable containing the context and the executable, that both of them are needed to cancel, and that cancellation needs to have side effect known by the `InboundExecutableRegistryImpl`

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
```diff
@@ -218,4 +219,91 @@ public void deactivateBatch(List<RegisteredExecutable> executables) {
...
+      InboundConnectorExecutable<InboundConnectorContext> newExecutable =
+          connectorFactory.getInstance(cancelled.context().getDefinition().type());
+      this.reactivationScheduler.schedule(
+          () ->
+              tryRestart(
+                      newExecutable,
+                      cancelled.context(),
+                      retryException.getBackoffDuration(),
+                      retryException.getRetries())
+                  .thenAccept(future::complete)
```
johnBgood, line 240: can't we just return the future from `tryRestart`?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
```diff
@@ -92,32 +80,30 @@ public void publishEvent(InboundExecutableEvent event) {
...
+    if (cancelled.throwable() instanceof ConnectorRetryException retryException) {
+      this.batchExecutableProcessor
+          .restartFromContext(executable, retryException)
+          .thenAccept(
+              registeredExecutable ->
+                  this.executables.replace(cancelled.uuid(), registeredExecutable))
+          .exceptionally(
+              throwable -> {
+                LOG.error(
+                    "The inbound connector could not be restarted, {}", throwable.getMessage());
```
johnBgood, line 104: I would log the exception here, to avoid the issue we have with Kafka and other connectors (exception message hidden)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
```diff
@@ -218,4 +219,91 @@ public void deactivateBatch(List<RegisteredExecutable> executables) {
...
   }
+
+  public CompletableFuture<RegisteredExecutable> restartFromContext(
+      RegisteredExecutable registeredExecutable, ConnectorRetryException retryException) {
+    CompletableFuture<RegisteredExecutable> future = new CompletableFuture<>();
+    LOG.warn(
+        "Inbound connector executable has requested its reactivation in {} {}",
+        retryException.getBackoffDuration().getSeconds(),
+        TimeUnit.SECONDS);
+    if (registeredExecutable instanceof RegisteredExecutable.Cancelled cancelled) {
```
sbuettner, line 230: Why dont we call this method only for `RegisteredExecutable.Cancelled`?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
```diff
@@ -218,4 +219,91 @@ public void deactivateBatch(List<RegisteredExecutable> executables) {
...
+                    .exceptionally(
+                        throwable -> {
+                          future.completeExceptionally(throwable);
+                          return null;
+                        }),
+            delay.getSeconds(),
+            TimeUnit.SECONDS);
+      }
+    }
+    return future;
```
sbuettner, line 288: What about using https://failsafe.dev/#getting-started which we already use in other places: https://github.com/camunda/connectors/blob/725dc7e61167f16aa676edcb3de3d264ac5b8dd9/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaConnectorConsumer.java#L117

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
```diff
@@ -218,4 +219,91 @@ public void deactivateBatch(List<RegisteredExecutable> executables) {
...
+  }
+
+  public RegisteredExecutable cancelExecutable(
+      RegisteredExecutable registeredExecutable, Throwable throwable) {
```
sbuettner, line 292: We could just the `Activated` type here and avoid the typecast.

sbuettner:
Added some feedback. Would be great if you can add some tests.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/BatchExecutableProcessor.java
mathias-vandaele: Because Schedulers returns `ScheduledFuture` which inherit from `Future`. `Future` is really inconvenient to work with and I prefer to work with `CompletableFuture`

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
mathias-vandaele: Done

johnBgood (approved):
Great! 🚒

johnBgood (approved):

Issue #3190, opened 2024-08-28
Add more logging during process definition importing


chillleader:
__What should we do?__

During process definition importing, the Connector runtime doesn't log its progress. We only have logs where latest versions are listed for each process, but the subsequent step of fetching the BPMN XML and analyzing it for inbound connectors can take some time, especially when there are many process definitions in the cluster. This stage is not transparent and the runtime doesn't log its progress while going through the process XML.

__Why should we do it?__

To make it easier to understand what's happening if the runtime takes more time to load.

PR #3512, opened 2024-10-18
others(operate-importer): add more logs during the import of process definitions

mathias-vandaele:
Adding logs during the importation of process definitions

## Description

This PR add logs tracing what is happening during the importation of the process definition from Operate.

## Related issues

closes https://github.com/camunda/connectors/issues/3190

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImporter.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImporter.java
@@ -65,6 +65,7 @@ public class ProcessDefinitionImporter {
     if (definitions.isEmpty()) {
       return;
     }
+    LOG.debug("Handle of the imported process definitions starts...");
     try {
       meter(definitions.size());
       var result =
@@ -78,6 +79,7 @@ public class ProcessDefinitionImporter {
                           definition ->
                               new ProcessDefinitionVersion(
                                   definition.getKey(), definition.getVersion().intValue()))));
+      LOG.info("Updating the store with retrieved process definitions");
       stateStore.update(result);
     } catch (Exception e) {
       LOG.error("Failed to handle imported definitions", e);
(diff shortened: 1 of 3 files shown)


sbuettner:
@chillleader Can you please take a look whether these log statements would help? It wasnt clear to me based on the ticket whether trace/debug will help as they are usually not shown using the default log levels.

mathias-vandaele:
> @chillleader Can you please take a look whether these log statements would help? It wasnt clear to me based on the ticket whether trace/debug will help as they are usually not shown using the default log levels.

From my understanding, when a lot of process are present, this can take lots of time, as we use the operate client to retrieve all of them page by page, then sorting, then filtering out etc..
Those logs add informations about the step we are on, depending on the severity of it, It is in `trace`, `debug` or `info`

sbuettner:
@mathias-vandaele If we want to get more information aboute "where we are in the process" we would benefit from some kind of "progress" or number of pages we already scanned.

chillleader (approved):

chillleader (dismissed):
> @chillleader Can you please take a look whether these log statements would help? It wasnt clear to me based on the ticket whether trace/debug will help as they are usually not shown using the default log levels.

I think these changes will be helpful - although they are not shown by default due to logging level (only info and above are shown), it's quite easy to increase the logging level if we want to gain access to this info. I was originally looking for some progress indication, as well as a way to know which process IDs are being processed - I think this PR solves this well.

The only suggestion from my side: I'd rather keep all detailed log entries under the `DEBUG` or `INFO` level (depending on whether we want to show them by default) and not introduce the `TRACE` level to keep things a little less complicated

mathias-vandaele:
> > @chillleader Can you please take a look whether these log statements would help? It wasnt clear to me based on the ticket whether trace/debug will help as they are usually not shown using the default log levels.
> 
> I think these changes will be helpful - although they are not shown by default due to logging level (only info and above are shown), it's quite easy to increase the logging level if we want to gain access to this info. I was originally looking for some progress indication, as well as a way to know which process IDs are being processed - I think this PR solves this well.
> 
> The only suggestion from my side: I'd rather keep all detailed log entries under the `DEBUG` or `INFO` level (depending on whether we want to show them by default) and not introduce the `TRACE` level to keep things a little less complicated

I will change the trace into debug 👍

chillleader (approved):
Thank you!

PR #3570, opened 2024-11-01
fix(generator): boolean conditions in template generator annotations

chillleader:
## Description

This PR will add the missing property in the template generator annotations to create a condition that references a boolean property.

This possibility already exists on the lower level in the classes that are mapped to element template JSON, but it has been missing in the annotation.

An enum is used because we need to have a null/missing value, as the annotation can lead to different behavior of the generator based on which property is set.

## Checklist


DenovVasil:
@chillleader Tested it, works great. Thank you very much, this update helped me a lot 🙂

DenovVasil (approved):

sbuettner (approved):

PR #3629, opened 2024-11-08
feat(document): zeebe-client-based document store impl

chillleader:
## Description

This PR will replace the in-memory document store implementation of the connector runtime with a real integration that's based on the Zeebe client

## Checklist


sbuettner:
@chillleader I though about our interface model regarding `Document`.  What about separating it into `DocumentOperations` and `DocumentReference`? `Document` would implement both of them. We could then remove the `metadata()` shortcut as well.

johnBgood (approved):

sbuettner:
@chillleader Feel free to merge it and we do the refactoring I suggested in another PR if that simplifies things.

chillleader:
Thanks! I created a follow-up task for refactorings: https://github.com/camunda/team-connectors/issues/964

PR #3701, opened 2024-11-22
feat(webhook): Add tests and fix activation condition check data used

johnBgood:
## Description

We need use the same object used in the correlation step when checking the activation result. Otherwise it would lead to valid conditions being evaluated as false.

## Related issues

https://github.com/camunda/team-connectors/issues/788

## Checklist


connectors-e2e-test/connectors-e2e-test-base/src/main/java/io/camunda/connector/e2e/BpmnFile.java
```diff
@@ -65,6 +65,15 @@ public BpmnFile writeToFile(File file) {
...
     try {
+      String command =
+          "element-templates-cli --diagram "
+              + bpmnFile.getPath()
+              + " --template "
+              + template.getPath()
+              + " --element "
+              + elementId
+              + " --output "
+              + output.getPath();
```
sbuettner, line 76: What is this about? Curious.

johnBgood: ooops, removed

sbuettner (approved):

Issue #3221, opened 2024-09-03
Missing error messages, during request mapping


DenovVasil:
##Case 1 @AssertTrue(message=...)
When get request from OutboundConnectorContex like outboundConnectorContext.bindVariables(T)
We are expecting the message that was the message parameter for @AssertTrue(message=...) but we get something else
![image (1)](https://github.com/user-attachments/assets/d0240e3e-9f05-4a6d-9247-b6dd57577ab1)
![image (2)](https://github.com/user-attachments/assets/827323c1-17d0-4b10-a987-46c934985857)

##Case 2 @AssertTrue that throw Exception 
When get request from OutboundConnectorContex like outboundConnectorContext.bindVariables(T)
and assert method throw exception, like here
![image](https://github.com/user-attachments/assets/e5769588-06df-4de1-b9eb-fb358d729ff3)
our original message is lost due to being wrapped in another message
![image](https://github.com/user-attachments/assets/e1ec3094-2e1b-4c09-8499-62d35600a60d)

__Describe the solution you'd like__

__Describe alternatives you've considered__

__Additional context__

PR #3704, opened 2024-11-25
fix(runtime): error messages are now delivered correctly to operate when input validation fails

mathias-vandaele:
## Description

Add wrapped error message to the overall input validation error

## Related issues

https://github.com/camunda/connectors/issues/3221


sbuettner:
As discussed in the architecture session
* Analyze the output for different constraint violations to see whether we leak any sensitive information
* Capture the replaced secrets in the outbound context and re-use them to filter out any potential secret values that might be part of a validation message.

PR #3757, opened 2024-12-05
feat(webhook): Replace DocumentReference by Document

johnBgood:
## Description

- Replace DocumentReference with Document so that we have the right format that can be used by subsequent connectors;
- Add the right Jackson modules to the ObjectMapper used by the AbstractFeelDeserializer

## Related issues

https://github.com/camunda/team-connectors/issues/788

## Checklist

### DIFF
--- a/connector-sdk/jackson-datatype-feel/src/main/java/io/camunda/connector/feel/jackson/AbstractFeelDeserializer.java
+++ b/connector-sdk/jackson-datatype-feel/src/main/java/io/camunda/connector/feel/jackson/AbstractFeelDeserializer.java
@@ -22,10 +22,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.deser.ContextualDeserializer;
 import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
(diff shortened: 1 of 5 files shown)


connector-sdk/jackson-datatype-feel/src/main/java/io/camunda/connector/feel/jackson/AbstractFeelDeserializer.java
```diff
@@ -46,6 +49,9 @@ protected AbstractFeelDeserializer(FeelEngineWrapper feelEngineWrapper, boolean
     super(String.class);
     this.feelEngineWrapper = feelEngineWrapper;
     this.relaxed = relaxed;
+    BLANK_OBJECT_MAPPER
+        .registerModule(new Jdk8Module())
+        .registerModule(new JacksonModuleDocument(new DocumentFactoryImpl(null), null));
```
sbuettner, line 54: Curious, as this is the deserializer. Why would it need to be document aware?

johnBgood: It's due to the `WebhookHttpResponse`that we're creating in `InboundWebhookRestController#buildSuccessfulResponse`

Issue #3795, opened 2024-12-13
SMTP outbound connector - html and body content is null 


itsnuyen:
__Describe the Bug__
<img width="350" alt="image" src="https://github.com/user-attachments/assets/36531bdc-0c84-419a-a679-4afbdf4770f9" />
<img width="500" alt="image" src="https://github.com/user-attachments/assets/11beb9a3-f602-4ad1-8d57-5d9bfee6a514" />

Because FEEL will take a `string` and a `data` and try to sum it, it will produce a `null` value. 
This null value will produce a nullpointer as an error but the process is still running and after the defined retries it will break.
If you have multiple process running, this behavior will block everything else  

__Steps to Reproduce__
1. Create a bpmn with a the outbound mail connector and use the SMTP feature
2. inside the mail content use following feel description
`"<h2>Test Test</h2>" + "hallo welt" + today()`
3. deploy the BPMN and start a new process

__Expected Behavior__

When having a faulty body content, the connector should fail instantly and create a incident.

__Environment__

- OS: OSX Sequoia 15.2 as-well as kubernetes environment 
- Camunda version: 8.6.5 (self-managed)

PR #3796, opened 2024-12-13
fix(mail-connector): check htmlBody and/or body is null and throw an error

itsnuyen:
## Description
Because FEEL will take a string and a data and try to sum it, it will produce a null value.
This null value will produce a nullpointer as an error but the process is still running and after the defined retries it will break.
If you have multiple process running, this behavior will block everything else

So check if the content of the body is null and throw a null pointer

## Related issues

closes #https://github.com/camunda/connectors/issues/3795

## Checklist


connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/outbound/JakartaEmailActionExecutor.java
```diff
@@ -301,16 +301,25 @@ private Multipart getMultipart(SmtpSendEmail smtpSendEmail) throws MessagingExce
...
+          throw new MessagingException("HTML Body is null");
+        }
         MimeBodyPart htmlPart = new MimeBodyPart();
         htmlPart.setContent(smtpSendEmail.htmlBody(), JakartaUtils.HTML_CHARSET);
         multipart.addBodyPart(htmlPart);
       }
       case MULTIPART -> {
+        if (smtpSendEmail.htmlBody() == null || smtpSendEmail.body() == null) {
+          throw new MessagingException("Text or HTML Body is null");
+        }
```
sbuettner, line 322: Hey @itsnuyen, instead of throwing a `MessagingException`, could you implement a custom validation method in the model that verifies that these values are not null?

itsnuyen: Hi @sbuettner, 
I have one questions. Do you mean adapting inside the `SmtpSendEmail` record the validation like this?
```java
switch (contentType) {
  case PLAIN:
    if (body == null) {
      throw new IllegalArgumentException("Email body is null");
    }
  case HTML:
    if (htmlBody == null) {
      throw new IllegalArgumentException("Email html body is null");
    }
  case MULTIPART:
    if (body == null || htmlBody == null) {
      throw new IllegalArgumentException("Email body or html body cannot be null");
    }
}
```
This could also running but as a error message it could be confusing for the end user:
image
As it state that an exception is raised and it is a JSON_PROCESSING_ERROR

Do you have an Idea how I could provide the end user proper feedback? 
Or do I even misunderstand your request?

sbuettner: Hey @itsnuyen I was thinking about something like this as part of the `SmtpSendEmail` record:
```
@AssertTrue(message = "Please provide a proper message body")
        private boolean isValidMessageBody() {
                return body != null || htmlBody != null;
        }
```

sbuettner:
Great work @itsnuyen, could you add a simple test that covers this new validation?

sbuettner (dismissed):

sbuettner (commented):
LGTM,
looks like the test is missing the right license information:
> Warning:  Missing header in: /home/runner/work/connectors/connectors/connectors/email/src/test/java/io/camunda/connector/email/outbound/protocols/actions/SmtpSendEmailTest.java

You can use the `license` plugin to make sure its added to all files.

mathias-vandaele:
LGTM, Great work 😄

mathias-vandaele (approved):

sbuettner:
Backport failed for `release/8.6`, because it was unable to cherry-pick the commit(s).

Please cherry-pick the changes locally and resolve any conflicts.
```bash
git fetch origin release/8.6
git worktree add -d .worktree/backport-3796-to-release/8.6 origin/release/8.6
cd .worktree/backport-3796-to-release/8.6
git switch --create backport-3796-to-release/8.6
git cherry-pick -x 657810f4bb25bf2da74bd19c0e7906a95aa934a3
```

PR #3822, opened 2024-12-19
fix(connector-runtime): add exclusion of spring-core creating dependencies incompatibility

mathias-vandaele:
## Description

Done during hacking session to make developper experience better while creating new connectors

## Related issues

closes #

## Checklist

### DIFF
--- a/connector-runtime/spring-boot-starter-camunda-connectors/pom.xml
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/pom.xml
@@ -36,6 +36,10 @@
       <version>${version.camunda}</version>
     </dependency>
     <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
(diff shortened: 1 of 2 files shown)


sbuettner:
@mathias-vandaele Great that you found something that improves the devx. Could you provide some information why it was done to get a better understanding?

mathias-vandaele:
> @mathias-vandaele Great that you found something that improves the devx. Could you provide some information why it was done to get a better understanding?

I am trying to provide a good experience for the dev, with the outbound template. there were an incompatibility within the camunda-sdk between spring-web and spring-core, the reason why it works in our repo is unknown as it seems that it brings `spring-core` `6.2` but when importing it from the template it brings `spring-boot` `6.1.14`.

We decided to exclude it from the `camunda-spring-sdk` so now the pom on the template side is very clean and does not requires any tricks to work

The PR is here btw https://github.com/camunda/connector-template-outbound/pull/149

sbuettner:
Thanks, please check via `dependency:tree` where and how the dependencies are coming from as excluding might also lead to surprises in the future. If there are issues we might need to raise them with the camunda sdk team.

sbuettner (approved):

connector-runtime/spring-boot-starter-camunda-connectors/pom.xml
```diff
@@ -27,14 +27,18 @@
     </dependency>
     <dependency>
       <groupId>io.camunda</groupId>
-      <artifactId>spring-boot-starter-camunda-sdk</artifactId>
+      <artifactId>zeebe-client-java</artifactId>
       <version>${version.zeebe}</version>
     </dependency>
```
sbuettner, line 32: We might actually be able to remove this here as its already defined in the runtime-core.

sbuettner (dismissed):

ztefanie (approved):

PR #3837, opened 2025-01-09
fix(rest): fix sending null params in a body map

johnBgood:
## Description

- Java's  `Collectors.toMap(` does not accept null value, hence producing a NPE when sending this kind of body:

```json
{
"myKey": null
}
```

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/utils/DocumentHelper.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/utils/DocumentHelper.java
@@ -19,6 +19,7 @@ package io.camunda.connector.http.base.utils;
 import io.camunda.document.CamundaDocument;
 import java.util.AbstractMap;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


sbuettner (approved):

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/utils/DocumentHelper.java
```diff
@@ -35,16 +36,20 @@ public Object parseDocumentsInBody(Object input, Function<CamundaDocument, Objec
     return switch (input) {
       case Map<?, ?> map ->
           map.entrySet().stream()
+              .filter(e -> e.getValue() != null)
```
sbuettner, line 39: Wouldnt this lead to non-inclusion of `null` fields?
Asking as we have customers that require to see null in their values.

johnBgood: Good point, `null` values will be there as we use the initial map as the initializer here:
` () -> new HashMap<>(map)));`

sbuettner:
Successfully created backport PR for `release/8.7`:
- #3839

PR #3840, opened 2025-01-10
fix(rest connector): add skip decoding field for rest outbound connector

### LINKED ISSUE #3694: REST Connector automatic URL decoding removes already encoded elements
__Describe the Bug__

We received a support request from a client: https://jira.camunda.com/browse/SUPPORT-24612?focusedCommentId=382201&atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258Y29tbWVudA%3D%3D

The REST Connector by default decodes incoming URLs to properly encoded them again. This removed already encoded elements of the incoming URL.

Example: "http://your.test.server/test%2Ftest" --> "http://your.test.server/test/test"

The server requires the last path segment to be encoded as "test%2Ftest" is the name of the resource.

One example where this behavior leads to issues is in the GitLab Connector as GitLab requires users to encode slashes in URLs when they are part of branch names: https://forum.gitlab.com/t/how-to-get-with-api-a-branch-name-containing-slashes/59347/5

__Steps to Reproduce__

1. Use an URL that contains encoded path segments: http://your.test.server/test%2Ftest
2. Pass it to the REST Connector as a process variable or a static string
3. The URL gets decoded and encoded again removing the previously encoded values
4. The final URL contains the decoded segments which lead to wring values as the path is now changed: "http://your.test.server/test/test instead of "http://your.test.server/test%2Ftest

One case this affects our users is the GitLab Connector. 

__Expected Behavior__

Already encoded elements of the URL should not be decoded and sent as is to the upstream server.

__Solutions__

- We could provided an optional property in the request mode to deactivate the automatic URL decoding. This flag disables the decoding on our side and would keep the URL as is. The flag could be marked as hidden defaulting to true and can optionally be disabled for example in the GitLab Connector for calls that support branch names.

- Update the Gitlab Connector accordingly to avoid URL encoding there.

An alternative would be to refactor our URL decoding in the HTTP base connector by keeping elements that dont require re-encoding as they are like it was done in the previous implementation in the Google HTTP Client library:  https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java

__Environment__

- Camunda Connectors 8.6

ztefanie:
## Description
- Added a skipEnding field to HttpCommonRequest
- Adjusted gitlab connector json to always set skipEncoding to true
- All github connectors now skip encoding
- in XML skip encoding can be set for rest connectors, defaults to false if not set

## Info
- Added a hint to the docs about this option, see PR here: https://github.com/camunda/camunda-docs/pull/4837
- not tested with the web modeller
- I adjusted the gitlab connector by editing the JSON directly

## Questions
- should I add tests for the UrlEncoder?
- is it possible to add a test for the gitlab connector?

## Related issues

closes #3694 

## Checklist

- [x] PR has a **milestone** or the `no milestone` label. (not sure which milestone i should set it too :thinking: )

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/UrlEncoder.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/UrlEncoder.java
@@ -28,11 +28,14 @@ import org.slf4j.LoggerFactory;
 public class UrlEncoder {
   private static final Logger LOG = LoggerFactory.getLogger(ApacheRequestUriBuilder.class);
 
-  public static URI toEncodedUri(String requestUrl) {
+  public static URI toEncodedUri(String requestUrl, Boolean skipEncoding) {
     try {
       // We try to decode the URL first, because it might be encoded already
       // which would lead to double encoding. Decoding is safe here, because it does nothing if
       // the URL is not encoded.
+      if (skipEncoding) {
+        return URI.create(requestUrl);
+      }
       var decodedUrl = URLDecoder.decode(requestUrl, StandardCharsets.UTF_8);
       var url = new URL(decodedUrl);
       // Only this URI constructor escapes the URL properly
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
@@ -111,6 +111,15 @@ public class HttpCommonRequest {
       description = "Store the response as a document in the document store")
   private boolean storeResponse;
 
+  @TemplateProperty(
+      label = "Skip URL encoding",
+      description = "Skip the default URL decoding and encoding behavior",
+      type = TemplateProperty.PropertyType.Hidden,
+      feel = Property.FeelMode.disabled,
+      group = "endpoint",
+      optional = true)
+  private String skipEncoding;
+
   public Object getBody() {
     return body;
   }
@@ -137,6 +146,14 @@ public class HttpCommonRequest {
 
   public void setQueryParameters(Map<String, String> queryParameters) {
     this.queryParameters = queryParameters;
+  }
+
+  public boolean getSkipEncoding() {
+    return Objects.equals(skipEncoding, "true");
+  }
+
+  public void setSkipEncoding(final String skipEncoding) {
+    this.skipEncoding = skipEncoding;
   }
 
   public boolean hasAuthentication() {
--- a/connectors/http/http-base/src/test/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClientTest.java
+++ b/connectors/http/http-base/src/test/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClientTest.java
@@ -320,6 +320,23 @@ public class CustomApacheHttpClientTest {
       assertThat(result).isNotNull();
       assertThat(result.status()).isEqualTo(200);
     }
+
+    @ParameterizedTest
+    @EnumSource(HttpMethod.class)
+    public void shouldKeepOriginalEscaping_whenSkipEscapingIsSet(
+        HttpMethod method, WireMockRuntimeInfo wmRuntimeInfo) {
+      stubFor(any(urlEqualTo("/path%2Fwith%2Fencoding")).willReturn(ok()));
+      stubFor(any(urlEqualTo("/path/with/encoding")).willReturn(badRequest()));
+      HttpCommonRequest request = new HttpCommonRequest();
+      request.setMethod(method);
+      request.setUrl(wmRuntimeInfo.getHttpBaseUrl() + "/path%2Fwith%2Fencoding");
+      request.setSkipEncoding("true");
+
+      HttpCommonResult result = customApacheHttpClient.execute(request);
+
+      assertThat(result).isNotNull();
+      assertThat(result.status()).isEqualTo(200);
+    }
   }
 
   @Nested
--- a/connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
+++ b/connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
@@ -7,7 +7,7 @@
     "keywords" : [ ]
   },
   "documentationRef" : "https://docs.camunda.io/docs/components/connectors/protocol/rest/",
-  "version" : 9,
+  "version" : 10,
   "category" : {
     "id" : "connectors",
     "name" : "Connectors"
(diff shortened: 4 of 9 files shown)


connectors/http/http-base/src/test/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClientTest.java
```diff
@@ -321,6 +321,23 @@
       assertThat(result).isNotNull();
       assertThat(result.status()).isEqualTo(200);
     }
+
+    @ParameterizedTest
+    @EnumSource(HttpMethod.class)
+    public void shouldKeepOriginalEscaping_whenSkipEscapingIsSet(
```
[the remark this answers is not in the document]
ztefanie: resolved

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -111,6 +111,13 @@ public class HttpCommonRequest {
       description = "Store the response as a document in the document store")
   private boolean storeResponse;
 
+  @TemplateProperty(
+      type = TemplateProperty.PropertyType.Hidden,
+      feel = Property.FeelMode.disabled,
+      group = "endpoint",
+      optional = true)
+  private String skipEncoding;
```
johnBgood, line 119: this should be a `boolean`, you can see an example above with `storeResponse` :)

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/UrlEncoder.java
```diff
@@ -28,11 +28,14 @@
 public class UrlEncoder {
   private static final Logger LOG = LoggerFactory.getLogger(ApacheRequestUriBuilder.class);
 
-  public static URI toEncodedUri(String requestUrl) {
+  public static URI toEncodedUri(String requestUrl, Boolean skipEncoding) {
     try {
       // We try to decode the URL first, because it might be encoded already
       // which would lead to double encoding. Decoding is safe here, because it does nothing if
       // the URL is not encoded.
+      if (skipEncoding) {
```
johnBgood, line 36: care if you use a `Boolean` it might lead to NPE I guess. It should be `boolean`

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -139,6 +146,14 @@ public void setQueryParameters(Map<String, String> queryParameters) {
     this.queryParameters = queryParameters;
   }
 
+  public boolean getSkipEncoding() {
+    return Objects.equals(skipEncoding, "true");
```
johnBgood, line 150: we don't need that anymore if it's a boolean

connectors/http/http-base/src/test/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClientTest.java
```diff
@@ -321,6 +321,23 @@ public void shouldReturn200_whenEscapedSpaceInPathAndQueryParametersInPath(
...
+        HttpMethod method, WireMockRuntimeInfo wmRuntimeInfo) {
+      stubFor(any(urlEqualTo("/path%2Fwith%2Fencoding")).willReturn(ok()));
+      stubFor(any(urlEqualTo("/path/with/encoding")).willReturn(badRequest()));
+      HttpCommonRequest request = new HttpCommonRequest();
+      request.setMethod(method);
+      request.setUrl(wmRuntimeInfo.getHttpBaseUrl() + "/path%2Fwith%2Fencoding");
+      request.setSkipEncoding("true");
+
+      HttpCommonResult result = customApacheHttpClient.execute(request);
```
johnBgood, line 336: nice that we have a test for that 👍

connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
```diff
@@ -407,6 +407,16 @@
       "type" : "zeebe:input"
     },
     "type" : "Boolean"
+  }, {
```
johnBgood, line 410: just to be sure: did you run the maven command or did you add these lines manually?

johnBgood (commented):
Great first one! Left a few comments 💯

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -111,6 +111,13 @@ public class HttpCommonRequest {
       description = "Store the response as a document in the document store")
   private boolean storeResponse;
 
+  @TemplateProperty(
```
sbuettner, line 114: Would be great to add a label & description to allow users to discover what this new property is all about.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
ztefanie: You are right, a boolean would make more sense. But I got an error. Apparently hidden properties must be String.

connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
ztefanie: This is added by the maven command. The changes in the gitlab connector json are done manually.

johnBgood: Maybe should we update the version in HttpJsonFunction ?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -111,6 +111,15 @@ public class HttpCommonRequest {
...
+      label = "Skip encoding",
```
sbuettner, line 115:
```suggestion
      label = "Skip URL encoding",
```

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -111,6 +111,15 @@ public class HttpCommonRequest {
...
+      description = "Skip the default decoding and encoding behavior",
```
sbuettner, line 116:
```suggestion
      description = "Skip the default URL decoding and encoding behavior",
```

connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
ztefanie: I updated the version and ran maven generate templates command, but now the changes to the ".json" files are removed? Is it correct to run the command in the `http/rest/` subfolder? I am confused, why they were added first and now they are removed :thinking:

sbuettner (approved):
Congrats to your first PR! Great work! 🚀

connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
ztefanie: @johnBgood Should i merge this now, although the .json files are not changed or how should I proceed? I am confused why the changes are no longer generated by the mvn command.

johnBgood: let me check locally

PR #3841, opened 2025-01-10
fix(kafka-inbound): log the exception when kafka throw on start

mathias-vandaele:
## Description

Log the exception when Kafka consumer cannot be started 

## Related issues

closes #https://github.com/camunda/team-connectors/issues/953

## Checklist

### DIFF
--- a/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaConnectorConsumer.java
+++ b/connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaConnectorConsumer.java
@@ -100,7 +100,7 @@ public class KafkaConnectorConsumer {
             consume();
             return null;
           } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


connectors/kafka/src/main/java/io/camunda/connector/kafka/inbound/KafkaConnectorConsumer.java
```diff
@@ -100,7 +100,7 @@ public void startConsumer() {
             consume();
             return null;
           } catch (Exception ex) {
-            LOG.error("Consumer loop failure, retry pending: {}", ex.getMessage());
+            LOG.error("Consumer loop failure, retry pending: {}", ex.getMessage(), ex);
```
johnBgood, line 103: can we also log it here:

 LOG.error("Failed to initialize connector: {}", ex.getMessage());

johnBgood (approved):

PR #3844, opened 2025-01-13
feat(runtime): fail without retries on ConnectorInputException

### LINKED ISSUE #3806: `ConnectorInputException` should lead to a failed job without retries (technical error)
__What should we do?__

`ConnectorInputException` should be handled in the `ConnectorJobHandler` in a way that they lead to a failed job without any retries (technical error).

__Why should we do it?__

This would allow Connector developers to raise errors without retrying them.

ztefanie:
## Description
Set retries to zero, when ConnectorInputException is thrown.

## Related issues

closes #3806 

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
@@ -25,6 +25,7 @@ import io.camunda.client.api.response.FailJobResponse;
 import io.camunda.client.api.worker.JobClient;
 import io.camunda.client.api.worker.JobHandler;
 import io.camunda.connector.api.error.ConnectorException;
+import io.camunda.connector.api.error.ConnectorInputException;
 import io.camunda.connector.api.error.ConnectorRetryException;
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 import io.camunda.connector.api.secret.SecretProvider;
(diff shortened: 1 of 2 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -212,10 +213,15 @@ public void handle(final JobClient client, final ActivatedJob job) {
       LOGGER.debug(
           "Exception while processing job: {} for tenant: {}", job.getKey(), job.getTenantId(), ex);
       String errorCode = null;
+      int retries = job.getRetries() - 1;
       if (ex instanceof ConnectorException connectorException) {
         errorCode = connectorException.getErrorCode();
+        Throwable cause = connectorException.getCause();
+        if (cause instanceof ConnectorInputException) {
```
johnBgood, line 220: maybe check if  `ex instanceof ConnectorInputException` because currently your test might if I replace jobHandler by :

```java
     var jobHandler =
          newConnectorJobHandler(
              context -> {
                throw new ConnectorInputException(
                    "expected Connector Input Exception", new RuntimeException("cause"));
              });
```

ztefanie: Good point :+1: Thanks for mentioning and explaining. I adjusted it and added it to the tests.

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
```diff
@@ -506,6 +506,34 @@ void shouldTruncateBpmnErrorMessage() {
       assertThat(result.getErrorMessage().length())
           .isLessThanOrEqualTo(ConnectorJobHandler.MAX_ERROR_MESSAGE_LENGTH);
     }
+
+    @ParameterizedTest
+    @MethodSource("provideInputExceptions")
+    void shouldNotRetry_OnConnectorInputException(Exception exception) {
```
johnBgood, line 512: I think this test is great, but I would have kept the previous one as it needs to work with both :)

johnBgood (approved):
Just 1 small comment to add the old test as well, other than that all good 💯

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
ztefanie: It is a parametrized test with both options ( the new and the previous)

johnBgood: oh god I read too fast, sorry about that :'(

PR #3848, opened 2025-01-14
fix(soap): remove logging of soap request and response

mathias-vandaele:
## Description

Remove logs that could be showing sensitive data (request, response)

## Related issues

closes #https://github.com/camunda/team-connectors/issues/965

## Checklist

### DIFF
--- a/connectors/soap/src/main/java/io/camunda/connectors/soap/SoapConnector.java
+++ b/connectors/soap/src/main/java/io/camunda/connectors/soap/SoapConnector.java
@@ -91,7 +91,7 @@ public class SoapConnector implements OutboundConnectorFunction {
               .withPreserveNamespaces(false)
               .build()
               .toJson(XmlUtilities.xmlStringToDocument(soapResponseMessage));
-      LOG.debug("Response to connector runtime: \n{}", response.toPrettyString());
+      LOG.trace("Response to connector runtime: \n{}", response.toPrettyString());
       return response;
     } catch (WebServiceException e) {
       if (e instanceof SoapFaultClientException soapFaultClientException) {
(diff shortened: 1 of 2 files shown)


sbuettner (dismissed):

connectors/soap/src/main/java/io/camunda/connectors/soap/client/SpringSoapClient.java
```diff
@@ -138,7 +138,6 @@ private ClientInterceptor[] buildInterceptors(
     List<ClientInterceptor> interceptorList = new ArrayList<>();
     handleAuthentication(authentication).ifPresent(interceptorList::add);
     handleSoapHeader(soapHeader, namespaces).ifPresent(interceptorList::add);
-    interceptorList.add(new LoggingInterceptor());
```
jonathanlukas, line 141: If I understand correctly, the customer does not want sensitive information to be logged.

However, the logging interceptor only uses the log level DEBUG, so the actual problem is on the customer side. I would vote for keeping this logging so that the debugging of connector issues is possible with the log.

jonathanlukas (approved):
lgtm

connectors/soap/src/main/java/io/camunda/connectors/soap/client/SpringSoapClient.java
johnBgood: Log/store PII is also forbidden (GDPR etc.), which might happen if we keep this line. Regardless of the log level, we just can't store this information. 
You might be right, but I just wanted confirmation here.

johnBgood: > All logging is TRACE
There is a parameter that you have to set so that the logging will happen.

Yea, I get this part :D But we can still log PII (using TRACE), which is from my knowledge simply forbidden (trace or info is not relevant). 

I might be wrong though, so feel free to ignore my comment.

jonathanlukas: Well, apache http logs everything when being set to log level debug, so it is not forbidden :) 

From my experience, the operations team has to ensure that they run compliant.

sbuettner (approved):

PR #3851, opened 2025-01-15
feat(jackson): enable 'accept single value as array' globally

### LINKED ISSUE #3849: [JSON Databinding] Support binding single objects to collections

Users currently have to use the FEEL array notation `=[myValue]` even if they only want to provide a single object which is in this case `myValue`. To simplify the usage of Connectors we should allow users to use a FEEL expression with a single object `=myValue` by mapping single objects to collections with a single element in our binding process on the runtime side.

We should enhance our JSON databinding in a generic way to allowing the binding of single objects to types of `Collection<?>` in our Jackson JSON mapping layer. This would enable this feature for all types of collections.

We could expect users to always use the array syntax even if they will only ever use a single entry for a collection.

__Additional context__

This discussion came up during a QA session as some of the document handling enabled Connectors support a document collection but users mainly want to provide a single document as an input and therefore might forget to add the brackets.

chillleader:
## Description

This PR will enable the "accept single value as array" feature in Jackon globally.

## Related issues

closes #3849 

## Checklist

### DIFF
--- a/connector-sdk/core/src/main/java/io/camunda/connector/api/json/ConnectorsObjectMapperSupplier.java
+++ b/connector-sdk/core/src/main/java/io/camunda/connector/api/json/ConnectorsObjectMapperSupplier.java
@@ -44,6 +44,7 @@ public class ConnectorsObjectMapperSupplier {
           .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
           .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
           .disable(SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS)
+          .enable(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
           .build();
 
   private ConnectorsObjectMapperSupplier() {}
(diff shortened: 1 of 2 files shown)


connector-sdk/core/src/main/java/io/camunda/connector/api/json/ConnectorsObjectMapperSupplier.java
```diff
@@ -44,6 +44,7 @@ public class ConnectorsObjectMapperSupplier {
           .enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS)
           .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
           .disable(SerializationFeature.WRITE_DURATIONS_AS_TIMESTAMPS)
+          .enable(DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY)
```
johnBgood, line 47: Do we need some tests with primitive types(Integer, ...), and custom ones (like Document)?

chillleader: Good point :) Added

johnBgood (dismissed):
Tests are failing, but all good for me :)

johnBgood (approved):

sbuettner (approved):
🚀

Issue #3886, opened 2025-01-29
[Congen] The generated template is not compatible with the web-modeler schema validation


markfarkas-camunda:
__Describe the Bug__

Generating element templates from the web-modeler UI fails (`/create-new-element-template`). Reason is the template validation.

__Steps to Reproduce__

1. Use this specification to generate a connector template: `https://raw.githubusercontent.com/revolutdev/business-api-spec/refs/heads/master/accounts.yaml`
2. Observe that the generated template is not compatible with the `zeebe-element-templates-schema.json` used in `web-modeler`

`zeebe-element-templates-schema.json`:
[zeebe-element-templates-schema.json](https://github.com/user-attachments/files/18586498/zeebe-element-templates-schema.json)

__Expected Behavior__

The generated template should have a compatible json structure.

PR #3888, opened 2025-01-29
fix(template-generator): use server url if description is not found

markfarkas-camunda:
## Description

Modified the template generator to use the server url as label if the description is null. This way we can avoid having choices without names, which would not be compatible with the template schema. 

## Related issues

closes https://github.com/camunda/connectors/issues/3886

## Checklist


johnBgood (approved):
💯

PR #3894, opened 2025-01-31
feat(rest-connector): Add support for proxy auth

ztefanie:
## Description
Added support for setting proxy via env var and to add proxy auth.

Proxy config can be set via system properties or via env vars. Setting via env vars is currently handled by using an RFC3986 string, e.g. http://my-user:demo@localhost:3128/
This means it is currently not possible to set nonProxyHosts via env vars and username and passwords must be url safe.
I think we could also change this, with some effort, to setting multiple env vars. I decided to go for this version, assuming setting via env vars is more for testing / fast setups, but not complex production setups, as setting passwords in ENV vars is worse than setting in system properties. If you disagree please let me know, then i will adjust it.

I tested different use cases manually, with the following results:
image

I tried to add Unit tests for nonProxyHosts but couldn't get it working as mocking only one host / localhost is possible with Wiremock. If you think nonProxyHost unit test should be added, please give me a hint how to achieve this.

Not tested for https.

I will add this to the docs, once this is approved or at least we agreed on how to set it with env vars.

## Related issues

closes https://github.com/camunda/team-connectors/issues/989

## Checklist


connectors/http/http-base/src/test/resources/squid.conf
```diff
@@ -6,6 +6,19 @@ acl Safe_ports port 80
...
 acl CONNECT method CONNECT
 
-# Allow all outgoing HTTP/HTTPS traffic
-http_access allow all
+# Define the authentication method
+auth_param basic program /usr/lib/squid/basic_ncsa_auth /etc/squid/passwords
+auth_param basic realm Squid Proxy
 
+# Allow unauthenticated users on /path
+acl no_auth_url url_regex -i path
```
ztefanie, line 14: so with this setting every path is needs username and password except "/path". I would rather have only "/protected" be protected and everything else allowed, but I am not sure if this is possible. So if you have an idea let me know.

sbuettner:
@ztefanie We currently only document environment variables for our component configuration and we should therefore continue to support using them also for this use-case: https://docs.camunda.io/docs/next/self-managed/connectors-deployment/connectors-configuration/

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,121 @@
...
+                Integer.parseInt(System.getProperty(protocol + ".proxyPort"))),
+            new UsernamePasswordCredentials(
+                System.getProperty(protocol + ".proxyUser"),
+                System.getProperty(protocol + ".proxyPassword").toCharArray()));
+      } else if (systemEnvIsSetForProtocol(protocol)) {
+        CustomProxy proxyData = getProxySettingsFromEnvVar(protocol);
+        provider.setCredentials(
+            new AuthScope(proxyData.host, proxyData.port),
+            new UsernamePasswordCredentials(proxyData.username, proxyData.password.toCharArray()));
+      }
```
sbuettner, line 38: Would it be possible to have a dedicated `ProxyConfig` data class and a single place (a factory) that is responsible for creating it? 

It would be easier to follow and having a single place where data is read either from System.properties and EnvVars would make it easier to debug.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+  private static boolean doesTargetMatchNonProxy(String protocol, String requestUri) {
+    String nonProxyHostsEnvVar =
+        System.getenv("CONNECTOR_" + protocol.toUpperCase() + "_PROXY_NON_PROXY_HOSTS");
+    return (nonProxyHostsEnvVar != null
+        && Arrays.stream(nonProxyHostsEnvVar.split("\\|"))
+            .anyMatch(
+                nonProxyHost ->
+                    requestUri.matches(nonProxyHost.replace(".", "\\.").replace("*", ".*"))));
+  }
```
sbuettner, line 74: This check should be done against a proxy config without requiring to fetch the data from the environment. Possibly using a dedicated `ProxyConfig` as mentioned above.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
sbuettner: I just saw that we have `CustomProxy` available. Why not use it here as well:
```
        provider.setCredentials(
            new AuthScope(
                System.getProperty(protocol + ".proxyHost"),
                Integer.parseInt(System.getProperty(protocol + ".proxyPort"))),
            new UsernamePasswordCredentials(
                System.getProperty(protocol + ".proxyUser"),
                System.getProperty(protocol + ".proxyPassword").toCharArray()));
```

ztefanie: You are right, that would make it easier to read. I will adjust it.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
```diff
@@ -97,14 +99,22 @@ private static PoolingHttpClientConnectionManager createConnectionManager() {
   @Override
   public HttpCommonResult execute(
       HttpCommonRequest request, @Nullable ExecutionEnvironment executionEnvironment) {
-    var apacheRequest = ApacheRequestFactory.get().createHttpRequest(request);
     try {
+      var apacheRequest = ApacheRequestFactory.get().createHttpRequest(request);
+      ProxyHandler proxyHandler =
+          new ProxyHandler(apacheRequest.getScheme(), apacheRequest.getUri().getHost());
```
sbuettner, line 105: Looks like we are creating a new proxy handler for every new request. I think this should be done once.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,118 @@
...
+
+  public CredentialsProvider getCredentialsProvider(String uncheckedProtocol) {
+    BasicCredentialsProvider provider = new BasicCredentialsProvider();
+    if (StringUtils.isNotBlank(host)
+        && StringUtils.isNotBlank(user)
+        && ArrayUtils.isNotEmpty(password)) {
+      provider.setCredentials(
+          new AuthScope(host, port), new UsernamePasswordCredentials(user, password));
+    }
+    return provider;
```
sbuettner, line 76: What happens if the password or another entry is empty? Should we log a warning or something?

Otherwise this will silently lead to a credentials provider without credentials.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  public HttpRoutePlanner getRoutePlanner(String uncheckedProtocol, HttpHost proxyHost) {
+    if (sourceIsSystemProperties) {
+      return new SystemDefaultRoutePlanner(
+          DefaultSchemePortResolver.INSTANCE, ProxySelector.getDefault());
+    } else if (proxyHost != null) {
+      return new DefaultProxyRoutePlanner(proxyHost);
+    }
+    return null;
```
sbuettner, line 100: Would be great to add some debug logs here. Otherwise it might be intransparent which config was applied.

ztefanie: Added debug logging.

connectors/http/http-base/src/test/resources/squid.conf
johnBgood:
```
acl protected_path url_regex -i ^/protected

auth_param basic program /usr/lib/squid/basic_ncsa_auth /etc/squid/passwords
acl authenticated_user proxy_auth REQUIRED

http_access deny protected_path !authenticated

http_access allow all
```

Not sure but you might try this approach

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,135 @@
...
+  record ProxyDetails(
+      String protocol,
+      String host,
+      int port,
+      String user,
+      char[] password,
+      String nonProxyHosts,
+      boolean sourceIsSystemProperties) {}
+
+  private Map<String, ProxyDetails> map = new HashMap<>();
```
johnBgood, line 33: `private final Map<String, ProxyDetails> map = new HashMap<>();`

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,135 @@
...
+  private static final Logger LOGGER = LoggerFactory.getLogger(ProxyHandler.class);
+
+  public ProxyHandler() {
+    try {
+      setFromEnvVars("http");
```
johnBgood, line 38: I think I would create a global variable `PROTOCOLS = http, https`

and here something like:
```java
Map<String, ProxyDetails> config = getConfigFromSystemProperties().orElse( () -> getConfigFromEnvVars());

// Set the 2 different CredentialsProvider as class members to avoid creating them on every request
```

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,135 @@
...
+              .orElseGet(() -> new char[0]);
+      String nonProxyHosts = System.getProperty(protocol + ".nonProxyHosts");
+      map.put(
+          protocol, new ProxyDetails(protocol, host, port, user, password, nonProxyHosts, true));
+      LOGGER.debug("Proxy settings applied from system properties for {}", protocol);
+    }
+  }
+
+  public CredentialsProvider getCredentialsProvider(String protocol) {
+    BasicCredentialsProvider provider = new BasicCredentialsProvider();
```
johnBgood, line 83: as I said here I would not create a `CredentialsProvider` I would just return the store value for this protocol.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,135 @@
...
+      provider.setCredentials(
+          new AuthScope(p.host(), p.port()),
+          new UsernamePasswordCredentials(p.user(), p.password()));
+    }
+    return provider;
+  }
+
+  public HttpHost getProxyHost(String protocol, String requestUrl) {
+    ProxyDetails p = map.get(protocol);
+    return p == null || doesTargetMatchNonProxy(protocol, requestUrl)
```
johnBgood, line 98: I'd add some logs here showing what is actually used

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
ztefanie: And the CredentialsProvider should then be created in the execute function of the CustomApacheHttpClient?

johnBgood: I would create them in the constructor, you can see my comment above (which is not super clear maybe sorry about that :D )

connectors/http/http-base/src/test/resources/squid.conf
ztefanie: Thanks for your suggestion! That was actually my first idea as well, but in my tests, it enforced authentication for all paths. I tweaked it a bit, and now it’s working as expected. Really appreciate your input!

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
ztefanie: Ah of course, sorry have missed that. Thanks for clarifying. I adjusted it.

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -0,0 +1,135 @@
...
+import org.apache.hc.client5.http.impl.DefaultSchemePortResolver;
+import org.apache.hc.client5.http.impl.auth.BasicCredentialsProvider;
+import org.apache.hc.client5.http.impl.routing.DefaultProxyRoutePlanner;
+import org.apache.hc.client5.http.impl.routing.SystemDefaultRoutePlanner;
+import org.apache.hc.client5.http.routing.HttpRoutePlanner;
+import org.apache.hc.core5.http.HttpHost;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProxyHandler {
```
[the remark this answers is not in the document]
ztefanie: Move it to a separate file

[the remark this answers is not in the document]
ztefanie: Really good point :fire:  I added more tests especially for nonProxyHost matching check and found some wrongly handed edge case.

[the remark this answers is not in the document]
johnBgood: Love it, nice!

johnBgood (approved):
💯

johnBgood:
# QA information
## Test Environment
  - [ ] SaaS
  - [ ] SM

## Test Scope

## Test Data

PR #3949, opened 2025-02-06
feat(element-templates): add RPA connector

marstamm:
## Description

This connector allows you to execute a linked script on the RPA Worker

image

## Related issues

related to https://github.com/camunda/product-hub/issues/2533

## Checklist

### DIFF
--- a/connectors/rpa/element-templates/rpa-connector.json
+++ b/connectors/rpa/element-templates/rpa-connector.json
@@ -1,0 +1,328 @@
+{
+  "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
+  "name": "RPA Connector",
+  "description": "Connects with the Camunda RPA Worker",
+  "id": "camunda.connectors.rpa",
+  "appliesTo": [
+    "bpmn:Task"
+  ],
+  "category" : {
+    "id" : "connectors",
+    "name" : "Connectors"
+  },
+  "engines": {
+    "camunda": "^8.7"
+  },
+  "elementType": {
+    "value": "bpmn:ServiceTask"
+  },
+  "entriesVisible": false,
+  "groups": [
+    {
+      "id": "worker",
+      "label": "Worker"
+    },
+    {
+      "id": "script",
+      "label": "Script"
+    },
+    {
+      "id": "prerun",
+      "label": "Pre-run"
+    },
+    {
+      "id": "postrun",
+      "label": "Post-run"
+    },
+    {
+      "id": "input",
+      "label": "Input"
+    }
+  ],
+  "properties": [
+    {
+      "type": "Hidden",
+      "binding": {
+        "type": "zeebe:linkedResource",
+        "property": "resourceType",
+        "linkName": "RPAScript"
+      },
+      "value": "RPA"
+    },
+    {
+      "group": "script",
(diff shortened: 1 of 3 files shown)


connectors/rpa/element-templates/rpa-connector.json
```diff
@@ -0,0 +1,145 @@
...
+      "type": "String",
+      "value": "default",
+      "constraints": {
+        "notEmpty": true
+      }
+    },
+    {
+      "group": "worker",
+      "label": "Timeout",
+      "description": "Timeout of the Script in seconds.",
```
marstamm, line 115: It's an ISO-8601 string, need to adjust

marstamm: fixed in [f89e610](https://github.com/camunda/connectors/pull/3949/commits/f89e61052586a697a56bcc1427462edc1b2d605f)

sbuettner:
@marstamm Looks great. Will we only host the element template or will the actual Connector be contributed as well?

marstamm:
Actually, good point. I added this under the "Connectors" category, but I'm not sure if this is right.

The "Connector" will be a [Customized Worker](https://github.com/camunda/rpa-worker/) that the User has to install on their machine, since you are executing custom scripts + need access to local files with it. If it is not using the connectors runtime, does it still make sense to have this under Connectors then?

marstamm:
Basically, we would like to reuse the established distribution mechanisms for connector templates for this new OOTB template as well, I hope this is the correct way to do this 😄

sbuettner:
Not sure TBH, for all the other element templates we host in this repo we have at least one implementation in our repository and other Connector repos like https://github.com/camunda/sap-rfc-connector host both the element template and the code. 

As they usually need to change together I think they should be located as close as possible.

marstamm:
I'm fine with renaming the Element template into "RPA Task" instead of Connector and moving it to another group.

I still think we should reuse the distribution mechanism for the element template that Connectors provides, as it allows us to easily add element templates to all modelers.

sbuettner:
@marstamm Please be aware that you will have to add the element template manually to the web-modeler as well as submitting it to market place. Hence the question whether the benefit of having the element template here stays the same :)

marstamm:
Oh, I was not aware of this. Then it does indeed not make sense to add them to this repo 😄 

So the way to get it into the marketplace would be to submit it here https://marketplace.camunda.com/en-US/pages/submitConnector ?

> will have to add the element template manually to the web-modeler

Are they not using the Marketplace then? The Desktop modeler fetches the Element templates from here: https://marketplace.cloud.camunda.io/api/v1/connectors?creatorType=camunda

marstamm:
Closing as we will version manage in RPA repo and submit to Marketplace for distribution

sbuettner:
> Are they not using the Marketplace then? The Desktop modeler fetches the Element templates from here: https://marketplace.cloud.camunda.io/api/v1/connectors?creatorType=camunda

Its support but if you want to have the element template available without requiring users to go to the Marketplace you would have to add it right to the web-modeler: https://github.com/camunda/web-modeler/blob/7dd8c22dd6b90594d5856a87e8bcd6d2e528018a/webapp/src/stores/OOTBConnectorsStore.saas.js#L83

PR #3972, opened 2025-02-07
feat(sql): Add JSON support

### LINKED ISSUE #3895: [SQL Connector] Add JSON support
__What should we do?__

We should add support for JSON types that are supported for most (if not all) databases we support with the SQL Connector. For Postgres this would be support for JSON & JSONB.

There seems to be support available via this module: https://jdbi.org/#jdbi3-json

__Why should we do it?__

We see some customer requests asking for JSON support:
https://forum.camunda.io/t/feel-expression/47520/7

ztefanie:
## Description
Added JSON parsing for string columns.
Added Uni Tests for all sql specific json columns

## Related issues

closes #3895 

## Checklist

### DIFF
--- a/connectors/jdbc/pom.xml
+++ b/connectors/jdbc/pom.xml
@@ -13,6 +13,10 @@
   <artifactId>connector-jdbc</artifactId>
   <packaging>jar</packaging>
 
+  <properties>
+    <version.jdbi>3.47.0</version.jdbi>
+  </properties>
+
   <licenses>
     <license>
       <name>Camunda Self-Managed Free Edition license</name>
@@ -36,7 +40,12 @@
     <dependency>
       <groupId>org.jdbi</groupId>
       <artifactId>jdbi3-core</artifactId>
-      <version>3.47.0</version>
+      <version>${version.jdbi}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jdbi</groupId>
+      <artifactId>jdbi3-jackson2</artifactId>
+      <version>${version.jdbi}</version>
     </dependency>
 
     <dependency>
--- a/connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
+++ b/connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.jdbc.model.client;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import org.jdbi.v3.core.qualifier.QualifiedType;
+import org.jdbi.v3.core.result.ResultIterable;
+import org.jdbi.v3.core.result.UnableToProduceResultException;
+import org.jdbi.v3.core.statement.Query;
+import org.jdbi.v3.json.Json;
+
+public class JdbiJsonHelper {
+  static Map<String, Set<String>> dbJsonTypeMapping =
+      Map.of(
+          "Microsoft SQL Server",
+          Set.of(),
+          "MySQL",
+          Set.of("JSON"),
+          "PostgreSQL",
+          Set.of("json", "jsonb"),
+          "MariaDB",
+          Set.of("JSON"));
+
+  public static ResultIterable<Map<String, Object>> mapToParsedMap(
+      String databaseProductName, Query query) {
+    return query.map(
+        (rs, ctx) -> {
+          Map<String, Object> row = new HashMap<>();
+          // Java SQL ResultSet and ResultSetMetadata columns start with index 1:
+          // https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/ResultSetMetaData.html#getColumnTypeName(int)
+          for (int i = 1; i <= rs.getMetaData().getColumnCount(); i++) {
+            String columnName = rs.getMetaData().getColumnLabel(i);
+            Object value = rs.getObject(i);
+            if (isJsonColumn(databaseProductName, rs.getMetaData().getColumnTypeName(i))) {
+              try {
+                value =
+                    ctx.findColumnMapperFor(QualifiedType.of(JsonNode.class).with(Json.class))
+                        .orElseThrow()
+                        .map(rs, i, ctx);
+              } catch (UnableToProduceResultException ignored) {
+                row.put(columnName, value);
+              }
+            }
+            row.put(columnName, value);
+          }
+          return row;
+        });
+  }
+
+  private static boolean isJsonColumn(String databaseProductName, String columnTypeName) {
+    return dbJsonTypeMapping.getOrDefault(databaseProductName, Set.of()).contains(columnTypeName);
+  }
+}
--- a/connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
+++ b/connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import org.jdbi.v3.core.mapper.ColumnMapper;
 import org.jdbi.v3.core.qualifier.QualifiedType;
 import org.jdbi.v3.core.result.ResultIterable;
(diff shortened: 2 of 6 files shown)


connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
```diff
@@ -0,0 +1,53 @@
...
+public class JdbiJsonHelper {
+
+  public static ResultIterable<Map<String, Object>> mapToParsedMap(Query query) {
+    return query.map(
+        (rs, ctx) -> {
+          Map<String, Object> row = new HashMap<>();
+          for (int i = 1; i <= rs.getMetaData().getColumnCount(); i++) {
+            String columnName = rs.getMetaData().getColumnLabel(i);
+            Object value = rs.getObject(i);
+            if (isPotentiallyJsonColumn(rs.getMetaData().getColumnType(i), value)) {
```
sbuettner, line 23: Did you try using `rs.getMetaData().getColumnTypeName`?

It might be more useful as it returns the
> type name used by the database. If the column type is a user-defined type, then a fully-qualified type name is returned.

Instead of the JDBC type returned by `getColumnType`.

ztefanie: Yes, @johnBgood and I have thought about this. We decided to go with the `getColumnType` function, because there we get a List of SQL types that are not depending on the SQL dialect but general for all SQL databases, which is what we want. Some SQL dialects do not have specific JSON datatypes, but use VARCHAR for JSONs. 

If we would use `getColumnTypeName` we need to check for all types of all dialects, which makes the check a lot more complex, without any additional benefit.

sbuettner: > If we would use getColumnTypeName we need to check for all types of all dialects, which makes the check a lot more complex, without any additional benefit.

But currently we try to parse simple strings as json just to figure out whether it works or not right?

connectors/jdbc/pom.xml
```diff
@@ -38,6 +38,11 @@
       <artifactId>jdbi3-core</artifactId>
       <version>3.47.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.jdbi</groupId>
+      <artifactId>jdbi3-jackson2</artifactId>
+      <version>3.47.0</version>
```
johnBgood, line 44: we can start using a variable for this version now that we have more than 1 occurrence

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
ztefanie: You can see in the test which dialect handels it how:

MSSQL: NVARCHAR
MYSQL: JSON
POSTGRESQL: JSON, JSONB
MARIADB: JSON

connectors/jdbc/src/test/java/io/camunda/connector/jdbc/integration/JdbiJdbcClientIntegrationTest.java
```diff
@@ -394,4 +398,26 @@ public void shouldReturnResultList_whenSelectQueryWhereInWithBindingParameters(
       }
     }
   }
+
+  @Nested
+  class JsonTests {
+
+    @ParameterizedTest
+    @MethodSource(PROVIDE_SQL_SERVERS_CONFIG)
+    public void shouldParseJson_whenPostgresJsonAndJsonBColumnTypes(IntegrationTestConfig config)
```
johnBgood, line 407: I think the naming is incorrect as it's used for all dbs, right?

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
```diff
@@ -0,0 +1,53 @@
...
+              }
+            }
+            row.put(columnName, value);
+          }
+          return row;
+        });
+  }
+
+  private static boolean isPotentiallyJsonColumn(int columnType) {
+    Set<Integer> jsonCompatibleTypes =
```
johnBgood, line 40: this could be a constant

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
```diff
@@ -0,0 +1,53 @@
...
+import org.jdbi.v3.core.statement.Query;
+import org.jdbi.v3.json.Json;
+
+public class JdbiJsonHelper {
+
+  public static ResultIterable<Map<String, Object>> mapToParsedMap(Query query) {
+    return query.map(
+        (rs, ctx) -> {
+          Map<String, Object> row = new HashMap<>();
+          for (int i = 1; i <= rs.getMetaData().getColumnCount(); i++) {
```
johnBgood, line 20: still not sure about this `1` , it might at least be worth a comment

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
sbuettner: >  put jsons in VARCHAR columns
Yes, but thats something that needs to be handled in userland from my perspective. We should only cover when json is managed in fields we know that contain json.

sbuettner: Otherwise we are just guessing and bruteforce json deserialization on every string (plus the other types that are defined in the method).

ztefanie: ok. Thanks for explaining. I adjusted it.

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
```diff
@@ -0,0 +1,49 @@
...
+              }
+            }
+            row.put(columnName, value);
+          }
+          return row;
+        });
+  }
+
+  private static boolean isJsonColumn(String databaseProductName, String columnTypeName) {
+    Map<String, Set<String>> dbTypeMapping = new HashMap<>();
```
johnBgood, line 42: minor detail, but it can be a static field in this class created with Map.of(....,...,....,...)

johnBgood (dismissed):
🥇

johnBgood:
# QA information
## Test Environment
  - [ ] SaaS
  - [ ] SM

## Test Scope

## Test Data

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
```diff
@@ -0,0 +1,60 @@
...
+          // https://docs.oracle.com/en/java/javase/17/docs/api/java.sql/java/sql/ResultSetMetaData.html#getColumnTypeName(int)
+          for (int i = 1; i <= rs.getMetaData().getColumnCount(); i++) {
+            String columnName = rs.getMetaData().getColumnLabel(i);
+            Object value = rs.getObject(i);
+            if (isJsonColumn(databaseProductName, rs.getMetaData().getColumnTypeName(i))) {
+              try {
+                value =
+                    ctx.findColumnMapperFor(QualifiedType.of(JsonNode.class).with(Json.class))
+                        .orElseThrow()
+                        .map(rs, i, ctx);
```
sbuettner, line 46: Do we have to do this lookup on every field? It seems like its always the same mapper we are using so why not do it once for all fields and rows?

ztefanie: Good point :+1: Fixed it.

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/client/JdbiJsonHelper.java
```diff
@@ -0,0 +1,61 @@
...
+            String columnName = rs.getMetaData().getColumnLabel(i);
+            Object value = rs.getObject(i);
+            if (isJsonColumn(databaseProductName, rs.getMetaData().getColumnTypeName(i))) {
+              try {
+                value = jsonMapper.map(rs, i, ctx);
+              } catch (UnableToProduceResultException ignored) {
+                row.put(columnName, value);
+              }
+            }
+            row.put(columnName, value);
```
sbuettner, line 52: Shouldnt this maybe in an `else` clause?

ztefanie: No, if I move this in a else clause, the value will not be written in the row after json parsing it.

sbuettner (approved):
🎉

PR #3978, opened 2025-02-10
fix(properties): correct properties for 8.6 docker images

chillleader:
## Description

This PR will set the correct config properties for our docker images.

## Related issues

closes #

## Checklist


bundle/default-bundle/src/main/resources/application.properties
johnBgood, line 1: do we need these params if we have them in the connector-runtime properties?

johnBgood (approved):

bundle/default-bundle/src/main/resources/application.properties
chillleader: Yes - we tested with @sbuettner that properties from connector-runtime are ignored if we run it as a docker image. We decided to keep them everywhere and synchronize

sbuettner (approved):

sbuettner (approved):

PR #3981, opened 2025-02-10
fix(box-connector): Avoid refetching data of deleted items

sbuettner:
## Description

Fixes the exception that happened after deleting a file or a folder when the Box client tried to refetch information from the now deleted object.

## Checklist

- [x ] PR has a **milestone** or the `no milestone` label.


sbuettner:
@chillleader It would be awesome if we could get this into the 8.7 release if its not final yet.

johnBgood (approved):

PR #3996, opened 2025-02-13
fix(email-connector): Remove unused dependency

sbuettner:
## Description

Remove unused dependency that ended up in the final jar.


mathias-vandaele (approved):
👍

PR #4001, opened 2025-02-14
fix(docs): Fix broken links in element templates

ztefanie:
## Description
- Fix broken docs links in element templates
- Some links are still broken, because they refer to newer versions of the docs. These are the links for s3, box and gemini connectors, that are not in the docs of version 8.6. They will work as soon as newer versions become the default.
- related issue: https://github.com/camunda/camunda-docs/pull/4992 there I created a new script that checks all element templates for broken links.

## Related issues

closes https://github.com/camunda/camunda-docs/issues/4995

## Checklist

### DIFF
--- a/connectors/aws/aws-base/README.md
+++ b/connectors/aws/aws-base/README.md
@@ -55,7 +55,7 @@ AWS-specific services.
 
 ### AWS DynamoDB Connector
 
-The [AWS DynamoDB Connector](https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/aws-dynamodb/)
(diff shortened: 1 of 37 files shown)


sbuettner:
@ztefanie Great initiative! 

As we already have the infrastructure (java libraries) in the Connectors repo to reliavbly parse and extract information from element-templates it might be more reliable to have this code in this repo rather than the other way around?

In addition to that we try (on the Connectors side) to have more complex logic written in Java for easier debuggability and maintenance (see https://github.com/camunda/connectors/tree/main/element-template-generator/uniquet)

sbuettner (approved):
Thanks for tackling this @ztefanie! 🥇

PR #4050, opened 2025-02-19
feat(rest): Improve http proxy

ztefanie:
## Description
Due to a bug I needed to adjust the handling of proxy configs set via env vars.

~~Once approved i will backport this to 8.7 with cherry-pick here: https://github.com/camunda/connectors/pull/4002~~
As this PR got merged, I added the backport label and a new PR will be created.

## Related issues
related to: https://github.com/camunda/team-connectors/issues/989

## Checklist


sbuettner (approved):
Great investigation and solution @ztefanie 🎉

johnBgood:
👍 Nice! One thing though, I can't see an attached issue (other than the original one). 
Can you describe the bug you fixed, just a few steps would be enough to get some context in the future?

Thank you again for fixing this 🙏

ztefanie:
> 👍 Nice! One thing though, I can't see an attached issue (other than the original one). Can you describe the bug you fixed, just a few steps would be enough to get some context in the future?
> 
> Thank you again for fixing this 🙏

Yes, sure. The problem was some sort of race condition, when running requests in parallel. So when you have a BPMN where two or more requests are executed in parallel and request1 goes to the nonProxyHosts and the others not, you will see that request1 will sometimes be proxied and sometimes not, same with the other requests.

Issue #3986, opened 2025-02-11
Email connector e2e tests are flaky


chillleader:
__Describe the Bug__

Email connector e2e tests sometimes fail randomly on CI. Here's one of the recent examples: https://github.com/camunda/connectors/actions/runs/13246828669/attempts/3

PR #4051, opened 2025-02-19
chore(email-tests): 3986 email connector e2e tests are flaky

mathias-vandaele:
## Description

All the e2e have been migrated to integration tests

## Related issues

closes https://github.com/camunda/connectors/issues/3986

## Checklist


johnBgood (approved):
Great but I would rename the package. These are not integration tests as we don't have any "real" mail server there. They look like regular UTs to me.
If you want integration tests, I recommend using [docker-mailserver](https://github.com/docker-mailserver/docker-mailserver) along with TestContainers. That would be great, but it's up to you 🇵🇭

PR #4128, opened 2025-02-25
feat(document): intrinsic operations

### LINKED ISSUE #3843: Intrinsic operations in the Connector SDK
### Is your feature request related to a problem? Please describe

The Connector SDK supports document handling capabilities since 8.7. However, the SDK is still missing a convenient way to apply transformations to documents in runtime without creating too much operational overhead.

**Example use-case:** The [GitHub connector](https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/github/) can support uploading files to GitHub via the API. GitHub API accepts [Base64-encoded files](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28).

Since Camunda now supports documents, I would like to be able to upload a Camunda document to GitHub, but for that I need to provide it in the Base64 format. I would like to do this in a low-code way that fits well into existing connectors machinery, and I would like to avoid creating a separate connector invocation to transform the document into Base64, as this will be overly complicated and will negate the benefits of seamless document mappings in the Connector SDK.

I therefore would like to have an option to add manipulations to the connector input variables that are executed in runtime, after the variable values are resolved and all document references are transformed to documents, but before the values are passed to the connector function.

### Describe the solution you'd like

The concept of an intrinsic operation can be implemented to instruct the connector runtime to perform a manipulation on values before they are passed to the connector function.

An intrinsic operation is a JSON object that can be provided as a connector input, for example:

```json
{
  "camunda.operation.name": "base64",
  "inputs": [
    {"camunda.document.type": "camunda", "documentId": "...", "storeId": "s3", "metadata": {}} 
  ]  
}
```

If the connector runtime receives this object as part of the job input, it will recognize the operation based on the presence of the marker field (`camunda.operation.name`) and execute the operation before the connector function object is invoked.

The connector can then bind the properties using `connectorContext.bindVariables(InputType.class)` - the result of the operation execution will be of type `String` for base64 content, but can be different for other operations.

This solution will also allow to retrofit existing connectors with document support without the need to modify the connector code - as long as an intrinsic operation transforms the document to a format that can be consumed by a legacy connector.

Further iterations may include FEEL functions to simplify the creation of operation objects for the most frequently used operations - for example, the example above can be simplified to `to base64(myDocument)`, where `to base64` is a custom FEEL function, and `myDocument` is a variable that contains the document reference.

### Describe alternatives you've considered

- Ad-hoc connectors can be implemented for the given transformations, but this would not be a low-code solution and would require significant effort to implement every new kind of transformation.

- Since this solution mainly addresses documents, we have initially considered including the operation into the document reference structure. However, with a separate concept of an operation, we will have more benefits such as: _1)_ an ability to pass multiple inputs into the operation, e.g. merge two documents into one; _2)_ room for future extensions of the operation concept and potentially applying it to other values (not only documents). The main downside of the suggested solution where operations are first-class citizens rather than part of document references is the need to introduce a new abstraction and overall higher maintenance effort.

### Additional context (🔒 limited access)

- This feature request evolved and branched off from the [Connector SDK document support epic](https://github.com/camunda/team-connectors/issues/873)
- [Initial design documentation](https://docs.google.com/document/d/1N1tAQjrgyZec-jWwUOQAiBwvKR3Q3w_ZM8KUf-MRR4A/edit?usp=sharing)

chillleader:
## Description

For an introduction into intrinsic operations concept, please see https://github.com/camunda/connectors/issues/3843

This PR adds intrinsic operation support. It consists of two parts:
- New operation structure (new `IntrinsicOperation`, `IntrinsicOperationExecutor` and `IntrinsicOperationProvider` interfaces with stricter input/output typing and hopefully a better developer experience)
- Refactored the Jackson module to have a better structure and to support operation deserialization

**Use-cases considered in the Jackson module:**
- Operation can output strings
- Operation can output documents which can be consumed in any way the document module allows (e.g. inputstream/bytearray)
- Operation can output a generic object
- Nested operations are also supported without any extra effort as we are reusing the same object mapper for parameter binding

## Related issues

closes #3843
related https://github.com/camunda/team-connectors/issues/1004

## Checklist


bundle/default-bundle/src/test/resources/application.properties
```diff
@@ -8,9 +8,9 @@ management.endpoint.health.show-details=always
 camunda.client.execution-threads=10
 
 camunda.client.grpc-address=http://localhost:26500
+camunda.client.rest-address=http://localhost:8089/b7696783-ce0b-4807-b8d6-ab53f0e7c54d
```
sbuettner, line 11: Intended?

connector-sdk/document/src/main/java/io/camunda/document/operation/Base64Operation.java
```diff
@@ -0,0 +1,43 @@
...
+    if (maybeDocument == null || !maybeDocument.isDocumentParameter()) {
+      return new ValidationFailure<>("Base64 operation expects a single document as argument");
+    }
+
+    try {
+      final var base64 = maybeDocument.asDocumentParameter().asBase64();
+      return new Success<>(base64);
+    } catch (Exception e) {
+      return new ExecutionFailure<>("Failed to read document as base64", e);
+    }
```
sbuettner, line 41: Would be cool we could support Strings as input as well.

connector-sdk/jackson-datatype-document/src/main/java/io/camunda/connector/document/jackson/OperationModel.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package io.camunda.connector.document.jackson;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+
+public record OperationModel(@JsonProperty(DISCRIMINATOR_KEY) String name, List<Object> params) {
+
+  public static final String DISCRIMINATOR_KEY = "camunda.operation.type";
```
sbuettner, line 24: Just food for thought, we could also add the name to the property like this: 
```camunda.operation::base64``` and we would have an object that is slightly more dense:
```
{"camunda.operation::base64": [...]``

connector-sdk/document/src/main/java/io/camunda/document/operation/Base64Operation.java
chillleader: done

bundle/default-bundle/src/test/resources/application.properties
chillleader: nope :)

connector-sdk/document/src/main/java/io/camunda/operation/DefaultIntrinsicOperationExecutor.java
```diff
@@ -0,0 +1,113 @@
...
+        if (operationSources.containsKey(entry.getKey())) {
+          throw new IllegalArgumentException(
+              "Operation with name: "
+                  + entry.getKey()
+                  + " duplicated in providers: "
+                  + provider.getClass().getName()
+                  + " and "
+                  + operationSources.get(entry.getKey()).provider.getClass().getName());
+        }
+        operationSources.put(entry.getKey(), new OperationSource(provider, entry.getValue()));
```
sbuettner, line 71: Can we log all discovered Operations?

connector-sdk/document/src/main/java/io/camunda/operation/DefaultIntrinsicOperationExecutor.java
```diff
@@ -0,0 +1,113 @@
...
+                    method -> method));
+    if (methodsByOperationName.isEmpty()) {
+      throw new IllegalArgumentException(
+          "No intrinsic operations found in provider "
+              + provider.getClass().getName()
+              + ". "
+              + "At least one method with @IntrinsicOperation expected.");
+    }
+    return methodsByOperationName;
+  }
```
sbuettner, line 112: We can refactor this later but I would extract the discovery and reflection part from the executor.

A `IntrinsicOperationRegistry` could do the discovery and do the reflection resulting in a dedicated `IntrinsicOperationInstance` record that hold all the necessary information for the executor to just "call" it.

connector-sdk/document/src/main/java/io/camunda/operation/IntrinsicOperation.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * <p>By default, all arguments are required. If an argument is nullable, it must be annotated with
+ * {@link javax.annotation.Nullable}.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface IntrinsicOperation {
+
+  String name();
+}
```
sbuettner, line 37: We might want to use the term [IntrinsicFunction](https://en.wikipedia.org/wiki/Intrinsic_function) as it seems to be the canonical name for this concept. WDYT?

connector-sdk/document/src/main/java/io/camunda/operation/IntrinsicOperationExecutor.java
```diff
@@ -14,8 +14,9 @@
...
  */
-package io.camunda.connector.document.annotation.jackson;
+package io.camunda.operation;
 
-import java.util.function.Supplier;
+public interface IntrinsicOperationExecutor {
 
-public interface DocumentOperationResult<T> extends Supplier<T> {}
+  <T> T execute(String operationName, IntrinsicOperationParams params, Class<T> resultType);
+}
```
sbuettner, line 22: The executor receives the `resultType`? Shouldnt that come from the operation itself?

connector-sdk/document/src/main/java/io/camunda/operation/impl/Base64Operation.java
```diff
@@ -0,0 +1,37 @@
...
+  public String execute(Object input) {
+    if (input instanceof Document) {
+      return ((Document) input).asBase64();
+    }
+    if (input instanceof String) {
+      return Base64.getEncoder().encodeToString(((String) input).getBytes());
+    }
+    throw new IllegalArgumentException(
+        "Unsupported input type: " + input.getClass() + ". Expected Document or String.");
+  }
```
sbuettner, line 36: Super nice and concise. Could be even nicer with a `switch` statement :)

connector-sdk/document/src/main/java/io/camunda/operation/impl/CreateLinkOperation.java
```diff
@@ -14,19 +14,21 @@
...
-
-  @Override
-  public Object execute(DocumentOperation operationReference, Document document) {
-    return document.asBase64();
+  @IntrinsicOperation(name = "createLink")
+  public String execute(Document document, @Nullable Duration ttl) {
+    if (ttl != null) {
+      return document.generateLink(ttl);
+    }
+    return document.generateLink();
```
sbuettner, line 32: Maybe `generateLink` could accept a `GenerateLinkRequest` that accepts a nullable TTL so we dont need two signatures.

connector-sdk/document/src/main/java/io/camunda/operation/impl/GetTextOperation.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class GetTextOperation implements IntrinsicOperationProvider {
+
+  @IntrinsicOperation(name = "getText")
+  public String execute(Document document, @Nullable String charset) {
+    final var bytes = document.asByteArray();
+    final var charsetInstance =
+        charset != null ? Charset.forName(charset) : Charset.defaultCharset();
+    return new String(bytes, charsetInstance);
+  }
```
sbuettner, line 33: Nice and clean!

connector-sdk/document/src/main/java/io/camunda/operation/IntrinsicOperationExecutor.java
chillleader: I thought of this as a way for the caller to say "I want the result of the function to be of this type, otherwise throw an exception", which is a common use case in our deserializers. But I'm with you here, we can remove this, and also I think the result type conversion via Jackson is problematic from the security perspective (typical injection attack if we use `parseJson` and the document contains a JSON which defines an operation)

connector-sdk/jackson-datatype-document/src/main/java/io/camunda/connector/document/jackson/OperationModel.java
sbuettner: Agree. Consistency is key.

connector-sdk/document/src/main/java/io/camunda/document/store/DocumentLinkCreationRequest.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package io.camunda.document.store;
+
+import io.camunda.document.reference.DocumentReference.CamundaDocumentReference;
+import java.time.Duration;
+import java.util.Optional;
+
+public record DocumentLinkCreationRequest(
+    CamundaDocumentReference reference, Optional<Duration> timeToLive) {}
```
sbuettner, line 24: Please consider https://www.baeldung.com/java-record-optional-param and maybe remove the `Optional` here.

ztefanie (commented):
Good work, nice feature :rocket: 

Why is the prefixed "chore" and not "feature"?

I tried to test this, but couldn't get it working. I used camunda platform `connectors-8.7.0-alpha2`. I did a maven clean install in connectors. I used desktop modeler with this BPMN:
image
but the issue that is created has not description:

chillleader:
> Why is the prefixed "chore" and not "feature"?

Right, let's change it back to feat, I initially decided to make it a chore because I was considering to split the implementation in several PRs and only get a QA approval for the last one. However, we added more functions in this PR so it makes sense to call it a feature and require QA approval.

> I tried to test this, but couldn't get it working. I used camunda platform connectors-8.7.0-alpha2. I did a maven clean 
install in connectors. I used desktop modeler with this BPMN:

This would not work as we decided not to implement the FEEL part yet. For now, if you want to invoke an operation, you need to provide the whole payload in the input, i.e.:

```
{
  
  "camunda.operation.type": "base64",
  "params": ["test123"]
}
```

johnBgood:
# QA information
## Test Environment
  - [ ] SaaS
  - [ ] SM

## Test Scope

## Test Data

connector-sdk/jackson-datatype-document/src/main/java/io/camunda/connector/document/jackson/deserializer/AbstractDeserializer.java
```diff
@@ -0,0 +1,51 @@
...
+
+  /**
+   * Deserialize from a JSON node. This method is used to cross-reference the deserialization logic
+   * from another deserializer, when the JSON node has already been read form the parser, because
+   * the parser can only be read once. All deserializer-specific logic should be implemented in this
+   * method.
+   */
+  protected abstract T handleJsonNode(JsonNode node, DeserializationContext context)
+      throws IOException;
+}
```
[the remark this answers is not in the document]
chillleader: We had something similar in the previous implementation of the document module. I'm a bit hesitant to re-introduce it again, because:
1. the complexity is currently quite manageable as is IMO, we get a pair of if-else in every deserializer, but on the other hand it's immediately visible what exactly each deserializer is doing; and
2. When we decide to introduce more possible return values of the intrinsic function, e.g. support for numeric return types, or maybe a supplier pattern to execute the operation lazily,  those new deserializers will not have a case where they need to deserialize something as a document - they will only be used for operations, so the abstraction may not be future-proof.

connector-sdk/document/src/main/java/io/camunda/intrinsic/IntrinsicFunctionParams.java
```diff
@@ -14,13 +14,13 @@
...
+import java.util.List;
 
-public interface DocumentOperationExecutor {
+public sealed interface IntrinsicFunctionParams {
 
-  boolean matches(DocumentOperation operationReference);
+  record Positional(List<Object> params) implements IntrinsicFunctionParams {}
 
-  Object execute(DocumentOperation operationReference, Document document);
+  // TODO: named parameters
```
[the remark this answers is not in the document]
chillleader: will add one 👍

connector-sdk/document/src/main/java/io/camunda/operation/DefaultIntrinsicOperationExecutor.java
chillleader: Done

chillleader:
@ztefanie it should work now, just a heads up that I also just changed the discriminator field name from operation to function so the payload should look like this now:
```
{
  
  "camunda.function.type": "base64",
  "params": ["test123"]
}
```

ztefanie (approved):
Nice, tested again and works as expected :100: :+1: 

Can the error handling be a bit improved? When using a non supported operation, e.g. base63, in operate I only get this error:
```	
Exception: JsonMappingException was raised

{"message":"Exception: JsonMappingException was raised","type":"io.camunda.connector.api.error.ConnectorException","code":"JSON_PROCESSING_ERROR"}
```

Would be nice, to also see this `No intrinsic function found with name: base63` message from the stacktrace in operate.

chillleader:
> Would be nice, to also see this No intrinsic function found with name: base63 message from the stacktrace in operate.

Created a follow-up ticket 👍 https://github.com/camunda/connectors/issues/4283
The message itself is provided but we don't expose such messages to the user at the moment. This is potentially related to https://github.com/camunda/connectors/issues/3221, I'll double check with Mathias

Szik (approved):

connector-sdk/document/src/main/java/io/camunda/intrinsic/functions/Base64Function.java
```diff
@@ -0,0 +1,37 @@
...
+package io.camunda.intrinsic.functions;
+
+import io.camunda.document.Document;
+import io.camunda.intrinsic.IntrinsicFunction;
+import io.camunda.intrinsic.IntrinsicFunctionProvider;
+import java.util.Base64;
+
+public class Base64Function implements IntrinsicFunctionProvider {
+
+  @IntrinsicFunction(name = "base64")
```
johnBgood, line 26: I didn't finished the review, so maybe it's done already: 
For the future, could we have some kind of generated docs showing all the OOTB functions with possible params? 
As a user I'd love to know that I can use base64 either with a document or a string for instance

chillleader: That's a great idea - I created a follow-up ticket: https://github.com/camunda/connectors/issues/4310

Issue #4145, opened 2025-02-26
IDP Connector should return response as Key/Value pair of extracted fields instead of json string


mathieu-stennier:
__What should we do?__

IDP Connector should return response as Key/Value pair of extracted fields instead of json string
Full conversation [here](https://camunda.slack.com/archives/C084ZLC2Z2R/p1740517057335689).

Response provided by a deployed IDP connector would be something like this

```
"response":"\n\n{\n  \"PolicyholderName\": \"Residential Builders Inc.\",\n  \"PolicyHolderAddress\": \"Parklaan 88 3500 Hasselt Belgie\",\n  \"PolicyholderEnterpriseNumber\": \"1234567890\",\n  \"PolicyholderCapacity\": \"Project Owner\",\n  \"TotalProjectValue\": \"5000000 EUR\",\n  \"ProjectDescription\": \"This project involves the construction of a new 5-story residential apartment building on a currently vacant lot. The building will have reinforced concrete construction, with one underground parking level and five above-ground floors, including a ground floor with retail spaces.\"\n}"
```

Which is not easily parseable as variables that can then be used in isolations in next steps of the process.

See related FEEL expression feature gap [here](https://github.com/camunda/product-hub/issues/2632).

__Why should we do it?__

I would argue we should parse the results we get from LLM already and provide variables that can be easily mapped as result for further activities in the process.

mathieu-stennier:
@sahilbhatoacamunda,
@sbuettner recommends us to invest into a proper schema in response that allows you to add more feature in the future. `ExtractionResult` cotaining extracted fields might be enough for now but then at least we can extend it later on.

mathieu-stennier:
Note: changing this output is ok before `8.7` because we don't offer support for things built with alphas.

sahilbhatoacamunda:
@mathieu-stennier @sbuettner 
The ExtractionResult will have a response which is of type `Map<String, Object>`. This map will be built based on the IDP extraction fields passed as `taxonomy items` in the request.

Use cases
1. The response from LLM is a valid JSON string (all taxonomy items available)> The Connector will parse the JSON string and build the final map by checking if all the `taxonomy items` are available in the response.
2. The response from LLM is a valid JSON string (some taxonomy items are missing)> The Connector will parse the JSON string and build the final map by checking if all the `taxonomy items` are available in the response. The system will log a warning `LLM model response is the following missing keys: ('missing key 1', 'missing key 2'})`.
[...]

PR #4147, opened 2025-02-26
feat(idp-extraction-connector): idp connector should return response as key/value pair of extracted fields

sahilbhatoacamunda:
## Description

Changes were made to the response from `ExtractionConnectorFunction`. The new method `buildResponseJsonIfPossible` ensures a specific structure based on the `taxonomyItems` input.

## Related issues

closes https://github.com/camunda/connectors/issues/4145

## Checklist

### DIFF
--- a/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
+++ b/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
@@ -6,4 +6,6 @@
  */
 package io.camunda.connector.idp.extraction.model;
 
-public record ExtractionResult(Object response) {}
+import java.util.Map;
+
+public record ExtractionResult(Map<String, Object> response) {}
--- a/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
+++ b/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
@@ -6,7 +6,6 @@
  */
 package io.camunda.connector.idp.extraction.model;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import java.util.Map;
 
-public record ExtractionResult(Map<String, JsonNode> response) {}
+public record ExtractionResult(Map<String, Object> response) {}
--- a/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
+++ b/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
@@ -6,6 +6,7 @@
  */
 package io.camunda.connector.idp.extraction.model;
 
+import com.fasterxml.jackson.databind.JsonNode;
 import java.util.Map;
 
-public record ExtractionResult(Map<String, Object> response) {}
+public record ExtractionResult(Map<String, JsonNode> response) {}
--- a/connectors/idp-extraction/src/test/java/io/camunda/connector/idp/extraction/ExtractionConnectorFunctionTest.java
+++ b/connectors/idp-extraction/src/test/java/io/camunda/connector/idp/extraction/ExtractionConnectorFunctionTest.java
@@ -138,7 +138,7 @@ class ExtractionConnectorFunctionTest {
   }
 
   @Test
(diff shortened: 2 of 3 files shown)


connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
```diff
@@ -6,4 +6,7 @@
  */
 package io.camunda.connector.idp.extraction.model;
 
-public record ExtractionResult(Object response) {}
+import com.fasterxml.jackson.databind.JsonNode;
+import java.util.Map;
+
+public record ExtractionResult(Map<String, JsonNode> response) {}
```
sbuettner, line 12: I dont know who will consume the output but I would recommend to have a structured response that is somewhat stable. `JsonNode` could be everything I guess 😄 but if that is required (for structured table data for example) and we dont want a schema in place it might be fine.

sahilbhatoacamunda: @sbuettner I can enforce `Map<String, String>`, but I don't want to limit it to just the `String` type. If LLM is able to return a number, then I think we shouldn't modify its output in any way. The consumer can cast it to string.

sbuettner: Yeah, `Map<String, String>` wouldnt be sufficient. I just wasnt sure whether you want to allow complex objects as well or wanted to limit it to primitive types.

sahilbhatoacamunda: Should I change it to `Map<String, Object>`?

sbuettner: @sahilbhatoacamunda That would work. Although there would still be no verification that your response doesnt contain complex objects like the `recipient` in this example: 

```
{
  "invoiceNumer": "abc123", 
  "amount": 400,
"recipient": {"name": "Simon"}
```

Maybe thats totally fine for your use-cases and if so `Map<String,Object>` would be sufficient.

For the future it could be interesting to look into something like this: https://docs.langchain4j.dev/tutorials/structured-outputs/#json-schema

sbuettner (approved):

connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/model/ExtractionResult.java
sahilbhatoacamunda: Done.

sbuettner: @sahilbhatoacamunda I could not find a test that works with a complex object for a single key like this: 
```
{
  "invoiceNumer": "abc123", 
  "amount": 400,
"recipient": {"name": "Simon"}
```

sahilbhatoacamunda: Added.

connectors/idp-extraction/src/test/java/io/camunda/connector/idp/extraction/ExtractionConnectorFunctionTest.java
```diff
@@ -33,20 +43,269 @@ class ExtractionConnectorFunctionTest {
...
+    var expectedResponseJson =
+        """
+            {
+              "sum": "$12.25",
+              "supplier": {
+                            "id": 12345,
+                            "name": "Camunda Inc.",
+                            "city": "Berlin",
+                            "country": "Germany"
+                          }
```
sbuettner, line 142: @sahilbhatoacamunda So you want to support embedded object? I thought you only want to support strings and numbers?

sahilbhatoacamunda: @sbuettner 
As of now, Web Modeller doesn't support it. I only added this scenario to keep the connector flexible as it all depends on the prompt that user passes in the `taxonomyItems`

sbuettner (approved):

PR #4191, opened 2025-03-05
fix(security): Disallow include in pebble engine

ztefanie:
Disallow include in pebble engine in element template builder core. See [here](https://www.cve.org/CVERecord?id=CVE-2025-1686) and [here](https://camunda.slack.com/archives/C02JLRNQQ05/p1741089936227569)

## Description
Applied the suggested fix. Tested element template generation for email connector and openAPI specs. Works as expected. Pebble engine is not used in [8.5.](https://github.com/camunda/connectors/blob/release/8.5/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java) so only needs backport to 8.7 and 8.6

## Checklist


sbuettner (approved):
Thanks for tackling this so quickly!

PR #4199, opened 2025-03-05
feat(message): Publish message outbound connector

ingorichtsmeier:
## Description

Support publishing messages and correlating messages directly from the BPMN model without writing your own worker for this.

## Related issues

No issues here, it's a new feature.

closes #

## Checklist

- [X] PR has a **milestone** or the `no milestone` label.

It would be great if the connector could be published with Camunda 8.7


Szik (dismissed):

connectors/message/src/main/resources/send.svg
sbuettner, line 1: @YanaSegal Can you check whether this logo can be used?

ingorichtsmeier: My first try was without a logo, as the send task has an icon in the BPMN spec and I don't see the requirement to override it, just for using a connector to publish the message.

But in the connector list in the modeler, it looks strange. It is the only entry without an icon. In the process model, it stays the send task.

Using the [original icon from the BPMN Font](https://github.com/bpmn-io/bpmn-font/blob/main/source/raw/activities/type/send.svg), the icon appeared much smaller than the original send task. There is a padding of 100 pixels around the envelope, which is scaled by unknown logic.

So I adjusted the SVG from the BPMN font to an icon without padding and smaller SVG footprint.

YanaSegal: Ingo, can you share screenshots, please?

ingorichtsmeier: Append menu:
image

Send task vs message connector send task:
image

All elements sending with or without connector:
image

All elements comparing the send task vs scaled BPMN font icon vs. original BPMN font icon:
image

connectors/message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
```diff
@@ -0,0 +1,131 @@
...
+  public Object execute(OutboundConnectorContext context) throws Exception {
+
+    SendMessageRequest messageRequest = context.bindVariables(SendMessageRequest.class);
+    LOG.debug(
+        "Invoke send message connector with name {} and correlation key {}",
+        messageRequest.messageName(),
+        messageRequest.correlationKey());
+
+    ZeebeClient zeebeClient = ZeebeClientContext.getZeebeClient();
+
```
sbuettner, line 72: This could be done once in the constructor.

connectors/message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
```diff
@@ -0,0 +1,131 @@
...
+      }
+      if (messageRequest.variables() != null) {
+        correlateMessageCommandStep3.variables(messageRequest.variables());
+      }
+      if (messageRequest.tenantId() != null) {
+        correlateMessageCommandStep3.tenantId(messageRequest.tenantId());
+      }
+      if (messageRequest.requestTimeout() != null) {
+        correlateMessageCommandStep3.requestTimeout(messageRequest.requestTimeout());
+      }
```
sbuettner, line 119: What about having a separate Mapper class the is repsonsible for mapping a `SendMessageRequest` to a `Command`? 

This woudl make this code more readable and keep the Connector function simple.

connectors/message/src/main/resources/send.svg
sbuettner: @YanaSegal Can you decide on the logo?

YanaSegal: Thanks for the screenshots - so, in general, it works.
It does not visually differentiate between the regular send task and the send task connector, but the user can find this information in the properties panel. (users can upload whatever icons they want to their connectors)

@ingo, if visual differentiation is not an issue on your side, we are good to go. 
otherwise, I can provide a different icon.

connectors/message/src/main/java/io/camunda/connector/message/SendMessageRequest.java
```diff
@@ -0,0 +1,43 @@
...
+            type = TemplateProperty.PropertyType.Dropdown,
+            choices = {
+              @TemplateProperty.DropdownPropertyChoice(
+                  label = "publish message (buffered)",
+                  value = "publish"),
+              @TemplateProperty.DropdownPropertyChoice(
+                  label = "correlate message (with result)",
+                  value = "correlate")
+            })
+        String mode,
```
chillleader, line 28: As @sbuettner mentioned, isually we define sealed hierarchies and rely on auto-generated dropdowns, here are a few examples:
- **[JDBC connector](https://github.com/camunda/connectors/blob/main/connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/request/connection/JdbcConnection.java)** - choice between connection types
- **[Box Connector](https://github.com/camunda/connectors/blob/6e4bef7581dfc5250f95f487ed023f9ab554dbf9/connectors/box/src/main/java/io/camunda/connector/box/model/BoxRequest.java#L135)** - choice between operation types.

In this case we could structure it like this:
- `record SendMessageRequest(messageName, correlationKey, tenantId, requestTimeout, correlationType)`
- `sealed interface CorrelationType permits CorrelateWithResult, Publish {}`
- `record CorrelateWithResult() implements CorrelationType {} // empty record as no additional props`
- `record Publish(timeToLive, messageId) implements CorrelationType {}`

But since the case is quite simple, I'm also fine to leave it as is if you prefer it this way.

connectors-e2e-test/connectors-e2e-test-message/src/test/java/io/camunda/connector/e2e/MessageTests.java
```diff
@@ -0,0 +1,221 @@
...
+Details from surefire report:
+2025-03-03T08:45:02.658+01:00 DEBUG 99400 --- [pool-7-thread-1] i.c.c.r.c.outbound.ConnectorJobHandler   : Exception while processing job: 2251799813685359 for tenant: <default>
+
+io.camunda.zeebe.client.api.command.ProblemException: Failed with code 401: 'Unauthorized'. Details: 'class ProblemDetail {
+    type: about:blank
+    title: Unexpected server response
+    status: 500
+    detail: {"message":"Full authentication is required to access this resource"}
+    instance: null
+}'
```
chillleader, line 135: interesting finding, do you already have an idea if this is a problem in `CamundaProcessTest`? We can follow up with the CPT team to clarify

connectors/message/src/main/resources/send.svg
ingorichtsmeier: Hi @YanaSegal, for this connector it is intended to use the same, original BPMN icon. There is no special technology involved. 

Only the sizing made me to tweak the icon.

connectors-e2e-test/connectors-e2e-test-message/src/test/java/io/camunda/connector/e2e/MessageTests.java
ingorichtsmeier: I don't have an idea yet. I hope that both sides: client implementation and server authorization, may change with further development and future alpha releases. 

It's worth to give it another try.

connectors/pom.xml
```diff
@@ -27,6 +27,7 @@
     <module>jdbc</module>
     <module>kafka</module>
     <module>microsoft-teams</module>
+    <module>message</module>
```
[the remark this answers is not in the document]
ingorichtsmeier: done

connectors/message/src/main/java/io/camunda/connector/message/SendMessageRequest.java
ingorichtsmeier: Done.

connectors-e2e-test/connectors-e2e-test-message/src/test/java/io/camunda/connector/e2e/MessageTests.java
```diff
@@ -0,0 +1,221 @@
...
+    properties = {
+      "spring.main.allow-bean-definition-overriding=true",
+      "camunda.connector.webhook.enabled=false",
+      "camunda.connector.polling.enabled=false"
+    },
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@CamundaSpringProcessTest
+@ExtendWith(MockitoExtension.class)
+@ComponentScan(basePackages = {"io.camunda.connector.runtime.app"})
+public class MessageTests {
```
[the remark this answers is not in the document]
ingorichtsmeier: Done

connectors/message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
ingorichtsmeier: Done. Looks better.

connectors/message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
ingorichtsmeier: Done

ingorichtsmeier:
Hi @sbuettner, @chillleader, I applied your proposed changes. If you agree, you can merge and take over. 

I'm happy to contribute to the docs.

connectors/camunda-message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
```diff
@@ -0,0 +1,147 @@
...
+      @ConnectorElementType(
+          elementType = BpmnType.SEND_TASK,
+          appliesTo = BpmnType.TASK,
+          templateIdOverride = "io.camunda.connectors.message.sendtask.v1")
+    })
+public class SendMessageConnectorFunction implements OutboundConnectorFunction {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SendMessageConnectorFunction.class);
+
+  ZeebeClient zeebeClient;
```
chillleader, line 60: Please migrate to `CamundaClient`: the old zeebe client is flagged as deprecated and should not be used anymore

connectors/camunda-message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
```diff
@@ -0,0 +1,147 @@
...
+      correlateMessageCommandStep3.variables(messageRequest.variables());
+    }
+    if (messageRequest.tenantId() != null) {
+      correlateMessageCommandStep3.tenantId(messageRequest.tenantId());
+    }
+    if (messageRequest.requestTimeout() != null) {
+      correlateMessageCommandStep3.requestTimeout(messageRequest.requestTimeout());
+    }
+    CorrelateMessageResponse correlateMessageResponse = correlateMessageCommandStep3.send().join();
+    return correlateMessageResponse;
```
chillleader, line 145: nit: consider inlining variable

chillleader (changes requested):
Hi @ingorichtsmeier! Thanks for making the changes, this already looks great.

I noticed tests are failing and took a deeper look. It seems that the approach used to get the zeebe client doesn't work correctly: the connector instance is [created](https://github.com/camunda/connectors/blob/93e283efe0b5b6a3240c341e29b329e1abe3f41f/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java#L65) at a point in time when the application context is not set yet in your component. Can you confirm if this is also the case for you when running the connectors bundle?

Our recommendation would be to declare the connector as a spring boot starter and define a custom configuration that provides this connector as a bean, rather than relying on the service loader and the `META-INF/services` definition. In this case, you'd be able to inject the Zeebe client directly into the connector function.

Here is an [example](https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java): the newly created Agentic AI connector that also uses a zeebe client.

Please reach out to me or to the team if you need assistance - happy to take a look together!

Besides that, please switch from using `ZeebeClient` to the new `CamundaClient` - the zeebe client is not deprecated.

ingorichtsmeier:
Hi @chillleader, I've tested my connector using a basic Spring Boot application connecting against a SaaS cluster, as described here: https://github.com/camunda/connectors/tree/main/connector-runtime#run-in-your-spring-boot-project
```
package io.camunda.connector.runtime.app;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.context.annotation.ComponentScan;

@SpringBootApplication
public class ConnectorTestRuntimeApplication {

  public static void main(String[] args) {
    SpringApplication.run(ConnectorTestRuntimeApplication.class, args);
  }
}
```

with this pom file:
```
<project xmlns="http://maven.apache.org/POM/4.0.0"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.camunda.consulting</groupId>
  <artifactId>connector-test-runtime</artifactId>
  <version>0.0.1-SNAPSHOT</version>
  <description>Testing own connectors</description>
  
  <properties>
    <version.connectors>8.8.0-SNAPSHOT</version.connectors>
    <version.connector-message>8.8.0-SNAPSHOT</version.connector-message>
    <maven.compiler.source>21</maven.compiler.source>
    <maven.compiler.target>21</maven.compiler.target>
  </properties>

  <dependencies>
    <dependency>
      <groupId>com.fasterxml.jackson.core</groupId>
      <artifactId>jackson-core</artifactId>
      <version>2.18.2</version>
    </dependency>

    <dependency>
      <groupId>io.camunda.connector</groupId>
      <artifactId>spring-boot-starter-camunda-connectors</artifactId>
      <version>${version.connectors}</version>
    </dependency>
    <dependency>
      <groupId>io.camunda.connector</groupId>
      <artifactId>connector-message</artifactId>
      <version>${version.connector-message}</version>
    </dependency>
  </dependencies>
</project>
```

It works in my IDE (Spring Tool Suite) as expected. The Zeebe client gets created before the connector is created.

What is the difference against the runtime bundle used in the docker container?

I'm still using the Zeebe client to simplify the backporting for Camunda 8.7.1.

connectors/camunda-message/src/main/java/io/camunda/connector/message/SendMessageConnectorFunction.java
ingorichtsmeier: Hi @chillleader, this is a leftover from a logging statement to print the result to the console. The compiler will inline it for us. But adding some log statements later on, will require a bigger refactoring then.

chillleader:
@ingorichtsmeier thanks for sharing more context, I checked it again and the issue was limited to the SaaS connectors bundle. The [Connector runtime spring boot app for SaaS](https://github.com/camunda/connectors/blob/main/bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/SaaSConnectorRuntimeApplication.java) bundle is in a different package, so it wasn't detecting your component. Adding a `@ComponentScan` with an extended base package solved the issue.

I took the liberty of pushing this fix directly into your branch.

> I'm still using the Zeebe client to simplify the backporting for Camunda 8.7.1.

Okay makes sense. I'm fine with merging as is and upgrading in a follow-up PR then.

johnBgood:
# QA information
## Test Environment
  - [ ] SaaS
  - [ ] SM

## Test Scope

## Test Data

chillleader (approved):
Thanks a lot for the contribution @ingorichtsmeier!

We can take care of upgrading to the `CamundaClient` after merging, but please create a follow-up issue, unless you plan to do this yourself 🙂 

I believe @Szik will now take over for QA

Szik (approved):

PR #4202, opened 2025-03-06
fix(rest-connector): Return always an array for set cookie headers

### LINKED ISSUE #4201: Return an array for set cookie header as a default
__Describe the Bug__

When updating from 8.5 to 8.6.0 no longer multiple set cookie headers are returned as an array, but only the first one.
With 8.6.7 it is possible to get all set cookie headers as an array when using the `groupSetCookieHeaders` flag. I just confirmed the output of 8.5 and latest with the flag is the same.

Purpose of this issue:
- 8.7 and 8.8. should return an array per default, without the need of the flag.
- next 8.6.10 / next minor should also return an array per default

| Behavior | Expected result |
| -- | -- |
| Users are on 8.5 and update to 8.6.11 |nothing should break |
| Users are on 8.5 and update to 8.6.7 | will break but should be fixed with manually editing the template |
| Users are on 8.6.7 and did edit their template and update to 8.6.11 | should work fine |
| Users are on 8.6.7 and did not edit their template and update to 8.6.11 | breaking change: if using a set cookie header in a feel result, this will fail, as now an array is returned |
| Users are on 8.6.7 or higher and update to 8.7 | breaking change: if using a set cookie header in a feel result, this will fail, as now an array is returned |

Both potentially breaking changes should be documented.

Support issues:
https://jira.camunda.com/browse/SUPPORT-25027
https://jira.camunda.com/browse/SUPPORT-26056

ztefanie:
## Description
Restore the default behavior of 8.5. Set cookie headers a always merged into an array. The flag for grouping is removed. This also means, if someone sets the group flag to false, this will be ignored after an update, but I think that is ok, as this is not really a helpful feature and I guess noone will do this. 

This will be documented [here](https://github.com/camunda/camunda-docs/pull/5137)

I did not document the "breaking change" in the last two lines of the table, because i consider this a bug fix and not a breaking change. Let me know if you disagree, then I will add it to the docs. (We could adjust the behavior so that if there is only one Set-Cookie header, it returns a simple string instead of an array. However, this approach is not ideal. For example, if an endpoint sometimes returns one cookie (e.g., when login fails) and sometimes two (e.g., when login succeeds), it would make handling cookies in FEEL more complex.)

| Behavior | Expected result |
| -- | -- |
| Users are on 8.5 and update to 8.6.11 |nothing should break |
| Users are on 8.5 and update to 8.6.7 | will break but should be fixed with manually editing the template |
| Users are on 8.6.7 and did edit their template and update to 8.6.11 | should work fine |
| Users are on 8.6.7 and did not edit their template and update to 8.6.11 | breaking change: if using a set cookie header in a feel result, this will fail, as now an array is returned |
| Users are on 8.6.7 or higher and update to 8.7 | breaking change: if using a set cookie header in a feel result, this will fail, as now an array is returned |

## Related issues

closes #4201 

## Checklist

### DIFF
--- a/connectors/http/rest/element-templates/http-json-connector.json
+++ b/connectors/http/rest/element-templates/http-json-connector.json
@@ -333,14 +333,14 @@
     },
     "type" : "Dropdown",
     "choices" : [ {
-      "name" : "DELETE",
-      "value" : "DELETE"
-    }, {
       "name" : "POST",
       "value" : "POST"
     }, {
       "name" : "GET",
       "value" : "GET"
+    }, {
+      "name" : "DELETE",
+      "value" : "DELETE"
     }, {
       "name" : "PATCH",
       "value" : "PATCH"
--- a/connectors/http/rest/element-templates/http-json-connector.json
+++ b/connectors/http/rest/element-templates/http-json-connector.json
@@ -333,14 +333,14 @@
     },
     "type" : "Dropdown",
     "choices" : [ {
+      "name" : "DELETE",
+      "value" : "DELETE"
+    }, {
       "name" : "POST",
       "value" : "POST"
     }, {
       "name" : "GET",
       "value" : "GET"
-    }, {
-      "name" : "DELETE",
-      "value" : "DELETE"
     }, {
       "name" : "PATCH",
       "value" : "PATCH"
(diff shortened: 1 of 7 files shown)


connectors/http/rest/element-templates/http-json-connector.json
```diff
@@ -333,14 +333,14 @@
     },
     "type" : "Dropdown",
     "choices" : [ {
+      "name" : "DELETE",
+      "value" : "DELETE"
+    }, {
```
sbuettner, line 338: Can we move `DELETE` down the list where it was before? We have a nice new order property :)

connectors/http/rest/element-templates/hybrid/http-json-connector-hybrid.json
```diff
@@ -338,14 +338,14 @@
     },
     "type" : "Dropdown",
     "choices" : [ {
+      "name" : "DELETE",
+      "value" : "DELETE"
+    }, {
```
sbuettner, line 343: see above

sbuettner (approved):
Minor comment regarding the dropdown order change

connectors/http/rest/element-templates/http-json-connector.json
ztefanie: Not sure, why this was changed. I reverted it.

PR #4223, opened 2025-03-07
chore(api): increase allowed headers size to 1MB

johnBgood:
## Description

- Increase headers size to 1MB, this is useful when Console calls our API with a huge JWT which might happen

## Checklist


ztefanie (approved):
LGTM

PR #4264, opened 2025-03-10
chore(deps): enable minor version upgrades on release branches

chillleader:
## Description

During the last CVE review, we discovered that Renovate doesn't automatically apply minor version upgrades on feature branches.

Sometimes libraries release fixes for major issues in minor versions instead of patches, so I think it is beneficial if we get minor version upgrades as well.

On the other hand, this might introduce more instability, so I'm open to suggestions/critic regarding this change.

## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -30,8 +30,7 @@
         "/^release/.*/"
       ],
       "matchUpdateTypes": [
-        "major",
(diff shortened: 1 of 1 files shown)


johnBgood (approved):
nice!

PR #4273, opened 2025-03-10
fix(proxy): Authentication username/password must be null if not provided

johnBgood:
## Checklist

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
@@ -22,14 +22,11 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.hc.client5.http.auth.AuthScope;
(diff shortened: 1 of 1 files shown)


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/ProxyHandler.java
```diff
@@ -89,10 +85,9 @@ private Optional<ProxyDetails> getConfigFromEnvVars(String protocol) {
                 Integer.parseInt(
                     System.getenv("CONNECTOR_" + protocol.toUpperCase() + "_PROXY_PORT")),
                 System.getenv()
-                    .getOrDefault("CONNECTOR_" + protocol.toUpperCase() + "_PROXY_USER", ""),
+                    .getOrDefault("CONNECTOR_" + protocol.toUpperCase() + "_PROXY_USER", null),
```
johnBgood, line 88: let me know if this `null` might be an issue somewhere. I don't think so, but wanted to flag this

ztefanie: I think you also need to adjust this in setSystemPropertyIfUnset method

ztefanie:
I tested again, sorry about the confusion yesterday. Everything works as expected :rocket:

ztefanie (approved):
Looks good, thanks for fixing this :100:

PR #4274, opened 2025-03-10
Add skipEndcoding flag for rest and gitlab connector (#3840)

ztefanie:
(cherry picked from commit 0c341672729da9d4e59fc716ec38d47f1331648d)

## Description
Backport the skip encoding feature to 8.6. Feel free to test this will change nothing, if not explicitly updating the element template.

## Checklist

### DIFF
--- a/connectors/http/rest/element-templates/http-json-connector.json
+++ b/connectors/http/rest/element-templates/http-json-connector.json
@@ -4,7 +4,7 @@
   "id" : "io.camunda.connectors.HttpJson.v2",
   "description" : "Invoke REST API",
   "documentationRef" : "https://docs.camunda.io/docs/components/connectors/protocol/rest/",
-  "version" : 8,
+  "version" : 10,
   "category" : {
     "id" : "connectors",
     "name" : "Connectors"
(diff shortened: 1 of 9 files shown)


connectors/http/rest/element-templates/http-json-connector.json
```diff
@@ -4,7 +4,7 @@
   "id" : "io.camunda.connectors.HttpJson.v2",
   "description" : "Invoke REST API",
   "documentationRef" : "https://docs.camunda.io/docs/components/connectors/protocol/rest/",
-  "version" : 8,
+  "version" : 10,
   "category" : {
     "id" : "connectors",
```
chillleader, line 9: I'm a little concerned that version 10 in this branch doesn't fully match version 10 in `release/8.7` (the difference is that the latter also has the `groupSetCookieHeaders` property).

I don't see how we could do this better as version 9 is also different, so maybe this is just something to keep in mind - version X for one release branch is not guaranteed to be the same as version X in another release branch.

chillleader (approved):
LGTM

PR #4295, opened 2025-03-11
other(tests): fix proxy tests, clear properties properly

johnBgood:
## Description

- We need to clear system properties properly, not setting them to empty string as we now handle unset properties.

## Checklist


ztefanie (approved):
LGTM. Thanks for fixing :+1:

PR #4303, opened 2025-03-12
fix(FEEL): Fix NPE for result expressions evaluating to null.

sbuettner:
* fix(FEEL): Fix NPE for result expressions evaluating to null.
* Remove unnecessary log statement in http response handler.

## Checklist


johnBgood (approved):
👍

PR #4307, opened 2025-03-13
feature(secrets): secrets are now hidden from the error message in operate, making error message more reliable

mathias-vandaele:
## Description

PR to hide secret from error  message sent to zeebe

- Secrets are being hidden from error message sent to zeebe
- The object mapper exception is way more precise
- Error from validation are more concise

## Related issues

closes https://github.com/camunda/connectors/issues/3221

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
@@ -24,9 +24,6 @@ import io.camunda.client.api.response.CompleteJobResponse;
 import io.camunda.client.api.response.FailJobResponse;
 import io.camunda.client.api.worker.JobClient;
 import io.camunda.client.api.worker.JobHandler;
-import io.camunda.connector.api.error.ConnectorException;
-import io.camunda.connector.api.error.ConnectorInputException;
-import io.camunda.connector.api.error.ConnectorRetryException;
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 import io.camunda.connector.api.secret.SecretProvider;
 import io.camunda.connector.api.validation.ValidationProvider;
@@ -41,12 +38,9 @@ import io.camunda.connector.runtime.core.outbound.ConnectorResult.SuccessResult;
 import io.camunda.connector.runtime.core.outbound.ErrorExpressionJobContext.ErrorExpressionJob;
 import io.camunda.connector.runtime.core.secret.SecretProviderAggregator;
 import io.camunda.connector.runtime.core.secret.SecretProviderDiscovery;
-import io.camunda.connector.runtime.core.secret.SecretUtil;
 import io.camunda.document.factory.DocumentFactory;
 import java.time.Duration;
 import java.time.format.DateTimeParseException;
-import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import org.slf4j.Logger;
@@ -67,15 +61,21 @@ public class ConnectorJobHandler implements JobHandler {
 
   protected ObjectMapper objectMapper;
 
+  private OutboundConnectorExceptionHandler outboundConnectorExceptionHandler;
+
   /**
    * Create a handler wrapper for the specified connector function.
    *
    * @param call - the connector function to call
+   * @param outboundConnectorExceptionHandler
    */
   public ConnectorJobHandler(
-      final OutboundConnectorFunction call, ValidationProvider validationProvider) {
+      final OutboundConnectorFunction call,
+      ValidationProvider validationProvider,
+      OutboundConnectorExceptionHandler outboundConnectorExceptionHandler) {
     this.call = call;
     this.validationProvider = validationProvider;
+    this.outboundConnectorExceptionHandler = outboundConnectorExceptionHandler;
   }
 
   /**
@@ -94,30 +94,8 @@ public class ConnectorJobHandler implements JobHandler {
     this.validationProvider = validationProvider;
     this.documentFactory = documentFactory;
     this.objectMapper = objectMapper;
-  }
-
-  protected static Map<String, Object> exceptionToMap(Exception wrappedException) {
-    Map<String, Object> result = new HashMap<>();
-    Throwable originalCause = wrappedException.getCause();
-    result.put("type", originalCause.getClass().getName());
-    var message = wrappedException.getMessage();
-    if (message != null) {
-      result.put(
-          "message", message.substring(0, Math.min(message.length(), MAX_ERROR_MESSAGE_LENGTH)));
-    }
-    if (originalCause instanceof ConnectorException connectorException) {
-      var code = connectorException.getErrorCode();
-      var variables = connectorException.getErrorVariables();
-
-      if (code != null) {
-        result.put("code", code);
-      }
-
-      if (variables != null) {
-        result.put("variables", variables);
-      }
-    }
-    return Map.copyOf(result);
+    this.outboundConnectorExceptionHandler =
+        new OutboundConnectorExceptionHandler(getSecretProvider());
   }
 
   protected static FinalCommandStep<CompleteJobResponse> prepareCompleteJobCommand(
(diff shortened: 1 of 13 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -95,13 +97,13 @@ public ConnectorJobHandler(
 
   protected static Map<String, Object> exceptionToMap(Exception exception) {
     Map<String, Object> result = new HashMap<>();
-    result.put("type", exception.getClass().getName());
+    result.put("type", exception.getCause().getClass().getName());
```
sbuettner, line 100: `exception.getCause()` could be null.

mathias-vandaele: @sbuettner It can't be null here, because we are always wrapping the cause exception into the `Exception` that contains the sanitized message

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -95,13 +97,13 @@ public ConnectorJobHandler(
...
     var message = exception.getMessage();
     if (message != null) {
       result.put(
           "message", message.substring(0, Math.min(message.length(), MAX_ERROR_MESSAGE_LENGTH)));
     }
-    if (exception instanceof ConnectorException connectorException) {
+    if (exception.getCause() instanceof ConnectorException connectorException) {
```
sbuettner, line 106: same as above

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -168,16 +170,23 @@ private static String truncateErrorMessage(String message) {
...
   }
 
+  public String hideSecretsFromMessage(String message, List<String> secrets) {
+    return secrets.stream().reduce(message, (s, s2) -> s.replace(s2, "***"));
+  }
+
   @Override
   public void handle(final JobClient client, final ActivatedJob job) {
     LOGGER.info("Received job: {} for tenant: {}", job.getKey(), job.getTenantId());
+    List<String> secrets = SecretUtil.retrieveSecretValues(job.getVariables(), getSecretProvider());
```
sbuettner, line 180: @mathias-vandaele Can we split `SecretUtil.extractSecrets(json)` and 
`SecretProvider.fetchAll()`. The second one can be added to the SecretProvider interface as "default" implementation. This way other implementations dont need to be adapted.

mathias-vandaele: So in the first function we retrieve all the secret key from the json, and then in the default `SecretProvider.fetchAll(List<String> keys)` we would actually retrieve the values ? @sbuettner

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandlerTest.java
```diff
@@ -862,6 +862,27 @@ void shouldCreateBpmnError_UsingExceptionCodeAndErrorVariables() {
...
+      var result =
+          JobBuilder.create()
+              .withVariables("{{secrets.FOO}}")
+              .executeAndCaptureResult(jobHandler, false);
+
+      // then
+      assertThat(result.getErrorMessage())
+          .isEqualTo("Something went wrong: *** is not the correct password");
+    }
+
```
sbuettner, line 885: Please add additional tests that:
* Show an example where we provide a real json object  with the variables
* Show how a validation exception looks like when the value(coming from a secret) is not passing validation and shows up obfuscated in the validation message.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -196,22 +205,27 @@ public void handle(final JobClient client, final ActivatedJob job) {
               job.getCustomHeaders().get(Keywords.RESULT_EXPRESSION_KEYWORD));
       result = new ConnectorResult.SuccessResult(response, responseVariables);
     } catch (ConnectorRetryException ex) {
+      Exception newException = new Exception(hideSecretsFromMessage(ex.getMessage(), secrets), ex);
       LOGGER.debug(
-          "ConnectorRetryException while processing job: {} for tenant: {}",
+          "ConnectorRetryException while processing job: {} for tenant: {}, error message: {}",
           job.getKey(),
           job.getTenantId(),
```
sbuettner, line 212: Would be cool to have all the exception handling code in a single method outside of the already too big: 
```handle``` method.

File: connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java (hunk 3 above)
Comment (sbuettner) [reply connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java:180]: Yes, exactly!

File: connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java (hunk 1 above)
Comment (sbuettner) [reply connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java:100]: But this is not clear in this code here. We assume that the exception we get here has a cause. Would be great to streamline this code a little bit.

File: connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java [hunk 6]
```diff
@@ -104,11 +104,8 @@ private <T> T mapJson(Class<T> cls) {
               .orElse("Unexpected Error, Further investigation is needed");
 
       throw new ConnectorException("JSON_FORMAT_ERROR", errorMessage);
-    } catch (MismatchedInputException e) {
-      throw new ConnectorException("JSON_MISMATCH_ERROR", e.getOriginalMessage());
     } catch (JsonProcessingException e) {
-      throw new ConnectorException(
-          "JSON_PROCESSING_ERROR", "Exception: " + e.getClass().getSimpleName() + " was raised", e);
+      throw new ConnectorException("JSON_PROCESSING_ERROR", e.getOriginalMessage(), e);
```
johnBgood, line 108: not sure, but should it be a ConnectorInputException (not retried)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -168,23 +172,20 @@ private static String truncateErrorMessage(String message) {
...
-    try {
-      retryBackoff = getBackoffDuration(job);
-    } catch (Exception e) {
-      ConnectorResult.ErrorResult result =
-          new ConnectorResult.ErrorResult(Map.of("error", exceptionToMap(e)), e, 0);
-      failJob(client, job, result);
-      return;
-    }
+    List<String> secrets =
+        getSecretProvider().fetchAll(SecretUtil.retrieveSecretKeysInInput(job.getVariables()));
```
johnBgood, line 183: are we happy with `getVariables`? Any gain to use `getVariablesAsMap` for instance ?

johnBgood (commented):
Few questions :)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
mathias-vandaele: `getvariables()` is the best way to do, we already have a regex ready and that is how it is done already to retrieve secret

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
mathias-vandaele: done

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -250,12 +234,66 @@ public void handle(final JobClient client, final ActivatedJob job) {
...
-      // failure while parsing the error expression
-      failJob(client, job, new ErrorResult(Map.of("error", exceptionToMap(ex)), ex, 0));
+      failJob(
+          client,
+          job,
+          new ErrorResult(Map.of("error", exceptionToMap(newException)), newException, 0));
     }
   }
 
+  private ConnectorResult handleGenericException(
```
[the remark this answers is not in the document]
mathias-vandaele: I extracted this logic from the `handle` function which now return a `connectorResult` even cleaner I believe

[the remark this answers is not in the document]
johnBgood: way better indeed 👍 I still think we should have a separate class for this error handling though. 
Maybe it's out of scope for this PR, up to you 👌

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -194,38 +198,18 @@ public void handle(final JobClient client, final ActivatedJob job) {
...
-      result = handleSDKException(job, ex, retries, errorCode, retryBackoff);
+      return handleGenericException(job, ex, secrets, retryBackoff);
     }
+  }
 
+  private void processFinalResult(
+      JobClient client, ActivatedJob job, ConnectorResult result, List<String> secrets) {
     try {
       final ConnectorResult finalResult = result;
       ConnectorHelper.examineErrorExpression(
```
johnBgood, line 215: this could be in the new component, for instance, it doesn't make sense in the Helper as it's used only there.
I'm not sure if we could also simplify the code below, as it's quite confusing, and we always have to spend some time reading it :/ Might be out of scope for this PR again.

Szik (approved):

johnBgood (approved):

PR #4320, opened 2025-03-14
refactor(rest): refactor the proxy support for the REST Connector

johnBgood:
## Description

This PR changes the previous behavior about how we configure a proxy (and non-proxy hosts).
- We can now set up a proxy for REST-like connectors **only** (REST, Github, etc). This is done using a specific set of **environment variables**: 

| Env. Var | Description | Default |
|--------|--------|--------|
| CONNECTOR_HTTP(S)_PROXY_SCHEME | The proxy **scheme** (`http`, or `https`). This allows the user to use https proxies which is not possible with other clients | `http` |
| CONNECTOR_HTTP(S)_PROXY_HOST | The **host** part of the proxy URL (do not include `http://`)  | |
| CONNECTOR_HTTP(S)_PROXY_PORT | The **port** part of the proxy URL | |
| CONNECTOR_HTTP(S)_PROXY_USER | If you need to authenticate, use this variable to set the **user** | |
| CONNECTOR_HTTP(S)_PROXY_PASSWORD | If you need to authenticate, use this variable to set the **password** | | 
| CONNECTOR_HTTP_NON_PROXY_HOSTS | Similar to the [Java](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) `http.nonProxyHosts` system property | 

- Note that these values are **not copied** to the System Properties anymore.
- If you want a global behavior, don't use these variables but use System Properties instead.

## Checklist

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
@@ -36,15 +36,15 @@ public class CustomApacheHttpClient implements HttpClient {
    * Converts the given {@link HttpCommonRequest} to an Apache {@link
    * org.apache.hc.core5.http.ClassicHttpRequest} and executes it.
    *
+   * <p>This method is thread-safe.
+   *
    * @param request the request to execute
    * @param executionEnvironment the {@link ExecutionEnvironment} we are in
    * @return the {@link HttpCommonResult}
    */
   @Override
   public HttpCommonResult execute(
-      HttpCommonRequest request,
-      ProxyHandler proxyHandler,
-      @Nullable ExecutionEnvironment executionEnvironment) {
+      HttpCommonRequest request, @Nullable ExecutionEnvironment executionEnvironment) {
     try {
       var apacheRequest = ApacheRequestFactory.get().createHttpRequest(request);
       var host = apacheRequest.getAuthority().getHostName();
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
@@ -21,7 +21,6 @@ import io.camunda.connector.http.base.ExecutionEnvironment;
 import io.camunda.connector.http.base.client.HttpClient;
 import io.camunda.connector.http.base.client.HttpStatusHelper;
 import io.camunda.connector.http.base.client.apache.proxy.ProxyAwareHttpClient;
-import io.camunda.connector.http.base.client.apache.proxy.ProxyHandler;
 import io.camunda.connector.http.base.exception.ConnectorExceptionMapper;
 import io.camunda.connector.http.base.model.HttpCommonRequest;
 import io.camunda.connector.http.base.model.HttpCommonResult;
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
@@ -20,71 +20,17 @@ import io.camunda.connector.api.error.ConnectorException;
 import io.camunda.connector.http.base.ExecutionEnvironment;
 import io.camunda.connector.http.base.client.HttpClient;
 import io.camunda.connector.http.base.client.HttpStatusHelper;
+import io.camunda.connector.http.base.client.apache.proxy.ProxyAwareHttpClient;
+import io.camunda.connector.http.base.client.apache.proxy.ProxyHandler;
 import io.camunda.connector.http.base.exception.ConnectorExceptionMapper;
 import io.camunda.connector.http.base.model.HttpCommonRequest;
 import io.camunda.connector.http.base.model.HttpCommonResult;
 import java.io.IOException;
 import javax.annotation.Nullable;
 import org.apache.hc.client5.http.ClientProtocolException;
-import org.apache.hc.client5.http.config.RequestConfig;
-import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
-import org.apache.hc.client5.http.impl.classic.HttpClients;
-import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
 import org.apache.hc.core5.http.HttpStatus;
-import org.apache.hc.core5.http.io.SocketConfig;
-import org.apache.hc.core5.util.Timeout;
 
 public class CustomApacheHttpClient implements HttpClient {
-
-  public static final CustomApacheHttpClient DEFAULT =
-      new CustomApacheHttpClient(createHttpClientBuilder());
-
-  private final HttpClientBuilder httpClientBuilder;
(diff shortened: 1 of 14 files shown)


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/ApacheRequestAuthenticationBuilder.java
```diff
@@ -63,7 +63,7 @@ public void build(ClassicRequestBuilder builder, HttpCommonRequest request) {
 
   String fetchOAuthToken(OAuthAuthentication authentication) {
     HttpCommonRequest oAuthRequest = oAuthService.createOAuthRequestFrom(authentication);
-    HttpCommonResult response = CustomApacheHttpClient.getDefault().execute(oAuthRequest);
+    HttpCommonResult response = new CustomApacheHttpClient().execute(oAuthRequest);
```
ztefanie, line 66: Could we create two clients (one for using a proxy and one for non-proxy) and then select the correct one for a request instead of creating new ones for every request?

ztefanie (commented):
Feel free to re-add these tests for matching non proxy hosts properly:
https://github.com/camunda/connectors/pull/3894/files#diff-0d190eecc67fcce2f29db41ca7e5fbe821ff94693ab1ab045393c42e148abb91R91-R126

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/ApacheRequestAuthenticationBuilder.java
johnBgood: Sadly not really, the timeouts are configured when creating the client and they are based on the request values :/ 
```java
  private RequestConfig getRequestTimeoutConfig(TimeoutConfiguration timeoutConfiguration) {
    return RequestConfig.custom()
        .setConnectionRequestTimeout(
            Timeout.ofSeconds(timeoutConfiguration.connectionTimeoutInSeconds()))
        .setResponseTimeout(Timeout.ofSeconds(timeoutConfiguration.readTimeoutInSeconds()))
        .build();
  }
```
If you find another way, I'll be more than happy though

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
```diff
@@ -20,103 +20,49 @@
...
-    connectionManager.setDefaultSocketConfig(SocketConfig.custom().setSoKeepAlive(true).build());
-
-    return connectionManager;
-  }
-
   /**
    * Converts the given {@link HttpCommonRequest} to an Apache {@link
    * org.apache.hc.core5.http.ClassicHttpRequest} and executes it.
    *
+   * <p>This method is thread-safe.
```
ztefanie, line 38: is this `<p>` tag ever closed?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
```diff
@@ -20,103 +20,49 @@
...
-              .build()
-              .execute(
-                  apacheRequest,
-                  new HttpCommonResultResponseHandler(
-                      executionEnvironment, request.isStoreResponse()));
-      if (HttpStatusHelper.isError(result.status())) {
-        throw ConnectorExceptionMapper.from(result);
+      var host = apacheRequest.getAuthority().getHostName();
+      var scheme = apacheRequest.getScheme();
+      try (var client =
```
ztefanie, line 51: Nice :rocket:

ztefanie (commented):
Thanks for this PR and fixing the proxy issues :green_heart: 

I reviewed it, the code looks good :+1: 

I tested this, but the nonProxyHost setting gets ignored for me
These are my settings `CONNECTOR_HTTP_PROXY_HOST=localhost;CONNECTOR_HTTP_PROXY_PORT=3128;CONNECTOR_HTTP_PROXY_NON_PROXY_HOSTS=www.gmx.de;CONNECTOR_HTTP_PROXY_USER=my-user;CONNECTOR_HTTP_PROXY_PASSWORD=demo` but requests to http://www.gmx.de still get proxied.

Please also do not forget to adjust the documentation, plus I think we need to mention that we introduced breaking changes, because now a setup with proxy settings with system variables will no longer work, right?

Please also do not forget to backport this to 8.7

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/ApacheRequestAuthenticationBuilder.java
sbuettner: @johnBgood Curious, can this be overriden on a request level like this:
```
final RequestConfig params = RequestConfig.custom().setConnectTimeout(3000).setSocketTimeout(3000).build();
httpGet.setConfig(params);
```
?

sbuettner: It seems like the proxy config can also be passed via the request config: https://hc.apache.org/httpcomponents-client-4.5.x/current/httpclient/apidocs/org/apache/http/client/config/RequestConfig.html

johnBgood: > @johnBgood Curious, can this be overriden on a request level like this:

Yes, but this is not how we create requests (we don't use HttpGet, HttpPost etc). This is because it induced other limitations or less generic code I don't remember exactly

sbuettner: Ok, should we maybe tackle it in a separate PR? As our Connectors are essentially request based we could potentially benefit from a more request focused configuration.

johnBgood: Yes we can have a look in another PR, I took a look and it's probably not that many changes

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java
johnBgood: no need to, it's usually not closed in Javadoc

ztefanie (approved):
Tested again, works as expected :rocket:

johnBgood:
# QA information
## Test Environment
  - [ ] SaaS
  - [ ] SM

## Test Scope

## Test Data

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/ApacheRequestAuthenticationBuilder.java
sbuettner: Created a follow up issue here: https://github.com/camunda/connectors/issues/4371

Szik (approved):

ztefanie (approved):

PR #4363, opened 2025-03-18
chore(ci): nightly e2e test slack notifications

chillleader:
## Description

This PR fixes the `SlowTest` condition - previously it never worked to run the tests if the flag is not present.

Additionally, I configured Slack notifications for test runs, and introduced a matrix to run e2e tests across all release branches too, not only main.

## Checklist


sbuettner (approved):
Nice!

PR #4413, opened 2025-03-26
chore(renovate.json): Use glob matching

sbuettner:
https://docs.renovatebot.com/string-pattern-matching/#glob-matching

## Related issues

closes #

## Checklist


ztefanie (approved):
Looks good. Thanks for fixing. I think the PR should be prefixed chore not fix, as we do not need this in the release notes as it does not affect customers and never was a "bug" visible to users.

PR #4419, opened 2025-03-26
fix(inbound-connector): do not apply number and boolean to inbound connector

mathias-vandaele:
## Description

ETG has a filter now

It will not apply `Number` and `Boolean` to inbound connector template

## Checklist

### DIFF
--- a/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
+++ b/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
@@ -183,7 +183,7 @@ public class TemplatePropertiesUtil {
     }
 
     PropertyBuilder propertyBuilder =
-        createPropertyBuilder(field, annotation)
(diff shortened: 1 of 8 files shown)


element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
```diff
@@ -291,7 +292,7 @@ private static PropertyBuilder createPropertyBuilder(Field field, TemplateProper
       type = PropertyType.Dropdown;
       Class<?> enumType = field.getType();
       dropdownChoices = createDropdownModelList(enumType);
-    } else if (isNumberClass(field.getType())) {
+    } else if (isNumberClass(field.getType()) && isOutbound(context)) { // To be removed
```
sbuettner, line 295:
```suggestion
    } else if (isNumberClass(field.getType()) && isOutbound(context)) { // TODO: To be removed
```

sbuettner (approved):

PR #4546, opened 2025-04-02
fix(json): Enhance json binding for single element arrays

### LINKED ISSUE #4538: Support binding single element arrays to objects

The form upload components of Tasklist always returns an array. To avoid users to extract the first element of an array `documents` like this: `documents[1]` using a FEEL expression we want to allow customers to simply pass the array `documents` and allow binding it to a single object `document` for example on the Connectors side.

Our jackson mapper should supporting binding arrays to single objects if the array only contains a single element.

Users can do the same by using the one-based array access of FEEL `myArray[1]` but this is inconvenient and error prone.

__Additional context__

https://camunda.slack.com/archives/C07CTRS99KM/p1743432209287149

sbuettner:
## Related issues

closes #4538

## Checklist


johnBgood (approved):

connectors/google/google-gemini/src/test/java/io/camunda/connector/gemini/mapper/PromptsMapperTest.java
```diff
@@ -56,16 +56,6 @@ void mapNullEntryShouldThrowEx() throws Exception {
...
-  @Test
-  void mapWithWrongValueTypeShouldThrowEx() throws Exception {
-    List<Object> input =
-        readValue("src/test/resources/prompts_with_wrong_value_type.json", List.class);
-
-    IllegalArgumentException ex =
-        assertThrows(IllegalArgumentException.class, () -> promptsMapper.map(input));
-    assertThat(ex).hasMessage(INVALID_PROMPT_MSG_FORMAT.formatted(Map.of(TEXT_KEY, List.of(1))));
-  }
-
```
chillleader, line 68: Should we keep this test case but use a different type of input data?

chillleader (approved):
LGTM

connectors/google/google-gemini/src/test/java/io/camunda/connector/gemini/mapper/PromptsMapperTest.java
sbuettner: Good question, I removed it as it tested something that is supported work now. There are additional tests that cover other illegal prompt input so I think its fine to just remove this.

Issue #4437, opened 2025-03-27
Provide access to the process definition XML or the CamundaClient from a connector implementation


maff:
__Is your feature request related to a problem? Please describe.__

We have the challenge that we need to access the process definition XML as part of an outbound connector. As the current interface of an outbound connector does not provide support for accessing existing beans (e.g. the `CamundaClient`) we can't use the C8 or Operate APIs without additional work/configuration.

__Describe the solution you'd like__

There are multiple ways which could make this possible for our use case, but which need consideration in terms of the API exposed to connectors. A couple of ideas:

- (very open): add DI/container support to individual connectors to allow autowiring needed dependencies via constructor injection as for normal Spring beans
- (pretty specific and might lead to follow-ups for additional use cases): create something like a `CamundaClientAware` interface exposing a `setCamundaClient(CamundaClient client)` method which can be implemented by a connector and which will be handled when creating the connector instance
- (specific to our use case): expose something like a `getProcessDefinitionXml()` method on the `JobContext` which would delegate the call to either the actual client or could make use of something similar to the `ProcessStateStore` to cache the information by process definition key

__Describe alternatives you've considered__

We might be able to not depend on the definition XML in the future.

__Additional context__

Discussed this with @sbuettner and @chillleader this morning.

sbuettner:
@saig0 Do you know whether the "adhoc subprocess worker" solution will come with the required process information as part of the job variables or not? Would be good to know before we invest time in this topic.

sbuettner:
@maff Just want to share an example by @ingorichtsmeier how one could expose a client by providing a static helper method:
https://github.com/camunda/connectors/blob/5a97c067fc49363ff1f51513cbc3b278c333a561/connectors/message/src/main/java/io/camunda/connector/runtime/app/ZeebeClientContext.java

This is just an intermediate solution (and will be changed) until we have a proper way forward on how Connectors can make use of Camunda client capabilities.

maff:
@sbuettner thank you, that definitely helps to unblock us for the moment! 🙏

saig0:
> Do you know whether the "adhoc subprocess worker" solution will come with the required process information as part of the job variables or not? 

We don't know yet. We need to decide if passing all request data as variable or custom headers is feasible.

sbuettner:
Thanks @saig0. We would like to wait until you decide on that before offering process definition access via the Connectors Runtime.

@maff Would waiting for you until we have this decision work for you?

maff:
[...]

PR #4548, opened 2025-04-02
feat(agentic-ai): foundational agentic AI & ad-hoc subprocess tools schema connectors

### LINKED ISSUE #4545: Ad-hoc subprocess tool schema connector

When designing LLM-based agentic AI use cases with tooling support it is crucial to supply a schema of each tool to the LLM. Typically this is name and description of an individual tool combined with a definition of input parameters. An example on the [Claude documentation](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#example-simple-tool-definition) outlines this for a `get_weather` tool accepting a location and a unit.

In Camunda terms, the recently added support for ad-hoc subprocesses (AHSP) allows to specify the "toolbox" a LLM can use by defining individual tools as activities within the AHSP. To avoid having to specify the schema manually when calling the LLM (especially when adding/changing tools), there is the need for an automatic tools/schema detection based on the activities defined within the AHSP.

~On the modeler side, it is currently planned to add support to define the input schema as a new extension property `camunda:adHocActivityInputSchema` in https://github.com/camunda/camunda-modeler/issues/4961.~

Update: based on recent discussions, we will use the `documentation` field for a first iteration to avoid dependencies on modeler changes. If the documentation field contains a valid JSON data structure in the following format it will be used as input schema:

```json
{
  "description": "Loads a user by ID",
  "inputSchema": {
    "type": "object",
    "properties": {
      "id": {
        "type": "number",
        "description": "The user ID"
      }
    },
    "required": [
      "id"
    ]
  }
}
```

If no JSON structure is present, the documentation will be used as description without defined input properties.

- a new outbound connector accepts the ad-hoc subprocess ID as configuration parameter
- when called, the connector reads the process definition and returns a list of tools + input parameters as its result
- the output format should be generic enough to be mapped into specific LLM formats as they slightly differ
- this result can be used by LLM connectors to provide a tool schema

### LINKED ISSUE #4547: Agentic AI connector (tool calling, history based on process variables)

While creating/modeling an agentic AI process is already possible today by leveraging existing LLM connectors (and will be easier when being able to automate a tool definition schema via #4545), it is still complex to create a fully fledged agent with a proper feedback loop, history, tool execution etc.

The goal of this issue is to provide a basic AI agent connector which does the heavy lifting of the feedback loop and tool execution:

- supports multiple LLM model providers
- allows to configure prompts and parameters needed to interact with the LLM
- handles conversation/short term memory (first implementation based on process variables)
- handles tool calling based on the schema definition implemented in #4545
  - providing tool definitions to LLMs + handling tool call requests/results

At later stages this can be enriched with more flexible configuration (e.g. different history storages, RAG, ...).

maff:
## Description

> [!NOTE]  
> The connectors implemented in this PR should be considered as experimental as they will be subject of further change.

This PR introduces a new set of outbound connectors acting as foundational building blocks for (agentic) AI integrations.

- An **ad-hoc tools schema** connector can resolve the "tools" defined in an ad-hoc subprocess to a list of schemas (based on JSON schema). See https://github.com/camunda/connectors/issues/4545
- An **AI agent** connector uses the functionality of the ad-hoc tools schema + handles a conversation flow with a LLM including tool calling and memory handling (currenty via process variables). This will be followed up with integration for documents (RAG) and more flexbile history handling. This connector uses Langchain4j under the hood to provide integrations with different providers and models. See https://github.com/camunda/connectors/issues/4547

### Implementation notes

- As the schema resolver acts on the process definition, it fetches the XML using the `CamundaClient`. To facilitate wiring of components and access to the `CamundaClient`, the new connectors are defined as Spring beans in contrast to the other core connectors using the SPI facility.
  - This means, the agentic AI connectors cannot be controlled using the `CONNECTOR_*_FUNCTION` env variables. Instead a `camunda.connector.agenticai.enabled` property can be set to true to disable the whole agentic AI connector integration. All beans provided by this integration are defined on `AgenticAiConnectorsAutoConfiguration` and can be switched off with the property.
- To avoid loading the process definition on every process instance, the resolved tool schema is cached when `camunda.connector.agenticai.cache.enabled` is set to `true` (default). The cache has a limited size and will auto-evict elements a certain duration. Both size and time can be adapted via config.
- The process definition loading/caching is most likely not used when integrating this on the ad-hoc subprocess itself (later), but it will still make sense to keep the schema resolver and the service task based connector for more flexible use cases.
- ❓ As the AI agent connector builds on top of the ad-hoc tools schema connector, both are currently implemented in the same Maven module. If we'd like to follow the one connector per module approach it would also be possible to split into multiple modules & have a Spring autoconfiguration module wiring everything together.

## Related issues

closes #4545 
closes #4547 

## Checklist


parent/pom.xml
```diff
@@ -111,8 +111,8 @@ limitations under the License.</license.inlineheader>
     <version.gson-extras>3.1.0</version.gson-extras>
 
     <version.httpcore>4.4.16</version.httpcore>
-    <version.httpcore5>5.3.4</version.httpcore5>
-    <version.httpclient5>5.4.3</version.httpclient5>
+    <version.httpcore5>5.3.3</version.httpcore5>
+    <version.httpclient5>5.4.2</version.httpclient5>
```
maff, line 115: This is temporary until #4519 is fixed.

johnBgood (commented):
I know  it's just a draft, so please feel free to ignore my comments if they don't apply in the future :)

connectors/agentic-ai/agentic-ai-connectors/src/main/java/io/camunda/connector/agenticai/core/AgentsApplicationContextHolder.java
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.core;
+
+public class AgentsApplicationContextHolder {
```
johnBgood, line 9: Could you try creating your Connector as a bean?

```java
@Component
public class TestComp {
  CamundaClient camundaClient;

  public Testcomp(CamundaClient camundaClient) {
    this.camundaClient = camundaClient;
  }
}
```

You can then merge the 2 projects (the connector and the spring project), and there's no need for the static context anymore.
You will be able to inject all required beans (objectMapper, CamundaClient, etc).

Your `@Component` (which will also be annotated with `@OutboundConnector`) will then be picked-up by the [OutboundConnectorAnnotationProcessor](https://github.com/camunda/connectors/blob/03689d65b3c8c95640235af18e0c7b42c5b72569/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/lifecycle/OutboundConnectorAnnotationProcessor.java#L62).

Can you try this approach and let me know if it worked?

maff: I was not aware of the bean support for connectors, thanks for that hint, this makes the handling way easier 💡 I'll update the PR accordingly.

connectors/agentic-ai/agentic-ai-connectors/src/main/java/io/camunda/connector/agenticai/aiagent/AiAgentFunction.java
```diff
@@ -0,0 +1,200 @@
...
+      @PropertyGroup(id = "parameters", label = "Parameters"),
+      @PropertyGroup(id = "context", label = "Context"),
+      @PropertyGroup(id = "prompt", label = "Prompt"),
+      @PropertyGroup(id = "tools", label = "Tools"),
+      @PropertyGroup(id = "history", label = "History"),
+      @PropertyGroup(id = "guardrails", label = "Guardrails"),
+    },
+    documentationRef = "https://example.com",
+    icon = "aiagent.svg")
+public class AiAgentFunction implements OutboundConnectorFunction {
```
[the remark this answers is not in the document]
maff: The main reason to have this one level deeper is that this module exposes 2 connectors:

- The ad-hoc tools schema connector as a primitive to resolve tools from an ad-hoc subprocess
- The actual AI agent connector which does the LLM orchestration and the tool calling. This one relies on the implementation of the other connector.

I tried to move things around to be a bit more consistent within the packages.

Would you prefer to have 2 separate modules (1 per connector), and potentially a third one for the Spring configuration?

connectors/agentic-ai/agentic-ai-connectors/src/main/java/io/camunda/connector/agenticai/aiagent/converter/AgentContextMessageSerializer.java
```diff
@@ -0,0 +1,64 @@
...
+import io.camunda.connector.api.error.ConnectorException;
+import java.util.List;
+import java.util.Map;
+
+public class AgentContextMessageSerializer {
+
+  private final ObjectMapper objectMapper;
+
+  public AgentContextMessageSerializer(ObjectMapper objectMapper) {
+    this.objectMapper = objectMapper;
```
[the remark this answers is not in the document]
maff: At the moment we don't - we just use Jackson to serialize the LLM messages as variables, but we'll add document support in a later iteration. I'll have a look at the DeSer support when we touch this as we might need a subset.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.springframework.test.context.bean.override.mockito.MockitoBean;
+
+public class AiAgentTests extends BaseAgenticAiTest {
```
maff, line 69: This test uses a prebuilt process to test the feedback loop of an AI agent using a mocked LLM model. This test only uses the lower part of the model, the upper part (schema resolving) is used from `AdHocToolsSchemaTests`.

It tests the basic interactions with the agent:

- A request can be fulfilled without additional user feedback (no loop)
- A request can be fulfilled with user feedback (loops back into the agent and refines the response based on feedback)
- A request can be fulfilled with tool support (tools from the ad-hoc subprocess are called to fulfill the request) + user feedback
- Guardrails (max model calls) are checked and raise an incident when exceeding configured values

image

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+            .handler(
+                (client, job) -> {
+                  jobWorkerCounter.incrementAndGet();
+                  client
+                      .newCompleteCommand(job.getKey())
+                      .variables(userFeedbackVariables.get())
+                      .send()
+                      .join();
+                })
+            .open();
```
maff, line 100: I struggled to complete the service task for "user feedback" within the test case/via API, so I created this worker which sets variables based on what is set in the `mockChatInteractions` setup. Happy to change this to a more concise method if possible.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+      final var initialUserPrompt = "Write a haiku about the sea";
+      final var expectedConversation =
+          List.of(
+              new SystemMessage(
+                  "You are a helpful AI assistant. Answer all the questions, but always be nice. Explain your thinking."),
+              new UserMessage(initialUserPrompt),
+              new AiMessage(
+                  "Endless waves whisper | moonlight dances on the tide | secrets drift below."));
+
+      mockChatInteractions(
```
maff, line 124: "chat interactions" is the mocked setup consisting of the expected AI response + the variables to return from the user feedback task. This list is used as a queue to return different responses as the test proceeds.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+                  .metadata(
+                      ChatResponseMetadata.builder()
+                          .finishReason(FinishReason.STOP)
+                          .tokenUsage(new TokenUsage(10, 20))
+                          .build())
+                  .aiMessage(
+                      new AiMessage(
+                          "Endless waves whisper | moonlight dances on the tide | secrets drift below."))
+                  .build(),
+              userSatisfiedFeedback()));
```
maff, line 136: User satisfied feeback = exit the loop and complete the process.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+                  .metadata(
+                      ChatResponseMetadata.builder()
+                          .finishReason(FinishReason.STOP)
+                          .tokenUsage(new TokenUsage(10, 20))
+                          .build())
+                  .aiMessage(
+                      new AiMessage(
+                          "Endless waves whisper | moonlight dances on the tide | secrets drift below."))
+                  .build(),
+              userFollowUpFeedback("Add emojis!")),
```
maff, line 179: User follow-up feedback = loop back into the agent with an updated request.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+                          .arguments("{\"a\": 5, \"b\": 3}")
+                          .build(),
+                      ToolExecutionRequest.builder()
+                          .id("bbb222")
+                          .name("Task_With_Text_Input_Schema")
+                          .arguments("{\"searchQuery\": \"Where does this data come from?\"}")
+                          .build())),
+              new ToolExecutionResultMessage("aaa111", "SuperfluxProduct", "24"),
+              new ToolExecutionResultMessage(
+                  "bbb222", "Task_With_Text_Input_Schema", "from follow-up task"),
```
maff, line 235: The results are modeled in the process (the "tools" are script tasks).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -0,0 +1,439 @@
...
+      assertThat(chatMemoryMessages).filteredOn(msg -> msg instanceof UserMessage).hasSize(10);
+    }
+  }
+
+  private void mockChatInteractions(ChatInteraction... chatInteractions) {
+    final var queue = new ArrayList<>(Arrays.asList(chatInteractions));
+    doAnswer(
+            invocationOnMock -> {
+              final var interaction = queue.removeFirst();
+              userFeedbackVariables.set(interaction.userFeedback());
```
maff, line 340: Updates the next job worker variables.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/expected-schema-result.json
maff, line 1: This is the expected schema resolved from the tasks modeled in the ad-hoc subprocess.

connectors-e2e-test/connectors-e2e-test-base/src/main/java/io/camunda/connector/e2e/ZeebeTest.java
```diff
@@ -53,15 +56,21 @@ public ZeebeTest deploy(BpmnModelInstance bpmnModelInstance) {
   }
 
   public ZeebeTest createInstance() {
+    return createInstance(Collections.emptyMap());
+  }
+
+  public ZeebeTest createInstance(Map<String, Object> variables) {
```
maff, line 62: Adds support to create an instance with variables.

connectors/agentic-ai/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+            <connector>
+              <connectorClass>io.camunda.connector.agenticai.adhoctoolsschema.AdHocToolsSchemaFunction</connectorClass>
+              <files>
+                <file>
+                  <templateId>io.camunda.connectors.agenticai.adhoctoolsschema.v0</templateId>
+                  <templateFileName>agenticai-adhoctoolsschema-outbound-connector.json</templateFileName>
+                </file>
+              </files>
+              <generateHybridTemplates>true</generateHybridTemplates>
+              <writeMetaInfFileGeneration>false</writeMetaInfFileGeneration>
```
maff, line 133: No META-INF file as we register the connectors as Spring beans.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CachingAdHocToolsSchemaResolver.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package io.camunda.connector.agenticai.adhoctoolsschema.resolver;
+
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
+import io.camunda.connector.agenticai.adhoctoolsschema.model.AdHocToolsSchemaResponse;
+import java.time.Duration;
+import java.util.Optional;
+
+public class CachingAdHocToolsSchemaResolver implements AdHocToolsSchemaResolver {
```
maff, line 15: Simple caching decorator holding resolve tool definition in memory - size and retention time is configurable and uses very small defaults, so I think we're not risking filling up memory.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CamundaClientAdHocToolsSchemaResolver.java
```diff
@@ -0,0 +1,134 @@
...
+import org.camunda.bpm.model.xml.instance.ModelElementInstance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CamundaClientAdHocToolsSchemaResolver implements AdHocToolsSchemaResolver {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(CamundaClientAdHocToolsSchemaResolver.class);
+
+  private static final String INPUT_SCHEMA_PROPERTY_NAME = "camunda:adHocActivityInputSchema";
```
maff, line 38: This is most likely subject to change as soon as we're able to decide to handle this best on the modeler side.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CamundaClientAdHocToolsSchemaResolver.java
```diff
@@ -0,0 +1,134 @@
...
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeProperty;
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+import org.apache.commons.lang3.StringUtils;
+import org.camunda.bpm.model.xml.instance.ModelElementInstance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CamundaClientAdHocToolsSchemaResolver implements AdHocToolsSchemaResolver {
```
maff, line 33: This is the supporting block of auto-resolving the available "tools" in an ad-hoc subprocess. Exposed as dedicated connector to interact with existing LLM connectors and directly called from the AI agent connector.

The basic logic is - given an ad-hoc subprocess ID:

- read the process definition XML
- extract all the "root elements" (no incoming flows) from the ad-hoc subprocess (activities, intermediate events)
- create a tool definition based on name, description and an optional input JSON schema which can be passed to an LLM an a tool calling setup ([Anthropic example](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview))

The JSON schema is currently managed in a custom extension property, but how the final definition will look like is subject to change as there are ongoing discussions.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CamundaClientAdHocToolsSchemaResolver.java
```diff
@@ -0,0 +1,134 @@
...
+        .orElseGet(
+            () -> new AdHocToolDefinition(element.getId(), documentation, JsonSchema.empty()));
+  }
+
+  private AdHocToolDefinition toolDefinitionFromInputSchema(
+      final FlowNode element, final String documentation, final String inputSchema) {
+
+    var inputSchemaJson = inputSchema;
+    if (inputSchemaJson.startsWith("=")) {
+      inputSchemaJson = feelEngineWrapper.evaluateToJson(inputSchemaJson);
```
maff, line 97: Basic FEEL support without external variables allows to construct the schema via an expression (e.g. looping over a list).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+      ToolCallingHandler toolCallingHandler) {
+    this.objectMapper = objectMapper;
+    this.chatModelFactory = chatModelFactory;
+    this.toolCallingHandler = toolCallingHandler;
+  }
+
+  @Override
+  public AgentResponse handleRequest(OutboundConnectorContext context, AgentRequest request) {
+    final AgentRequest.AgentRequestData requestData = request.data();
+    final AgentContext agentContext =
```
maff, line 52: The `AgentContext` is important as it allows agent re-entry. It contains the needed data for short-term memory (chat conversation) and for guardrails checks (amount of calls ...).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        Optional.ofNullable(requestData.context()).orElseGet(AgentContext::empty);
+
+    // initialize configured model
+    final ChatLanguageModel chatModel = chatModelFactory.createChatModel(request);
+
+    // set up memory and load from context if available
+    final AgentContextChatMemoryStore chatMemoryStore =
```
maff, line 59: Currently, we store the conversation short-term memory in a variable. In the future we might need to extend this to different storage backends to avoid running in to variable size limits.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        new AgentContextChatMemoryStore(objectMapper);
+    chatMemoryStore.loadFromAgentContext(agentContext);
+
+    final ChatMemory chatMemory =
+        MessageWindowChatMemory.builder()
```
maff, line 64: Keeps messages in chat memory according to the configured message window size. When it hits the limits if will first start to evict tool call requests + results and then continue evicting the oldest messages.

If we want to keep the whole conversation history (e.g. for RAG) we need to make sure to also keep track of that (will be a follow-up).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+
+    // check configured guardrails
+    checkGuardrails(requestData, agentContext);
+
+    // update memory with system + new user messages/tool call responses
+    addSystemPromptIfNecessary(chatMemory, requestData);
+    addUserMessagesFromRequest(agentContext, chatMemory, requestData);
+
+    // call LLM API with updates messages + resolved tool specifications
+    final var toolSpecifications =
```
maff, line 79: This is the output of the AdHocToolsSchemaResolver, mapped to the Langchain4j tool specification format.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        toolCallingHandler.loadToolSpecifications(
+            context.getJobContext().getProcessDefinitionKey(),
+            requestData.tools().containerElementId());
+
+    final var chatRequest =
+        ChatRequest.builder()
+            .messages(chatMemory.messages())
```
maff, line 86: The chat request always contains the whole message/conversation window to allow follow-up requests (every call to the LLM will contain a bigger list of messages).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+            .toolSpecifications(toolSpecifications)
+            .build();
+
+    final ChatResponse chatResponse = chatModel.chat(chatRequest);
+    final AiMessage aiMessage = chatResponse.aiMessage();
+    chatMemory.add(aiMessage);
+
+    // extract tool call requests from LLM response
+    final var toolsToCall = toolCallingHandler.extractToolsToCall(toolSpecifications, aiMessage);
+    final var nextAgentState =
```
maff, line 96: Very simple "state machine" to determine if we need to inject a user/follow-up prompt or a tool calling result in the next iteration (see `addUserMessagesFromRequest`).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+        chatMemoryStore
+            .storeToAgentContext(agentContext)
+            .withState(nextAgentState)
+            .withMetrics(
+                agentContext
+                    .metrics()
+                    .incrementModelCalls(1)
+                    .incrementTokenUsage(AgentMetrics.TokenUsage.from(chatResponse.tokenUsage())));
+
+    return new AgentResponse(updatedContext, updatedContext.memory().getLast(), toolsToCall);
```
maff, line 110: This will be emitted as process variable - here it is crucial that we re-inject it on for the next call. This is a manual configuration for now, but we'll work on making this easier in the future. 

image

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/AiAgentFunction.java
```diff
@@ -0,0 +1,52 @@
...
+
+import io.camunda.connector.agenticai.aiagent.agent.AiAgentRequestHandler;
+import io.camunda.connector.agenticai.aiagent.model.request.AgentRequest;
+import io.camunda.connector.api.annotation.OutboundConnector;
+import io.camunda.connector.api.outbound.OutboundConnectorContext;
+import io.camunda.connector.api.outbound.OutboundConnectorFunction;
+import io.camunda.connector.generator.java.annotation.ElementTemplate;
+import io.camunda.connector.generator.java.annotation.ElementTemplate.PropertyGroup;
+
+@OutboundConnector(
```
maff, line 17: For now, this is an outbound connector on a service task which interacts with the ad-hoc subprocess. In the future, there will also be the same logic applied to the ad-hoc subprocess directly (as soon as the AHSP supports job workers). I tagged this as alpha as this will definitely be subject of change in follow-up iterations (same for the tools  schema resolver).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+      ChatMemory chatMemory, AgentRequest.AgentRequestData requestData) {
+    // add system prompt if this is the first request
+    if (chatMemory.messages().isEmpty()) {
+      chatMemory.add(promptFromConfiguration(requestData.systemPrompt()).toSystemMessage());
+    }
+  }
+
+  private void addUserMessagesFromRequest(
+      AgentContext agentContext, ChatMemory chatMemory, AgentRequest.AgentRequestData requestData) {
+    if (agentContext.isInState(AgentState.WAITING_FOR_TOOL_INPUT)) {
```
maff, line 134: Based on the "state machine" we know which kind of messages to inject.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAiAgentRequestHandler.java
```diff
@@ -0,0 +1,158 @@
...
+      final var toolCallResults = requestData.tools().toolCallResults();
```
maff, line 135: `toolCallResults` is the multi-instance output collection which is injected as part of the connector config:

image
image

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/AgentContextChatMemoryStore.java
```diff
@@ -0,0 +1,103 @@
...
+
+    // step 2: convert JSON string to map for better variable inspection
+    try {
+      return objectMapper.readValue(json, STRING_OBJECT_MAP_TYPE_REFERENCE);
+    } catch (JsonProcessingException e) {
+      throw new ConnectorException(
+          "Failed to convert memory entry: " + humanReadableJsonProcessingExceptionMessage(e));
+    }
+  }
+
```
maff, line 89: We're using Langchain4j's serialization format for now, but that exposes only serializing to a JSON string which is not too nice to introspect in Operate. So we first serialize to string and then parse that JSON structure into a Map.

In the future we can improve this, but for now I would go this way for a first iteration. Examples how we could improve:

- using our own data structures for storage
- finding ways to serialize to a Map directly

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/AgentRequest.java
```diff
@@ -0,0 +1,121 @@
...
+import io.camunda.connector.generator.dsl.Property;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.Min;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import java.util.List;
+import java.util.Map;
+
+public record AgentRequest(ProviderConfiguration provider, AgentRequestData data) {
```
maff, line 20: This is the first shot of the parameters we want to support in the first place, but we already have some ideas how to improve usability. Very happy to accept any feedback regarding how this is set up, we should just keep in mind that this will be subject to change.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/ProviderConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+                group = "model",
+                description = "Specify endpoint if need to use a custom API endpoint",
+                type = TemplateProperty.PropertyType.Hidden,
+                feel = Property.FeelMode.disabled,
+                optional = true)
+            String endpoint,
+        OpenAiAuthentication authentication,
+        OpenAiModel model) {}
+
+    public record OpenAiAuthentication(
```
maff, line 126: Tried to align this with the existing OpenAI connector as much as possible.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/ProviderConfiguration.java
```diff
@@ -0,0 +1,198 @@
...
+        @NotBlank
+            @TemplateProperty(
+                group = "model",
+                label = "Model",
+                type = TemplateProperty.PropertyType.String,
+                feel = Property.FeelMode.optional,
+                defaultValue = "claude-3-5-sonnet-20240620",
+                defaultValueType = TemplateProperty.DefaultValueType.String)
+            String model,
+        @Valid ModelParameters parameters) {}
```
maff, line 77: We are using the same parameters record for all 3 configurations, but it might be needed to have a specific implementation, based on the model, so I kept this under the specific model structure.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tools/ToolCallingHandler.java
```diff
@@ -0,0 +1,111 @@
...
+      Map<String, Object> arguments =
+          objectMapper.readValue(inputJson, STRING_OBJECT_MAP_TYPE_REFERENCE);
+      return new AgentResponse.ToolToCall(id, name, arguments);
+    } catch (Exception e) {
+      throw new ConnectorException(
+          "Failed to parse tool call results for tool %s".formatted(name), e);
+    }
+  }
+
+  public List<ToolExecutionResultMessage> toolCallResultsAsToolExecutionResultMessages(
```
maff, line 80: Maps tool call results (from a tool in the Ad-hoc subprocess) to a message we can pass to the LLM. Right now, only string responses are supported by major providers.

In a follow-up we should add support for our document handling in here to allow better interactions with existing connectors (e.g. load a document from S3 and pass the contents to the LLM).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tools/ToolSpecificationConverter.java
```diff
@@ -0,0 +1,202 @@
...
+import dev.langchain4j.model.chat.request.json.JsonSchemaElement;
+import dev.langchain4j.model.chat.request.json.JsonStringSchema;
+import io.camunda.connector.agenticai.adhoctoolsschema.model.AdHocToolsSchemaResponse;
+import java.util.Map;
+
+/**
+ * Validates and converts a json schema to a tool specification. Conversion logic is based on
+ * dev.langchain4j.mcp.client.ToolSpecificationHelper.
+ */
+public class ToolSpecificationConverter {
```
maff, line 34: Unfortunately it's currently not straightforward to parse a JSON schema string into a Langchain4j tool specification object, so for now we need this converter doing this mapping. I'll follow-up on this in the future if we can make it easier to handle.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty(
+    value = "camunda.connector.agenticai.enabled",
+    havingValue = "true",
+    matchIfMissing = true)
+@EnableConfigurationProperties(AgenticAiConnectorsConfigurationProperties.class)
+public class AgenticAiConnectorsAutoConfiguration {
```
maff, line 34: As we're using some beans (e.g. the `CamundaClient`) for the connector functionality, the connectors and their dependencies are defined as beans as it also makes it easier to test/mock in e2e tests. To completely disable all the agentic AI integration, a property `camunda.connector.agenticai.enabled=false` can be set to disable this configuration.

This configuration is loaded into the Spring context with the `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports` file, so no need to configure component scanning etc. (it behaves like a normal Spring Boot starter/autoconfiguration).

In addition, this exposes additional config propertes (currently only used to configure the schema caching) under `camunda.connector.agenticai.*`.

Happy to discuss the base path for these configurations, but something like this could be a good pattern for other connectors as well. 

@sbuettner probably also related to https://github.com/camunda/connectors/issues/4437#issuecomment-2804825768

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Bean
+  @ConditionalOnMissingBean
```
maff, line 37: `@ConditionalOnMissingBean` on all the beans allows to override/define a custom implementation in tests or in client projects.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CamundaClientAdHocToolsSchemaResolver.java
maff: After recent discussions we'll read this from the documentation element as a first iteration - see updated description on https://github.com/camunda/connectors/issues/4545.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CamundaClientAdHocToolsSchemaResolver.java
maff: Removed this when moving to the documentation block.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+import io.camunda.connector.agenticai.aiagent.tools.ToolSpecificationConverter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConditionalOnProperty(
+    value = "camunda.connector.agenticai.enabled",
```
johnBgood, line 29: I think in general we use `camunda.connectors.....` plural

johnBgood (dismissed):
I left a few comments, but overall it's approved :) Amazing work there

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
maff: I tried to align this with the already existing config properties - would you still prefer to move this to `connectors`?
image

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/AdHocToolsSchemaFunction.java
```diff
@@ -0,0 +1,46 @@
...
+import io.camunda.connector.api.outbound.OutboundConnectorContext;
+import io.camunda.connector.api.outbound.OutboundConnectorFunction;
+import io.camunda.connector.generator.java.annotation.ElementTemplate;
+import io.camunda.connector.generator.java.annotation.ElementTemplate.PropertyGroup;
+
+@OutboundConnector(
+    name = "Ad-hoc subprocess tools schema (alpha)",
+    inputVariables = {"data"},
+    type = "io.camunda.agenticai:adhoctoolsschema:0")
+@ElementTemplate(
```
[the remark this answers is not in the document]
maff: The idea here was to keep the version to 0 as long as this is an ongoing development, as this will be subject to some more changes. Also I would prefer to omit creating docs until the functionality is stable enough. Would that work for you?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
johnBgood: oh all good then mb :)

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CachingAdHocToolsSchemaResolver.java
```diff
@@ -0,0 +1,43 @@
...
+  @Override
+  public AdHocToolsSchemaResponse resolveSchema(
+      Long processDefinitionKey, String adHocSubprocessId) {
+    return cache.get(new AdHocToolsIdentifier(processDefinitionKey, adHocSubprocessId));
+  }
+
+  private static LoadingCache<AdHocToolsIdentifier, AdHocToolsSchemaResponse> buildCache(
+      AdHocToolsSchemaResolver delegate, CacheConfiguration config) {
+    final var builder = Caffeine.newBuilder();
+    Optional.ofNullable(config.maxSize()).ifPresent(builder::maximumSize);
```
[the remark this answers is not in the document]
maff: I would prefer to keep the defaults on the config properties class as it will allow configuring them to `null` (the Caffeine builder is quite strict and does not allow overriding an already configured value with `null`). I refactored the handling a bit + added a comment where the defaults are initialized.

johnBgood (approved):

gbetances089 (approved):

PR #4565, opened 2025-04-03
chore(deps): pin httpcore5-h2 due to an incompatibility in camunda client

johnBgood:
## Description

Pin httpcore5-h2 dependency, as the Camunda client imports a different version that causes issues on our side when starting the Runtime. The runtime starts, but the httpclient thread dies, which means nothing can be done.

## Checklist


chillleader (approved):

ztefanie (approved):
Thank you

mathias-vandaele (approved):
Good!

PR #4623, opened 2025-04-15
fix(connector-runtime-core): SecretProviderAggregator should respect fetchAll function of each secret provider

jonathanlukas:
## Description

This PR fixes an issue that the `fetchAll` function of each secret provider is not invoked if the aggregator is used.

## Related issues

closes #

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretProviderAggregator.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretProviderAggregator.java
@@ -19,6 +19,7 @@ package io.camunda.connector.runtime.core.secret;
 import io.camunda.connector.api.secret.SecretProvider;
 import java.util.Collections;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -56,6 +57,13 @@ public class SecretProviderAggregator implements SecretProvider {
     }
     LOG.debug("Could not resolve secret '{}'", secretName);
     return null;
+  }
+
+  @Override
+  public List<String> fetchAll(List<String> keys) {
+    return secretProviders.stream()
+        .flatMap(provider -> provider.fetchAll(keys).stream())
+        .collect(Collectors.toList());
   }
 
   public List<SecretProvider> getSecretProviders() {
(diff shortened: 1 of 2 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretProviderAggregator.java
```diff
@@ -58,6 +59,13 @@ public String getSecret(String secretName) {
     return null;
   }
 
+  @Override
+  public List<String> fetchAll(List<String> keys) {
+    return secretProviders.stream()
+        .flatMap(provider -> provider.fetchAll(keys).stream())
+        .collect(Collectors.toList());
+  }
```
sbuettner, line 67: Usually we only need single secret value for a key. Wouldnt this implementation possibly lead to more secret values than keys?

Or did you find a problem with the current approach?

jonathanlukas: >Usually we only need single secret value for a key. Wouldnt this implementation possibly lead to more secret values than keys?

Potentially yes, but the use case here is the replacement of secrets for a secure response message.

>Or did you find a problem with the current approach?

This is the motivation here: I am currently working on a custom secret provider and I found out that I must not throw exceptions from the `getSecret()` method because the replacement part is not working then. My way around this would have been to use the `fetchAll()` method to collect secrets while suppressing exceptions so that building the response message works.

>Its also important to know that we need exactly the behaviour of getSecret to make sure we get the same secrets as they were resolved before when calling getSecret.

As I saw it when doing my research, I found that you would need _at least_ the secrets you would get with `getSecret()`.

sbuettner: > but the use case here is the replacement of secrets for a secure response message.
Can you elaborate?

> I am currently working on a custom secret provider and I found out that I must not throw exceptions from the getSecret() method because the replacement part is not working then. 

The `SecretProvider`  interface describes the contract as the following:
>  @return the secret's value for the given name, if it exists. Otherwise, <code>null</code> is returned.
Why do you want to throw an exception?

It would be great if you can provide more context here.

jonathanlukas: >Why do you want to throw an exception?

Because I did not read and understand that properly. I would say never mind.

And thank you for tackling this together with me.

sbuettner: All good 😄

PR #4635, opened 2025-04-23
feat(agentic-ai): support fromAi() FEEL function to define ad-hoc tools schema

### LINKED ISSUE #4634: Agentic AI: define ad-hoc tools schema via transparent fromAi function instead of a JSON schema

In https://github.com/camunda/connectors/issues/4545 we added a way to describe a JSON schema within the documentation element for a task within an ad-hoc subprocess. As this is pretty, technical and error-prone, we should update this to a more inline approach of defining tool inputs where they are used.

Input and output mappings can use a function `fromAi(<value>, "<description>", "<type>", <schema>, <options>)` to define a variable which should be provided by an LLM.

On the engine side (see https://github.com/camunda/camunda/issues/31258) this function will simply return the value passed as first parameter and ignore the remaining parameters.

The connector implementation is expected to parse the function calls from input and output mappings in order to internally convert its parameters to a JSON schema which can be used in combination with LLM tools calling. The function can be used standalone or in complex FEEL expressions defining multiple parameters.

The function exposes the following parameters:

- `value` (required): on the engine side it can be any value, on the connector side it needs to be a reference within an object, so the connector can derive the input variable name. E.g. `toolCall.searchQuery` ends up in a variable `searchQuery` on the tasks's input schema.
- `description` (string, optional, default `null`): describes the input variable in plain text for the LLM
- `type` (string, optional, default `string`): a [JSON schema type](https://json-schema.org/understanding-json-schema/reference/type) to use for this property.
- `schema` (context, optional): the actual JSON schema for more advanced use cases (e.g. to define object structures, enum values, ...). `type` and `description` will be merged into this schema if they are set.
- `options` (context, optional): open map of options for future use (e.g. to mark properties as optional, ...)

Valid usage examples:

```
fromAi(toolCall.aSimpleValue, "A simple value")
```

```
fromAi(toolCall.anEnumValue, "An enum value", "string", { enum: ["A", "B", "C"] })
```

```
{
  foo: "bar",
  bar: fromAi(toolCall.barValue, "A good bar value", "string"),
  combined: fromAi(toolCall.firstOne, "The first value") + fromAi(toolCall.secondOne, "The second value", "string")
}
```

- Defining the JSON schema as a whole in a dedicated extension property (needs modeler changes, very technical)
- Defining input properties as a dedicated list in modeler (similar to inputs). Clutters the UI & makes the properties panel complex + finding a good naming to distinguish from input mappings is hard.

__Additional context__

Usage example: 

Image

maff:
## Description

See #4634 for background. This PR moves the definition of inputs which need to be provided by an LLM in an agentic context from a JSON schema in the `documentation` block to individual `fromAi()` function calls which define the metadata/schema of the input properties.

Instead of parsing a JSON schema from the `documentation` element, the tools schema resolver now looks for `fromAi` FEEL function calls in input and output mappings and collects them to generate the tool schema. The first argument to the `fromAi()` function needs to reference the injected variable in the `toolCall` object which contains all LLM-generated inputs.

In addition, this PR renames `toolToCall`/`toolsToCall` to `toolCall`/`toolCalls` (reflected in the e2e tests).

**Note:** I added the current working use cases to the `examples` folder so we can develop them in parallel with the implementation. If needed we can also do this on a dedicated PR. In terms of review I think we can ignore them.

### Examples

A task with an input mapping containing `fromAi(toolCall.myVariable, "This is our first variable")` will result in the following JSON schema. If multiple `fromAi` calls are used in the same task, they are all added to the task schema.

```json
{
  "name": "MyTask",
  "description": "Some description.",
  "inputSchema": {
    "type": "object",
    "properties": {
      "myVariable": {
        "type": "string",
        "description": "This is our first variable"
      }
    },
    "required": [
      "myVariable"
    ]
  }
}
```

Mappings can define very simple inputs (defaulting to string):
```
fromAi(toolCall.userId)
```

Or more complex ones, including a schema:
```
fromAi(toolCall.anEnumValue, "An enum value", "string", { enum: ["A", "B", "C"] })
```

An optional last parameter `options` supports passing a map of options which are currently unused, but reserved for future use (e.g. to provide metadata, mark properties as optional, ...).

### Parameter name resolution

The examples above showcase the value being defined as `toolCall.myVariable` which will be a pattern implemented by this connector. The engine implementation of `fromAi` does not restrict the value format, but the parsing implementation requires the value to be an object dereference as doing so allows us to use the field name as variable name. 

In the example above, a variable `myVariable` is derived from the `fromAi` function call.

### Schema extraction logic

As all parameters besides the value are optional, the schema is built in the following way:

- If the `schema` property is defined, its value be used as the initial schema. Otherwise, an empty JSON schema object is initialized to `{}`.
- If the `type` property is set, it will be applied to the schema (`{"type": "number"}`), potentially overwriting a type which was directly defined on the schema object.
- If no type is set after reading `schema` and `type` properties, a default value of `{"type": "string"}` will be set.
- If the `description` property is set, it will be applied to the schema (`{"description": "My property"}`), potentially overwriting a description which was directly defined on the schema object.

This process is repeated for every parameter and combined to a task schema like in the example above. In this implementation, all properties are defined as required, but this can be adapted in a follow-up if needed.

### Repeated usage of the same input value name

A `fromAi()` function for a certain value can only be used once per task to avoid ambiguity, however as the value part accesses an existing value, a value can be directly referenced in case it is needed multiple times:

```
{
  first: fromAi(toolCall.first, "The first parameter", "number"),
  second: fromAi(toolCall.second, "The second parameter", "string", { enum: ["A", "B", "C"] }),
  allParams: [toolCall.first, toolCall.second]
}
```

The following will raise an incident due to the repeated usage of the same value ❌ 
```
{
  first: fromAi(toolCall.first, "The first parameter", "number"),
  first2: fromAi(toolCall.first, "Testme")
}
```

## Related issues

Based on https://github.com/camunda/connectors/pull/4548.
Depends on https://github.com/camunda/camunda/pull/31259 to implement `fromAi()` support in the engine ✅ 

closes #4634

## Checklist

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
@@ -362,7 +362,7 @@ public class AiAgentTests extends BaseAgenticAiTest {
       AgentResponse agentResponse, AgentMetrics expectedMetrics, List<ChatMessage> expectedMessages)
       throws JsonProcessingException {
     assertThat(agentResponse).isNotNull();
-    assertThat(agentResponse.toolsToCall()).isEmpty();
+    assertThat(agentResponse.toolCalls()).isEmpty();
     assertTrue(agentResponse.context().isInState(AgentState.READY));
     assertThat(agentResponse.context().metrics()).isEqualTo(expectedMetrics);
 
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
@@ -388,9 +388,9 @@ public class AiAgentTests extends BaseAgenticAiTest {
 
   private void assertToolSpecifications(ChatRequest chatRequest) {
     assertThat(chatRequest.toolSpecifications())
-        .hasSize(4)
         .extracting(ToolSpecification::name)
-        .containsExactly("GetDateAndTime", "SuperfluxProduct", "Search_The_Web", "An_Event");
+        .containsExactly(
+            "GetDateAndTime", "SuperfluxProduct", "Search_The_Web", "A_Complex_Tool", "An_Event");
   }
 
   private void assertIncident(ZeebeTest zeebeTest, ThrowingConsumer<Incident> assertion) {
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/expected-schema-result.json
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/expected-schema-result.json
@@ -47,6 +47,82 @@
       }
     },
     {
+      "name": "A_Complex_Tool",
+      "description": "A very complex tool",
+      "inputSchema": {
+        "type": "object",
+        "properties": {
+          "aSimpleValue": {
+            "type": "string",
+            "description": "A simple value"
+          },
+          "anEnumValue": {
+            "type": "string",
+            "description": "An enum value",
+            "enum": [
+              "A",
+              "B",
+              "C"
+            ]
+          },
+          "anArrayValue": {
+            "type": "array",
+            "description": "An array value",
+            "items": {
+              "type": "string",
+              "enum": [
+                "foo",
+                "bar",
+                "baz"
+              ]
+            }
+          },
+          "urlPath": {
+            "type": "string",
+            "description": "The URL path to use"
+          },
+          "firstValue": {
+            "type": "string"
+          },
+          "secondValue": {
+            "type": "integer",
+            "description": "The second value"
+          },
+          "thirdValue": {
+            "type": "string",
+            "description": "The third value to add"
+          },
+          "fourthValue": {
+            "type": "array",
+            "description": "The fourth value to add",
+            "items": {
+              "type": "string",
+              "enum": [
+                "foo",
+                "bar",
+                "baz"
+              ]
+            }
+          },
+          "outputValue": {
+            "type": "string",
+            "description": "An output value"
+          }
+        },
+        "required": [
+          "aSimpleValue",
+          "anEnumValue",
+          "anArrayValue",
+          "urlPath",
+          "firstValue",
+          "secondValue",
+          "thirdValue",
+          "fourthValue",
+          "outputValue"
+        ]
+      }
+    },
+    {
       "name": "An_Event",
       "description": "An event!",
       "inputSchema": {
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/expected-schema-result.json
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/expected-schema-result.json
@@ -48,7 +48,7 @@
     },
     {
       "name": "An_Event",
-      "description": "",
+      "description": "An event!",
       "inputSchema": {
         "type": "object",
         "properties": {},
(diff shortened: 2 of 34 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/schema/AdHocToolSchemaGenerator.java
maff, line 1: FYI this is singular `Tool Schema` opposed to plural `Tools Schema` like the function as it is only handling the schema for a single tool.

connectors/agentic-ai/examples/ad-hoc-tools-schema/ad-hoc-tools-schema-resolver.bpmn
maff, line 1: FYI I added the current working use cases to the `examples` folder so we can develop them in parallel with the implementation. If needed we can also do this on a dedicated PR. In terms of review I think we can ignore them.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/schema/AdHocToolSchemaGenerator.java
```diff
@@ -0,0 +1,15 @@
...
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.adhoctoolsschema.resolver.schema;
+
+import io.camunda.connector.agenticai.adhoctoolsschema.feel.FeelInputParam;
+import java.util.List;
+import java.util.Map;
+
+public interface AdHocToolSchemaGenerator {
```
maff, line 13: FYI this is singular Tool Schema opposed to plural Tools Schema like the function as it is only handling the schema for a single tool.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -362,7 +362,7 @@ private void assertAgentResponse(
       AgentResponse agentResponse, AgentMetrics expectedMetrics, List<ChatMessage> expectedMessages)
       throws JsonProcessingException {
     assertThat(agentResponse).isNotNull();
-    assertThat(agentResponse.toolsToCall()).isEmpty();
+    assertThat(agentResponse.toolCalls()).isEmpty();
```
maff, line 365: We renamed `toolsToCall` to `toolCalls` throughout the integration.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/expected-schema-result.json
```diff
@@ -46,9 +46,85 @@
         ]
       }
     },
+    {
+      "name": "A_Complex_Tool",
```
maff, line 50: Showcases extracting a complex schema from nested properties.

image

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/resolver/CamundaClientAdHocToolsSchemaResolver.java
```diff
@@ -66,29 +78,46 @@ public AdHocToolsSchemaResponse resolveSchema(
   }
 
   private AdHocToolDefinition mapActivityToToolDefinition(FlowNode element) {
-    final var documentation = getDocumentation(element);
-
-    if (!documentation.isBlank() && documentation.trim().startsWith("{")) {
```
maff, line 71: No JSON parsing on the documentation anymore - instead we extract `fromAi` function calls from input & output mappings.

connectors/agentic-ai/examples/ad-hoc-tools-schema/ad-hoc-tools-schema-resolver.bpmn
maff, line 1: FYI I added the current working use cases to the examples folder so we can develop them in parallel with the implementation. If needed we can also do this on a dedicated PR. In terms of review I think we can ignore them.

gbetances089 (approved):

johnBgood (approved):

Issue #4367, opened 2025-03-18
[Inbound Email connector] Fixing the identified issues


mathias-vandaele:
__Describe the Bug__

After a long meeting, we could make good progress.

Those are all of the issue our customers are facing when using email connector:

1. 2 instances are created when an email is received (already known)
2. When moving an email to another folder, the initial email is supposed to be deleted, but there is some caching system that makes it still there, it appears with the flag deleted, even the web UI of the email provider becomes broken, suggesting we might not deleting properly or purging properly?
3. When deploying a new version of a process definition using the inbound with a different mail provider, the entire connector pods seems to crash, even though it does not appears all the time (an exception not caught?)

related to https://github.com/camunda/connectors/issues/4313

__Steps to Reproduce__

1. do this
2. do that

__Expected Behavior__

__Environment__

- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]
- Camunda version: [e.g. 8.0.4 (self-managed)]

sbuettner:
@mathias-vandaele What about avoiding this issue by requiring the folder to be present when starting the Connector?

PR #4640, opened 2025-04-23
fix(email-connector): fix for inbound email connector when an email cant be moved

mathias-vandaele:
## Description

Some IMAP servers do not permit folder creation:

- Infomaniak allows it.
- Gmail does not.

When the folder did not exist on Gmail, the scheduler malfunctioned due to an exception being thrown. To prevent this issue, we now mark the email as 'read' before attempting to move it, ensuring that it is not processed again, thereby improving the system's efficiency.

If an exception occurs because the folder could not be created, and we are reading only unseen emails, the connector is canceled with a RetryException. If we are reading all emails, the connector is canceled with a ConnectorException.

## Related issues

closes https://github.com/camunda/connectors/issues/4367

## Checklist

### DIFF
--- a/connectors/email/pom.xml
+++ b/connectors/email/pom.xml
@@ -37,6 +37,12 @@
             <version>2.1.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>4.3.0</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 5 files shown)


sbuettner:
@mathias-vandaele What about avoiding this issue by requiring the folder to be present when starting the Connector?

mathias-vandaele:
> @mathias-vandaele What about avoiding this issue by requiring the folder to be present when starting the Connector?

@sbuettner I'm not sure about this, it doesn't seem right to remove a feature just because one email provider doesn't allow folder creation at the root level.

It should still be possible to create folders in Gmail if we prefix them with `[Gmail]`, rather than trying to create them directly at the root.

sbuettner:
@mathias-vandaele Please make sure this is properly documented in our docs for this operation.

connectors/email/pom.xml
```diff
@@ -37,6 +37,12 @@
             <version>2.1.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
sbuettner, line 40: Why do we need this in the email connector?

sbuettner (approved):

PR #4644, opened 2025-04-24
chore(element-templates): Bring all versions of element templates to connectors repo

### LINKED ISSUE #4567: Bring all Elements Templates versions in the Connectors repository
__What should we do?__

Currently, the [web-modeler](https://github.com/camunda/web-modeler/tree/f53f66c6f6734c019ce5567f52ee56e39540c3f1/webapp/src/App/Pages/Diagram/BpmnJSExtensions/connectorsExtension/.camunda/element-templates) contains all the versions of our element templates.

We also have [this file](https://github.com/camunda/connectors/blob/e9a042c7448a1cb50e8a8ddc789825ae78bea63e/connector-templates.json) that references the different versions of each connector.

💡 We could simplify both processes by leveraging our Element Template Generator.

We would bootstrap the system using the web-modeler to get older versions to create this kind of structure:

Image

We could then update [uniquet](https://github.com/camunda/connectors/blob/f085f0d5e6b80e6d810be148c5de15dc49f50571/element-template-generator/uniquet) to generate this kind of file:

```json
...,
"io.camunda.connector.IdpExtractionOutBoundTemplate.v1": [
    {
      "version": 1,
      "ref": "https://raw.githubusercontent.com/camunda/connectors/f614a32d460029f0b9719de5e02bf206c1157eaa/connectors/idp-extraction/element-templates/versioned/idp-extraction-outbound-connector-1.json",
      "engine": {
        "camunda": "^8.6"
      },
      "version": 2,
      "ref": "https://raw.githubusercontent.com/camunda/connectors/f614a32d460029f0b9719de5e02bf206c1157eaa/connectors/idp-extraction/element-templates/idp-extraction-outbound-connector.json",
      "engine": {
        "camunda": "^8.7"
      }
    }
  ],
...
```

One can then easily find the right template for version X and/or compatible with Engine Y. 

**Note**
We would keep the existing ET generation behavior in place. It's just an additional layer on top of it (versioning + single file generation).

__Why should we do it?__

- Simplify our process of updating Element Templates
- Remove the possibility of having outdated ET in the web-modeler repo

ztefanie:
## Description
This PR imports all element templates from the Web Modeler into the connectors repository.

### Summary of changes:
- The element template arrays from the Web Modeler for each connector have been split into individual files.
- Each template is saved in the versioned subdirectory, using version numbers in the filenames.
- Existing element templates in the connectors repo have received a version bump.
- The Element Template Generator was executed on all connectors to ensure that the templates are in sync with the annotations in the corresponding Java classes.
- adjust docs-next-link-checker to not include files in the versioned folder (if you think we should update the links there, let me know, then I will do it in a separate PR)

### Notes:
The scripts used to automate these tasks are not included in this PR, as they were one-off utilities and are unlikely to be reused. You can find them here: [scripts.zip](https://github.com/user-attachments/files/19886097/scripts.zip)

### Not handled yet:
The following templates are not yet integrated, as they are not currently part of the repo and will require special handling:
- kafka-inbound-connector with the id: "io.camunda.connectors.inbound.KAFKA.v1" (is deprecated, will not be copied)
- power-automate-connector (is deprecated, will not be copied)
- rpa-connector (will be disccued by @sbuettner with RPA team)

These will be addressed in a follow-up PR to keep this one focused and clean.

## Related issues
relates to #4567 

## Checklist


connectors/asana/element-templates/asana-connector.json
```diff
@@ -2,8 +2,8 @@
   "$schema": "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
   "name": "Asana Outbound Connector",
   "id": "io.camunda.connectors.Asana.v1",
-  "version": 2,
-  "engines":{
+  "version": 3,
```
johnBgood, line 5: here it's 3 but below it's 2 ?

johnBgood (approved):
Only 1 template is off I guess, I left a comment

connectors/asana/element-templates/asana-connector.json
ztefanie: What do you mean below it is 2? in the versioned folder? 
Because that is the expected pattern, the version in the element-template root folder is one higher than in the versioned folder.

johnBgood: no no in the same file, there's a section that's been replaced with the exact same content

ztefanie: ok thanks i will have a look at it

ztefanie: good find. Thanks. I fixed it here and in other connectors where the same happened.

johnBgood (approved):

PR #4652, opened 2025-04-24
feat(agentic-ai): add documents support to Agentic AI connector

### LINKED ISSUE #4641: Agentic AI: Add documents support to agentic AI connector

To integrate with other connectors (e.g. the Vector DB connector), we should add document handling support to the AI Agent connector introduced in #4547.

- a new `documents` input field (similar to the Bedrock integration) allows to input documents to the user prompt
- when handling tool responses, tools should be able to return document references which are transparently resolved to a string value by the connector
- as tool responses currently only accept strings as responses, we need to find a good serialization format to provide document contents as tool responses.

To reduce the scope we should initially limit the document support to the following content types:

- Plain text
- Text file formats: `text/*` + some selected text formats such as `application/json`
- Images: jpg, png, gif, webp (currently supported formats by OpenAI and Anthropic)
- PDF

___Tool Response Serialization___

This might be something we'll need to gain more experience, but as a first iteration it will be sufficient to serialize document data into an arbitrary format which is understood by the supported models. The tool response is already serialized to JSON if it is not a text value, so we can leverage JSON serialization and provide a custom serializer for documents which converts a document to a structure similar to what Anthropic uses for document uploads:

```
{
  "type": "base64",
  "media_type": "application/pdf",
  "data": "<base64 data>"
}
```

If the file is a text file, we can use a `text` type and pass the plaintext data directly:

```
{
  "type": "text",
  "media_type": "text/plain",
  "data": "Hello, world"
}
```

maff:
## Description

See linked issue for background and details. Updates the Agentic AI connector to support documents in 2 ways:

- As a `documents` input mapping which is read and applied to the user prompt/message. The `CamundaDocumentToContentConverter` introduced on this PR is able to convert text, image and PDF documents to Langchain4j content blocks which can be added to a user message. This allows reference previously added documents (e.g. user upload, RAG, ...) directly from the user prompt. Example: `Summarize this PDF for me`.
- When handling tool call results, the L4J API (and model providers) only allow passing back a string. To support arbitrary data structures, we already serialize objects to JSON if they are not string values. With this PR, the newly introduced `CamundaDocumentContentSerializer` adds support to include Camunda documents in the response and to have their contents transparently converted to text. 

## Related issues

Based on #4635

closes #4641 

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tools/CamundaDocumentSerializer.java
```diff
@@ -0,0 +1,98 @@
...
+import dev.langchain4j.data.message.PdfFileContent;
+import dev.langchain4j.data.message.TextContent;
+import dev.langchain4j.data.message.TextFileContent;
+import io.camunda.client.api.response.DocumentMetadata;
+import io.camunda.connector.agenticai.aiagent.document.CamundaDocumentToContentConverter;
+import io.camunda.document.Document;
+import io.camunda.document.reference.DocumentReference.CamundaDocumentReference;
+import java.io.IOException;
+
+public class CamundaDocumentSerializer extends JsonSerializer<Document> {
```
sbuettner, line 26: Can you provide some comments what this serializer is used for?

connectors/agentic-ai/pom.xml
```diff
@@ -51,6 +51,11 @@
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-autoconfigure</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-web</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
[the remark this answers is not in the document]
maff: I used Spring's `MediaType` for MIME type comparisons, but switched to HttpClient's implementation now as we already have that on the classpath from the Langchain4J dependencies.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tools/CamundaDocumentSerializer.java
maff: Sure, I updated the PR description + added class level docs & tests.

When handling tool call responses, currently it's only possible to pass back a string result to the LLM (opposed to the user prompt which supports content blocks of different types of mostly base64 serialized data). This is a limitation of Langchain4J, but also of the provider APIs.

To support arbitrary objects in these tool call results, we already serialize the result to JSON if it isn't a string. To additionally support documents, this serializer converts documents to a similar structure as the one supported in user prompts:

```
{
  "type": "base64",
  "media_type": "application/pdf",
  "data": "<base64 data>"
}
```

By using Jackson and a custom serializer we can also support documents in nested structures like in this test: https://github.com/camunda/connectors/pull/4652/files#diff-d36314036e10c4323614aaaabdc5db3d9344292600f139041cdcb4aa4cfb737dR100

Opposed to the already existing `JacksonModuleDocumentSerializer`, this serializer mainly serializes the content without keeping the camunda document data (e.g. reference ID, ...).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/AiAgentTests.java
```diff
@@ -286,13 +324,259 @@ void executesAgentWithToolCallingAndUserFeedback() throws Exception {
...
+      }
+
+      final var initialUserPrompt = "Go and download a document!";
+      final var expectedConversation =
+          List.of(
+              new SystemMessage(
+                  "You are a helpful AI assistant. Answer all the questions, but always be nice. Explain your thinking."),
+              new UserMessage(initialUserPrompt),
+              new AiMessage(
+                  "The user asked me to download a document. I will call the Download_A_File tool to do so.",
```
maff, line 357: `Download_A_File` is a REST connector using the "Store Response" option.

image

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tools/ToolCallResultConverter.java
```diff
@@ -0,0 +1,67 @@
...
+import io.camunda.connector.api.error.ConnectorException;
+import java.util.Map;
+
+/**
+ * Converts the result of a tool call to a {@link ToolExecutionResultMessage}.
+ *
+ * <p>If the result is not a string, it will be serialized to a JSON string, using the {@link
+ * CamundaDocumentContentSerializer} to serialize document contents.
+ */
+public class ToolCallResultConverter {
```
maff, line 27: Extracted from `ToolCallingHandler` + added custom module to objectmapper.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tools/CamundaDocumentSerializer.java
sbuettner: Thanks @maff for the elaborate answer. That helps!

sbuettner (approved):

gbetances089 (approved):

johnBgood (approved):

gbetances089 (approved):

Issue #4629, opened 2025-04-18
IDP Connector: Handle table content in structured form extraction


reiballa:
__What should we do?__

We need to parse table content and figure out how to map table content to key-value pairs.

__Why should we do it?__

We need to be able to extract data from table content when doing structured extractions.

PR #4684, opened 2025-04-29
fix(idp-extraction-connector): added table content support for structured extraction for aws

reiballa:
## Description

We need to parse table content and figure out how to map table content to key-value pairs.
This ticket adds support for this process on textract. for all table content it will structure the key as follows: 
- Every key that was extracted from tables will have a 'table' prefix
- After the prefix we use the column header as identifier
- we add a number for the row as suffix.

So for example lets say we have the following table inside a document:

| Name               | Profession.    |
| -------------- | ------------- |
| Stephen King  | Writer             |
| Roy Keane       | Footballer      |

The resulting key-value pairs would be (with a '_' as a delimiter)
| Key                               | Value              |
| --------------------- | ------------- |
| table_name_1             | Stephen King |
| table_name_2            | Roy Keane      |
| table_proffession_1   | Writer              |
| table_proffession_2  | Footballer       |

## Related issues

closes https://github.com/camunda/connectors/issues/4629

## Checklist


reiballa:
Closing this pr because this feature will be done thorugh [this](https://github.com/camunda/connectors/issues/4757) task

PR #4696, opened 2025-04-30
fix(rest): Fix removing of quotes in body

### LINKED ISSUE #4671: [REST Connector] Quotes are removed from the body when sending a POST request
__Describe the Bug__

Customers can't easily send a body containing quotes like: `"My body content"`. Quotes are removed by the REST Connector.

__Steps to Reproduce__

1. Send a POST request with this body: `"The body"`
3. Check in Beeceptor the received body, it will be `The body` (without quotes)

__Expected Behavior__

The quotes should be there.

This should be backported to 8.7 & 8.6

ztefanie:
## Description
Removing the @Feel annotation resolves the issue by preventing the FeelDeserializer from being applied to the body field. While this change fixes the bug, I’m not fully certain whether this is the correct long-term solution or what potential side effects it may introduce.

## Related issues
closes #4671 

## Checklist


connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/jackson/FeelDeserializerTest.java
```diff
@@ -111,6 +111,19 @@ void feelDeserializer_plainString_preserved() throws JsonProcessingException {
     assertEquals("foobar", mapper.readValue(json, TargetTypeString.class).props);
   }
 
+  @Test
+  void feelDeserializer_handleObjectString() throws JsonProcessingException {
```
johnBgood, line 115: can you add a text for text/plain body sending `"Some text"` without the JSON format?

ztefanie: I think even if the body is plain text, it would still be in a json like this:
```json
{
  "url" : "https://camunda.proxy.beeceptor.com/test",
  "body" : "\"hi\"",
  "method" : "POST",
  "storeResponse" : true,
  "authentication" : {
    "type" : "noAuth"
  },
  "ignoreNullValues" : false,
  "readTimeoutInSeconds" : 20,
  "connectionTimeoutInSeconds" : 20
}
```


I tried to add a test like this, but it is not working as a json is expected. Let me know if I miss-understood you.
```java
  @Test
  void feelDeserializer_handleObjectString() throws JsonProcessingException {
    // given, e.g. used in the REST connector body property
    String json =
        """
        "\\\"foobar\\\""
        """;

    // when && then
    assertDoesNotThrow(() -> mapper.readValue(json, TargetTypeObjectString.class));
  }
```

I added a test to CustomApacheHttpClientTests

connectors/http/http-base/src/test/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClientTest.java
```diff
@@ -662,6 +663,27 @@ public void shouldReturn200WithBody_whenGetWithBodyJSON(WireMockRuntimeInfo wmRu
...
     }
 
+    @Test
+    public void shouldReturn200WithBody_whenGetWithQuotedBodyString(
+        WireMockRuntimeInfo wmRuntimeInfo) throws Exception {
+      stubFor(post("/path").willReturn(ok().withBody("\"Hello, world\"")));
+
+      HttpCommonRequest request = new HttpCommonRequest();
+      request.setMethod(HttpMethod.POST);
+      request.setHeaders(Map.of("Accept", "application/json"));
```
johnBgood, line 673: perfect, maybe can Parametrize this test to use application/json and text/plain ?

johnBgood (dismissed):

connector-sdk/jackson-datatype-feel/src/test/java/io/camunda/connector/feel/jackson/FeelDeserializerTest.java
sbuettner: @ztefanie Does this implementation also support sending "raw" requests without any quotes at all?

ztefanie: @sbuettner 
Like this in the modeler?

Yes this is working.

Like without quotes in the json? 
```json
{
  "url" : "https://camunda.proxy.beeceptor.com/test",
  "body" :  hi,
....
}
```
No, because this is not a valid json.

johnBgood (approved):

PR #4716, opened 2025-05-05
fix(vector-db): multiple enhancements and tech debt addressing

igpetrov:
## Description

fix(vector-db): multiple enhancements and tech debt addressing


johnBgood (approved):

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/doc/parsing/source/PlainTextAsDocumentSource.java
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  @Override
+  public InputStream inputStream() {
+    return IOUtils.toInputStream(text, StandardCharsets.UTF_8);
+  }
+
+  @Override
+  public Metadata metadata() {
+    return Metadata.from(HttpHeaders.CONTENT_TYPE, MediaType.TEXT_PLAIN.toString());
```
maff, line 32: Minimal nitpick: we're using the  `key, value` constructor here while we use the `Map` one in the `CamundaDocumentSource`. Would it make sense to stick to one of the variants for consistency?

igpetrov: Lol, how did you even notice that? 😆 

I am usually not that perfectionist: by the end of the day the result is the same. I can change that, not a problem.

maff (approved):

maff (approved):
LGTM 👍

PR #4717, opened 2025-05-06
fix(rest-connector): Fix url encoding to handle edge cases correctly

### LINKED ISSUE #4676: [REST Connector]: Url encoding not working properly with + sign
__Describe the Bug__

`+` sign in URL path are not correctly encoded. 
They are replaced by a space due to the URLEncoder that we use. This space will in turn be escaped using `%20` (this is correct).

This sign should be replaced by `%2B`

__Steps to Reproduce__

1. configure a rest connector with URL as following:
`..../clusters/2585CZ342-1E + 2/test`
2. The URL called is: 
`/clusters/2585CZ342-1E%20%20%202/test`

__Expected Behavior__

The URL called should be: 
`/clusters/2585CZ342-1E%20%2B%202/test`

https://jira.camunda.com/browse/SUPPORT-27014

This should be backported to 8.7 & 8.6

ztefanie:
## Description
Use the correct functions for en- and decoding.

See from [Java docs](https://docs.oracle.com/javase/8/docs/api/java/net/URL.html): 

> Note, the URI class does perform escaping of its component fields in certain circumstances. The recommended way to manage the encoding and decoding of URLs is to use URI, and to convert between these two classes using toURI() and URI.toURL().
> 
>   The URLEncoder and URLDecoder classes can also be used, but only for HTML form encoding, which is not the same as the encoding scheme defined in RFC2396.

And the [URL Encoder docs:](https://docs.oracle.com/javase/8/docs/api/java/net/URLDecoder.html)

> The plus sign "+" is converted into a space character "   " .

Fr Refernce also see: [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)

## Related issues
closes #4676

## Checklist

https://stackoverflow.com/questions/43375916/urldecoder-is-converting-into-space
https://bugs.openjdk.org/browse/JDK-8179507

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/UrlEncoder.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/UrlEncoder.java
@@ -16,90 +16,25 @@
  */
 package io.camunda.connector.http.base.client.apache.builder.parts;
 
-import io.camunda.connector.api.error.ConnectorInputException;
+import com.google.api.client.http.GenericUrl;
 import java.net.MalformedURLException;
 import java.net.URI;
-import java.net.URISyntaxException;
 import java.net.URL;
-import java.util.Objects;
-import java.util.function.BiFunction;
-import java.util.function.Function;
-import org.apache.hc.core5.net.URIBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class UrlEncoder {
-
   private static final Logger LOG = LoggerFactory.getLogger(ApacheRequestUriBuilder.class);
-
-  private static String chooseEncoded(
-      String input,
-      Function<URL, String> getUrlPart,
-      Function<URI, String> get,
-      Function<URI, String> getRaw,
-      BiFunction<URIBuilder, String, URIBuilder> setPart)
-      throws MalformedURLException {
-    URL url = new URL(input);
-    try {
-      URI uri = url.toURI();
-      String rawPart = getRaw.apply(uri);
-      String part = get.apply(uri);
-
-      String fromDecoded = getRaw.apply(setPart.apply(new URIBuilder(), part).build());
-      String fromRaw = get.apply(setPart.apply(new URIBuilder(), rawPart).build());
-
-      return Objects.equals(fromDecoded, fromRaw) ? part : rawPart;
-    } catch (URISyntaxException e) { // if it is not a valid URI, we need to encode it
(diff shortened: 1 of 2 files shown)


johnBgood (commented):
I think this is great but I have 1 question/concern. This looks very manual, I wondered if we could just use a Library (from spring, or anything else) so that we don't have to handle this part ever again?
Maybe can we check how it was implemented by the Google HTTP client in release/8.5 and maybe use the library/code?

ztefanie:
> I think this is great but I have 1 question/concern. This looks very manual, I wondered if we could just use a Library (from spring, or anything else) so that we don't have to handle this part ever again? Maybe can we check how it was implemented by the Google HTTP client in release/8.5 and maybe use the library/code?

Hi Jonathan,

Thanks for your feedback - I looked into a few library-based options as you suggested:

- Spring's UriComponentsBuilder unfortunately doesn’t offer a decode method or any built-in functionality to check whether a URL is already encoded. That means we’d still need to determine encoding status beforehand manually.

- release/8.5 this was handled with using `GenericUrl` from the Google HTTP client, but there was a decision to move from this library to http-commons in 2023: https://github.com/camunda/connectors/issues/1069

And researched on google on how to figure out if a URL is encoded in Java and did not find anything helpful.

The solution I suggested also has the benefit, that only parts of the URL that are not yet encoded will be encoded. E.g. if you have a url specified as this in FEEL:
`="http://www.test.com/my%2CencodedPath?search=" + userInput`. Then the path will not be double encoded, while the user Input will be encoded if needed.

Let me know what you think or how you’d prefer to proceed — happy to adjust based on your input.

sbuettner:
@ztefanie The old client was deprecated as the whole project seemed stale but we could for example use the [GenericUrl](https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java) class from the previously used client library?

ztefanie:
Thanks for guiding me here. I replaced the URL encoding with the GenericUrl. 
Compared to main this changes:
- UTF-8 chars, like `ü` are now encoded. Previously this url: `http://localhost:8080/überuns` would stay the same, now it is encoded like this `http://localhost:8080/%C3%BCberuns`
- plus signs in paths are now not encoded, previously they where encoded as %20, which is space

As before plus signs in query params are encoded as %20, which seems correct.

sbuettner:
Kudos @ztefanie for your initial solution without using a library 👌 Just wanted to share some ideas.

@ztefanie & @johnBgood as I am off I would leave the review up to both of you.

Issue #4693, opened 2025-04-30
[ETG] modify the element template generator


mathias-vandaele:
__What should we do?__

As discussed in the architecture session, we should copy the generated element template to `versioned` when `versions` are enabled in the pom configuration.

* We dont need to copy it to versioned when there is no version change.

__Why should we do it?__

We need a way to keep track of the version history, so that they can be used by older runtime of Camunda

PR #4731, opened 2025-05-07
others(etg): versions history are now managed by the element template generator

mathias-vandaele:
## Description

versions history are now managed by the element template generator

## Related issues

closes https://github.com/camunda/connectors/issues/4693

## Checklist

### DIFF
--- a/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/BasicElementTemplate.java
+++ b/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/BasicElementTemplate.java
@@ -1,0 +1,19 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.generator.java.util;
+
+public record BasicElementTemplate(String id, Integer version) {}
--- a/element-template-generator/maven-plugin/src/main/java/io/camunda/connector/generator/ElementTemplateGeneratorMojo.java
+++ b/element-template-generator/maven-plugin/src/main/java/io/camunda/connector/generator/ElementTemplateGeneratorMojo.java
@@ -32,7 +32,7 @@ import io.camunda.connector.generator.api.GeneratorConfiguration.GenerationFeatu
 import io.camunda.connector.generator.dsl.ElementTemplate;
 import io.camunda.connector.generator.java.ClassBasedDocsGenerator;
 import io.camunda.connector.generator.java.ClassBasedTemplateGenerator;
-import io.camunda.connector.generator.java.util.BasicElementTemplate;
+import io.camunda.connector.generator.java.util.VersionedElementTemplate;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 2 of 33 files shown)


element-template-generator/maven-plugin/src/main/java/io/camunda/connector/generator/ElementTemplateGeneratorMojo.java
```diff
@@ -74,14 +84,21 @@ public class ElementTemplateGeneratorMojo extends AbstractMojo {
...
-                  .withObjectIndenter(new DefaultIndenter().withLinefeed("\n")));
+  @Parameter(
+      property = "versionedDirectory",
+      defaultValue = "${project.basedir}/element-templates/versioned")
+  private String versionedDirectory;
 
-  private static final String COMPILED_CLASSES_DIR = "target" + File.separator + "classes";
-  private static final String HYBRID_TEMPLATES_DIR = "hybrid";
+  @Parameter(property = "versions", defaultValue = "false")
+  private Boolean versions;
```
ztefanie, line 93: (optional) Can this get a more meaningful naming? Like hasVersioning or something like this?

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/BasicElementTemplate.java
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.generator.java.util;
+
+public record BasicElementTemplate(String id, Integer version) {}
```
sbuettner, line 19:
```suggestion
public record VersionedElementTemplate(String id, Integer version) {}
```

ztefanie (commented):
I have two questions:

- When do we set the version flag and when not?
- Is it still possible to make a small adjustment like fixing a typo in the description, without increasing the version?

element-template-generator/maven-plugin/src/main/java/io/camunda/connector/generator/ElementTemplateGeneratorMojo.java
mathias-vandaele: @ztefanie Would you like both the parameters and the variable ? or just the variable

element-template-generator/maven-plugin/src/main/java/io/camunda/connector/generator/ElementTemplateGeneratorMojo.java
```diff
@@ -274,12 +290,61 @@ private void generateElementTemplates(ConnectorConfig config, ClassLoader classL
...
       List<ElementTemplate> templates, boolean hybrid, List<FileNameById> fileNames) {
+    List<BasicElementTemplate> versionedElementTemplates = retrieveBasicElementTemplates();
     for (var template : templates) {
       var fileName = determineFileName(template, fileNames, hybrid);
+      if (versions && !hybrid) manageVersions(template, versionedElementTemplates, fileName);
       writeElementTemplate(template, hybrid, fileName);
     }
   }
 
+  private void manageVersions(
```
sbuettner, line 301: Would be great to have a more concise name here: What is this method doing?

element-template-generator/maven-plugin/src/main/java/io/camunda/connector/generator/ElementTemplateGeneratorMojo.java
```diff
@@ -274,12 +290,61 @@ private void generateElementTemplates(ConnectorConfig config, ClassLoader classL
...
+                  + "-"
+                  + latestBasicElementTemplate.version()
+                  + ".json"));
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private List<BasicElementTemplate> retrieveBasicElementTemplates() {
+    File classicFolder = new File(this.outputDirectory);
```
sbuettner, line 332: What about `latestTemplatesFolder`

ztefanie (approved):

mathias-vandaele:
> I have two questions:
> 
> * When do we set the version flag and when not?
> * Is it still possible to make a small adjustment like fixing a typo in the description, without increasing the version?

1- We will always use the flag, but it is rather not to force our customer to create versions
2- Yes of course, the current `element-template` will just be replaced

PR #4741, opened 2025-05-09
fix(agentic-ai): resilient optional configuration (tools, memory, limits) handling

maff:
## Description

- Make configuration of tools, memory settings and limits optional & do not fail if the request object is null
- Add e2e tests verifying that the connector still works when none of these settings is set and that the max model calls limit falls back to a default value of 10

## Checklist


reiballa (approved):
LGTM

Issue #4740, opened 2025-05-08
E-Mail Connector is not evaluating the result of the `correlate` call


sbuettner:
__Describe the Bug__

The E-Mail Connector is not evaluating the result of the `correlate` call and therefore marking mails as `processed` even if the `correlation` might have failed.

__Steps to Reproduce__

1. Send a mail when there is an correlation error
2. The mails are marked as read but not correlated

__Expected Behavior__

1. Mails are only marked as read when processed successfully.

__Environment__

- `main` and probably all previous versions (8.8, 8.7, 8.6)

### Tasks 
- [x] Fix the issue by evaluating the result of the `correlate` call [here](https://github.com/camunda/connectors/blob/a6ab7701090c62dbd85c86dda68c4ce004a0f2ea/connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java#L153).
- [x] The `canActivate` result we get [here](https://github.com/camunda/connectors/blob/a6ab7701090c62dbd85c86dda68c4ce004a0f2ea/connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java#L186) should be used to decide whether the should even be processed. It currently only leads to an empty list of documents but it could just skip the whole process.
- [x] Add a test that makes sure we dont mark mails as processed when the correlation did not work.
- [x] Add activity logs for the correlation results 
- [ ] Add documentation about the breaking change for 8.8 with the message id

sbuettner:
@mathias-vandaele Please dont close this until everything is backported, document and tests are properly added.

mathias-vandaele:
Here is the [PR](https://github.com/camunda/camunda-docs/pull/5757) for documentation

PR #4742, opened 2025-05-09
fix(email-connector): add correlation result management and and stop removing the brackets from messageID

mathias-vandaele:
## Description

Fix:

- CorrelationResult is now properly managed.
- MessageId is not stripped from its brackets

## Related issues

closes https://github.com/camunda/connectors/issues/4740

## Checklist

### DIFF
--- a/connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
+++ b/connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
@@ -52,6 +52,19 @@ public class PollingManager {
     this.store = store;
   }
 
+  private static ReadEmailResponse createResponse(Email email, List<Document> documents) {
+    return new ReadEmailResponse(
+        email.messageId(),
+        email.from(),
+        email.headers(),
+        email.subject(),
+        email.size(),
+        email.body().bodyAsPlainText(),
+        email.body().bodyAsHtml(),
+        documents,
+        email.receivedAt());
+  }
+
   public static PollingManager create(
       InboundConnectorContext connectorContext, JakartaUtils jakartaUtils) {
     Store store = null;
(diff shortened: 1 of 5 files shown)


sbuettner (approved):

connectors/email/src/test/java/io/camunda/connector/email/integration/BaseEmailTest.java
```diff
@@ -214,8 +214,7 @@ protected List<String> getTo(Message message) {
   protected String getLastMessageId() {
     Message message = getLastReceivedEmails()[0];
     try {
-      String messageId = message.getHeader("Message-ID")[0];
-      return messageId.trim().replaceAll("[<>]", "");
+      return message.getHeader("Message-ID")[0];
```
sbuettner, line 217: This method could use: https://github.com/camunda/connectors/pull/4742/files#diff-0cdc6a9ff56610a7c6af5e92b5942f5bef23354885215b037ae4bd82247fade6R313

mathieu-stennier:
This is a breaking change potentially if customers were relying on the messageId to be without brackets? If they somehow used it for correlation themselves?

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
```diff
@@ -150,62 +214,62 @@ private void processMail(IMAPMessage message, PollingConfig pollingConfig) {
...
-      return email.body().attachments().stream()
-          .map(
-              document ->
-                  connectorContext.create(
-                      DocumentCreationRequest.from(document.inputStream())
-                          .contentType(document.contentType())
-                          .fileName(document.name())
-                          .build()))
-          .toList();
+  private boolean process(Email email) {
```
ztefanie, line 232: I think this naming is confusing. This function is named `process` and has an Email as input parameter and above we have the function processEmail. Also it is unclear / confusing, what these booleans are that these mails return.

I think it would be clearer what is happening here, if we have functions that do the evaluation, like canActivate, canCorrelate. And the functions that do the actual workload.

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
```diff
@@ -150,62 +214,62 @@ private void processMail(IMAPMessage message, PollingConfig pollingConfig) {
...
+    this.connectorContext.log(
+        Activity.level(Severity.INFO)
+            .tag("new-email")
+            .message("Processing email: %s".formatted(email.messageId())));
+    ActivationCheckResult activationCheckResult =
+        this.connectorContext.canActivate(createResponse(email, List.of()));
+    return switch (activationCheckResult) {
+      case ActivationCheckResult.Failure failure ->
+          switch (failure) {
```
ztefanie, line 241: This nested switch case is also hard to follow.

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/utils/JakartaUtils.java
```diff
@@ -310,9 +310,16 @@ private void processBodyPart(
     }
   }
 
-  private String stripMessageId(String messageId) {
-    if (messageId == null) return null;
-    return messageId.trim().replaceAll("[<>]", "");
+  private String getMessageId(Message message) {
+    try {
+      String[] messageIds = message.getHeader("Message-ID");
```
ztefanie, line 315: This can throw a Nullpointer exception. so maybe this would be more robust: `return (messageIds != null && messageIds.length > 0) ? messageIds[0] : null;`

ztefanie (approved):
Few comments from my side, mostly about cleaner structure, but this does not to be done in a hotfix.

The nullpoint exception should be handled now tough, i think

sbuettner (approved):

ztefanie:
@mathias-vandaele do we need to backport this to 8.7 and 8.6?

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/utils/JakartaUtils.java
sbuettner: Good point! @mathias-vandaele was this fixed?

mathias-vandaele:
@ztefanie `8.8` it should not be backported

PR #4762, opened 2025-05-13
refactor(agentic-ai): Update model parameters to be provider specific

maff:
## Description

Initially we used a common set of model parameters (e.g. max output tokens, topP) for all providers, but the provider APIs differ in which parameters are actually supported. For example OpenAI and Bedrock do not support topK and the maximum tokens configuration varies by provider. 

This PR:

1. Adds a dedicated set of model parameters targeted to the specific provider API
2. Adds tooltips to each parameter referencing the actual provider documentation
3. Refactor the parameter handling in the `ChatModelFactory` to account for null values
4. Adds tests for `ChatModelFactory` verifying that the chat models are built as expected

## Checklist


reiballa (approved):
LGTM 🚀

PR #4764, opened 2025-05-14
feat(element-template,agentic-ai): Add support to configure a default result variable & result expression

### LINKED ISSUE #4729: Agentic AI: set a default value for agent context + for result expression/result variable

Right now, it is necessary to set an agent context input + a matching result expression/variable when setting up a new agent. To make setting up a new agent easier, it would be beneficial to set a default variable for both fields.

While this has some implications when using multiple agents in the same process as each agent needs its own context variable, this would reduce the amount of fields which need to be configured for every new agent.

Image

- set a default value for Agent Context
- set a default value for either result variable or result expression matching the configured Agent Context

__Additional context__

This seems to be currently not possible using the `@ElementTemplate` generator, so we might need to implement that first.

maff:
## Description

Adds support to configure a default `resultVariable` and/or `resultExpression` via `@ElementTemplate` for both inbound and outbound connectors & configures a default result variable on the Agentic AI connector.

Related discussion: https://camunda.slack.com/archives/C02JLRNQQ05/p1746616997970169

## Related issues

closes #4729
related to https://github.com/camunda/team-connectors/issues/1049

## Checklist

### DIFF
--- a/connectors/agentic-ai/element-templates/agenticai-aiagent-outbound-connector.json
+++ b/connectors/agentic-ai/element-templates/agenticai-aiagent-outbound-connector.json
@@ -635,6 +635,7 @@
     "label" : "Agent Context",
     "description" : "The agent context variable containing all relevant data for the agent to support the feedback loop between user requests, tool calls and LLM responses. Make sure this variable points to the <code>context</code> variable which is returned from the agent response.",
     "optional" : false,
+    "value" : "=agent.context",
     "constraints" : {
       "notEmpty" : true
     },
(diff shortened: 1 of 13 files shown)


johnBgood (approved):

Szik:
@maff do you need some QA here? if so, do you have API keys for testing? otherwise changes so far okay

maff:
@Szik I primarily tagged you as the change mostly affects the core element template generator. However as the change does not lead to any changes in existing generated templates + the changes in the Agentic AI template are reflected as expected I don't think we need additional QA (but the approval still 😉).

Szik (approved):

Issue #4765, opened 2025-05-14
IDP Connector: Handle duplicate keys on structured extraction


reiballa:
__What should we do?__

We will save all duplicate keys on the extracted data. They will have an additional numerator suffix in the end. 

__Why should we do it?__

We need to save all keys on structured extraction, even if they have the same value.

PR #4766, opened 2025-05-14
fix(idp-extraction-connector): added handling of duplicate keys and checkboxes on structured extraction

reiballa:
## Description

This code does 2 things: 
- Adds logic to save all duplicate keys as a new key with a numerator. So if we have three "name" keys they will be returned as "name" "name 2" and "name 3" then the space will be substituted with whatever the delimiter is.
- Handles checkboxes and maps their values to "true" or "false"
 
Additionally i added this fix:
- Fixes the code to use both text-extraction and document-analysis for textract

## Related issues

closes https://github.com/camunda/connectors/issues/4765

## Checklist

### DIFF
--- a/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/DocumentAiCaller.java
+++ b/connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/DocumentAiCaller.java
@@ -89,15 +89,26 @@ public class DocumentAiCaller {
   private StructuredExtractionResponse extractFormFieldsWithConfidence(Document document) {
     Map<String, String> keyValuePairs = new HashMap<>();
     Map<String, Float> confidenceScores = new HashMap<>();
+    Map<String, Integer> keyOccurrences = new HashMap<>();
 
     // Process form fields from Document AI response
     for (Document.Page page : document.getPagesList()) {
       for (Document.Page.FormField formField : page.getFormFieldsList()) {
         if (formField.hasFieldName() && formField.hasFieldValue()) {
-          String name = getTextFromLayout(document, formField.getFieldName().getTextAnchor());
-          String value = getTextFromLayout(document, formField.getFieldValue().getTextAnchor());
+          String originalKey = getTextFromLayout(formField.getFieldName().getTextAnchor()).trim();
+          String key = originalKey;
+          String value = getValueFromFormField(formField).trim();
 
-          if (name != null && !name.trim().isEmpty()) {
+          // Handle duplicate keys by adding a suffix
+          if (keyValuePairs.containsKey(key)) {
+            int count = keyOccurrences.getOrDefault(originalKey, 1) + 1;
+            keyOccurrences.put(originalKey, count);
+            key = originalKey + " " + count;
+          } else {
+            keyOccurrences.put(originalKey, 1);
+          }
+
+          if (!key.isEmpty()) {
             // Get confidence scores from both name and value fields
             float nameConfidence = formField.getFieldName().getConfidence();
             float valueConfidence = formField.getFieldValue().getConfidence();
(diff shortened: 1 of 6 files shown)


connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/DocumentAiCaller.java
```diff
@@ -89,75 +89,58 @@ public StructuredExtractionResponse extractKeyValuePairsWithConfidence(
...
         if (formField.hasFieldName() && formField.hasFieldValue()) {
-          String name = getTextFromLayout(document, formField.getFieldName().getTextAnchor());
-          String value = getTextFromLayout(document, formField.getFieldValue().getTextAnchor());
+          String originalKey = getTextFromLayout(formField.getFieldName().getTextAnchor()).trim();
+          String key = originalKey;
+          String value = getValueFromFormField(formField).trim();
+
+          // Handle duplicate keys by adding a suffix
+          if (keyValuePairs.containsKey(key)) {
+            int count = keyOccurrences.getOrDefault(originalKey, 1) + 1;
```
maff, line 104: ❓ Would this ever emit a `name 1` here or would it do `name`, `name 2`? Can we cover this with tests?

connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/PollingTextractCaller.java
```diff
@@ -56,56 +68,110 @@ public StructuredExtractionResponse extractKeyValuePairsWithConfidence(
...
 
     List<Block> allBlocks;
     try (ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor()) {
-      final String jobId = response.jobId();
-      final TextractTask firstTextractTask =
-          prepareTextractTextDetectionTask(jobId, textractClient);
-      final GetDocumentTextDetectionResponse firstDocumentResult =
-          executeTextDetectionTask(firstTextractTask, 0, executorService);
+      Object firstTask =
+          processType == TextractProcessType.DOCUMENT_ANALYSIS
```
maff, line 118: 💭 As we're doing a lot of these type comparisons, could we extract the type specific handling into 2 distinct handler classes implementing the type specific logic? E.g. a `TextDetectionHandler` and a `DocumentAnalysisHandler` implementing the type specific logic and returning the list of blocks? In this state I would consider it pretty hard to follow and to extend (e.g. will there be a third type at some point)?

connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/PollingTextractCaller.java
```diff
@@ -115,35 +181,75 @@ private List<Block> processDocument(
...
                     && block.entityTypes().contains(EntityType.KEY))
         .forEach(
             keyBlock -> {
-              String key = getTextFromRelationships(keyBlock, blockMap);
+              String originalKey = getTextFromRelationships(keyBlock, blockMap);
+              String key = originalKey;
+
+              // Handle duplicate keys by adding a suffix
+              if (keyValuePairs.containsKey(key)) {
+                int count = keyOccurrences.getOrDefault(originalKey, 1) + 1;
```
maff, line 246: ❓ Same as above regarding the sequence.

connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/PollingTextractCaller.java
```diff
@@ -115,35 +181,75 @@ private List<Block> processDocument(
...
 
+    return new StructuredExtractionResponse(keyValuePairs, confidenceScores);
+  }
+
+  private void extractKeyValueSet(
+      List<Block> blocks,
+      Map<String, Block> blockMap,
+      Map<String, String> keyValuePairs,
+      Map<String, Integer> keyOccurrences,
+      Map<String, Float> confidenceScores) {
```
maff, line 233: 🔧 It's pretty hard to follow that the maps which are passed into this method are mutable and being mutated within this method. Is there really a point in extracting this logic into its own method opposed to keeping it in `extractDataFromDocument` where the variables are initialized? Why does `extractDataFromDocument` need to know about `keyOccurrences` if they are only used inside `extractKeyValueSet`?

I would suggest to either:

- keep the logic in the same method
- or only pass the needed data (the blocks) to the `extractKeyValueSet` method and to return relevant results (`keyValuePairs`, `confidenceScores`) in a some data structure/record

connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/DocumentAiCaller.java
reiballa: if the first one is name, the next should be name 2 etc. added tests for it

connectors/idp-extraction/src/main/java/io/camunda/connector/idp/extraction/caller/PollingTextractCaller.java
reiballa: This is valid, if its ok with you i will rework this in a seprarte ticket, so this doesnt blow out of proportion

maff (approved):
LGTM with the discussed refactorings done in a follow-up 👍

Issue #4772, opened 2025-05-19
Handle unsupported multipart/form-data parts gracefully


ztefanie:
__What should we do?__
When an unsupported / invalid part is entered for a value in a multipart/form-data body, like this:
```
{
  "address" : {  //this is invalid as only string and file are allowed
    "city" : "Berlin", 
    "street" : "Teststr"
  },
  "id" : "1234",
  "profileImage" : result.documents[1]
}
```

A NullPointerException is throw. We should provided a more meaningful error message to the user

Suggestion:
```java
catch (IllegalArgumentException e) {
      LOG.error(
          "Failed to convert the provided map {} to CloudFunctionFilePart. A request with Content-Type: multipart/form-data can only contain maps that are convertible to CloudFunctionFilePart.",
          map,
          e);
    }
```
we can just throw the exception using a ConnectorInputException
and catch more exceptions there

PR #4774, opened 2025-05-19
fix(rest): Handle unsupported multipart/form-data parts gracefully

ztefanie:
## Description
Added NullpointerException to catch, as this is thrown for input like this:
```
{
  "address" : {  //this is invalid as only string and file are allowed
    "city" : "Berlin", 
    "street" : "Teststr"
  }
}
```

I decided to not throw a ConnectorInputExcpetion for this, as this would be a breaking change for process that currently have invalid input that leads to an `IllegalArgumentException` and therefore ignoring this part, but succeeding the process. Feel free to disagree, then I can change this.

## Related issues

closes https://github.com/camunda/connectors/issues/4772

## Checklist

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/DocumentAwareMultipartEntityBuilder.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/DocumentAwareMultipartEntityBuilder.java
@@ -108,7 +108,7 @@ public class DocumentAwareMultipartEntityBuilder {
           new ByteArrayInputStream(part.content()),
           getContentType(part.contentType()),
           part.fileName());
-    } catch (IllegalArgumentException e) {
+    } catch (IllegalArgumentException | NullPointerException e) {
       LOG.error(
           "Failed to convert the provided map {} to CloudFunctionFilePart. A request with Content-Type: multipart/form-data can only contain maps that are convertible to CloudFunctionFilePart.",
           map,


johnBgood (approved):

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/builder/parts/DocumentAwareMultipartEntityBuilder.java
```diff
@@ -108,7 +108,7 @@ private void handleFilePartContent(Map map) {
           new ByteArrayInputStream(part.content()),
           getContentType(part.contentType()),
           part.fileName());
-    } catch (IllegalArgumentException e) {
+    } catch (IllegalArgumentException | NullPointerException e) {
```
johnBgood, line 111: might be nice to add in the documentation that fields other than documents ( as json objects) and text will be ignored. 
I'm not sure if it makes sense as this is how Parts work so it's your call :)

ztefanie: ok, the hint is added to the docs with this PR: https://github.com/camunda/camunda-docs/pull/5851

johnBgood: nice! and should we backport this to 8.7 ?

PR #4793, opened 2025-05-22
refactor(agentic-ai): introduce custom data abstractions for messages and tool calling

### LINKED ISSUE #4777: Agentic AI: Abstraction layer for data for memory
__What should we do?__

__Why should we do it?__

maff:
## Description

Introduces custom abstractions for everything related to messages and tool calling to avoid directly relying on Langchain4J's data structures which do not work so well for how we serialize data in processes as they are not directly serializable. This PR does a couple of things to lay the foundations of future implementations:

- Framework abstraction: introduces a `framework` concept which is responsible for doing the actual framework interaction. The Langchain4J framework implementation accepts a list our message data models, converts them the L4J format, executes the LLM call and converts the response back to our message format.
- Memory handling: introduces a concept to handle our message structures in a `ConversationStore` and `ConversationContext` abstraction. This is implemented as in-process only, but can be extended to multiple providers in https://github.com/camunda/product-hub/issues/2882
- Response format handling. Users can now choose it they want to have only the response text as `responseText`, the whole assistant message including metadata as `responseMessage`, or both by configuring the `Response` section in the properties panel.

In addition to the functional changes, this PR moves all beans (which could be customized in a custom deployment) to implement an interface + to provide default impl.

## Related issues

closes #4777 

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AiAgentRequestHandlerImpl.java
```diff
@@ -0,0 +1,242 @@
...
+    addSystemPromptIfNecessary(runtimeMemory, requestData);
+    addUserMessagesFromRequest(agentContext, runtimeMemory, requestData);
+
+    // call framework with memory
+    AiFrameworkChatResponse<?> frameworkChatResponse =
+        framework.executeChatRequest(request, agentContext, runtimeMemory);
+    agentContext = frameworkChatResponse.agentContext();
+
+    final var assistantMessage = frameworkChatResponse.assistantMessage();
+    runtimeMemory.addMessage(assistantMessage);
```
maff, line 94: This is where we interact with the framework (Langchain4J): we send the messages in our format into the framework and expect a response with the assistant message and tool calls already being converted to our format.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/model/tool/ToolCallProcessVariable.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.Map;
+
+/**
+ * Tool call process variable representation which moves arguments to the upper level and id/name to
+ * a "_meta" object.
+ *
+ * <p>This is needed to allow accessing tool call arguments from the root of the object in
+ * fromAi(toolCall.myParameter) vs. fromAi(toolCall.arguments.myParameter).
+ */
+public record ToolCallProcessVariable(
```
maff, line 21: The in-process reference of a tool call, moving id and name to `_meta`.

reiballa (approved):
🚀

PR #4823, opened 2025-06-02
test(agentic-ai): fix e2e tests breaking due to renamed CPT variable method

maff:
Fixes e2e tests breaking du to renamed CPT method.


sbuettner (approved):
🚀

PR #4842, opened 2025-06-04
fix(saas): credentials cache path not applied

chillleader:
## Description

In https://github.com/camunda/connectors/pull/4836 we switched to the new config cache path variable.

This caused our SNAPSHOT to break on SaaS because we are constructing a custom instance of Camunda client in the SaaS environment, and we didn't explicitly apply the new credentials cache path when building it.

Before that change, it worked because we provided the lower-level env var defined in the Zeebe client which took effect even though the new property was not provided.

## Related issues

closes #

## Checklist


johnBgood (approved):

chillleader:
Backporting failure is expected - I added the label by mistake, no backport needed for 8.7 as we don't construct the client ourselves there.

Issue #4810, opened 2025-05-26
Plus sign encoding issues with Google HTTP client


chillleader:
__What should we do?__

- In the HTTP client, we have a [test](https://github.com/camunda/connectors/blob/9b4d2e764b3ec3df7ddb8049aba6fdd04ac3c77b/connectors/http/http-base/src/test/java/io/camunda/connector/http/base/client/apache/builder/parts/UrlEncoderTest.java#L113) that checks that '+' signs are URL encoded only when they are part of the query param. If they are part of the path, we don't expect to encode them.
- We use Google's HTTP client to take care of the encoding
- The HTTP client team at Google has recently decided they want to always encode the '+' sign no matter where in the URL - apparently due to their own API's behavior: https://github.com/googleapis/google-http-java-client/pull/2094
- This breaks our test.

To mitigate, we keep using Google HTTP client 1.45.2 instead of upgrading.

We should determine whether we want to stay compliant with the previous decision to encode the '+' sign only in particular cases
 - If yes => We should not rely on the HTTP client from Google for URL encoding.
 - If no => Update tests and remove the fixed Google client version fix

sbuettner:
What if we use this one: https://github.com/googleapis/google-http-java-client/blob/f77f40880e171ab6124b6fd34653a654d093577d/google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java#L35 ? 

As mentioned in the issue @chillleader  linked above: 
> Note we still do NOT want + signs to be encoded for SAFEPATHCHARS_URLENCODER, as we still want to utilize it for reserved expansions See issue

johnBgood:
I think their PR is valid.

> Currently we are not encoding + signs if it part of the URL path, while this [is to spec](https://www.rfc-editor.org/rfc/rfc1738.html)
> Thus, only alphanumerics, the special characters "$-_.+!*'(),", and
> reserved characters used for their reserved purposes may be used
> unencoded within a URL.

sbuettner:
As discussed in the grooming, we will investigate as the linked ticket mentions that `+` signs should still be supported for URL paths.

sbuettner:
As discussed with @chillleader. Lets add the Google HTTP Client library as a dependency to `http-base` as its currently implicitly imported by the `google-oauth2` library.

PR #4878, opened 2025-06-13
fix(deps): declare google http client dependency directly

chillleader:
## Description

We use Google's HTTP client for URL encoding. This PR declares it as a direct dependency instead of using it transitively.

This is a backported subset of changes introduced in #4877 

## Related issues

related #4810 

## Checklist


sbuettner:
Successfully created backport PR for `release/8.6`:
- #4879

PR #4899, opened 2025-06-17
WIP: Agentic AI: document storage suppoer

maff:
## Related issues

closes #

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/AiAgentFunction.java
```diff
@@ -23,7 +23,7 @@
     id = "io.camunda.connectors.agenticai.aiagent.v1",
     name = "AI Agent",
     description =
-        "Provides a generic AI Agent implementation handling the feedback loop between user requests, tool calls and LLM responses. Compatible with 8.8.0-alpha5 or later.",
+        "Provides a generic AI Agent implementation handling the feedback loop between user requests, tool calls and LLM responses. Compatible with 8.8.0-alpha6 or later.",
```
johnBgood, line 26: I think we should remove this part: "Compatible with ...." for different reasons:
- We talking about Connectors Runtime version here, when users reason in terms of "Camunda version" + we might release alpha6.1 or 6.2 if we discover some bugs before the release (it happens before).
- We now have the `engineVersion`, this information should be enough for the modeler to provide hints about this template usage.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/document/CamundaDocumentConversationStore.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CamundaDocumentConversationStore
+    extends BaseConversationStore<CamundaDocumentConversationContext> {
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(CamundaDocumentConversationStore.class);
+  private static final int DEFAULT_PREVIOUS_DOCUMENTS_RETENTION_SIZE = 2;
```
johnBgood, line 38: same question here

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AiAgentRequestHandlerImpl.java
```diff
@@ -26,9 +26,10 @@
 
 public class AiAgentRequestHandlerImpl implements AiAgentRequestHandler {
 
-  private static final int DEFAULT_MAX_MEMORY_MESSAGES = 20;
+  private static final int DEFAULT_CONTEXT_WINDOW_SIZE = 20;
```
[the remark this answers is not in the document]
maff: This one is configurable already, this value is just the default fallback in case it's not configured (also for backwards compatibility).

johnBgood (approved):
minor questions only :)

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/AiAgentFunction.java
maff: We had to introduce this comment when we added the AI Agent templates to the web modeler codebase for alpha5 to make it very explicit which alpha version the connector template is compatible with. We move a lot of things at the moment, features are missing/moved or will break/change behavior between alpha versions and we want early adopters to be aware which cluster version is needed for the template.

As the modeler version selection only targets minor versions (I can select 8.7 or 8.8) my understanding was that the engine version setting is not enough to restrict a template to a specific alpha version. However, this is definitely a temporary approach which we'll remove at some point before approaching the 8.8 release.

johnBgood: Usually it's ok to break things in alphas, but perfectly fine if it's removed later on

PR #4902, opened 2025-06-17
others(kafka-connector): introducing schema property group and others minor changes

mathias-vandaele:
## Description

Create a separate property group for "Schema"

Move Headers to the "Message" group for the Outbound Connector

Rename "Inline Schema" to "Avro Inline Schema"

Rename the Schema Registry Type from "Avro (alpha)" to just "Avro"

Rename the "Schema Registry" to "Confluent Schema Registry"

Refine the naming of the decode method from `decode` to `envelope`

## Related issues

closes #

## Checklist


sbuettner:
@johnBgood as you also worked on the Connector would be great to get your opinion here as well

PR #4908, opened 2025-06-18
chore(element-templates): Fix corrupted element templates without engine

mathias-vandaele:
## Description

Fixes all corrupted element templates without a runtime engine and fix uniquet file naming


sbuettner (commented):
Minor nit @mathias-vandaele but can we add the engine version to the top of the element templates? Now you have to scroll to the bottom to find this. For example after `elementType`?

PR #4918, opened 2025-06-19
other(preview-env): enable inbound polling again

johnBgood:
Enable Inbound polling for preview envs


ztefanie (approved):
Changes look valid to me, although i do not fully understand.

It would be nice to test before merging. Is the failing deploy-preview caused by the changes here?

Issue #4912, opened 2025-06-18
Add `@ElementTemplate` annotation to Connectors Missing It


mathias-vandaele:
__What should we do?__

Add the `@ElementTemplate` annotation to the HttpPollingConnector and audit other connectors to ensure none are missing it.

__Why should we do it?__

Having the `@ElementTemplate` annotation consistently applied ensures all connectors expose a standardized and reliable element template, improving usability and configuration consistency in the modeler.

PR #4953, opened 2025-06-24
others(etg): add `ElementTemplate` annotation to polling connector, add versioned templates

mathias-vandaele:
## Description

This pull request introduces the `@ElementTemplate` annotation to the polling connector.

The Element Templates have been standardized and a new version has been created due to lots of updates.

I have tested it, and the feature remains consistent. Regarding correlation and activation, we are now using the ETG-generated version instead of the custom implementation.

## Related issues

closes https://github.com/camunda/connectors/issues/4912

## Checklist


sbuettner (commented):
Nice!

Please test whether the newly generated template works for example with an 8.7 cluster in SaaS.

johnBgood (approved):

Issue #4938, opened 2025-06-23
Support `client_assertion` flow for Operate in Connectors 8.6 and 8.7


chillleader:
__What should we do?__

We need to ensure Connectors are capable of using the `client_assertion` flow in 8.6 and 8.7

8.6:
- We use Spring Zeebe 8.5.18 and it uses Identity SDK 8.5.5 to obtain an access token for Zeebe and Operate
- There is a new version of Identity SDK (8.6.15) that supports this flow out of the box
- We need to adjust the Spring Zeebe code to use the new Identity SDK with it, then release Spring Zeebe and use it in the new connectors release

8.7:
- Zeebe client support comes with the official Spring SDK (already available)
- Operate: we are using a different [client](https://github.com/camunda-community-hub/camunda-operate-client-java) that doesn't support this
  - We need to modify the Operate client to enable this feature; release the operate client and use it in the new connectors release
  - Draft implementation: https://github.com/camunda/connectors/pull/4924

__Why should we do it?__

http://github.com/camunda/product-hub/issues/2850

PR #4964, opened 2025-06-25
chore(deps): bump operate client to 8.7.1-rc1

chillleader:
## Description

This PR upgrades the Operate client to the new version where `client_assertion` flow is supported

## Related issues

related #4938 

## Checklist


chillleader:
Yes, we'll need to upgrade to a non-rc version later manually - but first we need to verify that the rc version works as expected and then give a green light to release the final one

PR #4965, opened 2025-06-25
Agentic AI: Set gateway handler type zeebe property on MCP element types

### LINKED ISSUE #4947: Agentic AI: Resolve MCP clients only by property, not by service task type

Currently, MCP gateway connectors/subflows are detected:

1. [by service task type](https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/discovery/McpClientGatewayToolDefinitionResolver.java#L51)
2. by checking if the tool is [tagged with a property](https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/discovery/McpClientGatewayToolDefinitionResolver.java#L47) `io.camunda.agenticai.gateway.type=mcpClient`
3. 
To make this solution more portable and to account for not having the service task type available when the AI Agent will be implemented on top of the ad-hoc sub-process we should remove option 1 and only analyze the property.

1. Configure the MCP client element templates to set the following property. This might imply adding support for properties to the element template generator.
    ```xml
          <zeebe:properties>
            <zeebe:property name="io.camunda.agenticai.gateway.type" value="mcpClient" />
          </zeebe:properties>
    ```
2. Remove discovery based on service task type

__Additional context__

Keeping the current approach, using 2 different ways for the same functionality and not working with the future approach.

maff:
## Description

To add a zeebe property of `io.camunda.agenticai.gateway.type=mcpClient` to via generated MCP client connector templates, it is necessary to add the following property binding to the generated element templates:

```
{
    "value" : "mcpClient",
    "binding" : {
      "name" : "io.camunda.agenticai.gateway.type",
      "type" : "zeebe:property"
    },
    "type" : "Hidden"
  }
```

This PR adds an additional way to define properties generated for an element template by referencing `@PropertySource` annotated static methods which can return a list of properties of any kind, making it possible to define arbitrary properties without having to define them as part of the input data.

## Related issues

closes #4947

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/McpClientFunction.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/McpClientFunction.java
@@ -25,6 +25,7 @@ import io.camunda.connector.generator.java.annotation.ElementTemplate;
     engineVersion = "^8.8",
     version = 0,
     inputDataClass = McpClientRequest.class,
+    propertySources = McpClientPropertySource.class,
     defaultResultVariable = "toolCallResult",
     propertyGroups = {
       @ElementTemplate.PropertyGroup(id = "client", label = "MCP Client"),
--- a/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
+++ b/element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
@@ -26,16 +26,22 @@ import io.camunda.connector.generator.api.GeneratorConfiguration.GenerationFeatu
 import io.camunda.connector.generator.dsl.*;
 import io.camunda.connector.generator.dsl.PropertyGroup.PropertyGroupBuilder;
 import io.camunda.connector.generator.java.annotation.ElementTemplate;
+import io.camunda.connector.generator.java.annotation.PropertySource;
 import io.camunda.connector.generator.java.util.ReflectionUtil;
 import io.camunda.connector.generator.java.util.TemplateGenerationContext;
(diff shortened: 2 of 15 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/McpClientFunction.java
```diff
@@ -25,6 +25,7 @@
     engineVersion = "^8.8",
     version = 0,
     inputDataClass = McpClientRequest.class,
+    propertySources = McpClientPropertySource.class,
```
maff, line 28: References a class containing methods annotated with `@PropertySource`.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
```diff
@@ -233,4 +244,64 @@ private List<PropertyGroup> addServiceProperties(
...
+                    "Failed to resolve template properties from method %s#%s annotated with @PropertySource"
+                        .formatted(method.getDeclaringClass().getTypeName(), method.getName()),
+                    e);
+              }
+            })
+        .flatMap(Collection::stream)
+        .toList();
+  }
+
+  private static List<Pair<Method, PropertySource>> findAllPropertySources(Class<?> type) {
```
maff, line 272: Similar to `ClassBasedDocsGenerator#findAllDataExampleMethods`.

johnBgood (dismissed):
Really nice!

gbetances089 (approved):

johnBgood (approved):

reiballa (approved):
LGTM 🚀

PR #4980, opened 2025-06-26
feat(csv): Add CSV Connector and introduce @Operation based Connectors

sbuettner:
## Description

Add CSV Connector based on a new `@Operation` annotated outbound Connector concept.

Future topics not considered for the first version:
* Support `@Variable` from the Camunda Spring Client

## Related issues

https://github.com/camunda/product-hub/issues/2851

## Checklist

### DIFF
--- a/connectors/csv/src/main/java/io/camunda/connector/csv/model/WriteCsvResult.java
+++ b/connectors/csv/src/main/java/io/camunda/connector/csv/model/WriteCsvResult.java
@@ -10,5 +10,5 @@ public sealed interface WriteCsvResult permits WriteCsvResult.Document, WriteCsv
 
   record Document(io.camunda.document.Document document) implements WriteCsvResult {}
 
-  record Value(String csv) implements WriteCsvResult {}
+  record Value(String content) implements WriteCsvResult {}
 }
--- a/connectors/csv/src/test/java/io/camunda/connector/csv/CsvConnectorTests.java
+++ b/connectors/csv/src/test/java/io/camunda/connector/csv/CsvConnectorTests.java
@@ -106,7 +106,7 @@ public class CsvConnectorTests {
     var result = (WriteCsvResult.Value) connector.writeCsv(request, context);
     assertNotNull(result);
     assertEquals(
-        "name,role\r\nSimon,Engineering Manager\r\nMathias,Backend Engineer\r\n", result.csv());
+        "name,role\r\nSimon,Engineering Manager\r\nMathias,Backend Engineer\r\n", result.content());
   }
 
   @Test
@@ -121,7 +121,7 @@ public class CsvConnectorTests {
             new CsvFormat(",", true, asList("name", "role")));
     var result = (WriteCsvResult.Value) connector.writeCsv(request, context);
     assertNotNull(result);
-    assertEquals("Simon,Engineering Manager\r\nMathias,Backend Engineer\r\n", result.csv());
+    assertEquals("Simon,Engineering Manager\r\nMathias,Backend Engineer\r\n", result.content());
   }
 
   @Test
@@ -137,6 +137,6 @@ public class CsvConnectorTests {
     var result = (WriteCsvResult.Value) connector.writeCsv(request, context);
     assertNotNull(result);
     assertEquals(
-        "name,role\r\nSimon,Engineering Manager\r\nMathias,Backend Engineer\r\n", result.csv());
+        "name,role\r\nSimon,Engineering Manager\r\nMathias,Backend Engineer\r\n", result.content());
   }
 }


connectors/csv/src/main/java/io/camunda/connector/csv/CsvConnector.java
```diff
@@ -0,0 +1,56 @@
...
+import io.camunda.connector.generator.java.annotation.ElementTemplate;
+import io.camunda.document.Document;
+import io.camunda.document.store.DocumentCreationRequest;
+import java.io.InputStreamReader;
+import java.io.StringReader;
+import java.util.Optional;
+
+/** Connector for reading and writing CSV files. */
+@OutboundConnector(name = "CSV Connector", type = "io.camunda:csv-connector")
+@ElementTemplate(name = "CSV Connector", id = "io.camunda.connectors.csv", engineVersion = "^8.7")
```
johnBgood, line 28: is it 8.7?

johnBgood (dismissed):
The introspection part is hard to review, I left a few questions. Would be nice to see the same CSV connector implemented with our current SDK to see the pros/cons

sbuettner:
> The introspection part is hard to review, I left a few questions. Would be nice to see the same CSV connector implemented with our current SDK to see the pros/cons

The main difference is how we will handle the different operations. They would need to live inside the model similar to how the Box Connector works: https://github.com/camunda/connectors/blob/6e4bef7581dfc5250f95f487ed023f9ab554dbf9/connectors/box/src/main/java/io/camunda/connector/box/model/BoxRequest.java#L22 and not as regular code and the dispatch of handling the different Operations would also need to live inside the Connector.

To me this `@Operation` based approach is way more streamligned and simpler to reason about than re-building Operation dispatching and modeling everytime again for each Connector.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java
```diff
@@ -68,7 +75,25 @@ private OutboundConnectorFunction createInstance(OutboundConnectorConfiguration
           if (c.customInstanceSupplier() != null) {
             return c.customInstanceSupplier().get();
           } else {
-            return ConnectorHelper.instantiateConnector(c.connectorClass());
+            if (OutboundConnectorFunction.class.isAssignableFrom(c.connectorClass())) {
```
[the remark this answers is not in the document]
sbuettner: Yes, as the Connector class itself is not implementing the interface but gets wrapped into the `OutboundConnectorOperationFunction`

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
```diff
@@ -86,4 +88,65 @@ public static InboundConnectorConfiguration getRequiredInboundConnectorConfigura
...
+  public static Set<String> getInputVariables(
+      List<ReflectionUtil.MethodWithAnnotation<Operation>> operations) {
+    Set<String> variables = new HashSet<String>();
+    operations.forEach(
+        method -> {
+          method
+              .parameters()
+              .forEach(
+                  parameter -> {
+                    if (parameter.getAnnotation(io.camunda.connector.api.annotation.Variable.class)
```
chillleader, line 123: Do we also want to handle the case when `@Variable` is not present? e.g. try to use the actual parameter names if they are available (if compiled with the `-parameters` flag).

If not, we can make this part easier to read by inverting the if statement
`if (parameter.getAnnotation(Variable.class) == null) { return; }`

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
```diff
@@ -86,4 +88,65 @@ public static InboundConnectorConfiguration getRequiredInboundConnectorConfigura
...
+                          declaredField.setAccessible(true);
+                          String fieldName = declaredField.getName();
+                          variables.add(fieldName);
+                        }
+                      } else {
+                        // If the variable name contains a dot, we take the part before the dot as
+                        // the variable name
+                        if (variableName.contains(".")) {
+                          String[] parts = variableName.split("\\.");
+                          variables.add(parts[0]);
```
chillleader, line 142: Can this be unexpected in certain cases? Like if someone tries to use `parent.child` to reflect nesting 🤔

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ReflectionUtil.java
```diff
@@ -0,0 +1,60 @@
...
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Parameter;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Utility class for reflection operations, specifically for retrieving methods annotated with a
+ * specific annotation. TODO Move to a reusable package so it can be shared with the element
```
chillleader, line 28: or maybe just move to `connector-core`

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/DefaultOutboundConnectorFactory.java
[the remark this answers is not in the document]
johnBgood: nope, but I misread anyway. I thought the Connector implemented some subclass of OutboundConnectorFunction, but it's just a marker interface; so all good

johnBgood (dismissed):
👍 I like how it looks for the customers!
I'm a bit worried about all the introspection part and how easy it'll be to maintain/improve this part of the SDK though 🤔 maybe it's not an issue

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
```diff
@@ -40,15 +42,15 @@ public static Optional<OutboundConnectorConfiguration> getOutboundConnectorConfi
...
-                  annotation.inputVariables(),
+                  getInputVariables(cls, annotation),
                   configurationOverrides.typeOverride().orElse(annotation.type()),
                   cls,
                   configurationOverrides.timeoutOverride().orElse(null));
             });
   }
 
   public static OutboundConnectorConfiguration getRequiredOutboundConnectorConfiguration(
-      Class<? extends OutboundConnectorFunction> cls) {
```
chillleader, line 51: Wouldn't `Class<? extends OutboundConnectorFunction>` also work with `OutboundConnectorOperationFunction`?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/OutboundConnectorOperationFunction.java
```diff
@@ -0,0 +1,138 @@
...
+      Object connector, ValidationProvider validationProvider) {
+    this.connector = connector;
+    this.validationProvider = validationProvider;
+    this.methods = ReflectionUtil.getMethodsAnnotatedWith(connector.getClass(), Operation.class);
+  }
+
+  @Override
+  public Object execute(OutboundConnectorContext context) throws Exception {
+
+    String operationId = context.getJobContext().getCustomHeaders().get("operation");
```
chillleader, line 59: Consider moving the constant to `io.camunda.connector.runtime.core.Keywords`

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/ReflectionUtil.java
```diff
@@ -53,4 +55,32 @@ public static List<Field> getAllFields(List<Field> fields, Class<?> type) {
     }
     return fields;
   }
+
+  public record MethodWithAnnotation<A extends Annotation>(
+      Method method, List<Parameter> parameters, A annotation) {}
```
chillleader, line 60: Do we need these values if they can be accessed via `method::getParameters` and `method::getAnnotation`?

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/OperationBasedConnectorUtil.java
```diff
@@ -0,0 +1,167 @@
...
+        .group("operation");
+  }
+
+  public static String getOperationName(Operation operation) {
+    return !operation.name().isBlank() ? operation.name() : operation.value();
+  }
+
+  public static String getOperationId(Operation operation) {
+    return !operation.id().isBlank() ? operation.id() : operation.value();
+  }
```
chillleader, line 59: this can still return an empty string if nothing is set, maybe we can validate it here

connector-sdk/core/src/main/java/io/camunda/connector/api/annotation/OutboundConnector.java
```diff
@@ -30,7 +30,7 @@
   String name();
 
   /** Input variables the connector reads */
-  String[] inputVariables();
+  String[] inputVariables() default "";
```
chillleader, line 33: Should be `default {}` perhaps? II'm not sure but I suspect one resolves to `[ "" ]`, the other to `[]` 🤔

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
```diff
@@ -99,9 +98,21 @@ public List<io.camunda.connector.generator.dsl.ElementTemplate> generate(
...
-            TemplatePropertiesUtil.extractTemplatePropertiesFromType(connectorInput, context));
+    List<PropertyBuilder> properties;
+    if (OutboundConnectorFunction.class.isAssignableFrom(connectorDefinition)
+        || InboundConnectorExecutable.class.isAssignableFrom(connectorDefinition)) {
+      // An outbound connector function or inbound connector executable
+      properties =
+          new ArrayList<>(
+              TemplatePropertiesUtil.extractTemplatePropertiesFromType(connectorInput, context));
+    } else {
+      // Probably an operation annotated connector
```
chillleader, line 109: We can also check if it's assignable from `OutboundConnectorProvider` to be sure, right?

sbuettner:
Thanks for the feedback @johnBgood & @chillleader I refactored the reflection and introspection part and improved the @Operation annotated method invocation. Let me know whether its clearer now and whether you have feedback how we can improve this further.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
sbuettner: Yes, this part is a leftover when I tried to implement it without the `OutboundConnectorProvider` marker interface solely on the classpath scanning of `@Operation` using https://github.com/classgraph/classgraph.  It works but seems to add some overhead. I will simplify this as suggested.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
sbuettner: Great question. My assumption was that the engine resolves only the top level variable name when fetching variables for jobs. Do you maybe have a reference for the path syntax is being used when requesting job variables?

connectors/csv/src/main/java/io/camunda/connector/csv/CsvConnector.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.InputStreamReader;
+import java.io.StringReader;
+import java.util.Optional;
+
+/** Connector for reading and writing CSV files. */
+@OutboundConnector(name = "CSV Connector", type = "io.camunda:csv-connector")
+@ElementTemplate(
+    name = "CSV Connector",
+    id = "io.camunda.connectors.csv",
+    engineVersion = "^8.7",
```
johnBgood, line 33: You can actually use 8.8 here, it's available in the Desktop modeler dropdown (you probably have an old version).

johnBgood (dismissed):
Looks great! Just left one comment about the connectors version.
thank you 💯

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
sbuettner: Nesting doesnt seem to be supported so "x.y.z" is a valid variable name.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ParameterResolver.java
```diff
@@ -0,0 +1,69 @@
...
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.camunda.connector.api.error.ConnectorInputException;
+import io.camunda.connector.api.outbound.OutboundConnectorContext;
+import io.camunda.connector.runtime.core.outbound.operation.ParameterDescriptor.VariableDescriptor;
+import java.io.IOException;
+
+public sealed interface ParameterResolver<T> {
```
chillleader: Really nice structural improvements 👍

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ConnectorUtil.java
chillleader: Ok good, I agree with the approach not to implement this for now 👍 Thanks for checking

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/ReflectionUtil.java
sbuettner: A method can have several annotations so its easier to have the non-nullable value as part of the result. For parameters yes, we could remove them

chillleader (dismissed):
There are a few minor style/nitpick comments left above but none are dealbreakers. Thanks for addressing the feedback!

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/ReflectionUtil.java
sbuettner: Did exactly that.

sbuettner:
@johnBgood @chillleader Would be great to get a final approval. 

Meanwhile we can start the QA process for the CSV Connector @Szik

Issue #4998, opened 2025-07-01
Agentic AI: Extension API for custom conversation stores


maff:
__Is your feature request related to a problem? Please describe.__

We currently support 2 conversation stores (in-process, camunda document) which are configured from the process definition, but we should allow creating custom conversation stores programatically for custom deployments.

__Describe the solution you'd like__

1. Enrich messages with a timestamp field, so it can be traced when a message was created
2. Introduce a storage "session" which will be created by the conversation store and be alive during the connector execution. This will allow custom implementations to build transactional logic for a conversation store.
3. Allow registering custom session storage on programatically on the connector runtime.

johnBgood:
@maff You can use the `agentic-ai` or `idp` labels so that the issues are not added to the Connectors project :)

PR #5032, opened 2025-07-09
refactor(agentic-ai): clean up/improve conversation/message data

### LINKED ISSUE #5036: Storage API: Refactor message properties and builders
- Rename `ConversationContext::id` to `ConversationContext::conversationId`
- Remove unnecessary builder methods/move test usage from models to test utils
- Add a timestamp to message metadata (all besides the system message)

maff:
## Description

Cleans up/improves the data structures used for a conversation:

- Rename `id` to `conversationId` for better intent (especially relevant when implementing custom storage adapters)
- Remove `properties` from conversation context interface as it is not needed
- Add a timestamp to message metadata (all besides the system message)
- Refactor message builder methods which are primarily used in tests to `TestMessageFixture`

## Related issues

closes #5036 
preparation for #4998

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/util/ClockProvider.java
```diff
@@ -0,0 +1,38 @@
...
+public class ClockProvider {
+  private static Clock clock = defaultClock();
+
+  private ClockProvider() {}
+
+  public static Clock getClock() {
+    return clock;
+  }
+
+  public static void setClock(final Clock newClock) {
```
mdii, line 22: I see we change the clock in the tests only but it would be safer to make the `clock` immutable.

maff: Immutable as in making it a final field? This primary reason for this change was to have the possibility to control the clock in tests by setting a fixed one which would be defeated by not being able to control the clock.

Initially this was part of the messages API and initialized by default for every message, which made way more tests affected. Now it's only 3 places which we can handle in a simpler way. I updated the implementation to directly rely on `ZonedDateTime.now()` without a clock provider in https://github.com/camunda/connectors/pull/5032/commits/d751dae8b544599851c57e27befd162ab7d69b5c.

maff: Agree, that would have been a good alternative approach! 💡 

I think it's OK for the moment to verifying on the closeness as it is now - can you re-review?

Issue #5047, opened 2025-07-10
Update ProcessDefinitionSearch to fetch only latest process definitions


johnBgood:
__What should we do?__

The backend part has been implemented [in core](https://github.com/camunda/camunda/pull/33160).
Now we need to update Connectors to use the new filter: `isLatestVersion`.

__Why should we do it?__

This will prevent Connectors from fetching all existing process definitions, as we currently filter them on the client side.

PR #5048, opened 2025-07-10
feat(inbound): Fetch only the latest process definitions

johnBgood:
## Description

The backend part has been implemented https://github.com/camunda/camunda/pull/33160.
Now we need to update Connectors to use the new filter: isLatestVersion.

## Related issues

closes https://github.com/camunda/connectors/issues/5047

## Checklist

### DIFF
--- a/bundle/default-bundle/src/test/java/io/camunda/connector/runtime/app/DataImportConfiguration.java
+++ b/bundle/default-bundle/src/test/java/io/camunda/connector/runtime/app/DataImportConfiguration.java
@@ -45,6 +45,8 @@ public class DataImportConfiguration {
 
   @EventListener(ApplicationReadyEvent.class)
   public void testClient() {
+    String processAsString = getProcessAsString();
+
     try (ExecutorService executorService = Executors.newFixedThreadPool(10)) {
       List<CompletableFuture<Void>> futures =
           IntStream.range(0, PROCESS_NUMBER)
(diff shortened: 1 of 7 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/search/SearchQueryClientImpl.java
```diff
@@ -25,24 +25,31 @@
...
   private final CamundaClient camundaClient;
 
   public SearchQueryClientImpl(CamundaClient camundaClient) {
     this.camundaClient = camundaClient;
   }
 
+  SearchQueryClientImpl(CamundaClient camundaClient, int limit) {
+    this.camundaClient = camundaClient;
+    if (limit <= 0) {
+      throw new IllegalArgumentException("Page limit must be greater than zero");
```
ztefanie, line 40: Why not log a warning "Set page limit was invalid. Must be greater than zero. Falling back to default page limit 200" and proceed?

bundle/default-bundle/src/test/java/io/camunda/connector/runtime/app/DataImportConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+  private CamundaClient client;
+
+  private static final int VERSIONS_PER_PROCESS = 100;
+  private static final int PROCESS_NUMBER = 500;
+
+  public DataImportConfiguration(CamundaClient client) {
+    this.client = client;
+  }
+
+  @EventListener(ApplicationReadyEvent.class)
```
ztefanie, line 46: Just a small note: the testClient() function is doing quite a lot, and I found it a bit hard to follow. I think readability could benefit from splitting it into a few well-named helper methods.

That said, since it’s test code and probably won’t change often, I totally understand keeping it as-is. Just wanted to share the thought in case you find it useful.

ztefanie (dismissed):
Nice feature :+1: :rocket:

bundle/default-bundle/src/test/java/io/camunda/connector/runtime/app/DataImportConfiguration.java
johnBgood: you're right, I refactored the class a bit. It doesn't hurt to make it clearer :) let me know if that's better

ztefanie: Nice :rocket: Thank you :)

Szik (approved):

PR #5065, opened 2025-07-14
fix(agentic-ai): make prompt parameter handling less invasive

### LINKED ISSUE #5034: Agentic AI: Templating syntax interfers with curly braces in system/user prompt text
__Describe the Bug__

As soon as the payload passed as system/user prompt contains `{{ }}` syntax, it is interpreted as placeholder and fails the request.

In the long run we're planning to remote the parameters in #4871, but for now we should make this as less invasive as possible.

See [Slack Thread](https://camunda.slack.com/archives/C08BHU6JQJG/p1752084695845409) for more information.

__Steps to Reproduce__

1. Add a payload to the user prompt containing curly braces syntax but not meant to be a placeholder. For example `{{ My custom content }}`.
2. Execution will fail with `Value for the variable ' My custom content ' is missing`.

__Expected Behavior__

1. It does not fail when the variable is not defined as placeholder.

__Proposed solution__

1. As long as we support the parameters, make them as less invasive as possible. Iterate a the list of defined parameters and do only string replacements on known parameter names instead of detecting all curly braces usage.
  a. This most likely means not using Langchain4j's prompt implementation anymore. As this will also remove the date parameters provided by default, remove the documentation around availability of `current_date` etc. from the generated template.
3. In the long run remove the parameter syntax (out of scope of this issue)

maff:
## Description

Fix prompt parameter handling to be less invasive and to not replace all curly brace usages:

- Only replace curly brace syntax for defined parameters, not for all usage
- Restrict parameter names to alphanumeric chars and underscores
- Loosen validation on system and user prompt text as it is not strictly required, but the validation assumes so

This introduces a breaking change in terms of the default parameters such as `currentDateTime` not being available anymore as we go away from the default Langchain4j implementation.

## Related issues

closes #5034

## Checklist

### DIFF
--- a/connectors/agentic-ai/CHANGELOG.md
+++ b/connectors/agentic-ai/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 - Added support to provide custom storage backends in a self-managed deployment:
   - https://github.com/camunda/connectors/pull/5035
+- Changed the way how prompt parameters are resolved to be less intrusive. Now only defined parameters
+  are considered in `{{curlyBraces}}` syntax while keeping other usages of curly braces in the prompt text intact.
+  - Breaking change: the default parameters for current date and time are not available by default anymore but can be
+    added on demand with FEEL's temporal functions.
+  - https://github.com/camunda/connectors/pull/5065
 
 ## 8.8.0-alpha6
 


reiballa (approved):
LGTM

PR #5069, opened 2025-07-14
feat(core): the secret provider is tenant aware

jonathanlukas:
## Description

This PR extends the secret provider to be context aware.

This will not change the current behaviour but allow secret providers to respect a context when resolving secrets.

## Related issues

closes #

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/secret/EnvironmentSecretProvider.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/secret/EnvironmentSecretProvider.java
@@ -67,7 +67,7 @@ public class EnvironmentSecretProvider implements SecretProvider {
   }
 
   /**
-   * returns the secret name in format (${prefix}_)${tenantId}_${name}
+   * returns the secret name in format ${prefix}${tenantId}_${name}
    *
    * @param name the secrets' name to find the value for
    * @param context the context of where the secret is originated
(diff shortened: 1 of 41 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/secret/EnvironmentSecretProvider.java
```diff
@@ -46,8 +49,31 @@ public void init() {
...
+   *
+   * @param name the secrets' name to find the value for
+   * @return the final secret name
+   */
+  private String composeSecretName(String name) {
+    return String.format("%s%s", StringUtils.hasText(prefix) ? prefix : "", name);
+  }
+
+  /**
+   * returns the secret name in format ${prefix}${tenantId}${name}
```
johnBgood, line 69: ${prefix}${tenantId}_${name}

johnBgood (dismissed):
Great! you probably already did it, but remember that we need to document this here https://docs.camunda.io/docs/self-managed/connectors-deployment/connectors-configuration/#secrets

jonathanlukas:
@johnBgood here is the docs PR, I added you as reviewer: https://github.com/camunda/camunda-docs/pull/6199

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/secret/EnvironmentSecretProvider.java
jonathanlukas: Done

jonathanlukas:
>perfect, do we need to backport this PR?

I would say we COULD as this is non-breaking. Which versions? Only 8.7 or also 8.6 and 8.5?

jonathanlukas:
The docs PR would then also be extended to the backported versions.

johnBgood:
> > perfect, do we need to backport this PR?
> 
> I would say we COULD as this is non-breaking. Which versions? Only 8.7 or also 8.6 and 8.5?

I recommend 8.7 as this is the stable one.

PR #4843, opened 2025-06-04
fix(agentic-ai,aws-auth): Prohibit usage of AWS default credential chain in SaaS in AI Agent connector

maff:
## Description

Add the same validation as done in https://github.com/camunda/connectors/pull/4835 to the AI Agent connector configuration.

## Related issues

relates to #4835 

## Checklist

- [x] PR has a **milestone** or the `no milestone` label.

PR #5089, opened 2025-07-18
feat(agentic-ai): support Google Vertex AI models

### LINKED ISSUE #4814: Agentic AI: Add support for Google Vertex AI models

Add support for Google Vertex AI models.

- Add provider configuration for Google Vertex AI models
- Authentication should be aligned with other Google Cloud connectors (e.g. GCS)
  - The [Google Cloud Storage Connector](https://docs.camunda.io/docs/next/components/connectors/out-of-the-box-connectors/google-cloud-storage/#authentication) supports a service account JSON credential which we should also support
  - If the Vertex AI API supports API keys we can consider adding support as well
  - We can support default credentials, but we need to make sure to not allow them in SaaS environments. See [respective AWS issue](https://github.com/camunda/connectors/issues/4824) and [how we fixed in for Agentic AI](https://github.com/camunda/connectors/pull/4843/files).
- implement chat model support for new models

__Additional context__

mdii:
## Description

This PR adds the support for Google Vertex AI models in the Agentic-AI connector.
[Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc) is used by L4J for authentication. That is why there is no configuration for authentication.

For testing locally I followed the instruction [here](https://docs.langchain4j.dev/integrations/language-models/google-vertex-ai-gemini#install-google-cloud-cli-optional) and [here](https://cloud.google.com/docs/authentication/set-up-adc-local-dev-environment).

Since `ProviderConfiguration.java` was becoming too large I moved each provider configuration impl into its own class.

## Related issues

closes #4814

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+              feel = Property.FeelMode.optional,
+              constraints = @TemplateProperty.PropertyConstraints(notEmpty = true))
+          String model,
+      @Valid GoogleVertexAiModelParameters parameters) {
+
+    public record GoogleVertexAiModelParameters(
+        @Min(0)
+            @TemplateProperty(
+                group = "model",
+                label = "Maximum tokens",
```
maff, line 64: 🔧 As this is called `maxOutputTokens` on the API docs, can we also call it "Maximum output tokens" here? We should try to align the parameters as much to the provider specific naming as possible to avoid ambiguity.

I know we have a similar construct for Bedrock where we call it `maxTokens` but call the `maxOutputTokens` builder, but there it's again called `maxTokens` on the API.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+                type = TemplateProperty.PropertyType.Number,
+                feel = Property.FeelMode.required,
+                optional = true)
+            Integer maxTokens,
+        @Min(0)
+            @TemplateProperty(
+                group = "model",
+                label = "Temperature",
+                tooltip =
+                    "Floating point number between 0 and 1. The higher the number, the more randomness will be injected into the response. <br><br>Details in the <a href=\"https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference\" target=\"_blank\">documentation</a>.",
```
maff, line 76: 🔧 Unfortunately the range is provider specific and for Google it seems to be between 0 and 2, but can be model specific. We need at least adapt it to what is stated in the docs, but in doubt we can also remove the range and reference the documentation.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+              label = "Model",
+              description =
+                  "Specify the model ID. Details in the <a href=\"https://cloud.google.com/vertex-ai/docs/generative-ai/models\" target=\"_blank\">documentation</a>.",
+              type = TemplateProperty.PropertyType.String,
+              feel = Property.FeelMode.optional,
+              constraints = @TemplateProperty.PropertyConstraints(notEmpty = true))
+          String model,
+      @Valid GoogleVertexAiModelParameters parameters) {
+
+    public record GoogleVertexAiModelParameters(
```
maff, line 60: 💭 Google seems also to support `topK`, would it make sense to add it here (we also have it for Anthropic for example)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/model/tool/ToolCall.java
```diff
@@ -10,10 +10,11 @@
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import io.camunda.connector.agenticai.model.AgenticAiRecord;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 @AgenticAiRecord
 @JsonDeserialize(builder = ToolCall.ToolCallJacksonProxyBuilder.class)
-public record ToolCall(String id, String name, Map<String, Object> arguments)
+public record ToolCall(@Nullable String id, String name, Map<String, Object> arguments)
```
maff, line 17: 💭 I fear Vertex AI not supporting tool call IDs might make some things complicated in the long run. I think it's OK for now, but we might need to think about if it is the right approach to have this nullable (vs. generating an ID on our side if it is missing).

maff (changes requested):
💭 The refactorings like moving the configs to individual classes make sense and are a good improvement, but it would be great if we could split refactorings from feature implementations at least on the commit level as it makes it pretty hard to distinguish the changes during a review.
For example a commit splitting up the existing functionality into the individual classes + a commit adding Vertex AI support on top of that. 

❌ Regarding authentication, the issue was not properly prepared (I updated it accordingly): we can support default credentials, but need to add additional validation for it on SaaS as we [did for AWS in this change](https://github.com/camunda/connectors/pull/4843).

As this is a Google Cloud connection, we should align it with other similar connectors such as the [Google Cloud Storage connector](https://docs.camunda.io/docs/next/components/connectors/out-of-the-box-connectors/google-cloud-storage/#authentication) and support a JSON service account credential for authentication. If Vertex AI has support for an API key, we can also consider adding support for that. We can keep default credentials as an option, but need to add additional validation around it as we did for AWS.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
mdii: It is actually between 0 and 1. The L4J doc is wrong. Since topP is a probability it cannot be more than 1. In the referenced documentation of Vertex AI the range is [0-1].

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/model/tool/ToolCall.java
mdii: I though of generating a random ID but was not sure about the implications.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
maff: I was referring to the temperature which [Google documents](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference#request) as being between 0-2.

mdii: Ah sorry I misread :man_facepalming:

gbetances089 (approved):

maff (approved):
LGTM 🎉

Issue #4399, opened 2025-03-24
External document support


chillleader:
__Is your feature request related to a problem? Please describe.__

Connectors support document handling, however, the functionality is currently limited to Camunda-native documents. By their nature, connectors are integration tools, so it would be useful to be able to reference documents from external systems in a similar way as Camunda documents.

Example: I have a document available for download using an HTTP URL. If a connector accepts documents, I want to be able to just provide the URL instead of doing the preparation work of downloading the document and saving it to the Camunda document store, then referencing it in my connector.

__Describe the solution you'd like__

A usual document payload looks like this:

```
{
    "camunda.document.type": "camunda",
    "storeId": "in-memory",
    "documentId": "b0f97b08-cc00-4f09-ba2d-42fbd658e607",
    "metadata": {
        "contentType": "application/pdf",
        "fileName": "Invoice-3139791273.pdf",
        "size": 36279,
        "customProperties": {}
    }
}
```

We can support a similar data structure for external documents that can be consumed by the Connector runtime, e.g.:

```
{
  "camunda.document.type": "external",
  "url": "https://random-bucket-49b2.s3.us-east-1.amazonaws.com/docs/2e19b51d-73b0-4df1-8b13-4a8b5094ba20"
}
```

For a better user experience, we can consider creating a custom FEEL function so that end users don't have to worry about creating this custom data structure:

```
= document("https://random-bucket-49b2.s3.us-east-1.amazonaws.com/docs/2e19b51d-73b0-4df1-8b13-4a8b5094ba20")
```

__Describe alternatives you've considered__

none so far

__Additional context__

https://github.com/camunda/product-hub/issues/2555 [🔒]

ztefanie:
@chillleader How should this handle authentication? Or is this only supported for unprotected endpoints?

PR #5106, opened 2025-07-22
feat(document-handling): Add external document feature

ztefanie:
## Description
Added the `documents` intrinsic function.
I tested this with the teams and gcs connector.

docs PR: https://github.com/camunda/camunda-docs/pull/6232

## Related issues

closes [#4399](https://github.com/camunda/connectors/issues/4399)

## Checklist

### DIFF
--- a/connector-sdk/document/src/main/java/io/camunda/document/ExternalDocument.java
+++ b/connector-sdk/document/src/main/java/io/camunda/document/ExternalDocument.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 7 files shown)


connector-sdk/document/src/main/java/io/camunda/document/ExternalDocument.java
```diff
@@ -0,0 +1,135 @@
...
+    try (InputStream inputStream = asInputStream()) {
+      return inputStream.readAllBytes();
+    } catch (IOException e) {
+      return null;
+    }
+  }
+
+  @Override
+  public DocumentReference reference() {
+    return null;
```
ztefanie, line 128: Is it ok to return null here?

sbuettner: @ztefanie @chillleader What about an exception?

connector-sdk/document/src/main/java/io/camunda/document/ExternalDocument.java
```diff
@@ -0,0 +1,135 @@
...
+        connection = (HttpURLConnection) urlObj.openConnection();
+        connection.setRequestMethod("GET");
+        connection.connect();
+        connection.getContentLengthLong();
+        connection.getContentType();
+      }
+      this.connection = connection;
+
+      metadata = createMetadata(name);
+    } catch (IOException e) {
```
sbuettner, line 49: Shouldnt this go through the regular proxy enabled http base client?

sbuettner: We should also be careful doing io heavy operations inside the constructor. If we decided to keep the whole document in memory we could do the fetching outside of this class and just pass the data as a byte[].

connector-sdk/document/src/main/java/io/camunda/document/ExternalDocument.java
johnBgood: can't we just create a `DocumentReference.ExternalDocumentReference`?

ztefanie:
closing this in favor of: https://github.com/camunda/connectors/pull/5259

PR #5108, opened 2025-07-22
refactor(agentic-ai): remove AgentState.WAITING_FOR_TOOL_INPUT

### LINKED ISSUE #5107: Agentic AI: Remove AgentState.WAITING_FOR_TOOL_INPUT in favor of a more dynamic logic
__What should we do?__

Relying on a previously set state enum makes it hard to handle in-between data not directly related to tool call results, such as events. Instead of flipping between READY/WAITING_FOR_TOOL_INPUT, the agent should inspect the history and decide based on the last message which kind of message (user, tool result) to add.

__Why should we do it?__

Better support for non-tool related content such as events.

maff:
## Description

Removes `AgentState.WAITING_FOR_TOOL_INPUT` and instead inspects the latest message in memory to decide if it needs to add a user or a tool result mesage. Relying on a previously set state enum makes it hard to handle in-between data not directly related to tool call results, such as events.

- Removes `WAITING_FOR_TOOL_INPUT` state
- Updates logic in `AgentMesagesHandler` to inspect the latest message in history in order to decide which kind of message to add
- Returns added messages from the messages handler
- Handles error conditions such as no user messages to add from the AgentRequestHandler instead of throwing an error in the messages handler as in the future not having a message might be a valid state in case of event handling.

## Related issues

closes #5107 

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandlerImpl.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandlerImpl.java
@@ -108,6 +108,7 @@ public class AgentMessagesHandlerImpl implements AgentMessagesHandler {
         .forEach(content::add);
 
     if (content.isEmpty()) {
+      LOGGER.debug("Not adding user message as no user content was found to add.");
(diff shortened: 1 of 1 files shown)


mdii (approved):
LGTM :+1:

Issue #5016, opened 2025-07-07
Generated Element Templates from OpenAPI Are Not Schema-Compliant


mathias-vandaele:
## What should we do?

The OpenAPI generator currently produces templates that are not valid against the [element template schema](https://jira.camunda.com/browse/SUPPORT-28009?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D), for reasons that are not yet fully understood.

We should investigate the root cause and ensure that the generator produces valid element templates. Additionally, we should introduce a validation step to automatically verify the generated templates against the schema.

Here is an example from the referenced Jira ticket that illustrates the issue:

```
{
  "id": "baseUrl",
  "label": "Server",
  "value": "https://httpbin.org/",
  "group": "server",
  "binding": {
    "name": "baseUrl",
    "type": "zeebe:input"
  },
  "type": "Dropdown",
  "choices": [
    {
      "name": null,
      "value": "https://httpbin.org/"
    },
    {
      "name": null,
      "value": "http://httpbin.org/"
    },
    {
      "name": null,
      "value": "https://eu.httpbin.org/"
    },
    {
      "name": null,
      "value": "http://eu.httpbin.org/"
    }
  ]
}
```

## Why should we do it?

To ensure a reliable and smooth user experience for those using the OpenAPI generator. Invalid templates can cause confusion and friction during modeling, and validating them upfront would help avoid such issues.

Related to: https://jira.camunda.com/browse/SUPPORT-28009?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

JacobiCamunda:
@mathias-vandaele, will this be picked up soon, so it can be incorporated with the Web Modeler `8.8.0-alpha7` release? Code freeze is next Tuesday (29.07)

PR #5122, opened 2025-07-23
fix(open-api): all generated templates will now validate the result against the JSON schema

mathias-vandaele:
## Description

This is the part of the fix for `8.8`, the other part has been corrected already on `8.8` but not in `8.7` with this [commit](https://github.com/camunda/connectors/commit/d605bbce62508074f2f693c95f17fcebf6434ec8) This old PR needs to be backported

This needs to be backported to `8.7` as well

## Related issues

related to: https://github.com/camunda/connectors/issues/5016

## Checklist


element-template-generator/congen-cli/src/main/java/io/camunda/connector/generator/cli/ReturnCodes.java
```diff
@@ -19,7 +19,8 @@
 public enum ReturnCodes {
   SUCCESS(0),
   GENERATION_FAILED(1),
-  INPUT_PREPARATION_FAILED(2);
+  INPUT_PREPARATION_FAILED(2),
+  JSON_SCHEMA_VALIDATION_FAILED(2);
```
johnBgood, line 23: shouldn't the code be `3` ?

element-template-generator/congen-cli/src/main/java/io/camunda/connector/generator/cli/command/Generate.java
```diff
@@ -76,27 +97,19 @@ public Integer call() {
       } else {
         resultString = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(templates);
       }
+      Set<ValidationMessage> errors = jsonSchema.validate(resultString, InputFormat.JSON);
+      if (!errors.isEmpty()) {
+        System.err.println("Validation failed:");
```
johnBgood, line 102: system.err ? expected?

johnBgood (dismissed):
left a few comments, also is the PR name correct? "will not validate", or should it be "will NOW validate" ?

element-template-generator/congen-cli/src/main/java/io/camunda/connector/generator/cli/ReturnCodes.java
mathias-vandaele: Corrected

Issue #5105, opened 2025-07-22
Agentic AI: Support OpenAI embeddings models for vector database


mathieu-stennier:
__Is your feature request related to a problem? Please describe.__

 Support OpenAI embeddings models for vector database

Issue #5081, opened 2025-07-15
Support Embedding Models via Google Vertex AI in Embeddings Provider


mathieu-stennier:
Technical Details:

**Goal**: Enable the use of Google Vertex AI embedding models for generating embeddings, accessible through the existing provider interface.

**Integration Points**:
Implement a new provider class (e.g., `GoogleVertexAIEmbeddingProvider`) that integrates with Vertex AI’s embedding endpoints.
Use Google Cloud’s Java SDK (or REST API) to invoke Vertex AI embedding models.
Support configuration for Google Cloud project, region, credentials, and model selection.
Ensure compatibility with the existing embeddings provider interface so Vertex AI models can be used interchangeably with other providers.

**References**:
[Vertex AI Embeddings API](https://cloud.google.com/vertex-ai/docs/generative-ai/embeddings/get-text-embeddings)
[Langchain4j documentation](https://github.com/langchain4j/langchain4j)

crobbins215:
@mathieu-stennier, this is showing on the connectors board, are you able to double check this ticket?

mdii:
@crobbins215 I removed it from the connectors project.

Issue #5080, opened 2025-07-15
Support Azure Vector Database as Embeddings Provider via Langchain4j


mathieu-stennier:
We should add support for Azure as a vector database embeddings provider in the connectors/embeddings-vector-database module, utilizing Langchain4j.

## Technical Details:

**Goal**: Enable storing and querying embeddings in Azure’s vector database services (e.g., Azure AI Search with vector search, or Azure Cosmos DB for vector data) through the Langchain4j integration.
Module Impacted: connectors/embeddings-vector-database

Implement a new provider class (e.g., `AzureVectorDatabaseProvider`) conforming to the existing embeddings provider interface.
Leverage Langchain4j’s Azure integration (if available) or implement a bridge between Langchain4j’s Embeddings API and Azure’s vector storage/query APIs.
Ensure configuration options for Azure credentials, endpoints, and index/container names.

References:
[Langchain4j documentation](https://github.com/langchain4j/langchain4j)
[Azure Embedding models](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/models?tabs=global-standard%2Cstandard-chat-completions#embeddings)

crobbins215:
@mathieu-stennier, this is showing on the connectors board, are you able to double check this ticket?

mdii:
@crobbins215 I removed it from the connectors project.

PR #5125, opened 2025-07-24
refactor(vector-db): add missing toString() overrides and small refactoring

mdii:
## Description

Add missing `toString()` overrides in order to redact secrets and passwords. 
Rename `BedrockEmbeddingModel` to `BedrockEmbeddingModelProvider` as it implements `EmbeddingModelProvider`.
Remove unused dependency on `langchain4j-ollama`.

## Related issues

This is a preparation for #5105, #5080, and #5081.

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/pom.xml
+++ b/connectors/embeddings-vector-database/pom.xml
@@ -40,10 +40,6 @@
     </dependency>
     <dependency>
       <groupId>dev.langchain4j</groupId>
-      <artifactId>langchain4j-pgvector</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>dev.langchain4j</groupId>
       <artifactId>langchain4j-elasticsearch</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/embedding/vector/store/AmazonManagedOpenSearchVectorStore.java
```diff
@@ -53,4 +53,10 @@ public record AmazonManagedOpenSearchVectorStore(
     implements EmbeddingsVectorStore {
   @TemplateProperty(ignore = true)
   public static final String STORE_AMAZON_MANAGED_OPENSEARCH = "STORE_AMAZON_MANAGED_OPENSEARCH";
+
+  @Override
+  public String toString() {
+    return "AmazonManagedOpenSearchVectorStore(accessKey=REDACTED, secretKey=REDACTED, serverUrl=%s, region=%s, indexName=%s)"
```
maff, line 59: 🔧 Nit: should we also format these as `'[REDACTED]'` for consistency?

connectors/embeddings-vector-database/pom.xml
```diff
@@ -33,16 +33,11 @@
       <groupId>dev.langchain4j</groupId>
       <artifactId>langchain4j</artifactId>
     </dependency>
-    
-    <dependency>
-      <groupId>dev.langchain4j</groupId>
-      <artifactId>langchain4j-ollama</artifactId>
-    </dependency>
```
maff, line 40: Good catch! Can we also remove the PGVector dependency? We had both Ollama + PGVector support in the beginning, but then decided to remove it for the moment, so probably that library is also a leftover.

maff (approved):
Great catch and refactoring! 🚀

maff (approved):

PR #5161, opened 2025-07-30
chore(http-client): Refactoring

ztefanie:
## Description
Moved everything that is not needed for element template generation from `http-base` to `http-client`. Everything that is still in `http-base` is mostly duplication of what is in`http-client`. No changes are required for existing connectors. Mappers are added to map modeles from `http-base` to models in `http-client.`
Http-client currently is java 21, but i think i might need to downgrade it to 17, but I will do this in a separate PR.

## Checklist


ztefanie:
close in favor of this: https://github.com/camunda/connectors/pull/5162

Issue #5117, opened 2025-07-23
Add exceptions to error logging


ztefanie:
__Is your feature request related to a problem? Please describe.__

Feedback we got, when using Connectors in CPT:

> When running tests where a connector raises an unexpected exception - my test fails as it doesn't go through. That's correct.
> I can also see on the console that there seems to be an incident - but it just says An error occurred while executing the request, or the connection was aborted
> How can I see all details of the incident? Ideally the full stack trace?
> How can I get more information on the exception of the connector? I feel kind of blind here to understand what's happening.

[Slack discussion 1](https://camunda.slack.com/archives/C072SGECL9W/p1753254746203099)
[Slack discussion 2](https://camunda.slack.com/archives/C06VC96GMNE/p1753254419689409)

After some investigation, this is caused by not appending the original exception to the logs, e.g. [here](https://github.com/camunda/connectors/blob/54b1889022ce8a53d3bfa164630bf895f1907e62/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java#L205). As we do not append the original exception, the logging in CPT cannot access the stacktrace.
I think this was removed to ensure secrets where not logged.

@berndruecker correctly highlighted that:

> I think this is a good example that we still need to expose enough information for developers to debug / root-cause.
Maybe we could reach the goal of not exposing any secrets differently? From experience I think a full stacktrace  can save days of headache even in single failure scenarios

__Describe the solution you'd like__

Append the original excpetion to error logging, like this:
`LOGGER.error("Error occurred", exception);` not like this: `LOGGER.error("Error occurred: {}", exception.getMessage());`

__Describe alternatives you've considered__

:one: we could dump entire stack trace to bedrock and make a beautiful message from it to send back to the console
:two: Append the actual exception message to the string returned by the ConnectorExcpetion to the Console, [here](https://github.com/camunda/connectors/blob/54b1889022ce8a53d3bfa164630bf895f1907e62/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/client/apache/CustomApacheHttpClient.java#L73).

PR #5172, opened 2025-07-31
other(logging): log the exception when an error occurs

johnBgood:
## Description

Log the exception in ConnectorJobHandler when an error occurs. Previously, we were logging the exception's message only.
Here's an example of the new log:
```
Exception while completing job: JobForLog[key=2251799813982047, customHeaders={elementTemplateId=io.camunda.connectors.HttpJson.v2, elementTemplateVersion=11, retryBackoff=PT0S}, tenantId=<default>, bpmnProcessId=Process_1wh3upd, type=io.camunda:http-json:1, processDefinitionKey=2251799813980804, processDefinitionVersion=1, processInstanceKey=2251799813982035], message: An error occurred while executing the request, or the connection was aborted
```
## Related issues

closes https://github.com/camunda/connectors/issues/5117

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
@@ -194,7 +194,8 @@ public class ConnectorJobHandler implements JobHandler {
           "Exception while completing job: {} for tenant: {}, message: {}",
           job.getKey(),
           job.getTenantId(),
-          errorResult.exception().getMessage());
+          errorResult.exception().getMessage(),
+          errorResult.exception());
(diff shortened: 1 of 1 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/ConnectorJobHandler.java
```diff
@@ -194,7 +194,8 @@ private void handleSuccessResult(
           "Exception while completing job: {} for tenant: {}, message: {}",
           job.getKey(),
           job.getTenantId(),
-          errorResult.exception().getMessage());
+          errorResult.exception().getMessage(),
```
sbuettner, line 197: Great! Can we also include the jobtype here? Could be useful for debugging.

ztefanie (approved):
Nice, thank you :rocket:

PR #5176, opened 2025-08-01
feat(agentic-ai): Implement AI Agent as job worker

### LINKED ISSUE #5175: Agentic AI: Implement a Job Worker acting on the ad-hoc sub-process

In addition to the service task based approach, we should support the AI Agent as implementation of a job worker acting directly on the ad-hoc sub-process.

The core processing logic can be the same as in the service task based approach, with some differences how data is processed:

- Available tools do not need to be fetched from the process definition XML, instead they can be read from a variable `adHocSubProcessElements` provided by the engine. This was already prepared in https://github.com/camunda/connectors/pull/5102, decoupling the process definition loading from the actual tools schema generation and introducing a `AdHocToolElement` model.
  - The job worker job based approach can directly read the variable and map it to the model. Schema processing can be done the same way as we do with the service task based approach.
- The job worker will need to handle the job completion manually and emit activated elements with the completion call rather than setting a toolCalls variable.
- As a job worker will be activated for every completed activity, the job worker will need to handle concurrent completions and only return to the LLM when all tool results are available.
  -  To do so, the JW can rely on a growing output collection collecting all the tool results. As long as no tool results are available, it can simply complete the job without calling the LLM.
  - As jobs will be cancelled by the engine as soon as a new job is activated, the JW needs to account for this and only call the LLM when really needed. For example, it should check if the job is still active before proceeding with the LLM call.
- A new element template and request structure will be needed for the JW approach.
  - No tools configuration as we don't need to reference the tool call results or the subprocess ID
  - The agent context will be a local variable, so it won't be needed to re-inject it via inputs on every iteration. However, it should be possible to configure an existing agent context as initial value to continue previous conversations.
  - Configuration for `outputElement` and `outputCollection`

Preparation to decouple the outbound connector specific logic from the core processing one has already been done in #5129, so we should be able to build on these changes.

__Additional context__

TBD

maff:
## Description

TBD

## Related issues

TBD

closes #5175 

## Checklist


connectors/agentic-ai/bin/transform-ai-agent-job-worker-template.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
maff, line 1: Temporary solution to transform the outbound connector template until implementing https://github.com/camunda/connectors/issues/5189.

- needs `jq` to be available
- is not automatically executed (needs to be run manually)

connectors/agentic-ai/element-templates/agenticai-aiagent-job-worker.json
```diff
@@ -0,0 +1,1472 @@
+{
```
maff, line 1: Note: this does not work yet as the modeler throws an error on the 2 `output` properties binding to `zeebe:adHoc`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
```diff
@@ -204,7 +213,64 @@ public AgentRequestHandler aiAgentRequestHandler(
...
+      matchIfMissing = true)
+  public AiAgentJobWorker aiAgentJobWorker(
+      SecretProviderAggregator secretProvider,
+      @Autowired(required = false) ValidationProvider validationProvider,
+      DocumentFactory documentFactory,
+      ObjectMapper objectMapper,
+      JobWorkerAgentRequestHandler agentRequestHandler) {
+    if (validationProvider == null) {
+      validationProvider = ValidationUtil.discoverDefaultValidationProviderImplementation();
+    }
```
maff, line 271: This is the same logic as implemented for outbound connectors.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAgentRequestHandler.java
```diff
@@ -132,9 +169,19 @@ private AgentResponse handleRequest(
...
     agentContext = session.storeFromRuntimeMemory(agentContext, runtimeMemory);
 
     return responseHandler.createResponse(
         executionContext, agentContext, assistantMessage, processVariableToolCalls);
   }
+
+  protected abstract boolean modelCallPrerequisitesFulfilled(
+      C executionContext, AgentContext agentContext, List<Message> addedUserMessages);
+
+  /** Handles job completion if needed. Agent response and conversation store may be null. */
```
mdii, line 182: I wanted to suggest adding explicit nullability annotations but which ones should/do we use?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAgentRequestHandler.java
```diff
@@ -6,26 +6,31 @@
...
 import io.camunda.connector.agenticai.model.tool.ToolCallProcessVariable;
 import io.camunda.connector.agenticai.model.tool.ToolCallResult;
-import io.camunda.connector.api.error.ConnectorException;
 import java.util.List;
 import java.util.Optional;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class DefaultAgentRequestHandler<C extends AgentExecutionContext>
```
mdii, line 30: :wrench: Now that it is an abstract class should we replace the `Default` prefix with something else, e.g., `AbstractAgentRequestHandler` or `BaseAgentRequestHandler`?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
```diff
@@ -0,0 +1,221 @@
...
+      JobWorkerAgentExecutionContext executionContext,
+      AgentResponse agentResponse,
+      ConversationStore conversationStore) {
+    if (agentResponse == null) {
+      LOGGER.debug(
+          "No agent response provided, completing job {} without response",
+          executionContext.job().getKey());
+      completeAsyncWithoutResponse(executionContext);
+    } else {
+      if (LOGGER.isDebugEnabled()) {
```
mdii, line 84: :question: what's the logic behind wrapping the debug log in the `if` as you don't do it everywhere? Does it depend on what we log?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/DefaultAgentRequestHandler.java
maff: Good point. I now added the Spring ones as we already use them in the records. For the future we could consider JSpecify as Spring will also start using them with the [next major version](https://docs.spring.io/spring-framework/reference/core/null-safety.html).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
```diff
@@ -0,0 +1,219 @@
...
+        prepareCompleteCommand(executionContext),
+        defaultExceptionHandlingStrategy);
+  }
+
+  private void completeAsyncWithResponse(
+      JobWorkerAgentExecutionContext executionContext,
+      AgentResponse agentResponse,
+      ConversationStore conversationStore) {
+    boolean completionConditionFulfilled = agentResponse.toolCalls().isEmpty();
+    boolean cancelRemainingInstances = false; // TODO JW check events for interruptions
```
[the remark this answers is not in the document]
maff: Reference: https://github.com/camunda/connectors/issues/5233

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
mdii: OK, it makes sense when the optimization is worth it so that we don't _pollute_ the code with the if blocks checking whether the log level is active :+1:

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/JobWorkerAgentResponse.java
```diff
@@ -0,0 +1,26 @@
...
+package io.camunda.connector.agenticai.aiagent.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import io.camunda.connector.agenticai.model.AgenticAiRecord;
+import io.camunda.connector.agenticai.model.message.AssistantMessage;
+import org.springframework.lang.Nullable;
+
+@AgenticAiRecord
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record JobWorkerAgentResponse(
```
mdii, line 16: :question: don't we need a `@JsonDeserialize` like in `AgentResponse`?

maff: Good catch! The main advantage of forcing the deserialization through the builders is to leverage features like the implicit null checks and that collections are always initialized vs. being null.

As this record has all nullable fields and no collections, it's not strictly necessary, but still I'll add it for consistency and for potential future changes.

mdii (commented):
That is a great work :clap: 
Just one :question: : should there be a test for `AiAgentJobWorkerErrorHandler`?

PR #5187, opened 2025-08-04
feat(sdk): Add @Header binding support

### LINKED ISSUE #5131: Support job header binding for `@Operation` annotated methods

With the newly introduced SDK support for `@Operation` we are currently missing support for binding models from job headers.

Introduce a new `@Header` annotation that can be used to bind a header value (a FEEL expression for example) to a model class (a `Function<A,B>` for example).  This should also be reflected in the generation of the element template.

The alternative is to inject an `OutboundConnectorContext` into the operation and manually extract the header values from the `JobContext`. This requires additional effort and requires manual management of the element template. 

__Additional context__

This was found during the development of the new CSV Connector which would benefit from a mapping function that would need to be passed as a header.

sbuettner:
## Related issues

closes #5131

## Checklist

- [x] Add SDK and Runtime support
- [x] Add basic element template generator support
- [x] Add example to CSV Connector that allows to transform rows based on provided FEEL function

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java
@@ -16,10 +16,10 @@
  */
 package io.camunda.connector.runtime.core.outbound.operation;
 
-import static io.camunda.connector.api.reflection.ReflectionUtil.getMethodsAnnotatedWith;
-import static io.camunda.connector.api.reflection.ReflectionUtil.getVariableName;
+import static io.camunda.connector.api.reflection.ReflectionUtil.*;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.camunda.connector.api.annotation.Header;
 import io.camunda.connector.api.annotation.Operation;
 import io.camunda.connector.api.outbound.OutboundConnectorContext;
 import io.camunda.connector.api.reflection.ReflectionUtil.MethodWithAnnotation;
@@ -73,6 +73,11 @@ public record ConnectorOperations(Object connector, Map<String, OperationInvoker
           getVariableName(variableAnnotation), parameter.getType(), variableAnnotation.required());
     } else if (parameter.getType().equals(OutboundConnectorContext.class)) {
       return new Context();
+    } else if (parameter.isAnnotationPresent(Header.class)) {
+      io.camunda.connector.api.annotation.Header headerAnnotation =
+          parameter.getAnnotation(io.camunda.connector.api.annotation.Header.class);
+      return new ParameterDescriptor.Header<>(
+          getHeaderName(headerAnnotation), parameter.getType(), headerAnnotation.required());
     } else {
       throw new IllegalArgumentException("Unsupported parameter type: " + parameter.getType());
     }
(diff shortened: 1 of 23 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java
```diff
@@ -73,6 +73,11 @@ private static ParameterDescriptor map(Parameter parameter) {
           getVariableName(variableAnnotation), parameter.getType(), variableAnnotation.required());
     } else if (parameter.getType().equals(OutboundConnectorContext.class)) {
       return new Context();
+    } else if (parameter.isAnnotationPresent(Header.class)) {
+      io.camunda.connector.api.annotation.Header headerAnnotation =
+          parameter.getAnnotation(io.camunda.connector.api.annotation.Header.class);
```
chillleader, line 78: nit: fully qualified name is redundant as we import it above

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ParameterDescriptor.java
```diff
@@ -56,4 +56,29 @@ public boolean isRequired() {
       return required;
     }
   }
+
+  final class Header<T> implements ParameterDescriptor {
+
+    private final String name;
+    private final Class<T> type;
+    private final boolean required;
```
chillleader, line 64: nit: this can be a record I suppose? but also fine to keep as is to match the style of `Variable` above

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/OperationInvoker.java
```diff
@@ -78,6 +81,25 @@ private Object resolveVariableValue(
...
+    String rawValue = headers.get(headerDescriptor.getName());
+    Object value;
+    try {
+      value = objectMapper.convertValue(rawValue, headerDescriptor.getType());
+    } catch (Throwable e) {
+      throw new RuntimeException(e);
+    }
+    if (headerDescriptor.isRequired() && value == null) {
+      throw new ConnectorInputException(
+          "Required variable '" + headerDescriptor.getName() + "' is missing in the job headers.");
```
chillleader, line 95: I guess we should not use the word 'variable' in this error message to avoid any confusion between different value sources

connector-sdk/jackson-datatype-feel/src/main/java/io/camunda/connector/feel/jackson/FeelFunctionDeserializer.java
```diff
@@ -71,10 +71,13 @@ protected Function<IN, OUT> doDeserialize(
 
   @Override
   public JsonDeserializer<?> createContextual(DeserializationContext ctxt, BeanProperty property) {
-    if (property.getType().containedTypeCount() == 2) {
-      var outputType = property.getType().containedType(1);
-      return new FeelFunctionDeserializer<>(outputType, feelEngineWrapper);
+    if (property != null) {
+      if (property.getType().containedTypeCount() == 2) {
```
chillleader, line 75: Just curious, when can this be null?

chillleader (commented):
Really nice and simple 🚀

connector-sdk/jackson-datatype-feel/src/main/java/io/camunda/connector/feel/jackson/FeelFunctionDeserializer.java
sbuettner: Its apparently null when I just want to convert it to an object like `String` and not as part of encapsulating object as a property.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java
sbuettner: Thanks, fixed!

chillleader (approved):

sbuettner:
@Szik Great finding, the default value of the record mapper was empty which is not a legal FEEL expression. I fixed the generator code and pushed a new element template version. Please give it another try.

PR #5191, opened 2025-08-05
feat(vector-db): add Google Vertex AI embedding models

### LINKED ISSUE #5081: Support Embedding Models via Google Vertex AI in Embeddings Provider

mdii:
## Description

Add support for Google Vertex AI embedding models. Application Default Credentials and Service Account Credentials are supported as authentication methods.

Note that contrary to other embedding model providers, the `dimensions` parameter is mandatory. This is because the supported embedding models by L4J are already deprecated, so we cannot rely on L4J to use the default dimensions. In addition, it is possible to use third-party models with Vertex AI APIs. In that case again we cannot know the default dimensions.

Also note that there is currently an [issue](https://camunda.slack.com/archives/C02JLRNQQ05/p1754381080999299) with specifying JSONs as secrets. 

## Related issues

closes #5081 

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/pom.xml
+++ b/connectors/embeddings-vector-database/pom.xml
@@ -28,7 +28,7 @@
       <groupId>io.camunda.connector</groupId>
       <artifactId>element-template-generator-core</artifactId>
     </dependency>
-    
+
     <dependency>
       <groupId>dev.langchain4j</groupId>
       <artifactId>langchain4j</artifactId>
@@ -71,8 +71,24 @@
       <groupId>dev.langchain4j</groupId>
       <artifactId>langchain4j-document-parser-apache-tika</artifactId>
     </dependency>
+
+    <dependency>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-test</artifactId>
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>org.junit.jupiter</groupId>
+        <artifactId>junit-jupiter-engine</artifactId>
+        <scope>test</scope>
+    </dependency>
+    <dependency>
+        <groupId>uk.org.webcompere</groupId>
+        <artifactId>system-stubs-jupiter</artifactId>
+        <scope>test</scope>
+    </dependency>
   </dependencies>
-  
+
 
   <build>
     <plugins>
--- a/connectors/embeddings-vector-database/src/test/java/io/camunda/connector/model/embedding/models/GoogleVertexAiEmbeddingModelProviderTest.java
+++ b/connectors/embeddings-vector-database/src/test/java/io/camunda/connector/model/embedding/models/GoogleVertexAiEmbeddingModelProviderTest.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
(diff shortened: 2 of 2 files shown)


connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/embedding/models/GoogleVertexAiEmbeddingModelProvider.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @TemplateProperty(ignore = true)
+  public static final String VERTEX_AI_DEFAULT_PUBLISHER = "google";
+
+  @AssertFalse(message = "Google Vertex AI is not supported on SaaS")
+  public boolean isUsedInSaaS() {
+    return System.getenv().containsKey("CAMUNDA_CONNECTOR_RUNTIME_SAAS")
+        && vertexAiAuthentication
+            instanceof GoogleVertexAiAuthentication.ApplicationDefaultCredentialsAuthentication;
+  }
```
maff, line 90: As this is pretty crucial, can we cover this validation in an integration test?

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/embedding/models/GoogleVertexAiEmbeddingModelProvider.java
```diff
@@ -0,0 +1,130 @@
...
+            description = "Google Cloud project ID",
+            constraints = @TemplateProperty.PropertyConstraints(notEmpty = true))
+        String projectId,
+    @NotBlank
+        @TemplateProperty(
+            group = "embeddingModel",
+            label = "Region",
+            description = "Google Cloud region for Vertex AI",
+            constraints = @TemplateProperty.PropertyConstraints(notEmpty = true))
+        String region,
```
maff, line 38: 💭 IIRC we called this `location` in the AI Agent Vertex integration. Do we have a common naming across all Google integrations we could use?

maff (commented):
Some minor comments, but looks good overall 🚀

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/embedding/models/GoogleVertexAiEmbeddingModelProvider.java
mdii: I see in other connectors that `region` is used. To my understanding `region` is the correct name to use. Maybe we should rename `location` in the Agentic-AI connector?

maff: Yep, if it's region in the other Google connectors, we should try to align it and rename `location` in the AI Agent connector 👍

maff (approved):

PR #5194, opened 2025-08-05
feat(vector-db): add Azure AI Search embeddings store

mdii:
## Description

Add Azure AI Search as a vector database. API key and client secret credentials are supported for authentication.

## Related issues

Part of #5080 

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/element-templates/hybrid/embeddings-vector-database-outbound-connector-hybrid.json
+++ b/connectors/embeddings-vector-database/element-templates/hybrid/embeddings-vector-database-outbound-connector-hybrid.json
@@ -1053,13 +1053,13 @@
     },
     "type" : "String"
   }, {
-    "id" : "vectorStore.aiSearchAuthentication.type",
+    "id" : "vectorStore.azureAiSearchAuthentication.type",
     "label" : "Authentication",
     "description" : "Specify the Azure OpenAI authentication strategy.",
     "value" : "apiKey",
     "group" : "embeddingsStore",
     "binding" : {
-      "name" : "vectorStore.aiSearchAuthentication.type",
+      "name" : "vectorStore.azureAiSearchAuthentication.type",
       "type" : "zeebe:input"
     },
     "condition" : {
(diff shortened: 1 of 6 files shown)


connectors/embeddings-vector-database/element-templates/hybrid/embeddings-vector-database-outbound-connector-hybrid.json
```diff
@@ -1239,7 +1431,7 @@
     },
     "type" : "String"
   }, {
-    "id" : "vectorStore.azureCosmosDbNoSql.consistencyLevel",
+    "id" : "vectorStore.consistencyLevel",
```
maff, line 1434: 🔧 Is this what we expect it to be? 🤔 As this property is cosmos DB specific, I would assume the property to contain cosmos DB in the ID.

Regarding the discussion in https://camunda.slack.com/archives/C02JLRNQQ05/p1754407737998059: assuming there would be another vector store configuring a consistency level, would that still work?

maff (commented):
The actual implementation looks good to me, some comments on the property structure. As we are seeing some property clashes, might it make sense to introduce another level of hierarchy with a unique name to the provider configurations (probably for both models & stores)?

In the AI Agent connector, we have this `Connection` level using a unique name per provider, e.g. `openai`: https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/OpenAiProviderConfiguration.java#L22

mdii:
> The actual implementation looks good to me, some comments on the property structure. As we are seeing some property clashes, might it make sense to introduce another level of hierarchy with a unique name to the provider configurations (probably for both models & stores)?

I did not do that to be consistent with the existing embedding model and vector store providers. But I agree that would be a proper solution. However, I think we need a dedicated PR for that, as there will be quite some changes. I can do it before or after merging this PR. What do you think @maff ?

maff:
That would be great! We can definitely do this in a follow-up PR after merging this one 👍

PR #5196, opened 2025-08-06
feat(servicenow): generic outbound + start-flow connector

vobu:
## Related issues

implements https://github.com/camunda/product-hub/issues/2887

## Checklist


Szik:
@vobu can we sync next week to get this setup?

vobu:
> @vobu can we sync next week to get this setup?
let's postpone this to the week after. This is still a draft. Need to complete the element template first.
After that, I'd also like to prep some automation around `element-templates-cli`, eventually even testing stuff with Josh's Node.js CPT implementation. I'll ping you here once the PR is ready.

vobu:
Hey @Szik and @sbuettner, this is conceptually done now.
What is different from the other element template-based connectors here is that an automated e2e test is included (only a sample one to illustrate the point, not (yet) an entire suite) *****.

This test sits one layer up from "integration tests" in the sense that what would usually require manual testing, this is done automatically: build a BPMN, use the element template, deploy it on a Camunda cluster, run the connector against a live ServiceNow instance. 

I'm aware that we break the pattern here with regard to test isolation in that we create a hard dependency to a ServiceNow system. But this is IMO justified b/c only then can we confidently guarantee
- working element template ↔︎ target Camunda version
- e2e functionality Camunda ↔︎ different ServiceNow versions

With this being the 2nd of more "Business Application Integrations" to come, I also expect more automated e2e tests of STP BPMN tests to come into play. So aligning on how to best handle this with the included ServiceNow example is a good thing™  IMO 😛 😄 


\***** assumptions:
- `env` containing `SN_INSTANCE`, `SN_USR`, and `SN_PWD`
- `application.yml` intentionally not included, but verified that the test is working with a C8.8-alpha7 based one à la
  ```yaml
  camunda:
    client:
      mode: saas
      auth:
        client-id: ...
        client-secret: ...
        token-url: https://login.cloud.dev.ultrawombat.com/oauth/token
        audience: zeebe.dev.ultrawombat.com
      cloud:
        cluster-id: ...
        region: ...
  
      grpc-address: https:// ... . ... .zeebe.dev.ultrawombat.com
      rest-address:  https:// ... .zeebe.dev.ultrawombat.com/...
  ```

connectors-e2e-test/connectors-e2e-test-servicenow/src/test/java/io/camunda/connector/e2e/ServiceNowTests.java
```diff
@@ -0,0 +1,123 @@
...
+    BpmnModelInstance updatedModel;
+    DeploymentEvent deploymentEvent;
+    ProcessInstanceResult processInstanceResult;
+
+    PipelineContext(String processId, String serviceTaskName, File filledTemplate) {
+      this.processId = processId;
+      this.serviceTaskName = serviceTaskName;
+      this.filledTemplate = filledTemplate;
+    }
+  }
```
sbuettner, line 56: Why not having a simple record class with methods like `getModel`, `updateModel`,... seems like an abstraction that (from a reader perspective) adds more complexity than it solves.

vobu: I just like the .andThen() chain this enables :) But I get your point - will refactor

sbuettner: @vobu `then()` you would enjoy https://effect.website/

vobu:
> @vobu (Just seeing the formatting problems) We have recently set up [pre-commit](https://pre-commit.com/) to catch exactly these kind of annoyance before they get rejected in CI

kk, will install and use for next PR - thanks for the hint!

connector-sdk/test/src/main/java/io/camunda/connector/test/ExternalSystem.java
```diff
@@ -0,0 +1,33 @@
...
+package io.camunda.connector.test;
+
+/**
+ * Enumeration of external systems for system integration tests. The {@link #id} value is also used
+ * as the expected environment variable name that must be present to enable tests annotated with
+ * {@link SystemIntegrationTest}.
+ */
+public enum ExternalSystem {
+  ServiceNow("ServiceNow"),
+  SAP("SAP");
```
sbuettner, line 26: Is this used here?

connector-sdk/test/src/main/java/io/camunda/connector/test/SystemIntegrationTestCondition.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+    Class<?> testClass = (Class<?>) context.getElement().orElse(null);
+    if (testClass == null || !testClass.isAnnotationPresent(SystemIntegrationTest.class)) {
+      return DISABLED;
+    }
+    SystemIntegrationTest annotation = testClass.getAnnotation(SystemIntegrationTest.class);
+    ExternalSystem externalSystem = annotation.with();
+    String envVar = System.getenv(externalSystem.id);
```
sbuettner, line 43: So then env var will be only something like `ServiceNow`? 

What about a prefix like = "INTEGRATION_TEST_"+ externalSystem.id.toUpperCase()?

connectors-e2e-test/connectors-e2e-test-servicenow/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+    <version>8.8.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>connectors-e2e-test-servicenow</artifactId>
+  <name>Archetype - connectors-e2e-test-servicenow</name>
+  <url>http://maven.apache.org</url>
+  <dependencies>
+<!--    <dependency>-->
+<!--      <groupId>io.camunda</groupId>-->
+<!--      <artifactId>camunda-process-test-spring</artifactId>-->
+<!--    </dependency>-->
```
sbuettner, line 16: Shouldnt CPT be used when running this on the CI?

vobu: from my understanding, the connector runtime is part of CPT, running in the isolated testcontainer environment - and thus, can not reach out (e.g. via the `REST` connector) to external systems. Which is what we'd need in order to connect to the live SN instance.

vobu: now works with CPT and bootstrapped Connectors runtime

connector-sdk/test/src/main/java/io/camunda/connector/test/ExternalSystem.java
vobu: gnarf, ofc, I was initially under the impression that we launch the tests here, pointing to the downstream pipeline in the `sap-connectors` repo. Which doesn't really make then but overcomplicates things. So will take it out here.

Szik (dismissed):

connectors-e2e-test/connectors-e2e-test-servicenow/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+  <dependencies>
+    <dependency>
+      <groupId>io.camunda</groupId>
+      <artifactId>camunda-process-test-spring</artifactId>
+    </dependency>
+<!--    <dependency>-->
+<!--      <groupId>io.camunda</groupId>-->
+<!--      <artifactId>spring-boot-starter-camunda-sdk</artifactId>-->
+<!--      <version>8.8.0-alpha2-rc1</version>-->
+<!--    </dependency>-->
```
sbuettner, line 21: Please remove the comments

vobu: done

vobu:
now this PR is complete code-wise from our end - please @Szik or @sbuettner approve and merge. 🙇

sbuettner (approved):

PR #5205, opened 2025-08-07
refactor(agentic-ai): Align `fromAi()` parameter extraction with core engine logic

### LINKED ISSUE #5204: Agentic AI: Align fromAi() tool call extraction to core engine logic
__What should we do?__

In the future we will do the `fromAi()` tool call extraction either in the engine (job worker) or on the connectors side. To avoid ambiguity, we should align the 2 approaches:

- [x] Extract the full property path (`toolCall.myValue`) instead of just using the last segment (`myValue`). We did this in the core engine to open possibilities to different job worker implementations which might need this information.
  - For the AI Agent connector, we can introduce a check for a tool call "namespace" `toolCall.*` which we can verify during schema generation. This will keep the same logic as before, only that the schema generation will be responsible for stripping the `toolCall.*` prefix and not the parameter extraction logic. When calling `fromAi()` with a different prefix than the expected one, the schema generation should fail.
- [x] Only allow constant values for the `fromAi()` arguments. Previously, it was theoretically possible for specify something like `fromAi(toolCall.myValue, "str" + "ing")` and it worked as long as the outcome was the expected type. However this gets ambiguous as soon as variables are involved as they would not be available during evaluation as this is done in a different phase. To avoid this ambiguity, we should not support any function calls or expressions within `fromAi()` function calls and rely on constant values (which is the normal case).
- [x] Support an option "required=false" to remove the parameter from the list of required params. 

__Why should we do it?__

Consistency with the core implementation in terms of behavior, but also with the potential to at some point reuse the same implementation.

maff:
## Description

Updates the `fromAi()` function extraction logic to use the `zeebe-feel-tagged-parameters` module implemented in https://github.com/camunda/camunda/pull/36508. This allows us to avoid having to maintain the function extraction multiple times. As the core module behaves a bit differently as the connector implementation, these are the high-level changes:

- The tagged parameter extractor returns the full property path to allow maximum flexibility of client implementations. Instead of just blindly using the last path segment of the property, our integration now expects a fixed prefix of `toolCall.` and raises an error otherwise. This is currently only configurable when providing a custom implementation in hybrid mode, but we could make it configurable down the line if we see the need to.
- The `fromAi()` function now only supports constant values as arguments . See #5204 for an example.
- As an unrelated change, this PR also adds support to set an `required=false` option on a property (happy to extract it to a dedicated PR if we want to keep this one smaller).

e2e tests still work without changes, so it should not introduce any regressions apart from the stricter argument handling.

## Related issues

depends on https://github.com/camunda/camunda/pull/36508
closes #5204 

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/schema/AdHocToolSchemaGeneratorImpl.java
```diff
@@ -86,4 +96,53 @@ public Map<String, Object> generateToolSchema(AdHocToolElement element) {
...
+    final var parameterName = parameter.name();
+    if (StringUtils.isBlank(parameterName)
+        || !parameterName.startsWith(expectedParameterNamespace)) {
+      throw new AdHocToolSchemaGenerationException(
+          "Failed to generate ad-hoc tool schema for element '%s'. Parameter name '%s' is not part of expected namespace '%s'."
+              .formatted(element.elementId(), parameter.name(), expectedParameterNamespace));
+    }
+
+    final var parameterNameWithoutNamespace =
+        parameterName.substring(expectedParameterNamespace.length());
```
maff, line 110: `zeebe-tagged-parameters` now returns the whole reference (e.g. `toolCall.myValue`) instead only the last segment. We're free how to handle it, but in this schema generator we're now expecting the "namespace" for a tool call to be under `toolCall.*` and failing otherwise.

This removes ambiguity by establishing a single, valid patterns for tool call parameters while still allowing alternative implementations to handle it differently.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/processdefinition/feel/AdHocToolElementParameterExtractorImpl.java
```diff
@@ -0,0 +1,54 @@
...
+      throw new IllegalArgumentException(
+          "Failed to parse FEEL expression: " + parseResult.failure().message());
+    }
+
+    return taggedParameterExtractor.extractParameters(parseResult.parsedExpression()).stream()
+        .map(this::asToolElementParameter)
+        .toList();
+  }
+
+  private AdHocToolElementParameter asToolElementParameter(final TaggedParameter parameter) {
```
maff, line 45: We could also directly use `TaggedParameter`, but it seemed cleaner to me to use our own model here.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/adhoctoolsschema/processdefinition/feel/AdHocToolElementParameterExtractorTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+import io.camunda.connector.agenticai.adhoctoolsschema.model.AdHocToolElementParameter;
+import io.camunda.zeebe.feel.tagged.impl.TaggedParameterExtractor;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class AdHocToolElementParameterExtractorTest {
```
maff, line 14: This is now just a high-level test that the integration is working. The actual, detailed `fromAi()` test cases are now tested in [`zeebe-feel-tagged-parameters`](https://github.com/camunda/camunda/blob/main/zeebe/feel-tagged-parameters/src/test/java/io/camunda/zeebe/feel/tagged/impl/TaggedParameterExtractorTest.java).

mdii (approved):
This is a great improvement :+1:

PR #5217, opened 2025-08-11
fix(vector-db): fix property IDs and use camel case IDs everywhere

mdii:
## Description

In the previous [PR](https://github.com/camunda/connectors/pull/5207) I added one level of hierarchy in the configuration of embedding model providers. This change was not reflected in property IDs referenced in `@PropertyCondition` annotations.
In addition there are still a few places where element IDs are not in camel case. This PR fixes them as well.

## Checklist


maff (approved):
Great work on constantly improving the consistency! 👍

PR #5228, opened 2025-08-13
refactor(sdk)!: move object mapper to a different package, move document API to core

chillleader:
## Description

Apologies for a huge and poorly reviewable PR 🙁 

List of significant changes:
1. Connector SDK core is now made lightweight, (almost) all external dependencies have been moved to other modules.
2. Document classes essential for the core SDK (`Document`, `DocumentFactory`) have been moved to the core SDK (previously they were in the separate module `connector-document`)
  - The document module now contains only implementations that talk to the Camunda API
  - Some classes that had been previously used directly from the Camunda client, like `DocumentMetadata`, have been recreated in the Core SDK to remove the dependency on Camunda Client.
3. For the sake of getting rid of Jackson dependency in the Core, I removed the `@JsonIgnore` annotations from `CorrelationResult` implementations. It seems they were used in the past because these classes were potentially exposed directly in activity logs, but I've tested this and there is no extra info in serialized payload after this change.
4. Restructured the `ProcessElement` classes, there were some classes with unclear boundaries and similar responsibilities, overused from Core
5. New high-level module `connector-commons` with submodules: `http-client` (moved from being top-level), `connectors-object-mapper`, `intrinsic-functions`. The new module is intended for things that are reused in this repo but probably not intended to be consumed elsewhere (e.g. external connectors).
6. Created an additional static instance of `ObjectMapper` for use in the `HttpClient` to avoid cyclic dependencies e.g. when the HttpClient is used in an intrinsic function.
  - Alternative considered: Avoid using static instance completely and make the `ObjectMapper` a parameter for the HttpClient. This seems nicer but I don't see real benefits as this `HttpClient` is not used outside of this repo.

## Related issues

closes #

## Checklist


connector-commons/http-client/src/test/java/io/camunda/connector/http/client/TestDocumentFactory.java
```diff
@@ -39,9 +41,43 @@ public Document resolve(DocumentReference reference) {
   @Override
   public Document create(DocumentCreationRequest request) {
     var reference = store.createDocument(request);
-    var metadata = new DocumentMetadata();
-    metadata.setContentType(reference.getMetadata().getContentType());
-    metadata.setFileName(reference.getMetadata().getFileName());
+    var metadata =
```
ztefanie, line 44: Why this change?

connectors/slack/src/test/java/io/camunda/connector/slack/outbound/caller/FileUploaderTest.java
```diff
@@ -196,8 +195,42 @@ private Document prepareDocument() {
     when(documentStore.getDocumentContent(any())).thenReturn(byteInput);
 
     DocumentMetadata documentMetadata =
-        new DocumentReferenceModel.CamundaDocumentMetadataModel(
-            "txt", OffsetDateTime.now(), 3000L, "fileName", "processId", 2000L, Map.of());
+        new DocumentMetadata() {
```
ztefanie, line 198: Can't we get rid of this?

ztefanie (dismissed):
Thanks for this refactorting. Looks good :rocket: 

Just some thoughts: 
- Do we need to somewhere document, that breaking changes have been introduced, or do we expect that customers do not use documents and processElement references in their custom connectors?
- Do we want to publish the commons under apache 2.0?

connector-commons/http-client/src/test/java/io/camunda/connector/http/client/TestDocumentFactory.java
chillleader: Side effect of switching to an implementation in our SDK instead of reusing the one from the camunda client. Good finding, cleaned it up

chillleader:
> Do we need to somewhere document, that breaking changes have been introduced, or do we expect that customers do not use documents and processElement references in their custom connectors?

Good point, I changed the PR name to contain a `!` which should lead to a note saying that it's a breaking change in the release notes. I think it's also a good idea to add an announcement to the public release notes on the docs website, I'll do that too 👍 

> Do we want to publish the commons under apache 2.0?

I don't think so, as this is the stuff we intend to use only in our repo, @sbuettner do you have an opinion?

PR #5230, opened 2025-08-13
fix(vector-db): add missing validation annotations

mdii:
## Description

There is no validation annotations on the fields of `EmbeddingsVectorDBRequest`. Hence the request data is not validated. 
This PR adds the required validation annotations and fixes the validation annotations that are incorrectly used.

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/EmbeddingsVectorDBRequest.java
+++ b/connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/EmbeddingsVectorDBRequest.java
@@ -9,8 +9,10 @@ package io.camunda.connector.model;
 import io.camunda.connector.model.embedding.models.EmbeddingModelProvider;
 import io.camunda.connector.model.embedding.vector.store.EmbeddingsVectorStore;
 import io.camunda.connector.model.operation.VectorDatabaseConnectorOperation;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
 
 public record EmbeddingsVectorDBRequest(
-    VectorDatabaseConnectorOperation vectorDatabaseConnectorOperation,
-    EmbeddingModelProvider embeddingModelProvider,
-    EmbeddingsVectorStore vectorStore) {}
+    @Valid @NotNull VectorDatabaseConnectorOperation vectorDatabaseConnectorOperation,
+    @Valid @NotNull EmbeddingModelProvider embeddingModelProvider,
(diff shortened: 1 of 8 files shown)


connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/embedding/models/BedrockEmbeddingModelProvider.java
```diff
@@ -44,7 +44,7 @@ public record Configuration(
               label = "Region",
               description = "AWS Bedrock region")
           String region,
-      @NotBlank
+      @NotNull
```
maff, line 47: 🔧 Can we also remove `constraints = @TemplateProperty.PropertyConstraints(notEmpty = true)` here or does it still make sense? Also for the dimensions field below and for other dropdowns in the other classes.

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/EmbeddingsVectorDBRequest.java
```diff
@@ -9,8 +9,10 @@
...
 import io.camunda.connector.model.embedding.vector.store.EmbeddingsVectorStore;
 import io.camunda.connector.model.operation.VectorDatabaseConnectorOperation;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotNull;
 
 public record EmbeddingsVectorDBRequest(
-    VectorDatabaseConnectorOperation vectorDatabaseConnectorOperation,
-    EmbeddingModelProvider embeddingModelProvider,
-    EmbeddingsVectorStore vectorStore) {}
+    @Valid @NotNull VectorDatabaseConnectorOperation vectorDatabaseConnectorOperation,
```
maff, line 16: This is nasty, I fell for that already many times 🙈 

💭 We also don't have it in the AI Agent codebase (at least there it is covered with e2e tests), but maybe it would be beneficial do have some e2e tests or a request validation test (maybe as a follow-up) testing the constraints.

maff (dismissed):
A small comment on the notEmpty constraints. Otherwise great improvement 👍

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/model/embedding/models/BedrockEmbeddingModelProvider.java
mdii: Most probably it doesn't make sense to keep the notEmpty constraints for dropdowns as there should always be one item selected.

PR #5232, opened 2025-08-13
fix(agentic-ai): release resources where necessary

mdii:
## Description

`EmbeddingStore` implementations often manage external resources (database connections, clients) but don't implement `AutoCloseable`, making it impossible to use try-with-resources for proper cleanup.

### Solution
* Create `ClosableEmbeddingStore`  that is a wrapper for `EmbeddingStore`, which implements `AutoCloseable`. This class delegates all embedding store operations to the wrapped instance.
* `ClosableEmbeddingStore`  supports optional cleanup actions via `Runnable closeAction` parameter and falls back to calling `close()` on the wrapped instance if it implements `AutoCloseable`.

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/src/main/java/io/camunda/connector/embeddingstore/AzureVectorStoreFactory.java
+++ b/connectors/embeddings-vector-database/src/main/java/io/camunda/connector/embeddingstore/AzureVectorStoreFactory.java
@@ -22,6 +22,7 @@ import com.azure.cosmos.models.IndexingPolicy;
 import com.azure.cosmos.models.PartitionKeyDefinition;
 import com.azure.identity.ClientSecretCredential;
 import com.azure.identity.ClientSecretCredentialBuilder;
+import dev.langchain4j.data.segment.TextSegment;
 import dev.langchain4j.model.embedding.EmbeddingModel;
 import dev.langchain4j.store.embedding.azure.cosmos.nosql.AzureCosmosDbNoSqlEmbeddingStore;
 import dev.langchain4j.store.embedding.azure.search.AzureAiSearchEmbeddingStore;
@@ -38,7 +39,7 @@ public class AzureVectorStoreFactory {
   public static final String COSMOS_DB_VECTOR_EMBEDDING_PATH = "/embedding";
   public static final String COSMOS_DB_PARTITION_KEY_PATH = "/id";
 
-  public ClosableEmbeddingStore createAiSearchVectorStore(
+  public ClosableEmbeddingStore<TextSegment> createAiSearchVectorStore(
       AzureAiSearchVectorStore azureAiSearchVectorStore,
       EmbeddingModel model,
       VectorDatabaseConnectorOperation operation) {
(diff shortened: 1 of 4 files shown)


connectors/embeddings-vector-database/src/main/java/io/camunda/connector/embeddingstore/ClosableEmbeddingStore.java
```diff
@@ -0,0 +1,131 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * A wrapper for {@link EmbeddingStore} that implements {@link AutoCloseable} to ensure proper
+ * resource management. This class delegates all embedding store operations to the wrapped instance.
+ *
+ * <p>Usage example:
+ *
+ * <pre>{@code
+ * try (ClosableEmbeddingStore store = ClosableEmbeddingStore.from(embeddingStore)) {
```
maff, line 27:
```suggestion
 * try (ClosableEmbeddingStore store = ClosableEmbeddingStore.wrap(embeddingStore)) {
```

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/embeddingstore/ClosableEmbeddingStore.java
```diff
@@ -0,0 +1,131 @@
...
+ *     store.add(embedding, textSegment);
+ *     // store will be automatically closed
+ * }
+ * }</pre>
+ */
+public class ClosableEmbeddingStore implements EmbeddingStore<TextSegment>, AutoCloseable {
```
maff, line 33: 💭 Just a thought, might not really make a difference: could this also be a generic `ClosableEmbeddingStore<T> implements EmbeddingStore<T>, AutoCloseable`, so it would work for other use cases than `TextSegment`?

maff (dismissed):
Awesome improvement 💪, pre-approving with 2 small comments.

Do you think it would make sense to raise this issue with Langchain4J so they'd make the stores `AutoClosable` on the framework level if needed?

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/embeddingstore/ClosableEmbeddingStore.java
mdii: I intentionally made it specific to `TextSegment` for simplicity as we currently only work with `TextSegment`. We can make it generic when we support multi-modal. But I am not against making it generic right now either.

connectors/embeddings-vector-database/src/main/java/io/camunda/connector/embeddingstore/AzureVectorStoreFactory.java
```diff
@@ -100,14 +99,18 @@ public EmbeddingStore<TextSegment> createCosmosDbNoSqlVectorStore(
...
+        AzureCosmosDbNoSqlEmbeddingStore.builder()
+            .cosmosClient(cosmosClient)
+            .databaseName(cosmosDbNoSql.databaseName())
+            .containerName(cosmosDbNoSql.containerName())
+            .cosmosVectorEmbeddingPolicy(embeddingPolicy)
+            .cosmosVectorIndexes(List.of(vectorIndexSpec))
+            .containerProperties(containerProperties)
+            .build();
+    //noinspection Convert2MethodRef
+    return ClosableEmbeddingStore.wrap(store, () -> cosmosClient.close());
```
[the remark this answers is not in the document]
mdii: Using method reference leads to NPE in tests, where we mock things, as the expression is evaluated eagerly.

maff (approved):

PR #5236, opened 2025-08-18
feat(agentic-ai): add support for event sub-processes

### LINKED ISSUE #5233: Agentic AI: Support events in ad-hoc sub-process job worker

With the AI Agent ad-hoc sub-process job worker, we're now able to support non-interrupting event sub-processes within the ad-hoc sub-process. This needs to be properly handled in the AI Agent codebase.

Image

For each handled event, a new job is created which is able to decide how to proceed. The worker needs to decide what to do when an event is handled while tools are currently being executed. On a high level, it has 2 possibilities:

- Wait for all tools to complete, then add the event data to the messages and call the LLM.
- "Abort" all tool calls by issuing a synthetic tool call result (e.g. "tool was canceled"), add the event data to the messages and call the LLM immediately.

Between these options, multiple nuances exist (e.g. only interrupt for some events, only interrupt on specific sub-processes), but we should start by adding generic event handling by supporting a configurable default behavior.

- [x] Add a new configuration property (only for the job worker) configuring event behavior to be either `WAIT_FOR_TOOL_CALL_RESULTS` (default when not configured) or `INTERRUPT_TOOL_CALLS`.
- [x] Detect "event data" as entries in the tool call results without ID/Name (not created from a tool call). In a later iteration, this could also be a dedicated data structure taken into account from the output element expression (e.g. `eventResult`).
- [x] When event data is available while processing tool call results and not all tool call results are available:
  - [x] When event behavior is `WAIT_FOR_TOOL_CALL_RESULTS`, just complete the job without calling the LLM. On the next completion event data is added as dedicated user message when all tool call results are available.
  - [x] When event behavior is `INTERRUPT_TOOL_CALLS`, add a synthetic "aborted" message for each missing tool call result, add the event data to the messages list and call the LLM. Complete the job including the `cancelRemainingInstances` flag set to `true`.

Not handling events.

__Additional context__

See discussion: https://camunda.slack.com/archives/C089AEDELPP/p1755204002479409

maff:
## Description

Introduces support to handle messages from events. With this implementation, an "event message" is a tool call result without an ID. An event sub-process can just write to `toolCallResult` and the element created by the `outputElement` mapping will be missing ID and name. As a follow-up, we could think of reading something like `eventResult` and updating the output element to take that into consideration.

For the handling of the events, this PR introduces a new setting on the Job Worker element template, allowing to configure the event behavior to either:

- Wait for tool call results, then add the event data as new user message and call the LLM
- Interrupt running tool calls (add a synthentic "Tool call was canceled" result), add the new event data as new user message and call the LLM immediately.

This configuration is valid for all event sub-processes in the AI Agent sub-process, but we can enhance this to be the default behavior which can be overridden from the actual event payload (e.g. by setting a variable `eventBehavior` which is handled by the output element) in a follow-up.

See #5233 for a detailed description.

## Related issues

closes #5233

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandlerImpl.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandlerImpl.java
@@ -10,7 +10,6 @@ import static io.camunda.connector.agenticai.aiagent.agent.AgentErrorCodes.ERROR
 import static io.camunda.connector.agenticai.model.message.MessageUtil.singleTextContent;
 import static io.camunda.connector.agenticai.model.message.content.ObjectContent.objectContent;
 import static io.camunda.connector.agenticai.model.message.content.TextContent.textContent;
-import static io.camunda.connector.agenticai.model.tool.ToolCallResult.PROPERTY_INTERRUPTED;
 import static io.camunda.connector.agenticai.util.PromptUtils.resolveParameterizedPrompt;
 
 import io.camunda.connector.agenticai.aiagent.memory.runtime.RuntimeMemory;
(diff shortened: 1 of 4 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/JobWorkerAgentExecutionContext.java
```diff
@@ -90,4 +97,12 @@ public JobClient jobClient() {
...
     return job;
   }
+
+  public boolean cancelRemainingInstances() {
+    return cancelRemainingInstances;
+  }
+
+  public void setCancelRemainingInstances(boolean cancelRemainingInstances) {
+    this.cancelRemainingInstances = cancelRemainingInstances;
+  }
```
maff, line 107: Not 100% happy with this one as it derives from the pattern of this being a read-only object, but it was the most obvious way to propagate this flag to the completion logic.

PR #5242, opened 2025-08-19
refactor(agentic-ai): integrate job worker template transformation into Maven build

### LINKED ISSUE #5189: Agentic AI: Generate and publish a job worker AI Agent template

With the job worker approach, we also need to generate and publish an element template providing the needed configuration values for the job worker approach which differs from the service task approach.

With https://github.com/camunda/connectors/issues/5175, this template is derived from the service task based template by transforming the JSON structure, but ideally the template can be generated as it can be done for normal connectors.

- The ad-hoc sub-process/job worker element template can be generated as part of the build process, including support for hybrid templates.
- The generated template is published in the same way as other element templates
- The versioning logic applied to other element templates is supported for the generated job worker template (when a version is changed, a file with the previous version is created in the `versioned` directory).

If this is not going to be supported by the element template generator, we should make the transformation part of the build pipeline. Maybe a Java based approach would work without needing the dependency on `jq`.

Keep the bash script based approach, which demands additional tooling (`jq`) being available and is not directly integrated into the build process.

__Additional context__

Also see https://github.com/camunda/connectors/issues/4430

maff:
## Description

Instead of providing a bash script relying on `jq`, this PR adds support to generate the AI Agent job worker template during the Maven build via a Groovy script.

- By using Jackson with the same settings as the `element-template-generator`, the outbound connector and job worker templates are easier to diff.
- Generates a hybrid template in addition to the normal one.
- Applies the same versioning logic as the element template generator.

## Related issues

closes #5189

## Checklist


mdii (approved):
:+1:

PR #5262, opened 2025-08-20
fix(deps): revert update com.microsoft.sqlserver:mssql-jdbc to v13 (main)"

chillleader:
Reverts camunda/connectors#5255

The original PR introduced a build failure due to a transitive dependency requiring Java 24


ztefanie (approved):
Thanks for fixing this :+1:

sbuettner (approved):

PR #5264, opened 2025-08-20
test(runtime): fix search query client test

chillleader:
## Description

Found another failing test that's only visible in the nightly runs. Not sure how the error was introduced, but the page limit is set to 0 as `@Value` inside the `SearchQueryClientImpl` does not have any effect during manual object creation.

## Related issues

closes #

## Checklist


sbuettner (approved):

sbuettner:
@chillleader Shouldnt we add the `@Value` annotation to the constructor parameter and remove the constructor without the parameter? 

``` SearchQueryClientImpl(CamundaClient camundaClient, int limit)```

chillleader:
Good point @sbuettner, I'll do a quick follow-up

PR #5269, opened 2025-08-21
chore(sdk): Refactor code from sdk to commons and create new test submodule

ztefanie:
## Description
- Moved `document` from sdk to runtime
- moved `intrinsic functions` from commons to runtime
- Http-client does no longer depend on document
- tests that use Document and http client stuff are moved to documents. Maybe we need to improve this.

New dependency graph is now:
image

## Related issues

Preparation for External document feature

## Checklist


sbuettner (approved):
I discussed this PR with @ztefanie and it goes into the right direction (moving the document implementations from the SDK to the runtime). Lets do a follow up to remove the intrinsic function handling (the registry etc) to the runtime together with other parts like the `DocumentFactoryImpl`.

chillleader:
Thanks @ztefanie this looks great! I'm happy we are slowly but steadily moving towards a layout that eliminates our dependency issues 🙂 

I will take a detailed look but here are some first thoughts on `common` vs. `sdk` vs. `runtime` for the affected components:
- I completely agree that we should move stuff out of the SDK as much as possible
>I am currently missing why we consider the document package as well as the intrinsic functions part of the "commons"
  - regarding this, I saw the `commons` package as a bunch of small single-responsibility modules that can be shared between OOTB connectors and runtime (example: HTTP client is used both in OOTB and the runtime via Document implementation)
  - I'm also not against putting things like `documents` and `intrinsic-functions` into Runtime. But now there is a high-level circular dependency between Runtime and `commons`: 
    - `document` in the Runtime depends on `http-client` from Commons
    - `object-mapper` from Commons depends on `jackson-datatype-*` from the Runtime
- This is probably not critical since it does not cause any build issues (there is no circle between individual modules, but IMO it is cleaner if all dependencies flow in one direction
- I'm also fine with the approach in this PR but we should be careful not to introduce new circles in the future

sbuettner (approved):
As discussed in the hacking session, this is the right way forward!

PR #5286, opened 2025-08-25
fix: `toolCallResult` bubbling up to process instance level

mathieu-stennier:
## Description

**When** I have `<zeebe:adHoc outputCollection="toolCallResults" outputElement="={&#10;  id: toolCall._meta.id,&#10;  name: toolCall._meta.name,&#10;  content: toolCallResult&#10;}" />` in my ad-hoc sub-process worker configuration.
**When** I send a request to 2 tools being the same
**Then** the tool results we have back are the same even if output is different on toolCalls themselves

```
"toolCalls": [
					{
						"id": "call_E4caD9aKY11wK3Lj7CFKTIyy",
						"name": "SuperfluxProduct",
						"arguments": {
							"b": 5,
							"a": 3
						}
					},
					{
						"id": "call_Hmdmehw42o3HxqTl302o3pc6",
						"name": "SuperfluxProduct",
						"arguments": {
							"b": 10,
							"a": 4
						}
					}
```
sent by the job worker.
results from engine:
```
{
				"role": "tool_call_result",
				"results": [
					{
						"id": "call_E4caD9aKY11wK3Lj7CFKTIyy",
						"name": "SuperfluxProduct",
						"content": 42
					},
					{
						"id": "call_Hmdmehw42o3HxqTl302o3pc6",
						"name": "SuperfluxProduct",
						"content": 42
					}
				],
				"metadata": {
					"timestamp": "2025-08-25T14:40:44.760538+02:00"
				}
			},
```

`call_E4caD9aKY11wK3Lj7CFKTIyy` should have resulted in `24 = (5+3)*3`.

Image

## Related issues

closes #

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
```diff
@@ -149,7 +149,13 @@ private void completeAsyncWithResponse(
...
                         adHocSubProcess
                             .activateElement(toolCall.metadata().name())
-                            .variables(Map.of(AiAgentJobWorker.TOOL_CALL_VARIABLE, toolCall));
+                            .variables(
+                                Map.ofEntries(
+                                    Map.entry(AiAgentJobWorker.TOOL_CALL_VARIABLE, toolCall),
+                                    // Creating empty toolCall result variable to avoid variable
+                                    // to bubble up in the upper scopes while merging variables on
+                                    // ad-hoc sub-process inner instance completion.
+                                    Map.entry(AiAgentJobWorker.TOOL_CALL_RESULT_VARIABLE, "")));
```
mdii, line 158: Is this a workaround or a definite fix?

mathieu-stennier: This would be the definite fix as this is the only way I know to prevent a local variable to propagate further up the execution tree. This is also the recommended approach in Camunda docs:

> To deactivate the variable propagation, the variables are set as local variables. This means the variables are created or updated in the given scope, regardless if they existed in this scope before.
...
> The propagation ends when a scope contains a variable with the same name. In this case, the variable value is updated.

https://docs.camunda.io/docs/components/concepts/variables/#output-mappings

PR #5291, opened 2025-08-26
feat(secret-provider): Add AWS Secret Provider

ztefanie:
## Description
I mostly moved the Code from GCP Secrets Provider to the abstract secrets provider and the only implented the custom load logic in gcp and aws secret providers.

add these properties to saas bundle application file to test or set them via env vars.

```
camunda.saas.cluster.provider=aws
camunda.client.cloud.clusterId=my-cluster-id
camunda.saas.secrets.useAwsSecretProvider=true
```

## Related issues
closes [#aws](https://github.com/camunda/team-connectors/issues/1086)

## Checklist


bundle/camunda-saas-bundle/pom.xml
```diff
@@ -65,6 +65,12 @@
       <artifactId>spring-security-test</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.camunda.connector</groupId>
+      <artifactId>secret-provider</artifactId>
+      <version>8.8.0-SNAPSHOT</version>
+      <scope>compile</scope>
+    </dependency>
```
sbuettner, line 73: We usually dont reference using the version but declare the version in the parent POM dependencies.

secret-providers/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Secret Provider that delegates to the GCP Security Manager.
+
+Requires the following properties being set,e.g. via application.properties:
+
+* `camunda.client.cloud.clusterId`: Cluster ID of Camunda SaaS cluster
+* `camunda.saas.secrets.projectId`: GCP project ID to use
+
+Optionally the following property can be set:
+* `camunda.saas.secrets.prefix`: If not set defaults to `connector-secrets`
```
sbuettner, line 15: I think the contents were focused on using the secret providers independently of our SaaS setup. We should there not mention:
> For AWS SaaS Clusters the AWS Secret Manager is used. For GCP SaaS Clusters the GCP Secret Manager is used.
but treat the secret providers as generic implementations that can also be used outside of our SaaS offering in self-managed setups.

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/SaaSSecretConfiguration.java
```diff
@@ -40,13 +42,24 @@ public class SaaSSecretConfiguration {
   @Value("${camunda.client.cloud.clusterId}")
   private String clusterId;
 
+  @Value("${camunda.saas.cluster.provider:gcp}")
+  private String clusterProvider;
+
+  @Value("${camunda.saas.secrets.useAwsSecretProvider:false}")
```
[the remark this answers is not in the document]
vringar: Should we have a ticket to remember?

secret-providers/aws/src/main/resources/io.camunda.connector.api.secret.SecretProvider
vringar, line 1: What does this file enable?

sbuettner: @vringar Its for the https://docs.oracle.com/javase/8/docs/api/java/util/ServiceLoader.html

vringar: @sbuettner but then it's not in the correct folder, right? The original file was also in `secret-providers/gcp-secret-provider/src/main/resources/META-INF/services/io.camunda.connector.api.secret.SecretProvider` (note the META-INF/services)

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/SaaSSecretConfiguration.java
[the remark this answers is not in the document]
johnBgood: I don't get why using the variable above (`clusterProvider` with `GCP` as the default value) is not enough?

secret-providers/aws/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+  </parent>
+
+  <artifactId>aws-secret-provider</artifactId>
+  <name>AWS secret provider</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>software.amazon.awssdk</groupId>
+      <artifactId>secretsmanager</artifactId>
+      <version>2.32.29</version>
```
johnBgood, line 20: similar to how it's done for [gcp](https://github.com/camunda/connectors/pull/5291/files#diff-da18030be8386ba7e507fad67f06eede500316d933a5d179dbb5d9622ce58c3dR22), we could add this version in our parent module

secret-providers/aws/src/main/java/io/camunda/connector/secret/providers/AwsSecretProvider.java
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  public AwsSecretProvider(
+      ObjectMapper mapper, String clusterId, String secretsProjectId, String secretsNamePrefix) {
+    super(mapper, clusterId, secretsProjectId, secretsNamePrefix);
+  }
+
+  @Override
+  protected String loadSecrets(
+      String clusterId, String secretsProjectId, String secretsNamePrefix, Logger LOGGER) {
```
johnBgood, line 46: :nit: `LOGGER` should be lowercase

johnBgood (commented):
Left a few questions :)

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/SaaSSecretConfiguration.java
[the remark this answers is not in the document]
johnBgood: Not sure I get it:

> if the information about the provider is set it would change the secrets behavior
This is what we want, right?

We don't need to involve the Controller team for the testing phase imo, we could just:
1. merge this PR with the provider param only
2. try it in dev using SNAPSHOT, add an ENV VAR manually set to AWS

Just like a regular feature, or is there a difference compared to other features that I'm missing?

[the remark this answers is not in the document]
ztefanie: Yes, I think the biggest difference here is, that we need to sync this with the controller and console team propertly, because they will have a time period, where console team will write secrets to gcp and aws secret stores (to ensure no secrets get lost) and there will be a time, where all secrets will (probably) be migrated. And this will happen for DEV, INT and PROD at different times.

(The controller team will also migrate the whole control plane of aws clusters to aws.)

As timing is critical here, i think it is easier and safer to keep it like that, as it is no big deal to remove this feature flag in the future or is there any other disadvantage of this approach?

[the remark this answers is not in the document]
johnBgood: As discussed, thank you a lot for the context, it helps :) 
I was confused by the `clusterProvider` variable, and maybe the naming. 

We can keep it this way, as long as we remember to remove the other property later 👍

PR #5307, opened 2025-08-29
chore(polling-connector): add @ElementTemplate to the polling connector

mathias-vandaele:
## Description

Add @ElementTemplate to the polling connector, it required the creation of a new PollingRequest Class

It also add a new api key authentication that was not present, which is managed by the http base.

Here is the new element template
image

Here is the one before this PR.
image

This ET is defined to be available from 8.8 only

I tested with this bpmn (github does not accept bpmn file so you will have to change the extension)
[polling_simple_test.xml](https://github.com/user-attachments/files/22041194/polling_simple_test.xml)

start the process with 

```
{ "code" : 200 }
```

## Related issues

closes #

## Checklist


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/HttpService.java
```diff
@@ -26,6 +27,17 @@ public HttpCommonResult executeConnectorRequest(HttpCommonRequest request) {
     return executeConnectorRequest(request, null);
   }
 
+  public HttpCommonResult executeConnectorRequest(HttpCommonInboundRequest request) {
```
sbuettner, line 30: `base` should not be aware of the inbound request. Can you do the mapping in the PollingConnector itself?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonInboundRequest.java
```diff
@@ -0,0 +1,180 @@
...
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import io.camunda.connector.http.base.model.auth.Authentication;
+import jakarta.validation.Valid;
+import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.Pattern;
+import java.util.Map;
+import java.util.Optional;
+
+public class HttpCommonInboundRequest {
```
sbuettner, line 20: This should live in the PollingConnector

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/HttpService.java
sbuettner: @johnBgood Yes, that was I was thinking about as well. Does that work for you @mathias-vandaele? We dont have any need for this in the `base` module itself.

mathias-vandaele: Yes, no issues on my side

Issue #5303, opened 2025-08-28
Refactor document code


ztefanie:
- move connector-document into runtime-core
- provide a test implementation of the DocumentFactoryInterface that should be in sdk/test that only depends on the core-sdk (similar to TestDocumentFactory but not depend on DocumentFactoryImpl)
- no connector should depend on CamundaDocument
- TestDocument class needs to be created to get rid of CamundaDocument dependencies in /connectors/
- remove     <dependency>
      <groupId>io.camunda.connector</groupId>
      <artifactId>connector-runtime-core</artifactId>
    </dependency> from connector-sdk/test

PR #5309, opened 2025-08-29
chore(refactoring): refactor document code

ztefanie:
## Description

- move connector-document into runtime-core/documents
- introduced new classes `TestDocument`, `TestDocumentFactory`, `TestDocumentMetadata` and `TestDocumentStore`
- moved interface CamundaDocumentStore from runtime to sdk
- fixed all connectors/tests so they do no longer use CamundaDocument

## Related issues

closes https://github.com/camunda/connectors/issues/5303

## Checklist


connector-sdk/test/src/main/java/io/camunda/connector/test/TestDocumentMetadata.java
```diff
@@ -0,0 +1,78 @@
+package io.camunda.connector.test;
```
ztefanie, line 1: This file is basically a copy of DocumentMetadataImpl.

Let me which option you like more:
:one: Copy DocumentMetadataImpl to TestDocumentMetadata
:two: Move DocumentMetadataImpl from runtime-core to sdk/core

sbuettner: @chillleader If its in the SDK I would prefer an interface as it probably should be constructed by client code.

chillleader: It's not supposed to be constructed by client code, the client passes all data in `DocumentCreationRequest`, and this class is just part of the document reference model

ztefanie: So  I would then remove the DocumentMetadata interface and create a class DocumentMetadata in the runtime, right?
But then how would I handle `DocumentReferenceModel` in sdk/jackson-datatype-document using the current DocumentMetadata interface?

sbuettner: Yes, I would prefer interfaces in the SDK as much as possible.

Yes, basically 1.

ztefanie:
I just realized that as we moved `connector-document` into `runtime-core/documents` we can no longer have dependency 
```
    <dependency>
      <groupId>io.camunda.connector</groupId>
      <artifactId>connector-document</artifactId>
    </dependency>
```
But would need to add:
```
    <dependency>
      <groupId>io.camunda.connector</groupId>
      <artifactId>connector-runtime-core</artifactId>
    </dependency>
```

But then we again get cyclic dependencies. I will need to have another look on how to resolve this.

chillleader (approved):
Great work ⭐ Just one small comment

connector-sdk/jackson-datatype-document/src/main/java/io/camunda/connector/document/jackson/deserializer/DocumentDeserializer.java
```diff
@@ -57,7 +57,8 @@ protected Document handleJsonNode(JsonNode node, DeserializationContext context)
       return documentFactory.resolve(reference);
     }
     if (node.isArray()) {
-      List<JsonNode> elements = Lists.newArrayList(node.elements());
+      List<JsonNode> elements = new ArrayList<>();
+      node.elements().forEachRemaining(elements::add);
```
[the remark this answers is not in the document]
ztefanie: Yes it was needed as we changed from `com.google.common.collect.Lists;` to `java.util.ArrayList;` and the array lists did not support adding directly as we got this error `Cannot resolve constructor 'ArrayList(Iterator<JsonNode>)'`

sbuettner (dismissed):

chillleader (dismissed):

sbuettner (approved):

PR #5324, opened 2025-09-01
fix(et): Remove idp base connectors from publishing

### LINKED ISSUE #5313: IDP Extraction outbound connector shows up in the OOTB element templates
__Describe the Bug__

When looking at the OOTB connectors we can select the IDP Extraction outbound connector. This is wrong, since it should not be selectable as am OOTB element template, but it should instead be published bu the IDP wizard in web modeler. The problem originates [here](https://github.com/camunda/connectors/blob/39dac6e90e76b29066a16f2a45d4f79882eb45e1/connector-templates.json#L9) in the `connector-templates.json`

__Steps to Reproduce__

1. Check the list of ootb connectors in wm

__Expected Behavior__

We should not see this:
Image

__Environment__

- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]
- Camunda version: [e.g. 8.0.4 (self-managed)]

ztefanie:
## Description
Removed idp connector from npm package and uniquET.

## Related issues

closes #5313 

## Checklist

### DIFF
--- a/connector-templates.json
+++ b/connector-templates.json
@@ -1,2498 +1,2510 @@
 {
   "camunda.connectors.rpa" : [ {
     "version" : 1,
-    "ref" : "https://raw.githubusercontent.com/camunda/connectors/1e71e3e62c1a3f31e51cf15618f01af262b5bfe1/connectors/rpa/element-templates/rpa-connector.json",
-    "engine" : {
-      "camunda" : "^8.7"
-    }
-  } ],
-  "io.camunda.connector.IdpExtractionOutBoundTemplate.v1" : [ {
-    "version" : 2,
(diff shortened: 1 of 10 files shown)


connector-templates.json
```diff
@@ -1,2498 +1,2510 @@
 {
   "camunda.connectors.rpa" : [ {
     "version" : 1,
-    "ref" : "https://raw.githubusercontent.com/camunda/connectors/1e71e3e62c1a3f31e51cf15618f01af262b5bfe1/connectors/rpa/element-templates/rpa-connector.json",
-    "engine" : {
-      "camunda" : "^8.7"
-    }
-  } ],
-  "io.camunda.connector.IdpExtractionOutBoundTemplate.v1" : [ {
```
ztefanie, line 9: So here the idp connector is removed

PR #5336, opened 2025-09-03
chore(local-setup): Fix rest address

ztefanie:
## Description

Fixing a issue with the local setup. Do we also need to make this change in the application.properties file? I am not sure if the port also changed in general.

## Checklist

### DIFF
--- a/bundle/default-bundle/src/test/resources/application-local.properties
+++ b/bundle/default-bundle/src/test/resources/application-local.properties
@@ -10,3 +10,4 @@ camunda.client.auth.token-url=null
 camunda.endpoints.cors.allowed.origins=http://localhost:6006
 camunda.endpoints.cors.mappings=/inbound-instances/**
 camunda.client.prefer-rest-over-grpc=false
+camunda.client.rest-address=http://localhost:8088


sbuettner (approved):

sbuettner:
> Do we also need to make this change in the application.properties file?

In which one @ztefanie? We have one for the default-bundle, saas and the application.

sbuettner:
@ztefanie should only be related to the local setup as other setups require the URLs to be set via env vars anyway.

sbuettner:
@ztefanie Do you have a different local setup? Asking as this 8080 seems to be the default: https://github.com/camunda/camunda-distributions/blob/main/docker-compose/versions/camunda-8.8/docker-compose-core.yaml#L54

sbuettner:
Good point, shouldnt we raise this with the distribution team? As the java client also defaults to 8080 but the spring config points to 8088 :) 


https://github.com/camunda/camunda/blob/cc1cfe793300cf1d9ef26dab8b8f9bb75002f825/clients/camunda-spring-boot-starter/src/main/resources/modes/self-managed.yaml#L5
```
rest-address: http://localhost:8088
```

https://github.com/camunda/camunda/blob/cc1cfe793300cf1d9ef26dab8b8f9bb75002f825/clients/java/src/main/java/io/camunda/zeebe/client/impl/ZeebeClientBuilderImpl.java#L83
```
public static final URI DEFAULT_REST_ADDRESS = getURIFromString("http://0.0.0.0:8080");
```

sbuettner:
@ztefanie do you want to merge it or wait until the zeebe client is released that comes with the new default?

ztefanie:
> @ztefanie do you want to merge it or wait until the zeebe client is released that comes with the new default? Happy to merge it and remove later

As this PR fixes it only on the local application file i think we would benefit of changing it with this pr.

PR #5339, opened 2025-09-03
other(logging): replace error logs with debug for saas auth type

chillleader:
## Description

- Replace error logs with warnings when SaaS auth type condition is checked
- Add logging to health checks for easier debugging

## Related issues

closes #

## Checklist


connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/ZeebeHealthIndicator.java
```diff
@@ -47,6 +51,10 @@ protected void doHealthCheck(Builder builder) {
     if (numBrokers > 0 && anyPartitionHealthy) {
       builder.up().withDetails(details);
     } else {
+      LOG.warn(
+          "Zeebe health check failed: numBrokers={}, anyPartitionHealthy={}",
+          numBrokers,
+          anyPartitionHealthy);
```
sbuettner, line 57: We probably wont catch it when 
```
var topology = camundaClient.newTopologyRequest().send().join();
```
throws an exception. Can we do this in a try catch where we log and rethrow?

PR #5366, opened 2025-09-10
feat(external-document): Add new external document type

### LINKED ISSUE #4399: External document support

ztefanie:
## Description
Added new document type `"camunda.document.type": "external"`

So documents can be used like this: (name is optional)

```
{
    "camunda.document.type": "external",
    "url": "https://www.antennahouse.com/hubfs/xsl-fo-sample/pdf/basic-link-1.pdf", 
    "name": "test2.pdf"
}
```

Example BPMN to test can be found [here](https://modeler.ultrawombat.com/diagrams/dc330357-6181-42e8-b7f6-31d4a936a18c--externaldocument?v=525,363,1)

## Related issues

closes #4399

## Checklist


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/document/ExternalDocument.java
```diff
@@ -0,0 +1,163 @@
...
+  private final String url;
+  private final String name;
+  private transient DocumentMetadata metadata;
+  private final HttpClientRequest request;
+  private final HttpClientService httpClientService;
+  private HttpClientResult result = null;
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ExternalDocument.class);
+
+  public ExternalDocument(String url, String name) {
```
sbuettner, line 48: Idea: To decouple the `ExternalDocument` from the actual HTTP handling and to avoid creating a new HttpClientService for every document you could add a parameter like `Supplier<HttpClientResult>` to the constructor and do the rest outside of this class.

connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/HttpCommonResultResponseHandler.java
```diff
@@ -141,6 +150,10 @@ private Object extractBody(byte[] content) throws IOException {
       return Base64.getEncoder().encodeToString(content);
     }
 
+    if (isShouldReturnRawBodySelected) {
+      return content;
```
johnBgood, line 154: please double-check, but I think this variable is not necessary. 

We could just return `content` instead of `bodyString` line 165 (previously line 152). 
then the apache client will handle the byte[], the String conversion is not necessary. 

Again, just a gut feeling :)

connector-commons/http-client/src/main/java/io/camunda/connector/http/client/document/HttpHeaderFilenameResolver.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  private static String getFilenameFromContentType(Map<String, Object> headers) {
+    Object ctObj = CustomApacheHttpClient.getHeaderIgnoreCase(headers, HttpHeaders.CONTENT_TYPE);
+    if (ctObj instanceof String ct && ct.contains("/")) {
+      String subtype = ct.substring(ct.indexOf('/') + 1);
+      return "." + subtype;
+    }
+    return "";
+  }
```
sbuettner, line 58: Is this resolving the filename or the file extension? 

If its for the latter we should rename this method appropriately and the approach might not work out all the time: https://stackoverflow.com/questions/13650372/how-to-determine-appropriate-file-extension-from-mime-type-in-java

sbuettner: @ztefanie My concern are mime types like this: [image/svg+xml](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Image_types#svg_scalable_vector_graphics) that come with a "plus" sign. I assume we would call a file with this subtype myImage.svg+xm? I think we should only use the first alphanumeric characters without any other signs.

If we take of that I would also be fine skipping the more elaborate solution for now but creating a follow-up issue to fix this holistically (same applies for store document response I assume)

sbuettner (approved):

sbuettner (approved):

Issue #5335, opened 2025-09-03
Refactor result handling of rest requests


ztefanie:
__What should we do?__

This code in the rest connector might need refactoring:

```
  @Override
  public HttpClientResult handleResponse(ClassicHttpResponse response) {
    int code = response.getCode();
    String reason = response.getReasonPhrase();
    Map<String, Object> headers =
        HttpCommonResultResponseHandler.formatHeaders(response.getHeaders());

    if (response.getEntity() != null) {
      try (InputStream content = response.getEntity().getContent()) {
        if (executionEnvironment instanceof ExecutionEnvironment.SaaSCluster) {
          return getResultForCloudFunction(code, content, headers, reason);
        }
        var bytes = content.readAllBytes();
        var documentReference = fileResponseHandler.handle(headers, bytes);
        return new HttpClientResult(
            code,
            headers,
            documentReference == null ? extractBody(bytes) : null,
            reason,
            documentReference);
      } catch (final Exception e) {
        LOGGER.error("Failed to parse external response: {}", response, e);
      }
    }
    return new HttpClientResult(code, headers, null, reason);
  }
```

Do we really want to return a result with a null body if something went wrong (an exception is thrown)?

__Why should we do it?__
tbd

PR #5396, opened 2025-09-12
fix(rest): Handle DocumentCreation exception in the rest client

johnBgood:
## Related issues

closes https://github.com/camunda/connectors/issues/5335

## Checklist

### DIFF
--- a/connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/HttpCommonResultResponseHandler.java
+++ b/connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/HttpCommonResultResponseHandler.java
@@ -79,11 +79,8 @@ public class HttpCommonResultResponseHandler
             documentReference == null ? extractBody(bytes) : null,
             reason,
             documentReference);
-      } catch (final DocumentCreationException e) {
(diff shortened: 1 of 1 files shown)


sbuettner:
A conceptual question: Why do we handle the document creation inside the HTTP client at all?

The feature (`storeResponse`) that allows users to create documents from responses is AFAIK only available and exposed via the REST Connector. As the REST Connector now supports returning the raw content - what about handling the Document creation outside of the HTTP Client in the REST Connector?

This would simplify the HTTP Client and remove its dependency on the `DocumentFactory`. WDYT?

connector-commons/http-client/pom.xml
```diff
@@ -125,6 +125,16 @@ limitations under the License.</license.inlineheader>
       <artifactId>junit</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>testcontainers</artifactId>
+      <scope>test</scope>
+    </dependency>
```
sbuettner, line 132: Why do we need testcontainers for these tests?

sbuettner (approved):

PR #5411, opened 2025-09-15
feat(agentic-ai): Support error expressions for error handling

### LINKED ISSUE #5237: Agentic AI: Support error expressions on AI Agent job worker implementation
__What should we do?__

Add support for the error expression field in the AI Agent job worker error handler and handle it the same way as an outbound connector.

__Why should we do it?__

Consistency with outbound connectors & possibility to throw a BPMN error depending on the result.

### LINKED ISSUE #5246: Agentic AI: (Optional) consolidate job worker completion and error handling
Depends on https://github.com/camunda/connectors/issues/5237 being implemented first.

__What should we do?__

Currently job completion and error handling is implemented in 2 different handlers:

- `JobWorkerAgentRequestHandler` handles the request and the completion
- `AiAgentJobWorkerErrorHandler` handles generic error/exception handling reusing the outbound connector handler logic.

Ideally, we can consolidate these 2 and handle all the completions/failures on the same place.

__Why should we do it?__

Clearer execution paths. Potentially better to extract a reusable handler.

maff:
## Description

Supports configuring an [error expression ](https://docs.camunda.io/docs/next/components/connectors/use-connectors/#error-expression) on the AI Agent job worker template similar to other connectors. The job completion evaluates the expression with either an exception thrown during the execution or the agent response as response object.

This will allow to implement modeling error paths, such as taking a different path when the agent logic throws an exception on reaching the max model calls limit:

image

In addition to implement the error expression, this PR refactors how the client API calls are made in order to do all the API calls in a central place, very similar how this is handled for outbound connectors in `SpringConnectorJobHandler` (see https://github.com/camunda/connectors/issues/5246).

- `JobWorkerAgentRequestHandler` does not directly interact with the API anymore. Instead, it returns a `JobWorkerAgentCompletion` result object containing all the needed information to complete the job (variables, completion condition, elements to activate, ...).
- `AiAgentJobWorkerHandler` implements the same error handling logic as the core outbound connector handling, while using the completion information returned in JobWorkerAgentCompletion.

## Related issues

closes #5237
closes #5246

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
@@ -128,7 +128,6 @@ public abstract class BaseAgentRequestHandler<C extends AgentExecutionContext, R
         executionContext, agentContext, runtimeMemory, executionContext.systemPrompt());
 
     // update memory with user messages/tool call responses
-    LOGGER.trace("Adding system message (if necessary)");
     if (LOGGER.isDebugEnabled()) {
       LOGGER.debug(
           "Adding user messages including {} tool call results for the following tool call results: {}",


sbuettner:
@maff Just saw this PR and was curious why we want to deviate here from the Connectors behaviour? Curious about the use-case. Is it just about recovery? The idea to support an error handler also in a successful case was that one can also raise an error in case of a business logic error.

maff:
@sbuettner sorry, I missed your comment. Very valid point and I fully agree - I'm currently looking into the needed refactorings to allow the same behavior as for normal connectors.

gbetances089 (approved):

PR #5446, opened 2025-09-19
fix(agentic-ai): adapt e2e tests for updated response handling (not explicitely setting TEXT)

### LINKED ISSUE #5434: Agentic AI: Anthropic provider throws response format execption
__Describe the Bug__

The Anthropic provider only supports the text response format, but despite the format being configured to text it throws the following error:

`JSON response format is not supported by Anthropic`

__Steps to Reproduce__

1. Configure an AI Agent (both variants) and use the Anthropic provider. Make sure to use the text response format.
2. Execute the process. It will raise an incident with the error message outlined above.

Background: the L4J Anthropic validation [throws this misleading error](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/InternalAnthropicHelper.java#L27) as soon as the response format is set, even if it is set to `TEXT`. We should create an upstream issue, but to work around it, we can omit setting the format to `TEXT` and only explicitely configure it to be `JSON` hen needed.

__Expected Behavior__

Anthropic provider works as expected with the response format configured to text.

__Environment__

- Camunda version: 8.8.0-alpha8

maff:
## Description

Follow up to #5434 to fix missed e2e tests.

## Related issues

closes #5434

## Checklist

### DIFF
--- a/connectors/agentic-ai/AI_AGENT.md
+++ b/connectors/agentic-ai/AI_AGENT.md
@@ -146,6 +146,7 @@ leading to the following result
     } ],
     "metadata" : {
       "framework" : {
+        "finishReason" : "STOP",
         "id" : "chatcmpl-123",
         "tokenUsage" : {
           "inputTokenCount" : 5,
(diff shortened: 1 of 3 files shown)


connectors/agentic-ai/AI_AGENT.md
```diff
@@ -146,6 +146,7 @@ leading to the following result
     } ],
     "metadata" : {
       "framework" : {
+        "finishReason" : "STOP",
```
maff, line 149: Generated from the response samples - I missed re-building after the last adaption.

mdii (approved):

PR #5463, opened 2025-09-23
test(vector-db): add a test for validation

mdii:
## Description

Provide a minimum set of tests to test the input validation is set up correctly.

## Related issues

closes #

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/src/test/java/io/camunda/connector/model/EmbeddingsVectorDBRequestValidationTest.java
+++ b/connectors/embeddings-vector-database/src/test/java/io/camunda/connector/model/EmbeddingsVectorDBRequestValidationTest.java
@@ -51,6 +51,13 @@ class EmbeddingsVectorDBRequestValidationTest {
     var request = new EmbeddingsVectorDBRequest(null, sampleModelProvider(), sampleVectorStore());
     var violations = validator.validate(request);
     assertThat(violations).hasSize(1);
+    assertThat(violations.iterator().next())
+        .satisfies(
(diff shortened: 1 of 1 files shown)


connectors/embeddings-vector-database/src/test/java/io/camunda/connector/model/EmbeddingsVectorDBRequestValidationTest.java
```diff
@@ -0,0 +1,130 @@
...
+        new EmbeddingsVectorDBRequest(
+            new EmbedDocumentOperation(
+                EmbedDocumentSource.PlainText,
+                "document content",
+                List.of(),
+                new RecursiveDocumentSplitter(-1, 0)),
+            sampleModelProvider(),
+            sampleVectorStore());
+    var violations = validator.validate(request);
+    assertThat(violations).hasSize(2);
```
maff, line 82: 💭 Maybe we could assert the actual field & error message (or parts of it), as it would make it a bit easier to understand which validation is triggered by the constellation when looking at this test class.

maff (dismissed):
Small suggestion on the assertion, otherwise LGTM 👍

connectors/embeddings-vector-database/src/test/java/io/camunda/connector/model/EmbeddingsVectorDBRequestValidationTest.java
mdii: done!

PR #5474, opened 2025-09-26
fix(etg): handle primitive operation parameters correctly

chillleader:
## Description

When working with operation connectors, I noticed that parameters of primitive types are not handled correctly by the ETG (either ignored or treated as composite classes, which in case of `String` or boxed types results in a lot of garbage properties).

This PR adds separate handling for primitive operation parameters.

❓ **Unresolved question:** Should Header parameters always be of type string in the element template? smth like Number or other types could also be useful. I haven't changed this yet.

## Related issues

closes #

## Checklist


chillleader:
I'd prefer to backport this to release/8.8 as well

sbuettner (approved):

PR #5492, opened 2025-10-01
fix: disable helm integration tests for main because we do not have a 8.9 helm chart yet

jessesimpson36:
## Description

There are noisy alerts lately due to these failing helm chart integration tests. These are not expected to pass on main, since main of connectors (and camunda core) is referring to 8.9 builds and the helm chart is only on 8.8 at the moment.

This PR adds a conditional disabling the helm integration tests on main.

## Related issues

closes #

## Checklist


johnBgood:
@jessesimpson36 do you know when we can enable it again?

PR #5504, opened 2025-10-03
feat(agentic-ai): create A2A outbound connector skeleton

### LINKED ISSUE #5498: Agentic AI: Create the A2A outbound connector skeleton
__What should we do?__

Define the basic input/output structure and add the required annotations on top of  `A2AClientFunction` in order to generate the element template for the connector. No request handling logic is included.

__Why should we do it?__

To prepare for the request handling logic.

mdii:
## Description

Define the basic input/output structure and add the required annotations on top of `A2AClientFunction` in order to generate the element template for the connector. No request handling logic is included.

Note that the element template will be updated in the subsequent PR, where more logic will be implemented.

## Related issues

closes #5498

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/result/A2AClientSendMessageResult.java
```diff
@@ -0,0 +1,59 @@
...
+package io.camunda.connector.agenticai.a2a.client.model.result;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import io.camunda.connector.agenticai.model.AgenticAiRecord;
+import io.camunda.connector.agenticai.model.message.content.Content;
+import java.util.List;
+
+@AgenticAiRecord
+public record A2AClientSendMessageResult(String responseId, List<Content> contents, TaskState state)
```
maff, line 16: Fine to do this in a follow-up PR, but should we already model this to return either `Task` or `Message` and align the properties with the spec? Something like

`Task`
- `id`
- `contextId`
- `status` (including state & message)
- `history` (list of messages - to be defined how many we'd include here)
- `artifacts`
- `metadata`

`Message`
- `role`
- `messageId`
- `taskId`
- `contextId`
- `referenceTaskIds`
- `parts` (mapped to our `Content` items)
- `metadata`
- `extensions` (not sure if needed already, but we could add it and configure it as `JsonInclude.Include.NON_EMPTY` to omit it from serialization)

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/result/A2AClientSendMessageResult.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+package io.camunda.connector.agenticai.a2a.client.model.result;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+import io.camunda.connector.agenticai.model.AgenticAiRecord;
+import io.camunda.connector.agenticai.model.message.content.Content;
+import java.util.List;
+
+@AgenticAiRecord
```
maff, line 15: I'd suggest we also add the `@JsonDeserialize` config to all the models to force using the builder as done for most of the models (example: https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/model/message/SystemMessage.java#L29). 

I raised an issue to add this to record-builder some time ago, but for the moment we need to manually create this setup. By forcing deserialization through the builder we have a couple of benefits without implementing it in the record constructor:

- ensures empty collections are initialized vs. being `null`
- ensures non-nullable fields are not null
- allows setting defaults if needed

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/A2AClientFunction.java
```diff
@@ -0,0 +1,43 @@
...
+
+@OutboundConnector(
+    name = "A2A Client",
+    inputVariables = {"data"},
+    type = "io.camunda.agenticai:a2aclient:0")
+@ElementTemplate(
+    id = "io.camunda.connectors.agenticai.a2a.client.v0",
+    name = "A2A Client (experimental)",
+    description =
+        "Agent-to-Agent (A2A) client, enabling discovering remote agents' Agent Cards as well as invoking remove agents. Compatible with 8.8.0-alpha8 or later.",
```
maff, line 24: Should we remove the "Compatible with ..." part?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/A2AClientFunction.java
```diff
@@ -0,0 +1,43 @@
...
+    engineVersion = "^8.8",
```
maff, line 25: I think we can assume this will be 8.9 only, right?

```suggestion
    engineVersion = "^8.9",
```

mdii: Yes, it should be `8.9` only :+1:

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/result/A2AClientSendMessageResult.java
mdii: I wanted to do it later when we implement multi-turn conversations but I agree it is clearer to already return `Task` or `Message`. However, to keep it simple, let's only return the strict minimum properties for the moment. For example, we can include `history` in a dedicated PR, or we can leave out `extensions` for later.

maff: Sounds good! Happy with any variant as we'll definitely iterate on the structure, so we can also keep it like it is in this PR and address it in follow-ups if you prefer.

maff (approved):
Nice!

PR #5520, opened 2025-10-07
feat(agentic-ai): Implement A2A send message operation

### LINKED ISSUE #5500: Agentic AI: Implement the Send Message operation
__What should we do?__

Implement the blocking **Send Message** operation using the [A2A Java SDK](https://github.com/a2aproject/a2a-java). Handle the response, whether it is a Message or a Task. The _Parts_ in response should be converted to the _Contents_ that are supported in the AI Agent connector:

- TextPart ➛ Text
- DataPart ➛ Text (assuming it is always JSON)
- FilePart (FileWithBytes, FileWithUri) ➛ Camunda Document

Return error when the task status is not supported.

mdii:
## Description

Implements the send message operation for the A2A (Agent-to-Agent) connector with significant architectural improvements.

### Key Changes

**✨ New Features**
- **Send Message Operation**: Full implementation for sending messages to A2A agents in a sync way.
- **Content Converters**: New converter layer for transforming between Camunda documents and A2A protocol formats
  - `A2aDocumentToPartConverter`: Converts Camunda documents to A2A parts
  - `A2aPartToContentConverter`: Converts A2A parts back to content objects with metadata preservation

**🏗️ Architecture Refactoring**
- Introduced clean separation with `api/` (interfaces) and `impl/` (implementations) package structure
- Created focused components with single responsibilities:
  - `A2aMessageSender`: Handles message sending and response polling
  - `A2aAgentCardFetcher`: Fetches agent card information
  - `A2aSendMessageResponseHandler`: Processes send message responses
  - `A2aSdkClientFactory`: Factory for SDK client creation

**📦 Naming Consistency**
- Renamed classes to clarify distinction between SDK clients and connector components:
  - `A2aClientFunction` → `A2aConnectorFunction`
  - `A2aClientRequest` → `A2aRequest`
  - `A2aClientResult` → `A2aResult`
  - `A2aClientOperationConfiguration` → `A2aOperationConfiguration`

## Related issues

closes #5500

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/convert/A2aSdkObjectConverterImpl.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/convert/A2aSdkObjectConverterImpl.java
@@ -50,6 +50,9 @@ public class A2aSdkObjectConverterImpl implements A2aSdkObjectConverter {
   }
 
   private List<A2aArtifact> convertArtifacts(Task task) {
+    if (task.getArtifacts() == null) {
+      return List.of();
+    }
     return task.getArtifacts().stream()
         .map(
             artifact ->


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/api/A2aMessageSender.java
```diff
@@ -0,0 +1,16 @@
...
+ */
+package io.camunda.connector.agenticai.a2a.client.api;
+
+import io.a2a.spec.AgentCard;
+import io.camunda.connector.agenticai.a2a.client.model.A2aOperationConfiguration.SendMessageOperationConfiguration;
+import io.camunda.connector.agenticai.a2a.client.model.result.A2aSendMessageResult;
+
+public interface A2aMessageSender {
+  A2aSendMessageResult sendMessage(
+      SendMessageOperationConfiguration sendMessageOperation, AgentCard agentCard);
```
maff, line 15: 💭 Nit: from a logical/hierarchical perspective, would make `(agentCard, sendMessageOperation)` more sense (thinking of I can send multiple message to a single agent card)? This would also align with `(agentCard, consumer)` in the client factory.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/A2aMessageSenderImpl.java
```diff
@@ -0,0 +1,111 @@
...
+          && task.status().state().isSubmittedOrWorking()) {
+        Duration timeSpent = Duration.between(startTime, Instant.now());
+        Duration updatedTimeout = sendMessageOperation.timeout().minus(timeSpent);
+        return taskPoller
+            .poll(task.taskId(), client, DEFAULT_POLL_INTERVAL, updatedTimeout)
+            .get(updatedTimeout.toMillis(), TimeUnit.MILLISECONDS);
+      }
+      return result;
+    } catch (TimeoutException e) {
+      // TODO: should be a ConnectorException with a specific error code?
```
maff, line 91: As mentioned above we should probably do something like that, but could also be a dedicated follow-up dealing with error handling.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/A2aSendMessageResponseHandlerImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        Message message = messageEvent.getMessage();
+        return new A2aSendMessageResult.A2aMessageResult(buildMessage(message));
+      }
+      case TaskEvent taskEvent -> {
+        Task task = taskEvent.getTask();
+        return handleTask(task);
+      }
+      default ->
+          throw new RuntimeException(
+              "Only message events and completed tasks are supported in the response yet.");
```
maff, line 47: The `completed` part does not apply anymore, right?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/configuration/A2aConnectorConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+  @ConditionalOnMissingBean
+  public A2aSendMessageResponseHandler a2aSendMessageResponseHandler(
+      PartsToContentConverter partsToContentConverter) {
+    return new A2aSendMessageResponseHandlerImpl(partsToContentConverter);
+  }
+
+  @Bean
+  @ConditionalOnMissingBean
+  public ScheduledExecutorService taskPollerExecutor() {
+    return Executors.newScheduledThreadPool(2);
```
maff, line 62: Should we make the value configurable by introducing an a2a specific config property structure with 2 as default value?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/A2aMessageSenderImpl.java
```diff
@@ -0,0 +1,111 @@
...
+      A2aSendMessageResult result =
+          response.get(sendMessageOperation.timeout().toMillis(), TimeUnit.MILLISECONDS);
+      if (result instanceof A2aSendMessageResult.A2aTaskResult(A2aTask task)
+          && task.status().state().isSubmittedOrWorking()) {
+        Duration timeSpent = Duration.between(startTime, Instant.now());
+        Duration updatedTimeout = sendMessageOperation.timeout().minus(timeSpent);
+        return taskPoller
+            .poll(task.taskId(), client, DEFAULT_POLL_INTERVAL, updatedTimeout)
+            .get(updatedTimeout.toMillis(), TimeUnit.MILLISECONDS);
+      }
```
maff, line 88: 💭 Especially with added support for an inbound connector, we also need a variant which returns a non-completed task without raising an error. If I understand correctly, this implementation starts to poll for the configured time frame, but returns with an error if the task did not complete in time.

I think it would make sense to make this somewhat configurable on the properties level:

- Use polling at all?
- Poll Timeout
- Fail when not completing within the timeout or just return the latest task state?
- (max poll retries)
- (poll interval)

WDYT of just returning the task as it is as part of this PR and introducing the polling logic including config in a dedicated one?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/A2aMessageSenderImpl.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    try {
+      Instant startTime = Instant.now();
+      A2aSendMessageResult result =
+          response.get(sendMessageOperation.timeout().toMillis(), TimeUnit.MILLISECONDS);
+      if (result instanceof A2aSendMessageResult.A2aTaskResult(A2aTask task)
+          && task.status().state().isSubmittedOrWorking()) {
+        Duration timeSpent = Duration.between(startTime, Instant.now());
+        Duration updatedTimeout = sendMessageOperation.timeout().minus(timeSpent);
```
maff, line 84: 💭 Can this become negative or very close to zero, making starting the polling operation obsolete (thinking it tool almost as long as the timeout)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/TaskPollerImpl.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public TaskPollerImpl(
+      ScheduledExecutorService scheduler,
+      A2aSendMessageResponseHandler sendMessageResponseHandler) {
+    this.scheduler = scheduler;
+    this.sendMessageResponseHandler = sendMessageResponseHandler;
+  }
+
+  @Override
+  public CompletableFuture<A2aSendMessageResult> poll(
```
maff, line 33: I ran this through Copilot and it raised a couple of issues regarding cancellation of the task on completion:

> The timeout implementation has a critical flaw: it will never cancel the scheduled task, which means the poller will keep running even after timing out.
> Here are the issues:
> - No task cancellation: When timeout occurs, the scheduled task continues running in the background
> - Race condition: The timeout check happens at the start of each poll iteration, so it could poll one more time even after timeout
> - No cleanup: The scheduled task is never cancelled, wasting resources

It suggests using a dedicated future to handle the timeout (which is cancelled when completing the task in  pollTask): 

```java
  java.util.concurrent.atomic.AtomicReference<java.util.concurrent.ScheduledFuture<?>> scheduledTaskRef =
      new java.util.concurrent.atomic.AtomicReference<>();

  // Schedule timeout
  ScheduledFuture<?> timeoutTask = scheduler.schedule(() -> {
    if (future.completeExceptionally(new TimeoutException("Polling timed out for task: " + taskId))) {
      // Cancel the polling task if timeout completes the future first
      ScheduledFuture<?> pollTask = scheduledTaskRef.get();
      if (pollTask != null) {
        pollTask.cancel(false);
      }
    }
  }, timeout.toMillis(), TimeUnit.MILLISECONDS);
```

Could you revisit the implementation regarding these issues?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/sdk/A2aSdkClientFactoryImpl.java
```diff
@@ -0,0 +1,34 @@
...
+import io.a2a.spec.AgentCard;
+import io.camunda.connector.agenticai.a2a.client.api.A2aSdkClientFactory;
+import java.util.function.BiConsumer;
+
+public class A2aSdkClientFactoryImpl implements A2aSdkClientFactory {
+
+  @Override
+  public Client buildClient(AgentCard agentCard, BiConsumer<ClientEvent, AgentCard> consumer) {
+    try {
+      // Disable streaming for the time being
```
maff, line 24: Do you think we'll ever be able to support streaming? Otherwise I'd suggest we can remove the comment.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/configuration/A2aConnectorConfiguration.java
```diff
@@ -0,0 +1,106 @@
...
+
+@Configuration
+@ConditionalOnBooleanProperty(
+    value = "camunda.connector.agenticai.a2a.client.enabled",
+    matchIfMissing = true)
+public class A2aConnectorConfiguration {
+
+  @Bean
+  @ConditionalOnMissingBean
+  public DocumentToPartConverter documentToPartConverter(ObjectMapper objectMapper) {
```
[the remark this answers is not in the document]
mdii: Yes, it makes sense :+1:

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/A2aAgentCardFetcherImpl.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public AgentCard fetchAgentCardRaw(ConnectionConfiguration connection) {
+    final var relativeCardPath =
+        isNotBlank(connection.agentCardLocation()) ? connection.agentCardLocation() : null;
+    try {
+      return A2A.getAgentCard(connection.url(), relativeCardPath, Collections.emptyMap());
+    } catch (A2AClientError e) {
+      throw new RuntimeException(e);
```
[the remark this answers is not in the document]
mdii: Yes, let's do in a follow up :+1:

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/sdk/A2aSdkClientFactoryImpl.java
mdii: We can remove this comment anyways :+1:

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/impl/A2aMessageSenderImpl.java
mdii: We should decide if we should keep this polling at all when we have the inbound connector. I am more and more in favor of removing it when we can poll using the inbound connector. It will reduce the complexity of the code.
If we decide to keep it then it makes sense to make it configurable and have a dedicated PR for the polling part. Otherwise, we can keep on with the a bare minimum working version until we remove it later. WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/result/A2aTask.java
maff, line 1: 🔧 Could we align the task model with the spec and call `taskId` just `id`? I was using `task` as a result object and by looking at the spec I guessed referencing the ID would be `task.id`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/configuration/A2aConnectorConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+
+  @Bean
+  @ConditionalOnMissingBean
+  public A2aDocumentToPartConverter a2aDocumentToPartConverter(ObjectMapper objectMapper) {
+    return new A2aDocumentToPartConverterImpl(objectMapper);
+  }
+
+  @Bean
+  @ConditionalOnMissingBean
+  public A2aPartToContentConverter a2aPartsToContentConverter() {
```
maff, line 46: Nit: `partToContent...`

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/configuration/A2aConnectorConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+    return new A2APartToContentConverterImpl();
+  }
+
+  @Bean
+  @ConditionalOnMissingBean
+  public A2aSdkObjectConverter a2aSdkObjectConverter(
+      A2aPartToContentConverter partsToContentConverter) {
```
maff, line 53: Nit: Same regarding plural

maff (approved):
Comments are non-blocking, fine to address later. Great work! 💪

PR #5524, opened 2025-10-09
feat(agentic-ai): A2A inbound polling connector

### LINKED ISSUE #5545: Agentic AI: Add A2A Inbound Task Polling connector
__What should we do?__

As the nature of A2A communication is mostly asynchronous, we should add an inbound polling connector which can be used in combination with the A2A Client connector. This will allow to model different A2A communication patterns, including asynchronous ones.

Example:
1. Client sends request to A2A server
2. Server responds with a task in state `working`
3. Inbound polling connector polls the get task endpoint using the task ID until the task reached a state which is not `submitted` or `working`.

Image

To avoid having to model the conditionals shown in the example above, the connector should additionally support directly correlating already finished tasks and messages without starting to poll. The above model can still be used for efficiency reasons (letting the engine handle the completion without triggering the inbound connector), but it is not strictly necessary for better modeling UX. The following constellation should also work and result in the connector not polling, but directly correlating a message with the input object as result. This would apply to:

- Server responds with a message
- Server responds with a task in `completed` state (or another, non-intermediary one)

Image

__Why should we do it?__

Supporting async flows without having to poll within the client connector.

maff:
## Description

Implements an A2A inbound connector able to directly correlate completed tasks/messages or polling tasks in an intermediary state. It can be configured with a **Client Response** input which is designed to reference the output element of the A2A client connector.

2 element templates as provided, both implementing the same functionality:

- for message catch events
- for receive tasks (allows to set for example a timer boundary event to stop polling)

In order to also support completed tasks and messages, the connector logic is as follows:

- If the client response is a message, it is directly correlated
  - the activation condition is preconfigured to activate all messages
  - the correlation key is preconfigured to match the message id with the message ID of the input object (uses a `msg-` prefix for more clarity)
- If the client response is a task which is not in `working` or `submitted` state, it is directly correlated
  - the activation condition is preconfigured to activate tasks which are not in `working`/`submitted` state
  - the correlation key is preconfigured to match the task id with the ID of the input object (uses a `task-` prefix for more clarity)
- if the client response is a task which needs polling, it start polling with the task ID and tries to correlate the polling result
  - the activation condition applies as soon as the state of the polled task is not `working` or `submitted`
  - same correlation key logic as the directly correlated one

To allow setting up the above logic in a reusable manner and to deliver it as default, the connector makes use of an input expression to bind the client response to a variable `internal_clientResponse` which can be referenced from the correlation key expression.

image

## Related issues

closes #5545 

## Checklist

### DIFF
--- a/connectors/agentic-ai/examples/a2a/a2a-polling/a2a-polling-example-receive-task.bpmn
+++ b/connectors/agentic-ai/examples/a2a/a2a-polling/a2a-polling-example-receive-task.bpmn
@@ -54,6 +54,21 @@
       <bpmn:outgoing>Flow_10hibj6</bpmn:outgoing>
     </bpmn:receiveTask>
     <bpmn:sequenceFlow id="Flow_10hibj6" sourceRef="Wait_For_Completion" targetRef="EndEvent_Async" />
+    <bpmn:boundaryEvent id="TimerBoundaryEvent" attachedToRef="Wait_For_Completion">
+      <bpmn:outgoing>Flow_10n8abt</bpmn:outgoing>
+      <bpmn:timerEventDefinition id="TimerEventDefinition_0ohpulx">
+        <bpmn:timeDuration xsi:type="bpmn:tFormalExpression">PT1M</bpmn:timeDuration>
+      </bpmn:timerEventDefinition>
+    </bpmn:boundaryEvent>
+    <bpmn:endEvent id="Event_03qzgbb" name="Polling timed out">
+      <bpmn:extensionElements>
+        <zeebe:ioMapping>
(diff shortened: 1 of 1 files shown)


maff:
@mdii Do I understand correctly that `polling = false` would make the server reject a task if it would be long-running?

Maybe something we could expose as config option on the client to explicitly model interactions without a polling connector as part of the process?

maff:
>  /** If true, the client will wait for the task to complete. The server may reject this if the task is long-running. */
>  blocking?: boolean;

If got it wrong (the server may reject if the client is not able to poll and waits for a sync response) & I think `polling=false` is a reasonable default. Still we can consider additing it as config option 👍

mdii (approved):

PR #5558, opened 2025-10-14
Delete queue connector

DomingossDias:
## Description

This PR adds a new Camunda Connector called RabbitMQ Delete Queue Connector, responsible for deleting RabbitMQ queues using the native API (com.rabbitmq.client).

The connector follows the same structure and conventions as the RabbitMQ Outbound Connector

Captura de tela 2025-10-15 002706

### DIFF
--- a/connectors/agentic-ai/bin/transform-ai-agent-job-worker-template.groovy
+++ b/connectors/agentic-ai/bin/transform-ai-agent-job-worker-template.groovy
@@ -57,7 +57,7 @@ if (outputFilePath.exists() && !isHybrid) {
 
 // Update template metadata
 json.id = "io.camunda.connectors.agenticai.aiagent.jobworker.v1"
-json.name = "AI Agent Subprocess"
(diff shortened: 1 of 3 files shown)


chillleader:
Hi @DomingossDias, thanks for your PR!

We usually try to avoid highly specific connectors that can only perform a single action. Instead, we cluster these actions based on the system we're integrating with. For example, take a look at the [S3 Connector](https://github.com/camunda/connectors/blob/main/connectors/aws/aws-s3/src/main/java/io/camunda/connector/aws/s3/S3ConnectorFunction.java) - it's capable of uploading, deleting, and downloading objects, all in one connector function and element template. This grouping provides better user experience.

I think deleting a queue is a really valuable feature but in line with our approach, we don't want to create a separate connector just for this. Would you be willing to change your implementation and add a new operation to the existing [RabbitMQ connector](https://github.com/camunda/connectors/tree/main/connectors/rabbitmq) (outbound)?

PR #5562, opened 2025-10-15
fix(embeddings-vector-db): Remove 200 char retrieval size limit

### LINKED ISSUE #5553: Remove 200 char limitation on query input (Embedding vector store connector)
__Describe the Bug__

A hard-coded input size limit in the AI-related connector prevents longer query inputs from being processed, causing user-visible failures for larger inputs. The limit was originally added in April for unclear reasons and not strictly validated until 8.8.0-alpha8, which is why the issue only surfaced recently. The team does not recall a strong rationale for this cap beyond generally limiting query length, and agrees it is an unnecessary limitation rather than an intended feature. Short-term, the proposal is to remove the limit to restore pre-8.8.0-alpha8 behavior, while long-term options include query chunking and better error handling when embedding inputs exceed embedding model token capacities (typically 512–8192 tokens).

__Steps to Reproduce__

1.    Configure the connector to accept a query/input whose size exceeds the current configured limit.
2.    Execute the operation that triggers embedding/search on this input.
3.    Observe that processing fails due to the enforced size constraint (seen after validation introduced in 8.8.0-alpha8).

__Expected Behavior__

The connector should accept larger inputs by either removing unnecessary hard limits or by automatically chunking queries similarly to how content is chunked for embedding. If inputs still exceed the underlying embedding model’s token constraints, the system should fail gracefully with a clear and actionable error message, not an opaque size limit failure.

__Environment__

•    Channel/thread: Slack [#ask-agentic-ai-team](https://camunda.slack.com/archives/C08BHU6JQJG) — https://camunda.slack.com/archives/C08BHU6JQJG/p1760354073732349.
•    Affected version: Behavior change visible since validation landed in Camunda 8.8.0-alpha8 (constraint existed earlier but became enforced here).
•    Component: Configurable size limit lives in the connector configuration; currently blocks larger inputs at the connector boundary.
•    Notes: Embedding model token limits (typically 512–8192) remain a natural ceiling; consider query chunking and improved error handling around these limits.

maff:
## Description

Removes 200 char size limit for retrieval query. Additionally bumps the min engine version to `8.8` as the previously set version `8.2` is misleading.

On hitting a provide size limit, it will return with the provider message, which I think is good enough for now. For the future we can look into more advanced patterns, such as splitting the retrieval query.

image

## Related issues

closes #5553 

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/element-templates/hybrid/embeddings-vector-database-outbound-connector-hybrid.json
+++ b/connectors/embeddings-vector-database/element-templates/hybrid/embeddings-vector-database-outbound-connector-hybrid.json
@@ -17,7 +17,7 @@
     "value" : "bpmn:ServiceTask"
   },
   "engines" : {
-    "camunda" : "^8.2"
+    "camunda" : "^8.8"
   },
   "groups" : [ {
     "id" : "taskDefinitionType",
--- a/connectors/embeddings-vector-database/element-templates/embeddings-vector-database-outbound-connector.json
+++ b/connectors/embeddings-vector-database/element-templates/embeddings-vector-database-outbound-connector.json
@@ -17,7 +17,7 @@
     "value" : "bpmn:ServiceTask"
   },
   "engines" : {
-    "camunda" : "^8.2"
+    "camunda" : "^8.8"
   },
   "groups" : [ {
     "id" : "operation",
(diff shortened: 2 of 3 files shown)


connectors/embeddings-vector-database/element-templates/hybrid/embeddings-vector-database-outbound-connector-hybrid.json
```diff
@@ -17,7 +17,7 @@
     "value" : "bpmn:ServiceTask"
   },
   "engines" : {
-    "camunda" : "^8.2"
+    "camunda" : "^8.8"
```
mdii: 👍

Issue #5568, opened 2025-10-15
[Polling connector] Feel expression for field in polling connector are resolved to null


mathias-vandaele:
__Describe the Bug__

When using a feel expression in the polling connector, they are resolved to null.

<img width="1131" height="470" alt="Image" src="https://github.com/user-attachments/assets/7afd8e21-4c80-4e5f-8757-dfe0356f5d41" />

__Steps to Reproduce__

1. Create a simple process with an intermediate polling connector
2. Use a url containing a process variable `"http://localhost:8080" + myVar`
3. It can not be resolved

__Expected Behavior__

__Environment__

- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]
- Camunda version: [e.g. 8.0.4 (self-managed)]

PR #5569, opened 2025-10-15
fix(polling-connector): split the model request to differentiate between Activation and Runtime

mathias-vandaele:
## Description

This PR splits the model request into two parts, Activation and Runtime, to allow FEEL expressions to be properly evaluated during runtime.

Previously, the model was bound only at activation time, which prevented FEEL expressions from being resolved correctly.

With this change, the model is now divided into two distinct sections:

Activation properties: evaluated at activation time; cannot resolve FEEL expressions

Runtime properties: evaluated at runtime; can resolve FEEL expressions using the current process variables as context

### Notes

I added some integration test, there must be a timing issue I don't get as the second test fails in ~10% of runs, I would like to discuss this with someone later to see what it could be

## Related issues

closes https://github.com/camunda/connectors/issues/5568

## Checklist

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/PollingConnectorTests.java
+++ b/connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/PollingConnectorTests.java
@@ -20,6 +20,7 @@ import static com.github.tomakehurst.wiremock.client.WireMock.*;
 import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig;
 import static io.camunda.connector.e2e.BpmnFile.Replace.replace;
 import static io.camunda.connector.e2e.BpmnFile.replace;
+import static io.camunda.process.test.api.CamundaAssert.assertThat;
 import static org.mockito.ArgumentMatchers.*;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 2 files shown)


connectors/http/polling/src/test/java/io/camunda/connector/http/polling/task/HttpRequestTaskTest.java
```diff
@@ -41,29 +40,57 @@ void init() {
   }
 
   @Test
-  public void shouldExecuteAndCorrelateHttpRequestOnRun() throws Exception {
+  public void shouldExecuteAndCorrelateHttpRequestOnRun() {
     // Given
+    var pollingRuntimeProperties = new PollingRuntimeProperties();
+    pollingRuntimeProperties.setUrl("http://dummyUrl.com");
```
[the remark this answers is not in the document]
sbuettner: > In order to properly test this feature, we need to use an integration test, since the failure occurs during the binding process, which is mocked in unit tests.

Yeah, but then lets include it otherwise we cant make sure its working. Meanwhile @Szik can start testing this PR.

connectors/http/polling/src/main/java/io/camunda/connector/http/polling/model/PollingActivationProperties.java
```diff
@@ -0,0 +1,55 @@
...
+package io.camunda.connector.http.polling.model;
+
+import com.fasterxml.jackson.annotation.JsonSetter;
+import com.fasterxml.jackson.annotation.Nulls;
+import io.camunda.connector.api.annotation.FEEL;
+import io.camunda.connector.generator.dsl.Property;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import java.time.Duration;
+
+public class PollingActivationProperties {
```
sbuettner: `PollingRequest` could be a better fit as activation is usually used during correlation.

sbuettner: What about this:
`PollingRequest` -> `HttpPollingConnectorProperties`
`PollingActivationProperties` -> `SchedulingProperties`
`PollingRuntimeProperties` -> `HttpPollingRequest`

johnBgood (approved):

sbuettner (approved):

PR #5577, opened 2025-10-17
fix(agentic-ai): Add support for null JSON schema type to define nullable properties

### LINKED ISSUE #5570: Agentic AI: MCP Connector Fails with "Unsupported JSON schema type 'JsonNullSchema'" for Nullable Parameters
__Describe the Bug__

The MCP Outbound Connector fails to parse the JSON schema of tools when an optional parameter is defined with a null type. When the connector fetches the tool list from an MCP server (like  [jupyter-mcp-server](https://github.com/datalayer/jupyter-mcp-server)), it encounters an error if a tool's inputSchema contains a property with `{"type": "null"}` as one of the `anyOf` options. This results in a `Failed to convert JSON schema for tool specification 'use_notebook': Unsupported JSON schema type 'JsonNullSchema'` error, preventing the tools from being used.

The issue stems from properties defined as optional with a default None value in Python, which translates to the following valid, but unsupported, JSON Schema construct:

```json
"notebook_path": {
  "anyOf": [
    {
      "type": "string"
    },
    {
      "type": "null"
    }
  ],
  "default": null,
  "title": "Notebook Path"
}
```

Relevant code snippet from the MCP tool used: https://github.com/datalayer/jupyter-mcp-server/blob/main/jupyter_mcp_server/server.py#L280

Changing the `None` to an empty string in the MPC server resolves the issue, but I think this is a parsing error on the connector side, not a bug in the server. 

__Steps to Reproduce__

1. Set up an instance of the jupyter MCP server. I used the [mpc-proxy](https://www.npmjs.com/package/mcp-proxy) nodejs package, since the MCP server only supports Streamable-HTTP/STDIO connections (not SSE).
2. Try to use the MCP in an adhoc process. Configure it to connect to the proxy.
3. Run the process
4. The operation fails with the following error:
> `Failed to convert JSON schema for tool specification 'use_notebook': Unsupported JSON schema type 'JsonNullSchema'`

__Expected Behavior__

The MCP Outbound Connector should correctly parse JSON schemas that include the null type, as this is a standard way to define optional/nullable fields. The connector should successfully fetch and list the tools from the Jupyter MCP server without throwing a `JsonNullSchema` error, allowing developers to configure and use the tools in their processes.

__Environment__

- OS: Arch Linux (via Docker compose)
- Library version: e89f9ea
- Camunda version:  8.8.0 (self-managed)

maff:
## Description

Adds support for `JsonNullSchema` while serializing/deserializing JSON schema from the L4J data structure.

## Related issues

closes #5570

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/jsonschema/JsonSchemaElementModuleTest.java
+++ b/connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/jsonschema/JsonSchemaElementModuleTest.java
@@ -16,6 +16,7 @@ import dev.langchain4j.model.chat.request.json.JsonArraySchema;
 import dev.langchain4j.model.chat.request.json.JsonBooleanSchema;
 import dev.langchain4j.model.chat.request.json.JsonEnumSchema;
 import dev.langchain4j.model.chat.request.json.JsonIntegerSchema;
+import dev.langchain4j.model.chat.request.json.JsonNullSchema;
 import dev.langchain4j.model.chat.request.json.JsonNumberSchema;
 import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
 import dev.langchain4j.model.chat.request.json.JsonReferenceSchema;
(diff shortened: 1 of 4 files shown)


connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/jsonschema/JsonSchemaElementModuleTest.java
```diff
@@ -371,6 +372,54 @@ void shouldHandleSchemaIncludingArray() throws Exception {
...
+          "properties": {
+            "myProperty": {
+              "anyOf": [
+                {
+                  "type": "string"
+                },
+                {
+                  "type": "null"
+                }
+              ],
```
mdii: Is this equivalent to `"type": ["string", "null"]`?

PR #5625, opened 2025-10-28
feat(agentic-ai): Add support polling flag to A2A element template

### LINKED ISSUE #5622: Agentic AI: Add a Support Polling flag for the A2A Outbound Connector
__What should we do?__

When we send an A2A message we can ask the remote A2A server to block until the final result is ready. This should be OK when the server returns rather quickly. Otherwise the job worker will become busy for a long time and it might time out. In the case of long running tasks, we can ask the remote server to return immediately and then use the A2A polling connector to poll for the task completion.
For that we need to have a __Support Polling__ flag on the Outbound connector's element template.

__Why should we do it?__

So that the user decides whether polling is used not.

mdii:
## Description

Add a `Support Polling` boolean property to the A2A outbound connector configuration. This flag tells the A2A client if the send message call should be blocking or not.

## Related issues

closes #5622

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/A2aCommonSendMessageConfiguration.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/A2aCommonSendMessageConfiguration.java
@@ -28,7 +28,7 @@ public record A2aCommonSendMessageConfiguration(
             description =
                 "Whether to enable polling for the remote agent's response instead of waiting for it synchronously.",
             defaultValueType = TemplateProperty.DefaultValueType.Boolean,
-            defaultValue = "false",
+            defaultValue = "true",
             optional = true)
         Boolean supportPolling,
     @TemplateProperty(
--- a/connectors/agentic-ai/element-templates/agenticai-a2a-client-outbound-connector.json
+++ b/connectors/agentic-ai/element-templates/agenticai-a2a-client-outbound-connector.json
@@ -279,7 +279,7 @@
     "label" : "Support polling",
     "description" : "Whether to enable polling for the remote agent's response instead of waiting for it synchronously.",
     "optional" : true,
-    "value" : false,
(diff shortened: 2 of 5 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/model/A2aCommonSendMessageConfiguration.java
```diff
@@ -22,6 +22,15 @@ public record A2aCommonSendMessageConfiguration(
             defaultValueType = TemplateProperty.DefaultValueType.Number,
             defaultValue = "3")
         Integer historyLength,
+    @TemplateProperty(
+            group = "operation",
+            label = "Support polling",
+            description =
+                "Whether to enable polling for the remote agent's response instead of waiting for it synchronously.",
+            defaultValueType = TemplateProperty.DefaultValueType.Boolean,
+            defaultValue = "false",
```
maff: Reconsidering, I think I got it wrong when we discussed it [here](https://github.com/camunda/connectors/pull/5524#issuecomment-3435448492). To summarize:

- setting `polling = false` results in `blocking = true` (at least in the Java SDK in [Client.java](https://github.com/a2aproject/a2a-java/blob/main/client/base/src/main/java/io/a2a/client/Client.java#L159))
- `blocking = true` might result in the server rejecting the message as defined [here](https://a2a-protocol.org/latest/specification/#712-messagesendconfiguration-object):

>   /** If true, the client will wait for the task to complete. The server may reject this if the task is long-running. */
>  blocking?: boolean;

Assuming most agent operations would be long running and our job model not being well suited for blocking tasks, should we default to `blocking = false` (equals `polling = true`) here?

mdii: Yes it makes more sense to set polling=true by default. But Let's discuss it with the team!

PR #5630, opened 2025-10-29
feat: add RabbitMQ delete queue feature

DomingossDias:
Following @chillleader's guidelines, this PR adds a new type of operation to the Camunda Connector without changing its default operating method, Delete Queue, responsible for deleting RabbitMQ queues using the RabbitMQ native API.
image

### DIFF
--- a/.github/workflows/TEST_FEATURE_BRANCH.yml
+++ b/.github/workflows/TEST_FEATURE_BRANCH.yml
@@ -18,8 +18,6 @@ permissions:
 
 jobs:
   test-branch:
-    # skip pull_request trigger for internal branches (only use for forks)
-    if: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository }}
 
     runs-on: ubuntu-latest
     steps:


DomingossDias:
Thanks for the feedback, @chillleader! 🙏
I’ve recreated the branch from a fresh clone, synced my fork with the latest main branch, and reapplied only the relevant changes for the Delete Queue feature.
The diff is now clean and only includes the necessary files.
Please let me know if you’d like me to adjust anything else.

connectors/rabbitmq/element-templates/rabbitmq-outbound-connector.json
```diff
@@ -1,13 +1,13 @@
 {
   "$schema" : "https://unpkg.com/@camunda/zeebe-element-templates-json-schema/resources/schema.json",
-  "name" : "RabbitMQ Outbound Connector",
-  "id" : "io.camunda.connectors.RabbitMQ.v1",
-  "description" : "Send message to RabbitMQ",
+  "name" : "RabbitMQ Outbound Connector Manager",
```
chillleader: let's not rename it

chillleader:
Thanks @DomingossDias looks much better now. I will take a detailed look and provide feedback by end of week

DomingossDias:
Thank you very much, I'll be waiting!

connectors/rabbitmq/src/main/java/io/camunda/connector/rabbitmq/outbound/model/RabbitMqRequest.java
```diff
@@ -15,9 +16,15 @@
 import jakarta.validation.constraints.NotNull;
 
 public record RabbitMqRequest(
+    @TemplateProperty(
+            id = "operationType",
+            label = "Operation Type",
+            description = "Select the desired RabbitMQ operation.")
+        @NotNull
+        String operationType,
```
chillleader: The problem with this approach is that we loose strict typing and validation (that's why you had to remove the `NotNull`/`NotEmpty` constraints. This will be difficult to maintain.

It is also incorrect to reuse the `routingKey` property for queue name in your new operation.

The proper way to implement operations here would be to introduce a sealed interface with different operations as implementations, smth like this:

```
public sealed interface RabbitMqOperation permits SendMessage, DeleteQueue {}

public record SendMessage {
    @TemplateProperty(
            group = "message",
            feel = Property.FeelMode.required,
            optional = true,
            defaultValue = "={}",
            type = TemplateProperty.PropertyType.Text,
            description = "Properties for the message, routing headers, etc")
        Object properties,
    @NotNull
        @TemplateProperty(
            group = "message",
            label = "Message",
            type = TemplateProperty.PropertyType.Text,
            description = "Data to send to RabbitMQ")
        Object body) {}

public record DeleteQueue(
  @NotNull
  @TemplateProperty(
    group = "queue",
    label = "Queue name",
    description = "Queue to delete")
  String queueName) {}
```
  
For another example of this pattern, please see the [S3 connector.](https://github.com/camunda/connectors/blob/9b775d18f4cabb00bc0957ca0b0e9fee1502ab38/connectors/aws/aws-s3/src/main/java/io/camunda/connector/aws/s3/model/request/S3Request.java)

When you apply this pattern, the element template generator plugin will generate the operation structure for you without the need to explicitly add `operationType` to your model.

connectors/rabbitmq/src/test/java/io/camunda/connector/rabbitmq/integration/RabbitMqIntegrationTest.java
```diff
@@ -211,6 +213,46 @@ void publishMessageWithOutboundConnectorAndConsumeMessageWithInboundConnector()
     assertEquals("Hello “ ” \"World\"", body.get("value"));
   }
 
+  @Test
+  @Order(4)
+  void deleteQueueWithOutboundConnector() throws Exception {
```
chillleader: Thanks for adding a new test 👍

chillleader (changes requested):
The main issue I see now is how the operation model is implemented - we should not explicitly add `operationType` to the input data model because we have a tool (element template generator) that supports the operation pattern and creates this meta-property for us when we use the right pattern in the code. I provided an example below, let me know if you need more context or get stuck and need help 🙂 

Just FYI we also support a different pattern for generating connectors with multiple operations (see [CSV Connector](https://github.com/camunda/connectors/blob/ad69662511ffb107c2fe7d1027379822230c4159/connectors/csv/src/main/java/io/camunda/connector/csv/CsvConnector.java)) - but here, we should rather stick with the one I described, because otherwise we would have to refactor too many things in this connector.

I also left comments on some organizational stuff (versioning/id/naming) to match our conventions.

vringar:
Closing this after a month of inactivity.
Feel free to reopen when you find time to address the feedback.

PR #5682, opened 2025-11-04
fix: make sure `value` of `feel: required` templates starts with `=`

jarekdanielak:
### Context
In the core team, we want to ensure that with `element-templates-json-schema`, if a field is marked as `feel: required`, the value starts with `=`.

We currently don't throw an error if that happens, for backward compatibility, but we mark such use cases as `deprecated`:

image

### Changes

We identified a couple of OOTB templates that suffer from this, and this PR attempts to fix it.

Please assist me in ensuring your contribution standards.

I moved the current versions of AI Agent Task and AI Agent Sub-process to `versioned` directory, bumped their versions and fixed the thing.

For HubSpot Outbound Connector, I only changed the "current" file.

## Related issues

Related to https://github.com/camunda/camunda-modeler/issues/4967

## Checklist

### DIFF
--- a/connectors/agentic-ai/AI_AGENT.md
+++ b/connectors/agentic-ai/AI_AGENT.md
@@ -230,6 +230,6 @@ leading to the following result
 | Connector Info            |                                                                       |
 | ---                       | ---                                                                   |
 | Type                      | io.camunda.agenticai:aiagent:1                                                            |
-| Version                   | 4                                                         |
+| Version                   | 5                                                         |
 | Supported element types   |     |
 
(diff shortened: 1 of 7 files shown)


connectors/hubspot/element-templates/hubspot-outbound-connector.json
```diff
@@ -27,7 +27,7 @@
     ]
   },
   "id": "io.camunda.connectors.HubSpot.v1",
-  "version": 1,
+  "version": 2,
```
ztefanie: Can you please create a versioned subfolder and move the version: 1 template there, as we want to keep track of all versions in our repository.

chillleader:
Looks ok as a quick fix, but it might be worth adding this as a default behavior into out element template generator (if `feel: required`, add a `=` to the default value if it's not already there

ztefanie (approved):

PR #5690, opened 2025-11-06
fix(ObjectMapper): qualify ObjectMapper to prevent accidental overrides

### LINKED ISSUE #5675: Qualify the connectors object mapper so it's harder to overwrite

A customer recently broke their connectors runtime because they wanted to customize
their own object mapper bean, but didn't realize that providing a bean would overwrite the object mapper
used by connectors

Use a Qualifier [here](https://github.com/camunda/connectors/blob/a2b58982a49c88118e943e05c871db3c6c87d0c0/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java#L81) and [here](https://github.com/camunda/connectors/blob/a2b58982a49c88118e943e05c871db3c6c87d0c0/connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java#L162) so that the customer has to be more explicit about overwriting the bean

vringar:
## Description

By adding the qualifier, a customer specifying a default bean doesn't accidentally override our object mapper and connectors continues to work

## Related issues

closes #5675

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/search/ProcessInstanceClientConfiguration.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/search/ProcessInstanceClientConfiguration.java
@@ -17,6 +17,7 @@
 package io.camunda.connector.runtime.inbound.search;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import io.camunda.connector.runtime.annotation.ConnectorsObjectMapper;
 import io.camunda.connector.runtime.core.inbound.ProcessInstanceClient;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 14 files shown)


vringar:
@maff do you think a better strategy for validating that everything is wired correctly is to inject the "hostile" object mapper into every test and see what breaks?

connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java
```diff
@@ -38,7 +39,8 @@
 public class OutboundConnectorsAutoConfiguration {
 
   @Bean
-  @ConditionalOnMissingBean
+  @Qualifier("connectorObjectMapper")
```
maff: To make this a bit less verbose/error-prone, do you think we could introduce a custom `@ConnectorsObjectMapper` annotation using `@Qualifier`, so that the custom OM could be injected with something like:

```java
@ConnectorsObjectMapper
private ObjectMapper objectMapper;
```

maff:
> @maff do you think a better strategy for validating that everything is wired correctly is to inject the "hostile" object mapper into every test and see what breaks?

Actually I would hope that this would be catched by e2e tests. As the agentic AI ones are green, I added 2 suggestions on usability/test setup, but I think we could go forward with the change and adapt agentic AI if necessary in a follow-up.

PR #5692, opened 2025-11-06
fix(element-template-generator): reproducible parameter order for OpenAPI

tfaller:
## Description

Currently the output order of parameters in generated templates from OpenAPI schemas are not reproducible. Even tho the OpenAPI schema is unchanged, the generated template will have often differently ordered parameters for each generator run.

This PR makes sure the template will be the same, by just using a `LinkedHashSet`, instead of a `HashSet`.


johnBgood (commented):
Thank you for this contribution, much appreciated :) 

Could you add a test in `OperationUtilTest`? 

Could you also check `extractOperation` in `PostmanOperationUtil` while tackling this? I guess there's a simple Set as well.

Thank you!

PR #5693, opened 2025-11-07
feat: make the secret provider optionally process definition id aware

jonathanlukas:
## Description

This PR makes the secret provider for outbound secrets aware for the process definition id.

There are no changes to the actual behaviour as the default secret provider are not sensitive to this.

The reason is that the behaviour cannot be unified across inbound and outbound connectors.

Context: https://camunda.slack.com/archives/C02JLRNQQ05/p1762355301979309

## Related issues

closes #

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/secret/EnvironmentSecretProvider.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/secret/EnvironmentSecretProvider.java
@@ -29,11 +29,14 @@ public class EnvironmentSecretProvider implements SecretProvider {
   private final Environment environment;
   private final String prefix;
   private final boolean tenantAware;
+  private final boolean processDefinitionAware;
 
-  public EnvironmentSecretProvider(Environment environment, String prefix, boolean tenantAware) {
+  public EnvironmentSecretProvider(
+      Environment environment, String prefix, boolean tenantAware, boolean processDefinitionAware) {
     this.environment = environment;
     this.prefix = prefix;
     this.tenantAware = tenantAware;
+    this.processDefinitionAware = processDefinitionAware;
   }
 
   @PostConstruct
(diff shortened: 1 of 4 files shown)


jonathanlukas:
@vringar @chillleader thank you for the feedback.

In adjusted the PR to reflect these assumptions. While I also got some good insights in slack, I am confident that the feature will work as expected.

A question going forward: Should the "EnvironmentSecretProvider" have a "processDefinitionIdAware" flag and implement this awareness, just like with the tenants?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/details/InboundConnectorDetailsUtil.java
```diff
@@ -42,16 +42,30 @@ static InboundConnectorDetails create(
...
   }
 
+  private static String extractProcessDefinitionId(List<InboundConnectorElement> elements) {
+    if (elements.stream()
+            .map(InboundConnectorElement::element)
+            .map(ProcessElementWithRuntimeData::bpmnProcessId)
+            .distinct()
+            .count()
+        > 1) {
+      throw new IllegalArgumentException("All elements in a group must have the same tenant ID");
```
johnBgood: the error message is not correct, it should mention processId not tenantId

jonathanlukas: Good catch!

jonathanlukas:
>A question going forward: Should the "EnvironmentSecretProvider" have a "processDefinitionIdAware" flag and implement this awareness, just like with the tenants?

@johnBgood any opinion on this?

johnBgood:
> @johnBgood any opinion on this?

Sorry I missed this. 
Can't the`EnvironmentSecretProvider` act depending on what's in the `SecretContext`? Or do we see an issue if, for example, we provide a `tenantId` and it shouldn't be taken into account by the provider? If not, we could remove the old `tenantIdAware` flag, and use the `SecretContext` directly.

+ I think you update the PR description as we don't have any differences between inbound/outbound now

jonathanlukas:
@sbuettner thank you for flagging me here. I created tests to verify the secret provider is working as expected.

Regarding the tests on the InboundConnectorDetails, I am not 100% sure how and where I would test this as the only thing to assert here would be the data mapping.

johnBgood (approved):

PR #5721, opened 2025-11-11
feat(agentic-ai): Support agent card in the A2A polling connector

### LINKED ISSUE #5691: Agentic AI: Support AgentCard type correlation in the polling connector
__What should we do?__

Currently the A2A polling connector only supports A2AMessage and A2ATask as input. We need to support all the A2A outbound connector's output types in the A2A polling connector so that the polling connector can be used in the AHSP as part of an A2A tool.

__Why should we do it?__

Otherwise the user always need to add an exclusive gateway before the polling connector.

mdii:
## Description

Add support for correlating `A2aAgentCardResult` in the polling connector so that this connector can be used together with the outbound connector as a tool for the AI Agent without the need to add a gateway in between the A2A outbound connector and the A2A polling connector.
The JSON type annotations are moved on top of `A2aResult`, with `A2aAgentCardResult` being one of the sub-types. To be able to correctly correlate when the client response is an agent card, an `id` field is added to `A2aAgentCardResult`.

## Related issues

closes #5691 

## Checklist


maff (approved):
LGTM (small comment on naming) 👍

PR #5771, opened 2025-11-20
refactor(http-webhook): extract HMAC verification logic

mdii:
## Description

The A2A webhook connector is very similar to the HTTP webhook connector. The major difference is that we need to transform the payload to our A2A model. That's why the agentic-ai module has a dependency on the `connector-webhook`. In order to avoid duplicating the logic for HMAC verification we need to extract this logic in a class so that it can be used from the A2A code. 

## Checklist


maff (dismissed):
LGTM 👍

Maybe somebody from @camunda/connectors could have a look as well as it affects the core webhook logic.

PR #5773, opened 2025-11-20
fix(vector-db): Mark v1 element template as deprecated and set engine version compatibility to 8.8

### LINKED ISSUE #5772: Wrong Camunda engine version range on Vector DB v1 template
__Describe the Bug__

The Vector DB template in version 1 introduced with Camunda 8.8.0-alpha4 defines a compatible version range of `^8.2` which is wrong. Furthermore, the v1 version should be deprecated as only version 2 is compatible with the connector implementation in Camunda 8.8 and onwards.

Related discussion: https://camunda.slack.com/archives/C08BHU6JQJG/p1763459800391919

maff:
## Description

Sets the engine compatibility version to `^8.8` and marks the v1 version as deprecated.

## Related issues

closes #5772

## Checklist

### DIFF
--- a/connectors/embeddings-vector-database/element-templates/versioned/embeddings-vector-database-outbound-connector-1.json
+++ b/connectors/embeddings-vector-database/element-templates/versioned/embeddings-vector-database-outbound-connector-1.json
@@ -8,7 +8,9 @@
   },
   "documentationRef" : "https://docs.camunda.io/docs/8.8/components/connectors/out-of-the-box-connectors/embeddings-vector-db/",
   "version" : 1,
-  "deprecated": true,
+  "deprecated": {
+    "message": "Replace with version >= 2 of the Embeddings Vector DB Outbound Connector template."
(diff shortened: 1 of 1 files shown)


connectors/embeddings-vector-database/element-templates/versioned/embeddings-vector-database-outbound-connector-1.json
```diff
@@ -17,7 +20,7 @@
     "value" : "bpmn:ServiceTask"
   },
   "engines" : {
-    "camunda" : "^8.2"
+    "camunda" : "^8.8"
```
mdii: Is it OK to modify the engine version?

maff: I think so, `8.2` was definitely wrong, so I would not see a good use case to keep it.

PR #5839, opened 2025-11-27
fix(agentic-ai): Partially support process instance migration for running AI Agents

maff:
## Description

This PR introduces partial support for process instance migration in running AI Agents by implementing migration detection and tool definition reconciliation mechanisms.

### Key Changes

**Migration Detection**
- Added `AgentMetadata` record to track `processDefinitionKey` and `processInstanceKey` for each agent execution
- The agent now compares the stored process definition key with the current execution context to detect migrations
- When a migration is detected, tool definitions are automatically re-loaded and reconciled

**Tool Definition Reconciliation**
- New `AgentToolsResolver` interface and `AgentToolsResolverImpl` implementation to handle tool definition updates after migration
- Support for updating existing tool definitions (e.g., description changes, schema updates)
- Support for adding new tool definitions to running agents

**Validation & Error Handling**
- New error codes: `MIGRATION_MISSING_TOOLS` and `MIGRATION_GATEWAY_TOOL_DEFINITIONS_CHANGED`
- Prevents removing tool definitions from running agents (would cause race conditions/stuck executions)
- Prevents adding/removing gateway tool definitions (MCP, A2A) as these require re-issuing tool discovery calls
- Clear error messages with actionable feedback for unsupported migration scenarios

**Gateway Tool Handler Registry Extensions**
- Added `isGatewayManaged(String toolName)` method to check if a tool is managed by a gateway handler
- Added `handlerForToolDefinition(String toolName)` to find the responsible gateway handler
- Added `resolveUpdatedGatewayToolDefinitions()` to detect changes in gateway tool definitions
- New `GatewayToolDefinitionUpdates` record to track added/removed gateway tool definitions

### Supported Migration Scenarios
- ✅ Changing AI Agent parameters (system prompt) on AI Agent Task
- ✅ Adding new tool definitions
- ✅ Updating existing tool definitions (description, schema changes)

### Unsupported Migration Scenarios (with clear error feedback)
- ❌ Removing tool definitions
- ❌ Renaming tool definitions (detected as removal + addition)
- ❌ Adding/removing gateway tool definitions (MCP, A2A)

## Related issues

part of #5809

## Checklist

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/BaseAgenticAiTest.java
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/BaseAgenticAiTest.java
@@ -62,6 +62,10 @@ public abstract class BaseAgenticAiTest {
   }
 
   protected ZeebeTest deployModel(BpmnModelInstance model) {
+    return deployModel(model, 1);
+  }
+
+  protected ZeebeTest deployModel(BpmnModelInstance model, int expectedVersionCount) {
     final var process =
         model.getDefinitions().getRootElements().stream()
             .findFirst()
@@ -81,7 +85,7 @@ public abstract class BaseAgenticAiTest {
                       .send()
                       .join();
 
-              assertThat(processDefinitions.items()).hasSize(1);
+              assertThat(processDefinitions.items()).hasSize(expectedVersionCount);
             });
 
     return zeebeTest;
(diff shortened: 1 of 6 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
```diff
@@ -21,49 +20,64 @@
...
-    final var toolElements = executionContext.toolElements();
 
-    // skip tool discovery if no tool elements are defined
-    if (CollectionUtils.isEmpty(toolElements)) {
-      return new AgentContextInitializationResult(
-          agentContext.withState(AgentState.READY), toolCallResults);
+    final var agentMetadata = agentContext.metadata();
+    final var executionMetadata = executionContext.jobContext().metadata();
+    if (agentMetadata == null
+        || !executionMetadata.processDefinitionKey().equals(agentMetadata.processDefinitionKey())) {
```
maff: This is basically the check if a migration has been applied. This will trigger for all running process instances (also without migration) as the data is missing, but as the resolving logic is cheap (using the variable on the sub-process and relying on cached data for the task) this should not be an issue.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/discovery/McpClientGatewayToolHandler.java
```diff
@@ -50,13 +52,15 @@ public String type() {
     return GATEWAY_TYPE;
   }
 
+  @Override
+  public boolean isGatewayManaged(String toolName) {
+    return McpToolCallIdentifier.isMcpToolCallIdentifier(toolName);
+  }
```
maff: Maybe we would have the possibility to not rely on the naming convention for these lookups by adding some metadata to the tool definitions upon resolving (e.g. which gateway it is managing it). I'll look into that in a follow-up.

PR #5940, opened 2025-12-17
chore(soap-connector): refactor element template, signature parts, interceptors, readme, and removed logging

aletutto:
## Description
This PR will improve several parts of the SOAP connector which were found while testing.

### Element template
- Remove binary security token and canonicalization algorithm as they are not used
- Update description of encryption parts
- Rename certificate type
- Add null validation on keystore properties

### WSS Signature: signature parts
The SOAP connector has a property to specify signature parts when using the authentication method "WSS signature". This property is marked as optional in the element template, but if they are not specified, the request cannot be signed and a NPE will be thrown. [The library we are using adds by default the signature to the SOAP body if no signature parts are specified](https://ws.apache.org/wss4j/config.html#:~:text=The%20SOAP%20body%20is%20signed%20by%20default.). Therefore a null check was inserted on the signature parts.

### Header and authentication interceptor
As there is a need to sign also elements present in the header, the header and authentication interceptor were swapped. Before, the headers were not present in the XML because they were added just after the security interceptor was creating the WSS signature.

### Logging 
The traces added with #5868 are not needed and removed, as there is already a LoggingInterceptor in place. 

### README
The README was aligned with parts of the [SOAP connector documentation](https://docs.camunda.io/docs/components/connectors/protocol/soap/), and an example and general information for using the WSS signature was added. Besides that, some general information on how the XML template and the XML compatible JSON can be used was added. The goal is to add these enhancements as well to our public documentation.

### E2E Tests
Two new e2e tests were added in order to verify that after swapping the header and authentication interceptor the headers are still added as expected, and a second one to verify that adding a WSS signature will lead to the expected request.

## Related issues

There is no issue for that, all enhancements are the result of testing the connector based on requirements of a customer.

## Checklist

### DIFF
--- a/connectors/soap/src/main/java/io/camunda/connectors/soap/client/SpringSoapClient.java
+++ b/connectors/soap/src/main/java/io/camunda/connectors/soap/client/SpringSoapClient.java
@@ -146,8 +146,9 @@ public class SpringSoapClient implements SoapClient {
 
   private ClientInterceptor[] buildInterceptors(
       String soapHeader, Authentication authentication, Map<String, String> namespaces) {
-    List<ClientInterceptor> interceptorList = new ArrayList<>(handleAuthentication(authentication));
+    List<ClientInterceptor> interceptorList = new ArrayList<>();
     handleSoapHeader(soapHeader, namespaces).ifPresent(interceptorList::add);
+    interceptorList.addAll(handleAuthentication(authentication));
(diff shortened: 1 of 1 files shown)


jonathanlukas:
@sbuettner I am fine with this.

sbuettner:
@jonathanlukas Can you give this another review?

PR #6002, opened 2026-01-05
fix(GHA): install all required tools

vringar:
## Description

On self-hosted runners, maven isn't present by default. We need to add it.

## Checklist


vringar:
Not backporting this as I will add the commit manually to all the backport PRs for the initial change

PR #6009, opened 2026-01-06
test(connectors): Add test for unknown enum values using default value

azan-baloch:
## Description

Added a unit test to verify that unknown enum values in JSON are deserialized to the default enum value when using the ObjectMapper from ConnectorsObjectMapperSupplier. 

It covers the case where the ObjectMapper has:

    .enable(DeserializationFeature.READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE)

For example, this enum:

    private enum TestStatus {
      ACTIVE,
      INACTIVE,
      @JsonEnumDefaultValue
      UNKNOWN
    }

and a record:

    private record TestRecordWithEnum(TestStatus status) {}

if the JSON contains an unknown value like:

    {"status":"PAUSED"}

the ObjectMapper will map it to TestStatus.UNKNOWN.

This test make sure that this behavior does not break in future changes and does not affect production runtime behavior and also prevents regressions in the `ObjectMapper` configuration

## Related issues

N/A

## Checklist

### DIFF
--- a/connector-commons/connector-object-mapper/src/test/java/io/camunda/connector/jackson/ConnectorsObjectMapperSupplierTest.java
+++ b/connector-commons/connector-object-mapper/src/test/java/io/camunda/connector/jackson/ConnectorsObjectMapperSupplierTest.java
@@ -18,6 +18,7 @@ package io.camunda.connector.jackson;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.fasterxml.jackson.annotation.JsonEnumDefaultValue;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.exc.MismatchedInputException;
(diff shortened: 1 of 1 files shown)


vringar (approved):

azan-baloch:
Hey @vringar 

My PR is not merged yet because some workflows are failing, my PR is actually from forked repo, so vault secrets etc aren't available.

The change itself is only a test and will not cause any issue to production code.

Let me know if you want me to update the branch or do anything else from my side.

sbuettner:
@azan-baloch Can you provide some more details why you think this tests adds value to the project? It seems like its testing Jackson. Thanks!

azan-baloch:
Hi @sbuettner , thanks for the question

This test is not to verify Jackson's internal logic, but rather to make sure that our configuration in `ConnectorsObjectMapperSupplier` doesn't change by accident.

Actually, we have enabled `READ_UNKNOWN_ENUM_VALUES_USING_DEFAULT_VALUE` and if a developer changes this class in future and accidentally removes that configuration line, it may cause some silent failures or exceptions in OOTB connectors because some OOTB connectors depends on this behaviour when working with external apis.

So this test is **regression** guard for our global configuration,

Issue #3227, opened 2024-09-04
Intermediate Inbound Connectors need to be activated for all running process instances


johnBgood:
__Describe the Bug__

Currently, Inbound Connectors are activated only for the **[latest](https://github.com/camunda/connectors/blob/8f6f6da56488f25caf6252cdef4f314879d5b4ab/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionSearch.java#L51)** version of each process definition.

This is incorrect, see [this](https://camunda.slack.com/archives/C07KF9VFH0D/p1725455304263799) discussion.

The customer had a process running in different versions:
- V1, this instance was blocked on a Polling Connector that is polling example.com for instance
- The user changed this connector's configuration, say the polled URL is now: example-other.com 
- V3, working instances

As we currently fetch only the **latest** **process** **definitions**, we will start a Polling connector that polls example-other.com as this is the configuration in V3.

__Steps to Reproduce__

1. Create a new BPMN diagram, using an Intermediate Polling connector
2. Configure the connector's activation condition so that it blocks there (body.status = "Awesome!NOTCOMPLETED", if your endpoint returns a body with {status: "Awesome!"})
3. Deploy (V1), and start the process
4. Make some changes (change the polled URL) and deploy+start (V2)
5. Change the response body (use beeceptor) to "Awesome!NOTCOMPLETED"
6. The V1 instance doesn't complete because the connector is not activated anymore for this configuration.

__Expected Behavior__

Connectors are activated if they're part of a running process instance.

See https://jira.camunda.com/browse/SUPPORT-23440
See https://jira.camunda.com/browse/SUPPORT-25398

sbuettner:
Related: https://github.com/camunda/operate/issues/6415

mathias-vandaele:
A customer has reported this incident 
https://jira.camunda.com/browse/SUPPORT-25398?focusedCommentId=392187&atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258Y29tbWVudA%3D%3D

johnBgood:
assigning @chillleader as he's now the epic's DRI

PR #6056, opened 2026-01-13
feat: fetch active process definitions

chillleader:
## Description

This PR adds a mechanism to keep inbound connectors running as long as there are active process instances waiting on a connector message subscription.

Main idea: Add a parallel import flow that fetches message subscriptions, then activate inbound connectors in process versions that define these subscriptions.

I also refactored the internal state of the runtime accordingly, broke it down into two components (used to be `ProcessStateStore`, now broken down into `ProcessStateContainer` and `ProcessStateManager`).

Auxiliary changes:
- Removed a useless metric (counter of process definitions checked, I don't see how it's useful anymore)
- Changed how setting `connectors.inbound.polling.enabled=false` affects the runtime. Now it only disables importers (i.e. polling) and other inbound connector beans are still present - the result is the same but easier to test in some cases)

## Related issues

closes [#3227](https://github.com/camunda/connectors/issues/3227)

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspectorUtilTests.java
+++ b/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspectorUtilTests.java
@@ -121,7 +121,7 @@ public class ProcessDefinitionInspectorUtilTests {
   private List<InboundConnectorElement> fromModel(String fileName, String processId) {
     try {
       var searchQueryClientMock = mock(SearchQueryClient.class);
-      var inspector = new ProcessDefinitionInspector(searchQueryClientMock);
+      var inspector = new ProcessDefinitionInspector(searchQueryClientMock, 100);
       var modelFile = ResourceUtils.getFile("classpath:bpmn/" + fileName);
       var model = Bpmn.readModelFromStream(new FileInputStream(modelFile));
       var processDefinitionID = new ProcessDefinitionRef(processId, "tenant1");


connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/InboundConnectorsAutoConfiguration.java
```diff
@@ -21,17 +21,11 @@
...
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Import;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 @AutoConfiguration
 @AutoConfigureAfter({CamundaAutoConfiguration.class, OutboundConnectorsAutoConfiguration.class})
-@ConditionalOnProperty(
-    prefix = "camunda.connector.polling",
-    name = "enabled",
-    havingValue = "true",
```
johnBgood: do we really to remove this? We are then creating all the beans in InboundConnectorRuntimeConfiguration even if polling is disabled, or did I miss something?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ImportSchedulers.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private boolean ready = true;
+
+  public ImportSchedulers(ProcessStateManager stateStore, Importers importers) {
+    this.stateStore = stateStore;
+    this.importers = importers;
+  }
+
+  @Scheduled(fixedDelayString = "${camunda.connector.polling.interval:5000}")
+  public synchronized void scheduleLatestVersionImport() {
```
johnBgood: Not sure the synchronized is necessary, the default TaskScheduler is single threaded. We can keep it, just wanted to highlight it

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainerImpl.java
```diff
@@ -0,0 +1,146 @@
...
+   */
+  private final Map<ProcessDefinitionId, Map<Long, MutableProcessVersionState>> processStates =
+      new HashMap<>();
+
+  @Override
+  public StateUpdateResult compareAndUpdate(ImportResult importResult) {
+    final Set<ProcessDefinitionIdAndKey> toActivate = new HashSet<>();
+    final Set<ProcessDefinitionIdAndKey> toDeactivate = new HashSet<>();
+
+    for (var importEntry : importResult.processDefinitionKeysByProcessId().entrySet()) {
```
johnBgood: 1. let's say we previously had PDK1 for processDefinition1.
2. processDeifnition1 is deleted, so it's not in the import result
3. we will never deactivate the corresponding connectors

or did I miss something?
I guess after this loop, we should find all the processDefIds in state that are not in the import, and deactivate them

Issue #5841, opened 2025-11-27
Agentic AI: Add prompt + resource fetching suport to MCP client connectors


maff:
__Is your feature request related to a problem? Please describe.__

The MCP client connectors are currently only capable of listing and calling tools. To enable MCP interactions outside of an AI Agent scope, the clients should also support MCP operations related to `prompts` and `resources`:

[**Prompts**](https://modelcontextprotocol.io/specification/2025-11-25/server/prompts):
- List prompts
- Get a single prompt

[**Resources:**](https://modelcontextprotocol.io/specification/2025-11-25/server/resources)
- List resources
- List resource templates
- Get a single resource

__Describe the solution you'd like__

The new operations should be supported in both standalone mode (providing an operation dropdown with operation-specific input fields) and tool-calling mode (accepting a raw tool call request). 

**Challenges:**

Both prompts and resources can potentially return binary data, such as images or blobs. To avoid returning these as process variables the client should transform them into Camunda Documents first (see REST connector) and return a content part containing the Document reference instead of the binary data.

__Describe alternatives you've considered__

Not supporting additional MCP client features potentially blocks usage outside of agentic AI.

__Additional context__

PR #6069, opened 2026-01-15
feat(agentic-ai): Add read resource capability to MCP connector standalone mode

### LINKED ISSUE #5972: Implement in Standalone Mode: Get a single resource
__What should we do?__

* Implement Get Resource method call to our MCP outbound connectors (gateway and remote) in Standalone Mode

__Why should we do it?__

* To increase capabilities of our MCP outbound connectors, we require accessing an arbitrary known resource from a configured MCP server

nikonovd:
## Description

* Add read resource capability to MCP connector standalone mode
* Convert blob resource data to Camunda documents with default system TTL

## Related issues

relates to #5841
closes #5972

## Checklist


nikonovd:
> Some minor comments, otherwise looks great :)
> 
> I think in the `McpMethod` enum we can now remove the `supportedMethods()` methods and rely on all enum values, right?

@maff  Technically yes, but i'd still prefer now to have at least the methodNames in the exception message so i still need some transformation.

PR #6093, opened 2026-01-20
feat(http): improve HTTP status incident message

### LINKED ISSUE #5820: REST connector: Improve HTTP status incident message
__What should we do?__

When an HTTP call returns a bad status, the REST connector creates an incident with a vague message that only contains the HTTP status name and mostly doesn't shed light onto the root cause.

Image

Instead, we should provide a more helpful error message, for example:

```
HTTP request failed with status code 422 (Unprocessable Entity). Response body: \"{"key": "value"\"}
```

The body is helpful to include in the incident message. It can be truncated if too long.

__Why should we do it?__

This will provide better UX as it is currently not obvious where to find additional error details (you need to look into the variables)

azan-baloch:
## Description

This PR improves HTTP status incident messages by including the response body in the error message, as requested in #5820

camunda-err-msg-fix

Now it easier to debug when a connector fails. Before, if an API gave an error, you only saw "Unauthorized" but didn't know *why* and we have too look error variables but now I've updated the code so it also shows the response body from the API. Now if the API says "Invalid API Key" or "Expired Token," the user can see that directly in the error message.

## Related issues

closes #5820 

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
+++ b/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
@@ -1148,7 +1148,7 @@ class SpringConnectorJobHandlerTest {
               .withErrorExpressionHeader(errorExpression)
               .executeAndCaptureResult(jobHandler, false, false);
       // then
-      assertThat(result.getErrorMessage()).isEqualTo("MyError");
+      assertThat(result.getErrorMessage()).startsWith("MyError");
       assertThat(result.getVariables()).containsEntry("myVar", "myVal");
       assertThat(result.getVariables()).containsEntry("error", "MyError");
     }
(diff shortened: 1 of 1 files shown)


azan-baloch:
@johnBgood Thanks for the follow-up and discussing it with team, Your approach makes sense and is straightforward, I was doing it in hard way, This is how i am implementing it now.

```
var baseMessage = result.exception().getMessage();
var errorMessage =
    truncateErrorMessage(
        baseMessage
            + (result.responseValue() != null
                ? " | Error variables: " + result.responseValue()
                : ""));
```

I am currently changing it but i have some questions. As we are changing the error msg format so existing tests will fail because they are using `isEqualTo` for assertion and we are appending `Error variables` suffix.

For example, in `shouldPrioritizeSecretNotFoundException` Test:
**Expected**: "Secret with name 'FOO2' is not available"
**Actual**: "Secret with name 'FOO2' is not available | Error variables: {error={type=..., message=...}}"

Now tell me should i use `.startsWth` or should I hardcode the full expected string in each test? 

Also, I noticed that sometimes the **Error variables** just repeat the same error message we already have. Should I add a quick check to skip the suffix if it's just a copy of the exception, or should we just keep it simple and show everything anyway?

johnBgood (approved):
Awesome, thank you :)

PR #6308, opened 2026-02-09
chore: actuator endpoint for clearing the cache

### LINKED ISSUE #6267: Actuator endpoint to reset connector runtime state
__What should we do?__

The connector runtime maintains a cache of inbound connector details found in process definitions. The cache keys are process definition keys.

In CPT tests, the state of the engine is reset between every test, so new deployments in subsequent tests reuse the same process definition keys.

We already expose `clearCache()` in `ProcessDefinitionInspector`; the following changes are needed:

1. It would be great to make it available externally via an actuator endpoint.

2. Additionally, we should also purge the process instance / executable state with this action, otherwise there will be a delay of a few seconds until the connector runtime picks up the new empty state of the engine. This will allow the tests to run quicker.

3. Lastly, when this is done, we should revisit existing e2e tests that use CPT because we had to work around this limitation by [manually calling](https://github.com/camunda/connectors/blob/d89ab06e03b1ce368aec69018a25e92cb55a68f9/connectors-e2e-test/connectors-e2e-test-base/src/main/java/io/camunda/connector/e2e/inbound/InboundConnectorTestConfiguration.java#L56) `clearCache()` and adding a wait in every test's setup method. We can at least remove the wait when 2. is implemented, or maybe we find a better way to handle this.

__Why should we do it?__

CPT tests are unstable without this.

chillleader:
## Description

This PR will add an actuator endpoint to reset the process definition cache of the connector runtime.

The endpoint is conditional on property and disabled by default. This is intended to be used in CPT tests (see linked issue for more context).

Update after iteration 2:
- Replaced the previous ad-hoc cache impl with Spring's `CacheManager`, it allows to expose an actuator endpoint automatically

## Related issues

closes #6267 

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/pom.xml
+++ b/connector-runtime/connector-runtime-spring/pom.xml
@@ -238,7 +238,6 @@
     <dependency>
       <groupId>org.springframework</groupId>
       <artifactId>spring-context-support</artifactId>
-      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>com.github.ben-manes.caffeine</groupId>


connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionCacheEndpoint.java
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.inbound.state;
+
+import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
+
+@Endpoint(id = "processDefinitionCache")
```
ztefanie: (nit) other actuator endpoints are all lower case for consistency it might make sense to also change it here to all lower case.

ztefanie (approved):
Look good 🚀 

Just wondering: Have you thought about using the [existing acutator evict cache endpoint](https://docs.spring.io/spring-boot/api/rest/actuator/caches.html#caches.evict-all) by annotating our cache as a cache?

PR #6320, opened 2026-02-10
feat: AI Agent supports AWS AgentCore Memory as `MemoryStorageConfiguration`

### LINKED ISSUE #6852: AWS AgentCore Short-Term Memory as Custom Memory Store for AI Agent Sub-process Connector
## Summary

Implement **AWS AgentCore Short-Term Memory** as a custom memory store option for Camunda's AI Agent Sub-process connector, allowing agents to persist conversation context within a session and across process steps without losing state between LLM calls.

## Motivation

Camunda's AI Agent Sub-process connector delegates reasoning to an LLM that orchestrates tool calls inside an ad-hoc sub-process. By default, the agent is stateless — each invocation starts from scratch unless conversation history is manually threaded through process variables. AgentCore Short-Term Memory provides a managed, session-scoped event store that captures raw interaction data (user/assistant/tool messages) in chronological order. Integrating it as a pluggable memory store means the agent automatically retains multi-turn context, enabling more coherent reasoning across complex tool-calling sequences — without burdening the process designer with manual state management.

## AWS Setup (High-Level)

1. **Open the Amazon Bedrock AgentCore console** → navigate to **Memory**.
2. **Create a memory resource** — provide a name and optionally configure event expiry duration (up to 365 days).
3. **Note the Memory ID** — this will be referenced by the connector.
4. **Create an IAM policy** granting permissions for `bedrock-agentcore:CreateMemorySession`, `bedrock-agentcore:WriteMemoryEvents`, `bedrock-agentcore:ReadMemoryEvents`, and related actions.
5. **Attach the policy** to the IAM user or role used by the Camunda connector runtime.

Alternatively, use the AgentCore starter toolkit CLI: `agentcore configure` with memory enabled.

### Integration Point

This is **not** a standalone connector. It extends the existing **AI Agent Sub-process connector** by adding a selectable memory store backend (alongside the default in-process-variable approach).

### Configuration Fields (in the AI Agent Sub-process properties panel)

- **Memory Store** (dropdown): `Default (Process Variables)` | `AWS AgentCore Short-Term Memory`
- **Memory Resource ID** (required when AgentCore is selected)
- **Session ID** (optional — FEEL expression; auto-generated if omitted, allowing multi-turn continuity)
- **Namespace** (optional — for scoping memory to specific actors or contexts)
- **AWS Region** (required)
- **Authentication** — reuse existing AWS auth pattern (Credentials / Default Credentials Chain)

### Behavior

- On agent step start: read all events for the session from AgentCore Memory and inject them as conversation history into the LLM request.
- After each tool call completion: write the tool call and result as new events to AgentCore Memory.
- On agent completion: write the final assistant response. Optionally close the session.

## References

- [AgentCore Memory documentation](https://aws.amazon.com/blogs/machine-learning/amazon-bedrock-agentcore-memory-building-context-aware-agents/)
- [AgentCore Memory — short-term walkthrough](https://dev.to/aws-heroes/amazon-bedrock-agentcore-runtime-part-6-using-agentcore-short-term-memory-with-strands-agents-sdk-55d4)
- [AI Agent Sub-process connector (Camunda)](https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-subprocess/)
- [Camunda AI Agent connector customization](https://docs.camunda.io/docs/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-subprocess/)

mathieu-stennier:
## Summary

Adds AWS Bedrock AgentCore Memory as a conversation storage backend for the AI Agent Connector, alongside the existing in-process and Camunda Document stores.

image

## Storage Architecture

Messages are stored as AgentCore events scoped by `memoryId`, `actorId`, and `sessionId`. Each event contains one or more payloads:

- **Conversational payloads** — text content with a role (USER, ASSISTANT, TOOL). These flow into AgentCore's long-term memory extraction.
- **Blob envelopes** — versioned JSON wrappers (`camunda.toolCalls`, `camunda.toolCallResults`, `camunda.messageContent`, `camunda.messageMetadata`) for structured data that needs exact round-trip fidelity.

Content order is preserved by interleaving conversational and blob payloads in a single pass on both write and read paths.

### Branch-per-turn Strategy

Each agent turn after the first writes to a new AgentCore branch, forking from the last event of the previous turn. The branch name and last event ID are stored in the Zeebe process context.

This prevents split-brain on failed job completions: if a job completes in AgentCore but fails in Zeebe, the orphaned branch is invisible to the retry (which loads from the previous branch stored in the stale context). If no events are actually written (e.g. all messages produce empty payloads), the branch name is reverted to prevent saving a phantom branch in context.

### Canonical Role in Metadata Properties

The message role (USER, ASSISTANT, TOOL) is always stored in the `camunda.messageMetadata` blob envelope's `properties` section. This is the **canonical source** for role during deserialization — the conversational payload's role is only used by AgentCore for long-term memory extraction, not for message reconstruction. This ensures that blob-only events (e.g. a UserMessage with only DocumentContent and no text) are correctly reconstructed.

### System Message Handling

AgentCore doesn't support the SYSTEM role. System messages are preserved in `AwsAgentCoreConversationContext` and restored into runtime memory on load.

### Metadata & Properties Handling

Message metadata is stored as a `camunda.messageMetadata` blob envelope (not in AWS event metadata, which only supports restricted string values). The envelope has two sections:
- `metadata` — user-facing metadata map (round-trips exactly)
- `properties` — framework-internal data (`role`, `userName`) for message reconstruction

### Configuration Alignment

AgentCore and Bedrock LLM provider settings are aligned: same field names, labels, ordering, annotations, and SaaS restrictions (default credentials chain blocked on SaaS).

## Authentication

- **Static credentials** — access key + secret key
- **Default credentials chain** — EC2 instance profile, environment variables, etc. (blocked on SaaS)

## Other

- HTTP proxy support via the same mechanism as the Bedrock LLM client
- `memoryId` and `actorId` consistency validation between iterations
- Client-assigned event timestamps with turn-local `seq` metadata for deterministic ordering
- `UserMessage.name` preserved via metadata blob properties
- `ToolCallResult.content` serialized as JSON (not `Object.toString()`) for tool-call summaries
- Reference documentation at `docs/reference/aws-agentcore-memory.md`

## Testing/QA

Testing can be done by:

- On AWS, create new AgentCore Memory instance. Configure it with the desired long-term memory extraction strategies (e.g. tick all default ones for testing).
- Configure the AI Agent to use AgentCore Memory + configure the credentials and the memory ID of the memory instance
- Run an AI Agent and confirm it works and can loop

To visualize the AWS storage side, `agentcore-memory-browser` helps accessing the data. See https://dev.to/aws/visualizing-ai-agent-memory-building-a-web-browser-for-amazon-bedrock-agentcore-memory-3571


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/AwsAgentCoreConversationSession.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    final var conversationContextBuilder =
+        previousConversationContext != null
+            ? previousConversationContext.with()
+            : AwsAgentCoreConversationContext.builder(sessionId);
+
+    final var conversationContext =
+        conversationContextBuilder
+            .memoryId(config.memoryId())
```
nikonovd: 💭 is there a reason to overwrite those for each job that we work on? Are those expected to change or can we preserve the ones from the `previous context`?

nikonovd (commented):
Here's my first feedback so far.  Generally looks good and readable, i just had some questions in mind about simplifying the session implementation.

Will look into tests later!

connectors/agentic-ai/element-templates/agenticai-aiagent-job-worker.json
```diff
@@ -1363,6 +1366,155 @@
...
+      "type" : "zeebe:input"
+    },
+    "condition" : {
+      "property" : "data.memory.storage.type",
+      "equals" : "aws-agentcore",
+      "type" : "simple"
+    },
+    "type" : "String"
+  }, {
+    "id" : "data.memory.storage.endpointOverride",
```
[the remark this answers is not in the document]
mathieu-stennier: Production Use Cases:

1. **VPC Endpoint/PrivateLink Configurations**:
   - Organizations using AWS PrivateLink for private connectivity need to specify custom endpoint URLs
   - Example: `https://vpce-1234567-abcdefgh.bedrock-agent-core.us-east-1.vpce.amazonaws.com`
   - This is a common enterprise security requirement to avoid traffic over public internet

2. **AWS GovCloud and Special Partitions**:
   - GovCloud and other AWS partitions have different endpoint patterns
   - Standard SDK region resolution may not work correctly
   - Example: `https://bedrock-agent-core.us-gov-west-1.amazonaws.com`

It's advanced but required in most SM setup.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/AwsAgentCoreConversationSession.java
```diff
@@ -0,0 +1,263 @@
...
+          config.memoryId(),
+          config.actorId(),
+          e.getMessage(),
+          e);
+      throw new IllegalStateException(
+          "Unexpected error while loading conversation history from AgentCore Memory", e);
+    }
+  }
+
+  private void storeMessagesToAgentCore(
```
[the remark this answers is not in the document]
mathieu-stennier: If the connector job fails after storing some (but not all) messages to AgentCore Memory,
the retry behavior depends on when the failure occurred:

**Failure during message storage**:
The `clientToken` mechanism provides some protection against duplicate events ([idempotency](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_CreateEvent.html#API_CreateEvent_RequestSyntax)). However, if the LLM is re-invoked on retry, it may generate different responses. This can lead to inconsistent conversation history.

This is a general limitation of the Connector Runtime's retry mechanism and affects
any connector with multiple external API calls within a single execution.

Maybe we could think about [deleting](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_DeleteEvent.html) all events that are > stored index. But maybe we don't need that right away?

connectors/agentic-ai/element-templates/agenticai-aiagent-job-worker.json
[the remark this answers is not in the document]
nikonovd: Thanks for the clarification. Maybe let's switch the order in the description then, because right now it feels more like a testing opportunity. WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/AwsAgentCoreConversationSession.java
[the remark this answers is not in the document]
mathieu-stennier: @nikonovd / @maff Maybe we could use event branching specifically in this AWS implementation
https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/short-term-create-event.html#short-term-event-branching

connectors/agentic-ai/element-templates/hybrid/agenticai-aiagent-job-worker-hybrid.json
```diff
@@ -277,7 +277,7 @@
   }, {
     "id" : "provider.bedrock.authentication.secretKey",
     "label" : "Secret key",
-    "description" : "Provide a secret key of a user with permissions to invoke specified AWS Lambda function",
+    "description" : "Provide the secret key for the IAM access key",
```
maff: ℹ️ This was wrong in the first place for the bedrock connection - fixed it as part of the alignment between Bedrock and AgentCore configs.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/InMemoryBedrockAgentCoreClientFactory.java
```diff
@@ -0,0 +1,184 @@
...
+      if (includeParents) {
+        // Walk up the branch chain to find parent branches
+        var currentBranch = targetBranch;
+        while (currentBranch != null) {
+          var branchInfo = branches.get(currentBranch);
+          if (branchInfo == null) break;
+
+          // Find which branch the rootEventId belongs to
+          var rootEvent = events.get(branchInfo.rootEventId());
+          if (rootEvent == null) break;
```
maff: 🔧 Let's not use inline conditionals here (and across the PR).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/mapping/AwsAgentCoreConversationMapper.java
```diff
@@ -0,0 +1,407 @@
...
+  // ==================== Payloads to Messages ====================
+
+  /**
+   * Extracts messages from payloads by processing them in order, preserving the original content
+   * interleaving between conversational and blob payloads.
+   *
+   * <p>Special blob types (metadata, toolCalls, toolCallResults) are extracted separately. Content
+   * payloads (conversational text and messageContent blobs) are collected in their original order.
+   */
+  private List<Message> extractMessagesFromPayloads(
```
maff: 🔧 Do we ever return more than one message here? If not, can we simplify the return type to a single message?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/AwsAgentCoreConversationSession.java
```diff
@@ -0,0 +1,326 @@
...
+  private boolean isStorableMessage(Message message) {
+    return message instanceof UserMessage
+        || message instanceof AssistantMessage
+        || message instanceof ToolCallResultMessage;
+  }
+
+  /**
+   * Validates that memoryId and actorId have not changed between iterations. Changing these would
+   * silently write to a different memory resource while the context still references the old one.
+   */
```
maff, line 182: ℹ️ Theoretically this could also happen on change to region or different credentials, but I think that's rather an edge case (a same memory ID must exist in a different account or region). memory & actor ID are FEEL-enabled fields which can easily be changed.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/DefaultBedrockAgentCoreClientFactory.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @Override
+  public BedrockAgentCoreClient createClient(AwsAgentCoreMemoryStorageConfiguration config) {
+    BedrockAgentCoreClientBuilder builder = BedrockAgentCoreClient.builder();
+
+    // apply authentication
+    builder.credentialsProvider(createCredentialsProvider(config.authentication()));
+
+    builder.region(Region.of(config.region()));
```
maff: 🔧  Let's merge that into the initial builder line - no need to keep it separate.

maff:
@nikonovd @mathieu-stennier I think it should now be ready for another review round. PR description and the reference docs should include all design decisions & updates. What primarily changed from the original PR:

- branching to cope with completion errors
- added HTTP proxy support & e2e tests
- some changes on the data stored (properties, metadata) to ensure round-trip consistency
- a bunch of refactorings and changes to the element template for consistency with the other AWS integrations

connectors/agentic-ai/docs/reference/ai-agent.md
```diff
@@ -378,6 +379,16 @@ ConversationStore (registered backend)
...
 - Supports transparent migration from `InProcessConversationContext`: if the context is in-process, it reads messages directly (no document to load)
 
+**AwsAgentCoreConversationStore** (`type = "aws-agentcore"`):
+- Stores messages as events in AWS Bedrock AgentCore Memory
+- Uses a **branch-per-turn** strategy for isolation: each agent turn writes to a fresh branch, so failed job completions leave orphaned branches that are invisible on retry
+- `AgentContext.conversation` contains `AwsAgentCoreConversationContext` with branch pointer (`branchName`, `lastEventId`) and system message
+- On load: `ListEvents` with branch filter + `includeParentBranches=true` returns the full conversation chain
+- On store: new messages written to a new branch forked from the previous turn's last event
+- Conversational payloads feed AWS long-term memory extraction; structured data (tool calls, results) stored as versioned blob envelopes
+- System messages preserved in context (AgentCore has no SYSTEM role)
```
mathieu-stennier, line 389: ℹ️ Also not to pollute the long term memory with system messages.

mathieu-stennier (commented):
LGTM

maff (approved):

gbetances089 (approved):

gbetances089:
Ticket for E2E tests: https://github.com/camunda/camunda/issues/50857

PR #6337, opened 2026-02-12
refactor: Extract element template generator annotations

sbuettner:
## Description

Extracted the element template generator annotations into a separate module to avoid consumers to import the whole generator including all its dependencies when they want to use the annotations for their connectors.

@chillleader I removed some Jackson annotations from  `BpmnType` (now to be found under `ElementTemplate.BpmnType`) to make it free from the Jackson dependency. Do you think its fine like that?

cc @johnBgood please also take a look.

## Checklist


element-template-generator/core/src/main/java/io/camunda/connector/generator/dsl/ElementTemplateBuilder.java
```diff
@@ -43,7 +44,7 @@ public class ElementTemplateBuilder {
   protected Engines engines;
   protected Metadata metadata;
   protected Set<String> appliesTo;
-  protected BpmnType elementType;
+  protected io.camunda.connector.generator.java.annotation.ElementTemplate.BpmnType elementType;
```
johnBgood: and here

element-template-generator/core/src/test/java/io/camunda/connector/generator/dsl/ElementTypeSupportTest.java
```diff
@@ -30,12 +30,19 @@ void serviceTask() {
             .id("id")
             .name("name")
             .type("type", false)
-            .appliesTo(BpmnType.TASK)
-            .elementType(BpmnType.SERVICE_TASK)
+            .appliesTo(io.camunda.connector.generator.java.annotation.ElementTemplate.BpmnType.TASK)
```
johnBgood: and here

chillleader (approved):

PR #6394, opened 2026-02-16
fix(agentic-ai): override default /mcp and /sse endpoint default settings to avoid breaking the configured URL

### LINKED ISSUE #6393: MCP Client connector can't connect to servers not served from default endpoint path
__Describe the Bug__

When connecting to an MCP server which is served from another path than the default endpoint `/mcp` or `/sse`, the default endpoint is wrongly added to the configured server URL.

__Steps to Reproduce__

1. Configure a MCP Remote Client connector with a streamable HTTP transport and configure the URL to an MCP server served under a path different from `/mcp`
  - For example `https://my-datacenter/clusters/3a9825a1-7647-4328-8c70-e1364354cff2/mcp/cluster` 
2. Try to execute the process. Despite configuring the correct url and authentication, the connector fails to execute the MCP request.

The same applies to SSE connections.

__Expected Behavior__

MCP connector can successfully connect to the configured MCP server.

__Environment__

- Camunda version: 8.9.0-SNAPSHOT

maff:
## Description

The MCP SDK HTTP transport builders accept both a base URI and an endpoint. When the endpoint is not explicitly configured, a default value of `/mcp`  or `/sse` will be used and will override any path parts in the configured base URI.

This change overrides the endpoint part to the same value as the base URI which will reconfigure the SDK to use only the endpoint part and to validate it against being a part of the base URI.

## Related issues

closes #6393 

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/mcp/client/framework/mcpsdk/McpSdkClientFactoryTest.java
+++ b/connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/mcp/client/framework/mcpsdk/McpSdkClientFactoryTest.java
@@ -12,6 +12,7 @@ import static com.github.tomakehurst.wiremock.client.WireMock.postRequestedFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.awaitility.Awaitility.await;
 import static org.mockito.Mockito.when;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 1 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/framework/mcpsdk/McpSdkClientFactory.java
```diff
@@ -80,6 +80,8 @@ private HttpClientStreamableHttpTransport createStreamableHttpTransport(
     var headerSupplier = headersSupplierFactory.createHttpHeadersSupplier(streamableHttpConfig);
 
     return HttpClientStreamableHttpTransport.builder(streamableHttpConfig.url())
+        .endpoint(
+            streamableHttpConfig.url()) // see https://github.com/camunda/connectors/issues/6393
```
maff: Internally, [Utils.java](https://github.com/modelcontextprotocol/java-sdk/blob/main/mcp-core/src/main/java/io/modelcontextprotocol/util/Utils.java#L71) will use the endpoint if it is an absolute URI and only validate if it is part of the base URI.

This should fix our use case and be covered by tests if the SDK behavior ever changes.

PR #6494, opened 2026-02-24
fix(deps): downgrade elasticsearch client

mdii:
## Description

Spring Boot 4 has upgraded the elasticsearch java client to 9.x. This new client is not compatible with elasticsearch 8.x that is supported in the embedding vector database. We need to downgrade the ES client to 8.x. 

## Related issues

closes #

## Checklist

### DIFF
--- a/bundle/pom.xml
+++ b/bundle/pom.xml
@@ -6,11 +6,8 @@
   <packaging>pom</packaging>
 
   <name>Connectors Bundle</name>
-   <properties>
-     <elasticsearch-java.version>8.19.11</elasticsearch-java.version>
-   </properties>
 
-   <parent>
+  <parent>
     <groupId>io.camunda.connector</groupId>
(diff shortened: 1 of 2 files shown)


connectors/embeddings-vector-database/pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <properties>
     <version.apache-tika>3.2.3</version.apache-tika>
-    <elasticsearch-rest-client.version>9.3.0</elasticsearch-rest-client.version>
+    <version.elasticsearch-rest-client>8.19.11</version.elasticsearch-rest-client>
```
vringar: Versions shared across multiple dependencies should use `parent/pom.xml` to define the version number and the dependencyManagement section. Then all users should declare it with the explicit version, unless this doesn't work with the BOM, in which case please still define the variable once in `parent/pom.xml`

PR #6536, opened 2026-02-27
fix(vector-db): downgrade OpenSearch client

mdii:
## Description

Langchain4j requires OpenSearch client `>= 2.9`. We get an error at runtime with version `3.x`. 
In addition `httpclient5` has introduced a change in `5.6` that enables content compression by default. This leads to an error when the server does not compress the content. We need to explicitly disable the content compression on the HTTP client. The Amazon-managed OpenSearch is not impacted by this behavior.

## Related issues

closes #

## Checklist

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -105,6 +105,16 @@
         "digest"
       ],
       "changelogUrl": "https://github.com/camunda/infra-global-github-actions/compare/{{currentDigest}}..{{newDigest}}"
+    },
+    {
+      "description": "Disable major version updates to prevent breaking transitive dependency compatibility with langchain4j.",
+      "matchUpdateTypes": [
+        "major"
+      ],
+      "matchPackageNames": [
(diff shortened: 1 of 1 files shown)


parent/pom.xml
```diff
@@ -107,7 +107,8 @@ limitations under the License.</license.inlineheader>
     <version.localstack>2.0.3</version.localstack>
 
     <version.okhttp>4.12.0</version.okhttp>
-    <version.opensearch-client>3.6.0</version.opensearch-client>
+    <!-- Langchain4j requires a version >= 2.9.0 < 3.x. Do not bump to 3.x unless Langchain4j supports it! -->
```
maff: Do we potentially need to configure renovate to not bump the major version?

mdii: Done!

maff (approved):

PR #6546, opened 2026-03-02
feat(http-connector): added followRedirects option for 3xx responses

### LINKED ISSUE #6421: REST Connector: Configurable redirect handling (follow 3xx redirects) , also applies to HTTP Polling Connector
### Is your feature request related to a problem? Please describe.

When using the Camunda 8 REST Connector to call APIs that respond with HTTP redirects (3xx), the connector does not follow the redirect automatically. This differs from common tooling such as Postman, where following redirects is the default behaviour. As a result, calls that succeed in testing can fail or require extra process logic when implemented with the connector.
This also impacts the HTTP Polling Connector, as it is built on top of the REST Connector, so the same redirect behaviour affects polling use cases.

### Describe the solution you'd like

Add support for following HTTP redirects (3xx) in the REST Connector, and make it configurable (for example, a “Follow redirects” toggle/setting). Ideally the setting should allow users to enable/disable redirect following, and apply consistently anywhere the REST Connector is used, including the HTTP Polling Connector.

### Describe alternatives you've considered

Avoid redirected endpoints by hardcoding the final URL (when possible).

Handle redirects manually by detecting 3xx responses and issuing a second request to the Location header within the process.

Both alternatives add unnecessary complexity and are error-prone.

### Additional context

The HTTP Polling Connector is built on top of the REST Connector, so implementing redirect handling at the REST Connector level would likely cover both connectors. Users expect redirect-following behaviour because it is the default in tools like Postman, and many APIs rely on redirects for URL normalisation, moved resources, or edge/front-door routing.

azan-baloch:
## Description

Added support for Http redirects(3xx), previously if user sent request to url that returned a 3xx response with location header, than connector did not follow the redirect. Now user have option to enable/disable redirects from **Follow redirects** option.

. When option is  **enabled** than connector automatically follows redirects like Postman.
. When disabled, connector will return the original 3xx response(default behaviour).

Here are screenshots of responses when redirect is disabled and enabled

**when disabled:**
3xx-default

**when enabled:**
3xx-follow-redirect

## Related issues

closes #6421

## Checklist

### DIFF
--- a/connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/CustomApacheHttpClient.java
+++ b/connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/CustomApacheHttpClient.java
@@ -52,10 +52,9 @@ public class CustomApacheHttpClient implements HttpClient {
     try (var client =
         new ProxyAwareHttpClient(
             new ProxyAwareHttpClient.TimeoutConfiguration(
-                request.getConnectionTimeoutInSeconds(),
-                request.getReadTimeoutInSeconds(),
-                request.isFollowRedirects()),
-            new ProxyAwareHttpClient.ProxyContext(scheme, host))) {
+                request.getConnectionTimeoutInSeconds(), request.getReadTimeoutInSeconds()),
+            new ProxyAwareHttpClient.ProxyContext(scheme, host),
+            request.isFollowRedirects())) {
 
       var apacheResponseHandler =
           new CustomResponseHandler<>(responseMapper, request.isFollowRedirects());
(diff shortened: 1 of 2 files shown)


ztefanie:
> @ztefanie Should I bump engineVersion for both REST and polling connectors, or only polling or none? Currently rest has engine version 8.3 and polling is on 8.8.

Please bump versions for both. The reason for this that then modelers can check for compatible versions of element templates. So if a user runs on engine 8.8 we do not want to suggest using this new template, which supports "follow redirects", as the runtime would not be able to handle this properly. So we only want to distribute this new element templates to users on >=8.9.

So just move the previous version of the element template to the "versioned" subfolder and increase the version to the next number and set the engine version. Thank you :)

PR #6591, opened 2026-03-05
[Release 8.9 ] Release branch to base branch

slolatte:
Sync results of release process back into appropriate base branch


ztefanie:
ok sure, we can also merge it, there are no conflicts. Or do you want to merge this onto main then?

@maff As the test is flaky this quick fix should stay in place until a permanent stabilization is implemented, correct? so it should not be an issue to merge this.

PR #6659, opened 2026-03-12
chore(docker): create a way to have health check inside hardened image

mathias-vandaele:
This pull request introduces a Docker health check mechanism for the connector runtime images, ensuring reliable readiness detection without relying on external tools like `curl` or `wget`. The health check is implemented via a new lightweight Java command that queries the Spring Boot Actuator readiness endpoint. The Dockerfiles for all relevant bundles are updated to use this new health check.

Health check implementation:

* Added a new `HealthCheckCommand` Java class in `connector-runtime/connector-runtime-application/src/main/java/io/camunda/connector/runtime/healthcheck/HealthCheckCommand.java` that performs an HTTP GET to the `/actuator/health/readiness` endpoint and exits with code 0 on success, or 1 on failure. It supports configurable ports and base paths via environment variables.

Dockerfile updates for health check:

* Updated `bundle/camunda-saas-bundle/Dockerfile` to include the new `HEALTHCHECK` instruction, invoking the `HealthCheckCommand` for readiness checks.
* Updated `bundle/default-bundle/Dockerfile` to add the same health check mechanism.
* Updated `connector-runtime/connector-runtime-application/Dockerfile` to add the health check using the new Java command.

# TEST

It was locally tested with success 

image

### DIFF
--- a/bundle/camunda-saas-bundle/Dockerfile
+++ b/bundle/camunda-saas-bundle/Dockerfile
@@ -19,4 +19,9 @@ USER 1001:1001
 ENV CAMUNDA_CONNECTOR_RUNTIME_SAAS=true
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
 
+HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
+  CMD java -cp /opt/app/*-with-dependencies.jar \
+      -Dloader.main=io.camunda.connector.runtime.healthcheck.HealthCheckCommand \
+      org.springframework.boot.loader.launch.PropertiesLauncher
+
 ENTRYPOINT ["java", "--add-opens=java.base/java.util.concurrent=ALL-UNNAMED", "-cp", "/opt/app/*", "org.springframework.boot.loader.launch.PropertiesLauncher"]
--- a/bundle/camunda-saas-bundle/Dockerfile
+++ b/bundle/camunda-saas-bundle/Dockerfile
@@ -19,6 +19,7 @@ USER 1001:1001
 ENV CAMUNDA_CONNECTOR_RUNTIME_SAAS=true
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
 ENV MANAGEMENT_SERVER_PORT=9080
+ENV HEALTHCHECK_URL=http://localhost:9080/actuator/health/readiness
 
 HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
   CMD java -cp /opt/app/*-with-dependencies.jar \
--- a/bundle/camunda-saas-bundle/Dockerfile
+++ b/bundle/camunda-saas-bundle/Dockerfile
@@ -18,6 +18,7 @@ USER 1001:1001
 
 ENV CAMUNDA_CONNECTOR_RUNTIME_SAAS=true
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
+ENV MANAGEMENT_SERVER_PORT=9080
 
 HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
   CMD java -cp /opt/app/*-with-dependencies.jar \
--- a/bundle/camunda-saas-bundle/Dockerfile
+++ b/bundle/camunda-saas-bundle/Dockerfile
@@ -18,7 +18,6 @@ USER 1001:1001
 
 ENV CAMUNDA_CONNECTOR_RUNTIME_SAAS=true
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
-ENV MANAGEMENT_SERVER_PORT=9080
 ENV HEALTHCHECK_URL=http://localhost:9080/actuator/health/readiness
 
 HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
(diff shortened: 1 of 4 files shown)


connector-runtime/connector-runtime-application/src/main/java/io/camunda/connector/runtime/healthcheck/HealthCheckCommand.java
```diff
@@ -0,0 +1,87 @@
...
+ * </ul>
+ */
+public class HealthCheckCommand {
+
+  public static void main(String[] args) {
+    HttpURLConnection connection = null;
+    try {
+      String port = resolvePort();
+      String basePath = resolveBasePath();
+      String url = "http://localhost:" + port + basePath + "/health/readiness";
```
sbuettner: What if we just provide the full URL directly as a command argument?

bundle/camunda-saas-bundle/Dockerfile
```diff
@@ -18,5 +18,11 @@ USER 1001:1001
 
 ENV CAMUNDA_CONNECTOR_RUNTIME_SAAS=true
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
+ENV HEALTHCHECK_URL=http://localhost:9080/actuator/health/readiness
+
+HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
+  CMD java -cp /opt/app/*-with-dependencies.jar \
+      -Dloader.main=io.camunda.connector.runtime.healthcheck.HealthCheckCommand \
+      org.springframework.boot.loader.launch.PropertiesLauncher
```
johnBgood, line 26: as @sbuettner suggested can we have something self-contained like:
```
CMD java -cp /opt/app/*-with-dependencies.jar \
    -Dloader.main=io.camunda.connector.runtime.healthcheck.HealthCheckCommand \
    org.springframework.boot.loader.launch.PropertiesLauncher \
    http://localhost:9080/actuator/health/readiness
```
and we args[0] in the command?

bundle/default-bundle/Dockerfile
```diff
@@ -27,4 +27,11 @@ RUN addgroup --gid 1001 camunda && adduser -S -G camunda -u 1001 --no-create-hom
 USER 1001:1001
 
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
+ENV HEALTHCHECK_URL=http://localhost:8080/actuator/health/readiness
+
+HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
+  CMD java -cp /opt/app/*-with-dependencies.jar \
```
johnBgood, line 33: Claude suggested this:

> Classpath glob *-with-dependencies.jar: This is fragile. If the jar naming convention changes or multiple jars match, the healthcheck silently breaks. Is there a more stable way to reference it? Some teams use a symlink like healthcheck.jar at image-build time to make this deterministic.

bundle/camunda-saas-bundle/Dockerfile
```diff
@@ -18,5 +18,11 @@ USER 1001:1001
 
 ENV CAMUNDA_CONNECTOR_RUNTIME_SAAS=true
 ENV CAMUNDA_CLIENT_AUTH_CREDENTIALS_CACHE_PATH=/tmp/connectors
+ENV HEALTHCHECK_URL=http://localhost:9080/actuator/health/readiness
+
+HEALTHCHECK --interval=30s --timeout=5s --retries=5 --start-period=30s \
+  CMD java -cp /opt/app/*-with-dependencies.jar \
```
johnBgood, line 24: so we spawn a full app every 30s, are we ok with the cost?

bundle/default-bundle/Dockerfile
johnBgood: maybe we can create a healthcheck.jar that contains only the command and we can execute it using java jar ?

bundle/camunda-saas-bundle/Dockerfile
mathias-vandaele: We can't pass it as an argument, otherwise if we decide to change the management port or the base path service through env variables or via configMap, we have no way to change how the image has been built

bundle/camunda-saas-bundle/Dockerfile
mathias-vandaele: `HealthCheckCommand` does minimal work (one HTTP call then exit) so the cost will be extremely minimal, and this is the standard practice for hardened image as we have no other choices of tools.

bundle/default-bundle/Dockerfile
mathias-vandaele: @johnBgood We could actually use a symlink, but that's already an "issue" if we can call it that way to copy the file like this without renaming it properly `COPY target/*-with-dependencies.jar /opt/app/`

johnBgood: no but I mean it would just be a Java class with a Main, not a full spring boot app.

bundle/camunda-saas-bundle/Dockerfile
johnBgood: why? If we pass the env var as an argument, wouldn't it be ok?

mathias-vandaele: Passing the env var as an argument would still bake the resolved value into the image at build time, so it wouldn't be overridable at runtime.

bundle/default-bundle/Dockerfile
mathias-vandaele: That's actually already the case 👍🏼  The `HealthCheckCommand` is not a full Spring Boot app, it's just a tiny main class. We're reusing the existing `-with-dependencies.jar` purely for convenience since it already has everything on the classpath, we just point `-Dloader.main` at that specific class and only that runs.

Creating a separate `healthcheck.jar` with just that one class would work too, but it adds build complexity for little gain since the class is already there.

johnBgood: all good I didn't notice this, perfect then

johnBgood (approved):
great! thx

bundle/camunda-saas-bundle/Dockerfile
sbuettner: @mathias-vandaele Last time you mentioned that users can override the environment variables.

PR #6709, opened 2026-03-17
fix(graphql): throw ConnectorException on GraphQL 200 responses containing errors

### LINKED ISSUE #6702: GraphQL connector should consider GraphQL Errors
__Describe the Bug__

The GraphQL spec mentions [errors with 200 OK](https://graphql.org/learn/debug-errors/#graphql-errors-with-200-ok), however these currently don't trigger the error expression, which feels like a bug in the GraphQL Connector.

__Steps to Reproduce__

1. Query `https://countries.trevorblades.com/` with query
    ```GraphQL
    query TestError {
      country(code: "DE") {
       code
       name
      thisFieldDoesNotExist
      }
   }
   ``` 
   and resultVariable `response`
2. See that response has status 200 and response.body only contains the errors arrray

__Expected Behavior__

This should result in an Exception/ErrorExpression should be triggered

vringar:
## Description

A GraphQL server can return HTTP `200 OK` while still reporting errors in the response body (e.g. `{"data": null, "errors": [{"message": "..."}]}`). This is valid per the GraphQL over HTTP spec, but the connector was silently treating these responses as successes, making it impossible to trigger `errorExpression` in a process.

This is a first pass at implementing the fix — I welcome any edits you see fit.

**Root cause:** `CustomResponseHandler` only throws for HTTP status ≥ 400. A `200 + errors` response bypassed it entirely and was returned as a `SuccessResult`.

**Fix (`GraphQLFunction.java`):**
- After the HTTP call, check if the response body is a `Map` with a non-empty `errors` list
- If so, throw a `ConnectorException` with:
  - Error code `GRAPHQL_ERROR`
  - The first GraphQL error's `message` as the exception message
  - `errorVariables = {response: {body: ..., headers: ...}}` — consistent with `ConnectorExceptionMapper` so existing `errorExpression` patterns work
- Safe `instanceof HttpCommonResult` guard (avoids a blind cast that would break when `storeResponse` is enabled and returns a `Document`)

**Tests added/strengthened (`GraphQLFunctionTest.java`):**
- `shouldThrowConnectorException_WhenResponseContains200WithErrors` — asserts `errorCode`, exception message, and `errorVariables` shape (`response.body` + `response.headers` present)
- `shouldReturnResult_WhenResponseContains200WithOnlyData` — regression test: clean `200` with only `data` (no `errors` key) still returns `HttpCommonResult` without throwing

## Related issues

closes #6702

## Checklist

### DIFF
--- a/connectors/http/graphql/src/main/java/io/camunda/connector/http/graphql/GraphQLFunction.java
+++ b/connectors/http/graphql/src/main/java/io/camunda/connector/http/graphql/GraphQLFunction.java
@@ -8,14 +8,19 @@ package io.camunda.connector.http.graphql;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import io.camunda.connector.api.annotation.OutboundConnector;
+import io.camunda.connector.api.error.ConnectorExceptionBuilder;
 import io.camunda.connector.api.outbound.OutboundConnectorContext;
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
 import io.camunda.connector.generator.java.annotation.ElementTemplate;
 import io.camunda.connector.http.base.HttpService;
 import io.camunda.connector.http.base.model.HttpCommonRequest;
+import io.camunda.connector.http.base.model.HttpCommonResult;
 import io.camunda.connector.http.graphql.model.GraphQLRequest;
 import io.camunda.connector.http.graphql.utils.GraphQLRequestMapper;
 import io.camunda.connector.jackson.ConnectorsObjectMapperSupplier;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


ztefanie (approved):
Thanks for your contribution, looks good. 🚀 

I suggest we do not backport this, as this is a breaking change (same processes will have different behavior) and it there is a workaround of having a response expression where checking for an error can be done. WDYT?

mathias-vandaele (changes requested):
Very good job! 

I see that we are only returning the first error message, couldn't we do something like 

```
throw new ConnectorExceptionBuilder()
    .errorCode(GRAPHQL_ERROR_CODE)
    .message("GraphQL returned errors")
    .errorVariables(Map.of(
        "response", responseVariables,
        "errors", errors  
    ))
    .build();
```

or even 

```
throw new ConnectorExceptionBuilder()
    .errorCode(GRAPHQL_ERROR_CODE)
    .message("GraphQL returned errors")
    .errorVariables(Map.of(
      "graphqlErrors", errors,
      "responseBody", result.body(),
      "responseHeaders", result.headers()
))
    .build();
```

So that the client has full visibility over the error ?

mathias-vandaele (approved):

PR #6730, opened 2026-03-18
feat(element-template-generator): improve HTTP DSL auth deduplication and server URL visibility

vringar:
## Description

This PR improves the OpenAPI-to-element-template generation pipeline in the HTTP DSL in two areas:

### 1. `baseUrl` always editable

Previously, when an OpenAPI spec contained exactly one server, the generator emitted a `Hidden` property for `baseUrl`, pre-filled with the server URL. This meant users could not override the URL from the Modeler without editing the template.

The single-server branch now emits a `StringProperty` (visible, editable) with the server URL as the pre-filled default — matching the behaviour of the no-servers branch. Multi-server specs (dropdown) are unchanged.

### 2. Auth deduplication across operations

**Problem:** The previous implementation emitted one full set of auth properties per operation, regardless of whether multiple operations shared identical authentication requirements. For an API with 50 operations all using the same Bearer token, this produced 50 copies of the same auth fields — making templates unwieldy and the Modeler panel difficult to navigate.

**Approach:** Operations are grouped by their *effective auth* — `authenticationOverride` if set, else the global `authentications` list. The grouping key is the ordered list of auth IDs. A `LinkedHashMap` preserves first-appearance order.

Two cases arise:

- **Unconditional** (all ops share a single auth group): Auth properties are emitted once, with no conditions. The auth group is placed *before* the operation dropdown in the properties panel, since no property needs to reference `operationId` to scope its visibility.
- **Conditional / multi-group** (ops have different auth): Each group gets a scoped discriminator ID (`<firstOpId>_authType`) and field IDs prefixed with `<firstOpId>_` to prevent duplicate property ID collisions. Each group's properties carry an `OneOf(operationId, [...ops in group])` condition. In this case, auth properties are placed *after* the operation dropdown, because the Modeler resolves conditions top-to-bottom and the `operationId` discriminator must already be resolved.

**Why ordering matters:** The Modeler evaluates conditions in property order. A condition referencing `operationId` on a property that appears *before* `operationId` in the template will always be unresolved. The dynamic reordering ensures the correct evaluation order for both the conditional and unconditional cases.

**Edge case:** Callers that pass an empty `authentications` list (e.g. tests) and operations with no override produce an empty effective-auth list. These operations are now skipped in the grouping loop rather than crashing with `NoSuchElementException`.

### Files changed

- `PropertyUtil.java`: single-server `StringProperty` (Change 1); `authPropertyGroup` rewrite with deduplication (Change 2)
- `HttpOutboundElementTemplateBuilder.java`: dynamic auth/operation group ordering (Change 3)
- `HttpOutboundElementTemplateBuilderTest.java`: `singleServer_noVisibleExpected` → `singleServer_visibleStringExpected`, updated assertion
- `PropertyUtilTest.java`: 10 new unit tests covering unconditional, conditional, single-auth, multi-auth, and empty-auth scenarios

## Related issues

closes #

## Checklist

### DIFF
--- a/element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpAuthentication.java
+++ b/element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpAuthentication.java
@@ -242,6 +242,10 @@ public interface HttpAuthentication {
       return new BasicAuth(key);
     }
 
+    public String key() {
+      return key;
+    }
+
     @Override
     public String label() {
       return "Basic";
--- a/element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpOutboundElementTemplateBuilder.java
+++ b/element-template-generator/http-dsl/src/main/java/io/camunda/connector/generator/dsl/http/HttpOutboundElementTemplateBuilder.java
@@ -128,21 +128,29 @@ public class HttpOutboundElementTemplateBuilder {
     if (operations == null || operations.isEmpty()) {
       throw new IllegalStateException("Could not find any supported operations");
     }
-    return builder
-        .propertyGroups(
-            List.of(
-                // Property order is important, parameters must come before their targets (URL,
-                // headers, or body)
-                // otherwise they will not be resolved correctly by the FEEL engine in Zeebe
-                PropertyUtil.serverDiscriminatorPropertyGroup(servers),
-                PropertyUtil.operationDiscriminatorPropertyGroup(operations),
-                PropertyUtil.authPropertyGroup(authentication, operations),
-                PropertyUtil.parametersPropertyGroup(operations),
-                PropertyUtil.requestBodyPropertyGroup(operations),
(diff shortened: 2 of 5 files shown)


crobbins215:
> > Keeping it hidden makes it cleaner.
> 
> In my case it was about generating a template from https://console.cloud.dev.ultrawombat.com/external/openapi/docs/#/ which only gets "/" as a URL. And I need to make it generic over environments by putting a secret there. I guess I can do that as a manual edit. Maybe that isn't a common use case

Yeah i'm not against showing it if we have a usecase, I just haven't seen it asked. If your hitting into it that's a usecase. 

> @crobbins215 Having control over the base URL sounds like a reasonable thing to me. Customers usually have to fine tune the template anyway and can later decide whether they want to hide it or not.

Lets go with the change then

mathias-vandaele (approved):
This PR only changed the 0-server and 1-server cases to be editable strings but the N-servers case remains non-editable, is that what we want @sbuettner ?

Approved ✅

PR #6784, opened 2026-03-24
refactor(agentic-ai): redesign conversation storage SPI

### LINKED ISSUE #7019: Redesign conversation storage SPI
## What should we do?

Redesign the `ConversationStore` SPI from a callback-based pattern (`executeInSession`) to a factory-based pattern (`createSession`) with explicit session lifecycle management.

Key design decisions:
- Replace `executeInSession(handler)` with `createSession()` returning an `AutoCloseable` session
- Decouple `ConversationSession` from `RuntimeMemory` — sessions work with plain message types, not runtime internals
- Sessions return only storage artifacts (conversation cursor), not assembled agent state
- Remove `ConversationSessionHandler` and the deprecated `compensateFailedJobCompletion`

### Breaking changes

This is a breaking change for custom `ConversationStore` implementations. A migration guide will be provided in the breaking changes doc.

## Why should we do it?

The current `executeInSession` pattern has several problems:

- **False transactional semantics**: The callback pattern looks like a transaction boundary but isn't one. A DB-backed store would need to hold a transaction open for the entire LLM call (seconds to minutes), which is impractical.
- **Lifecycle ambiguity**: The implicit session lifecycle makes it hard to reason about resource cleanup, especially for stores managing external resources (e.g., AWS clients, DB connections).
- **Unnecessary coupling**: The session knows about `RuntimeMemory`, pulling runtime concerns into the persistence layer.
- **Testing friction**: All test logic must be wrapped inside lambda callbacks, making tests harder to read and debug.

maff:
## Description

Redesigns the conversation storage SPI from a callback-based pattern (`executeInSession`) to a factory-based pattern (`createSession`) for clearer lifecycle management and decoupling from `RuntimeMemory`.

**SPI changes:**
- `ConversationStore.executeInSession(handler)` → `createSession()` returning `AutoCloseable` session
- `ConversationSession` decoupled from `RuntimeMemory`: `loadMessages()` / `storeMessages()` work with `List<Message>` via wrapper types (`ConversationLoadResult`, `ConversationStoreRequest`)
- `storeMessages()` returns only `ConversationContext` (storage cursor); caller assembles `AgentContext`
- `ConversationSessionHandler` deleted, `compensateFailedJobCompletion` removed (was deprecated, never invoked)
- All 3 implementations migrated (InProcess, CamundaDocument, AwsAgentCore)
- AWS AgentCore client lifecycle now managed via `session.close()`

**Response type refactoring:**
- Introduce `AiAgentTaskConnectorResponse` implementing `StandardConnectorResponse` for the outbound task connector flavor
- Rename `AiAgentSubProcessResponse` to `AiAgentSubProcessConnectorResponse` to align with the `ConnectorResponse` hierarchy
- Tighten generic bound on `AgentRequestHandler` / `BaseAgentRequestHandler` to `R extends ConnectorResponse`
- Rename `completeJob` → `buildConnectorResponse` (the method only assembles the response, actual job completion happens in the runtime)

**Documentation:**
- ADR 003: [003-conversation-storage-spi-redesign.md](https://github.com/camunda/connectors/blob/agentic-ai-storage-refactoring/connectors/agentic-ai/docs/adr/003-conversation-storage-spi-redesign.md)
- [Storage Contract](https://github.com/camunda/connectors/blob/agentic-ai-storage-refactoring/connectors/agentic-ai/docs/reference/ai-agent.md#storage-contract) section in ai-agent.md — documents the write-ahead with pointer-based visibility pattern all store implementations must follow
- [Breaking changes](https://github.com/camunda/connectors/blob/agentic-ai-storage-refactoring/connectors/agentic-ai/docs/breaking-changes.md) with migration guide for custom implementations
- Javadoc on all SPI interfaces documenting the storage contract
- Updated `AGENTS.md` and `ai-agent.md` to reflect new API

## Related issues

closes #7019

## Checklist

### DIFF
--- a/connectors/agentic-ai/AGENTS.md
+++ b/connectors/agentic-ai/AGENTS.md
@@ -74,8 +74,8 @@ framework/
 
 memory/
 ├── conversation/
-│   ├── ConversationStore       # Pluggable storage: executeInSession() callback pattern
-│   ├── ConversationSession     # Per-invocation: loadIntoRuntimeMemory/storeFromRuntimeMemory
+│   ├── ConversationStore       # Pluggable storage: createSession() factory pattern
+│   ├── ConversationSession     # Per-invocation: loadMessages/storeMessages (AutoCloseable)
 │   ├── ConversationContext     # Persistent reference (conversationId)
 │   ├── inprocess/              # In-process store (messages in agentContext variable)
 │   └── document/               # Camunda Document Storage backend
@@ -133,14 +133,15 @@ complete — handled via `CommandWrapper` retries. For detailed mechanics, see
 
 ### Conversation Session Lifecycle
 
-`ConversationStore.executeInSession()` wraps agent processing in a callback:
+`ConversationStore.createSession()` returns an `AutoCloseable` session used via try-with-resources:
 
 ```
-executeInSession(ctx, agentContext, session -> {
-    session.loadIntoRuntimeMemory(agentContext, runtimeMemory)
-    [add messages, call LLM, etc.]
-    session.storeFromRuntimeMemory(agentContext, runtimeMemory)
-}) → completeJob
+try (var session = store.createSession(ctx, agentContext)) {
+    var loaded = session.loadMessages(agentContext)
+    [add messages to runtime memory, call LLM, etc.]
+    var cursor = session.storeMessages(agentContext, request)
+    agentContext = agentContext.withConversation(cursor)
+} → completeJob
 ```
 
 For backend-specific behavior and failure handling, see [ai-agent.md §6](docs/reference/ai-agent.md#6-conversation-memory).
--- a/connectors/agentic-ai/AGENTS.md
+++ b/connectors/agentic-ai/AGENTS.md
@@ -83,9 +83,6 @@ memory/
     ├── RuntimeMemory           # Transient working memory for single execution
     └── MessageWindowRuntimeMemory  # Sliding window filter (keeps last N messages)
 
-jobworker/
-└── AiAgentSubProcessResponse  # AdHocSubProcessConnectorResponse for AHSP directives (element activation, completion condition)
-
 tool/
 ├── GatewayToolHandler          # Interface for gateway tools (MCP, A2A)
 └── GatewayToolHandlerRegistry  # Registry of gateway tool handlers
@@ -105,7 +102,9 @@ tool/
 
 **`ToolCallProcessVariable`** — flattened tool call for process variables: `{_meta: {id, name}, ...args}`.
 
-**`AiAgentSubProcessResponse`** — job completion directives: AHSP done/continue, cancel flags, element activations, variables. Implements `AdHocSubProcessConnectorResponse` — the runtime translates it into the Zeebe complete command with ad-hoc sub-process result configuration.
+**`AiAgentSubProcessConnectorResponse`** — job completion directives: AHSP done/continue, cancel flags, element activations, variables. Implements `AdHocSubProcessConnectorResponse` — the runtime translates it into the Zeebe complete command with ad-hoc sub-process result configuration.
+
+**`AiAgentTaskConnectorResponse`** — wraps `AgentResponse` as a `StandardConnectorResponse` for the task connector flavor. The runtime evaluates result expressions against the wrapped response value.
 
 For full record definitions, see [ai-agent.md §5](docs/reference/ai-agent.md#5-data-model).
 
@@ -141,7 +140,7 @@ try (var session = store.createSession(ctx, agentContext)) {
     [add messages to runtime memory, call LLM, etc.]
     var cursor = session.storeMessages(agentContext, request)
     agentContext = agentContext.withConversation(cursor)
-} → completeJob
+} → buildConnectorResponse
 ```
 
 For backend-specific behavior and failure handling, see [ai-agent.md §6](docs/reference/ai-agent.md#6-conversation-memory).
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
@@ -22,6 +22,7 @@ import io.camunda.connector.agenticai.aiagent.tool.GatewayToolHandlerRegistry;
 import io.camunda.connector.agenticai.model.message.Message;
 import io.camunda.connector.agenticai.model.tool.ToolCallProcessVariable;
 import io.camunda.connector.agenticai.model.tool.ToolCallResult;
+import io.camunda.connector.api.outbound.ConnectorResponse;
 import java.util.List;
 import java.util.Optional;
 import org.apache.commons.lang3.tuple.Pair;
@@ -29,7 +30,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.lang.Nullable;
 
-public abstract class BaseAgentRequestHandler<C extends AgentExecutionContext, R>
+public abstract class BaseAgentRequestHandler<
+        C extends AgentExecutionContext, R extends ConnectorResponse>
     implements AgentRequestHandler<C, R> {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(BaseAgentRequestHandler.class);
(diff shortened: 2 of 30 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/ConversationLoadResult.java
```diff
@@ -0,0 +1,21 @@
...
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.aiagent.memory.conversation;
+
+import io.camunda.connector.agenticai.model.message.Message;
+import java.util.List;
+
+public record ConversationLoadResult(List<Message> messages) {
```
maff, line 14: ℹ️ I opted to wrapper objects here as we might need to transport some metadata if we potentially enhance this with support for message reconciliation (handling partial results).

connectors/agentic-ai/AGENTS.md
```diff
@@ -74,9 +74,10 @@ framework/
 
 memory/
 ├── conversation/
-│   ├── ConversationStore       # Pluggable storage: executeInSession() callback pattern
-│   ├── ConversationSession     # Per-invocation: loadIntoRuntimeMemory/storeFromRuntimeMemory
-│   ├── ConversationContext     # Persistent reference (conversationId)
+│   ├── ConversationStore       # Pluggable storage: createSession() factory
+│   ├── ConversationSession     # Per-invocation lifecycle: loadMessages/storeMessages/close
+│   ├── ConversationLoadResult  # Load result: messages + reconciledFromStore flag
+│   ├── ConversationContext     # Persistent reference with version tracking
```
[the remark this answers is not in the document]
maff: Stale — this references content from a previous iteration of the PR. The current AGENTS.md has no version tracking references.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/memory/conversation/inprocess/InProcessConversationStoreTest.java
```diff
@@ -54,15 +44,11 @@ void storeTypeIsAlignedWithConfiguration() {
...
-        session -> {
-          session.loadIntoRuntimeMemory(agentContext, memory);
-          return agentResponse(agentContext);
-        });
+    final var session = store.createSession(executionContext, agentContext);
+    final var loadResult = session.loadMessages(agentContext);
 
-    assertThat(memory.allMessages()).isEmpty();
+    assertThat(loadResult.messages()).isEmpty();
+    assertThat(loadResult.reconciledFromStore()).isFalse();
```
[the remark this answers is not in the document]
maff: Addressed — all InProcess and CamundaDocument tests now use try-with-resources.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/memory/conversation/document/CamundaDocumentConversationStoreTest.java
```diff
@@ -147,15 +126,11 @@ void throwsExceptionForUnsupportedConfiguration() {
...
-        session -> {
-          session.loadIntoRuntimeMemory(agentContext, memory);
-          return agentResponse(agentContext);
-        });
+    final var session = store.createSession(executionContext, agentContext);
+    final var loadResult = session.loadMessages(agentContext);
 
-    assertThat(memory.allMessages()).isEmpty();
+    assertThat(loadResult.messages()).isEmpty();
+    assertThat(loadResult.reconciledFromStore()).isFalse();
```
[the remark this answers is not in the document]
maff: Addressed — all CamundaDocument tests now use try-with-resources.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/ConversationLoadResult.java
```diff
@@ -0,0 +1,22 @@
...
+/** Result of loading a conversation from a {@link ConversationSession}. */
+public record ConversationLoadResult(List<Message> messages) {
+
+  public static ConversationLoadResult of(List<Message> messages) {
+    return new ConversationLoadResult(List.copyOf(messages));
+  }
+
+  public static ConversationLoadResult empty() {
+    return new ConversationLoadResult(List.of());
+  }
```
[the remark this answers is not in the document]
maff: Addressed — added compact constructors to both `ConversationLoadResult` and `ConversationStoreRequest` enforcing non-null and `List.copyOf()` on all instantiation paths.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/memory/conversation/inprocess/InProcessConversationStoreTest.java
```diff
@@ -37,32 +33,22 @@ class InProcessConversationStoreTest {
...
-        agentContext,
-        session -> {
-          session.loadIntoRuntimeMemory(agentContext, memory);
-          return agentResponse(agentContext);
-        });
+    var session = store.createSession(executionContext, agentContext);
+    var loadResult = session.loadMessages(agentContext);
 
-    assertThat(memory.allMessages()).isEmpty();
+    assertThat(loadResult.messages()).isEmpty();
```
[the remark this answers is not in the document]
maff: Addressed — all tests now use try-with-resources.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -110,11 +111,11 @@ private R handleRequest(
     return completeJob(executionContext, agentResponse);
```
nikonovd: 🔧 we should put the entire operations into the try-with-resources scope to avoid having a potentially unitilaized variable to the `completeJob` call.

❓ what about potential exceptions during createSession? Maybe it's even worth to consider a checked exception on the interface level, i.e.

`public ... createSession(...) throws SessionCreationException`

...when considering that we might expose this interface for custom implementations on self-managed instances?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd, line 134: ❓ ⛏️ not related to the changes here, but we could probably put the validation first to avoid loading messages if the model call limits are already reached.

nikonovd (approved):
Great improvement! Much simpler.

Generally i think we can improve the code flow a bit and make it more domain-oriented, flattening some of the handlers. But it's generally unrelated to this PR.

I left two minor comments, but will approve already as they are not blocking.

🚀

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
maff: Regarding the uninitialized variable: `AgentResponse` can be `null`, for example in the short-circuiting jobs still waiting on additional tool call results. We could potentially transport that information with a dedicated result object instead of null, but I refrained from introducing additional structures at the moment.

Would you prefer to initialize the `agentResponse` to `null` before the try block?

Regarding the checked exception: right now any exception in session creation would bubble up to the SDK and be treated as error & handled with job retries. What would you expect from the checked exception? To catch it & wrap it in a `ConnectorException` with a defined error code?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
maff: Yes that's a good point. There might be additional limits as part of the real time visibility, but I think none of them will need the messages. I'd still consider this part of a follow-up if you agree.

nikonovd: Sure!

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: Yes you're right about that. Generally i'd prefer using a Monad-ish type like `Maybe<T>` for highlighting optionality, but that's going to far.

In general, i'd just prefer moving everything inside the session-scope block what's feasible to flatten the branching complexity of the code.

i.e.:

```java
try (var session = store.createSession(executionContext, agentContext)) {
      var agentResponse = processConversation(executionContext, agentContext, toolCallResults, session);

    LOGGER.debug(
        "Request processing completed {} agent response, completing job",
        agentResponse == null ? "without" : "with");

    return completeJob(executionContext, agentResponse);
}
```

as the null is anyways produced inside the `processConversation` call and anyways passed to `completeJob`

maff: Agree! I now updated it as suggested and added some more improvements to get rid of the ambiguity of `completeJob`:

- both subprocess and task flavors now return an implementation of `ConnectorResponse` - this allows us to narrow down on `extends ConnectorResponse` on the `AgentRequestHandler`
- `completeJob` -> `buildConnectorResponse`

nikonovd (approved):
🚀 Awesome job! LGTM.

PR #6824, opened 2026-03-27
fix(agentic-ai): fix agent variable leaking into upper scope

### LINKED ISSUE #6821: Agentic AI: `agent` variable propagates to global scope from AI Agent processes
__Describe the Bug__

The AI Agent Process stores its final result as `agent` variable which is then mapped via an output mapping to the variable configured as result variable. By default, the result variable is configured to `agent`, leading to a single `agent` variable being created in the upper scope (e.g. the global scope).

When the result variable is configured to something else (e.g. `myAgentResponse`), both `agent` and `myAgentResponse` are created in the upper scope with the same content.

Image

The `agent` variable set by the agent is an internal implementation detail and only the output mapping should lead to a variable being created in the outer scope.

__Steps to Reproduce__

1. Model an AI Agent Process
2. Configure the result variable to `myAgentResponse` or something else than `agent`

⏩ When the agent completes, both `agent` and `myAgentResponse` are created in the global process scope.

__Expected Behavior__

Only `myAgentResponse` is created in the global process scope. `agent` is only set inside the ad-hoc sub-process scope and its contents are propagated to the process scope via the output mapping creating `myAgentResponse`.

__Environment__

- Camunda version: 8.9.x

maff:
## Description

Adds empty input mapping for `agent` to make sure the internal `agent` variable local and to make sure it is propagated to the parent scope only via the output mapping.

## Related issues

closes #6821

## Checklist


nikonovd (approved):
Looks good! 🏖️

PR #6799, opened 2026-03-26
feat(aws): add Bedrock Knowledge Base connector

tshepomaredi:
## Description

Adds a new outbound connector for AWS Bedrock Knowledge Base Retrieve API. Enables semantic search over documents indexed in a Bedrock Knowledge Base.

- BedrockAgentRuntimeClient-based implementation
- Returns individual results per chunk with documentReference + content
- Each result includes content text, relevance score, source URI, and metadata
- Template excluded from publishing via ignore-templates.json (pending 8.10)
- 5/5 unit tests passing (including paginationToken assertion)

## Related issues

closes #6851

## Checklist

- [ ] Backport labels are added if these code changes should be backported. No backport label is added to the latest
  release, as this branch will be rebased onto main before the next release. Example backport labels:
    - `backport stable/8.8`: for changes that should be included in the next 8.8.x release.
    - **or** `backport release-8.8.7`: for changes that should be included in the specific release 8.8.7, and this
      *release has already been created*. The release branch will be merged back into stable/8.8 later, so the change
      will be included in future 8.8.x releases as well.
- [ x] Tests/Integration tests for the changes have been added if applicable.

PR #6909, opened 2026-04-08
feat(aws): add Bedrock AgentCore Runtime connector

### LINKED ISSUE #6854: AWS AgentCore Runtime — Invoke Deployed Agent
## Summary

Implement an outbound Camunda connector that uses the **AWS AgentCore SDK** to invoke agents deployed on **AgentCore Runtime** via the `InvokeAgentRuntime` API. This allows Camunda processes to delegate complex, multi-step reasoning tasks to externally hosted AI agents on AWS, then consume the results as process variables.

## Motivation

Camunda's built-in AI Agent connectors are excellent for orchestrating agent behavior *inside* BPMN. But many organizations also have specialized agents already deployed on AWS AgentCore Runtime — built with Strands, LangGraph, CrewAI, or other frameworks. These agents may have their own tool integrations, memory configurations, and fine-tuned behaviors that are impractical to replicate as BPMN tool tasks. A dedicated connector bridges these worlds: Camunda handles end-to-end process orchestration, governance, and human-in-the-loop, while AgentCore Runtime handles the heavy lifting of agent execution in isolated, secure microVMs with up to 8 hours of runtime.

## Use Cases in Camunda

- **Delegate to a specialist agent:** A Camunda process handles overall loan origination, but delegates the underwriting risk analysis to a specialized agent deployed on AgentCore Runtime that has access to proprietary models and tools.
- **Multi-agent orchestration:** A BPMN process coordinates multiple AWS-hosted agents (e.g. a research agent, a compliance agent, and a drafting agent) in sequence or parallel, collecting and routing their outputs through gateways.
- **Hybrid architecture:** Keep deterministic steps (approvals, SLA timers, escalations, notifications) in Camunda while offloading non-deterministic AI reasoning to purpose-built agents on AWS.

## AWS Setup (High-Level)

1. **Build your agent** using any supported framework (Strands SDK, LangGraph, CrewAI, etc.) with an entrypoint script.
2. **Install the AgentCore SDK:** `pip install bedrock-agentcore`
3. **Configure and deploy:**
   ```bash
   agentcore init
   agentcore configure -e my_agent.py
   agentcore deploy
   ```
4. **Note the Agent Runtime ARN** from the deployment output (or from `.bedrock_agentcore.yaml`).
5. **Create an IAM policy** granting `bedrock-agentcore:InvokeAgentRuntime` permission on the agent ARN.
6. **Verify deployment:** `agentcore status` — ensure the agent is in `ACTIVE` state.

### Authentication

Reuse the existing AWS authentication pattern (Credentials or Default Credentials Chain).

### Input Fields

- **Agent Runtime ARN** (required) — the ARN of the deployed agent
- **Input message** (required) — the prompt or payload to send to the agent (FEEL expression)
- **Session ID** (optional) — for multi-turn interactions with the same agent session. Enables filesystem and memory persistence across calls.
- **AWS Region** (required)
- **Timeout** (optional) — max wait time for the agent response (agents can run up to 8 hours asynchronously)
- **Response mode** (optional): `SYNC` (wait for full response) or `STREAM` (for long-running agents, poll for completion)

### Output

- Agent response text / payload
- Session ID (for follow-up calls)
- Execution metadata (trace ID, duration)

### Streaming Consideration

AgentCore Runtime supports streaming responses. For synchronous use cases (most BPMN processes), the connector should buffer the full response. For long-running agents, consider supporting an async pattern where the connector creates a message correlation event upon completion.

## References

- [AgentCore Runtime documentation](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/agents-tools-runtime.html)
- [AgentCore SDK (Python)](https://github.com/aws/bedrock-agentcore-sdk-python)
- [Runtime quickstart](https://aws.github.io/bedrock-agentcore-starter-toolkit/user-guide/runtime/quickstart.html)
- [InvokeAgentRuntime API](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/agents-tools-runtime.html)
- [AgentCore GA announcement](https://aws.amazon.com/about-aws/whats-new/2025/10/amazon-bedrock-agentcore-available/)

tshepomaredi:
## Description

Adds a new outbound connector for invoking external agents deployed on AWS Bedrock AgentCore Runtime.

- Sync client using invokeAgentRuntimeAsBytes()
- Supports agentRuntimeArn, prompt (FEEL), and optional sessionId for multi-turn conversations
- Added to default-bundle
- Template excluded from publishing via ignore-templates.json (pending 8.10)
- 5/5 unit tests passing

## Related issues

Issue #6854

## Checklist

- [x ] Tests/Integration tests for the changes have been added if applicable.


nikonovd (changes requested):
Overall good looking, i left some remarks. Please look into them 🐥

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
```diff
@@ -0,0 +1,61 @@
...
+        builder.runtimeSessionId(input.getSessionId());
+      }
+
+      var responseBytes = client.invokeAgentRuntimeAsBytes(builder.build());
+      return new AgentCoreRuntimeResponse(
+          responseBytes.asUtf8String(),
+          responseBytes.response().runtimeSessionId(),
+          responseBytes.response().statusCode());
+
+    } catch (BedrockAgentCoreException e) {
```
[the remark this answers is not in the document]
tshepomaredi: For error handling, should we follow the same approach as KB/CI (let all exceptions go through SDK retry as @maff suggested [here](https://github.com/camunda/connectors/pull/6799#discussion_r3028486498)) or distinguish transient from non-transient errors like an `AccessDeniedException`? Just want to make sure its consistent across all three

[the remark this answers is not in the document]
nikonovd: Did some quick investigation.

The OutboundConnectorException handler has three branches for exception handling:

1. `ConnectorRetryException` -> maps a job command with retries from the exception data (we could use this as a shortcut)
2. Generic exception -> `job.getRetries() - 1`
3. `InvalidBackOffDurationException` -> immediately set retries to 0 and fail the job. Technically this exception seems to be only used at one location for a parse error. We could refactor here a bit with the connectors team to create sth. like an `ExhaustRetriesException` where the mentioned one would also extend to fit into a generic concept.

Anyways we can also create a followup issue here and proceed with the simple approach of retries promoted by the SDK, fine with that as well 👍

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/model/request/AgentCoreRuntimeInput.java
```diff
@@ -0,0 +1,78 @@
...
+      optional = true)
+  private String sessionId;
+
+  @TemplateProperty(
+      group = "agentConfig",
+      label = "Content type",
+      description = "Content type of the payload. Defaults to application/json.",
+      defaultValue = "application/json",
+      optional = true)
+  private String contentType;
```
maff: 🔧 As this refers to the payload, can we name it `payloadContentType` and put it after the payload field?

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/model/request/AgentCoreRuntimeInput.java
```diff
@@ -0,0 +1,78 @@
...
+public class AgentCoreRuntimeInput {
+
+  @NotBlank
+  @TemplateProperty(
+      group = "agentConfig",
+      label = "Agent Runtime ARN",
+      description = "The ARN of the AgentCore Runtime agent to invoke.")
+  private String agentRuntimeArn;
+
+  @FEEL
```
maff: 🔧 `@FEEL` can be removed in outbound connectors (I know it's there historically in others).

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+          input.getContentType() != null ? input.getContentType() : DEFAULT_CONTENT_TYPE;
+
+      var payloadBytes = objectMapper.writeValueAsBytes(input.getPayload());
+
+      var builder =
+          InvokeAgentRuntimeRequest.builder()
+              .agentRuntimeArn(input.getAgentRuntimeArn())
+              .payload(SdkBytes.fromByteArray(payloadBytes))
+              .contentType(contentType)
+              .accept(contentType);
```
maff: ❓ Is there any case where it would make sense to use different content types? Let's say we input a text, but expect a JSON response?

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+
+      if (input.getSessionId() != null && !input.getSessionId().isBlank()) {
+        builder.runtimeSessionId(input.getSessionId());
+      }
+
+      var responseBytes = client.invokeAgentRuntimeAsBytes(builder.build());
+      var responseText = responseBytes.asUtf8String();
```
maff, line 48: 🔧 Just to confirm the runtime always returns a text response or do we need to build some special handling around other responses (e.g. binary)?

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
```diff
@@ -0,0 +1,78 @@
...
+      var sessionId = responseBytes.response().runtimeSessionId();
+      var statusCode = responseBytes.response().statusCode();
+
+      Object response = tryParseJson(responseText);
```
maff: 💭 Should we only try to parse JSON when the `accept` config is actually `application/json` or a compatible MIME type (think for example `application/schema+json`)?

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/model/request/AgentCoreRuntimeInput.java
tshepomaredi: Done

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/model/request/AgentCoreRuntimeInput.java
tshepomaredi: Removed

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
tshepomaredi: Good point. For now both use the same value since the AgentCore Runtime API typically mirrors the content type. If we see a use case for split request/response types, we can add a separate `acceptContentType` field in a follow-up.

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
tshepomaredi: The AgentCore Runtime API returns the response as bytes, and we read it as UTF-8 text. For the current use case (invoking deployed agents), responses are always text-based (JSON or plain text). If binary response support is needed in the future, we can add Document-based handling as a follow-up? WDYT?

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
tshepomaredi: Done

connectors/aws/aws-bedrock-agentcore-runtime/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+              <files>
+                <file>
+                  <templateId>io.camunda.connectors.aws.bedrock.agentcore.runtime.v1</templateId>
+                  <templateFileName>aws-bedrock-agentcore-runtime-outbound-connector.json</templateFileName>
+                </file>
+              </files>
+              <generateHybridTemplates>true</generateHybridTemplates>
+            </connector>
+          </connectors>
+          <versionHistoryEnabled>true</versionHistoryEnabled>
```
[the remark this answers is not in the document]
maff: This is invalid, versioned template will only be created when bumping the version.

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
maff: Would it make sense to generally fix the content type to `application/json` and remove the configuration options for the moment? So we'd be sure the output is JSON and we're able to pass primitives (a simple string is valid JSON) and complex structures as input. In that case we could also be sure that the `asUtf8String` is actually a string and we don't need to deal with binary data. WDYT?

connectors/aws/aws-bedrock-agentcore-runtime/src/test/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutorTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @Test
+  void shouldInvokeAgentAndReturnParsedJsonResponse() {
+    mockResponse("{\"result\": \"Fraud risk is LOW\"}", SESSION_ID, 200);
+    var result = executor.invoke(createInput(Map.of("inputText", PROMPT), null));
+
+    assertThat(result.response()).isInstanceOf(Map.class);
+    @SuppressWarnings("unchecked")
+    var responseMap = (Map<String, Object>) result.response();
+    assertThat(responseMap).containsEntry("result", "Fraud risk is LOW");
```
maff: Could be a shortened a bit to get rid of the suppress warnings:

```suggestion
    assertThat(result.response())
        .asInstanceOf(InstanceOfAssertFactories.map(String.class, Object.class))
        .containsExactly(entry("result", "Fraud risk is LOW"));
```

connectors/aws/aws-bedrock-agentcore-runtime/src/test/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutorTest.java
```diff
@@ -0,0 +1,145 @@
...
+  void shouldSerializePayloadWithObjectMapper() {
+    mockResponse("{}", SESSION_ID, 200);
+    var captor = ArgumentCaptor.forClass(InvokeAgentRuntimeRequest.class);
+
+    executor.invoke(createInput(Map.of("inputText", "Hello \"world\""), null));
+
+    verify(client).invokeAgentRuntimeAsBytes(captor.capture());
+    var payloadJson = captor.getValue().payload().asUtf8String();
+    assertThat(payloadJson).contains("inputText");
+    assertThat(payloadJson).contains("Hello \\\"world\\\"");
```
maff: 🔧 Can we use `JSONAssert` on the value here?

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
tshepomaredi: That was the initial implementation but I updated this after this comment https://github.com/camunda/connectors/pull/6909#discussion_r3080141740 to align with supporting different content types. Happy to remove it though until the use case emerges, I do think it would be cleaner that way.

maff: Ok, then lets remove it and stick to JSON as it simplifies a couple of things 👍

connectors/aws/aws-bedrock-agentcore-runtime/src/test/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutorTest.java
tshepomaredi: Done

connectors/aws/aws-bedrock-agentcore-runtime/src/main/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutor.java
tshepomaredi: Done

connectors/aws/aws-bedrock-agentcore-runtime/src/test/java/io/camunda/connector/aws/bedrock/agentcore/runtime/AgentCoreRuntimeExecutorTest.java
tshepomaredi: Done

maff (approved):
LGTM! 🚀

nikonovd (approved):
Good work! 🚀

gbetances089 (approved):

Szik (approved):

ztefanie (approved):

johnBgood (approved):

PR #6912, opened 2026-04-08
feat(bpmnError function): Accept only one argument

### LINKED ISSUE #6734: bpmnError with invalid input in connector errorExpression is silently ignored instead of failing
When having an error expression with `bpmnError("one")`, this is invalid, because the function takes two parameters. But when executed, the user does not get any information about this misconfiguration. 

Should we give more clearer feedback to the user about that?

ztefanie:
## Description

bpmnError function now also works with only one param.

Docs PR is here: https://github.com/camunda/camunda-docs/pull/8510

## Related issues

closes https://github.com/camunda/connectors/issues/6734

## Checklist


johnBgood (approved):
nice one!

PR #6989, opened 2026-04-17
feat: add /outbound endpoint to get information about our outbound connectors

johnBgood:
## Add `/outbound` endpoint to expose outbound connector state

  ### Summary

  - Adds `GET /outbound` and `GET /outbound/{type}` endpoints to the connector runtime, mirroring the existing `/inbound` API
  - Each entry reports connector metadata (name, job type, input variables, timeout, enabled state) enriched with live job-stream connectivity observed from the Zeebe gateway and broker actuators
  - Supports multi-instance deployments via the same `X-Camunda-Forwarded-For` forwarding and reduction pattern used by inbound endpoints

  ### API structure

  `GET /outbound` returns a list of `OutboundConnectorResponse`:

  ```json
  [
    {
      "name": "HTTP REST",
      "type": "io.camunda:http-json:1",
      "inputVariables": ["url", "method", "headers"],
      "timeout": 300000,
      "enabled": true,
      "runtimeId": "pod-abc123",
      "gatewayConnectivityState": "CONNECTED",
      "brokerConnectivityState": "ALL_CONNECTED",
      "streamIds": ["stream-1a2b"]
    }
  ]
  ```

  **`gatewayConnectivityState`** (always present when a `GatewayJobStreamClient` bean is configured):

  | Value | Meaning |
  |---|---|
  | `UNKNOWN` | No `GatewayJobStreamClient` configured — stream state cannot be determined |
  | `UNREACHABLE` | Gateway monitoring endpoint is down or returned an error |
  | `NOT_CONNECTED` | Gateway is reachable but no client stream exists for this job type |
  | `CONNECTED` | Client stream registered on the gateway — see `brokerConnectivityState` |

  **`brokerConnectivityState`** (only present when `gatewayConnectivityState` is `CONNECTED`):

  | Value | Meaning |
  |---|---|
  | `NONE` | No broker has this connector's stream as a consumer |
  | `PARTIALLY_CONNECTED` | Stream exists on the gateway but not propagated to all brokers |
  | `ALL_CONNECTED` | Stream is visible as a consumer on all broker remote streams |

  > `gatewayConnectivityState`, `brokerConnectivityState`, and `streamIds` are omitted when no `GatewayJobStreamClient` is configured, keeping the response backward-compatible for runtimes that don't expose the gateway actuator.

  ### Test plan

  - [x] `GET /outbound` returns all registered connectors
  - [x] `GET /outbound/{type}` returns matching connectors; 404 when type is unknown
  - [x] `gatewayConnectivityState` / `brokerConnectivityState` are omitted when no gateway client is configured
  - [x] Multi-instance: responses are correctly forwarded and reduced across runtime pods

## Related issues

closes #

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/common/AbstractConnectorFactory.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/common/AbstractConnectorFactory.java
@@ -20,7 +20,11 @@ import io.camunda.connector.runtime.core.ConnectorFactory;
 import io.camunda.connector.runtime.core.config.ConnectorConfiguration;
 import io.camunda.connector.runtime.core.discovery.DisabledConnectorEnvVarsConfig;
 import java.text.MessageFormat;
-import java.util.*;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 /**
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/common/AbstractConnectorFactory.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/common/AbstractConnectorFactory.java
@@ -37,7 +37,11 @@ public abstract class AbstractConnectorFactory<T, C extends ConnectorConfigurati
       new DisabledConnectorEnvVarsConfig();
 
   public record ConnectorRuntimeConfiguration<T extends ConnectorConfiguration>(
-      T config, boolean isActive) {}
+      T config, boolean isActive) {
+    public Optional<T> getActiveConfig() {
+      return isActive ? Optional.of(config) : Optional.empty();
+    }
+  }
 
   private Map<String, ConnectorRuntimeConfiguration<C>> configurations = new HashMap<>();
 
@@ -49,15 +53,13 @@ public abstract class AbstractConnectorFactory<T, C extends ConnectorConfigurati
   @Override
   public List<C> getActiveConfigurations() {
     return configurations.values().stream()
-        .filter(ConnectorRuntimeConfiguration::isActive)
-        .map(ConnectorRuntimeConfiguration::config)
+        .flatMap(configuration -> configuration.getActiveConfig().stream())
         .collect(Collectors.toList());
   }
 
   protected Optional<C> getActiveConfiguration(String type) {
     return Optional.ofNullable(configurations.get(type))
-        .filter(ConnectorRuntimeConfiguration::isActive)
-        .map(ConnectorRuntimeConfiguration::config);
+        .flatMap(ConnectorRuntimeConfiguration::getActiveConfig);
   }
 
   /**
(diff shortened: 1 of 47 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/common/AbstractConnectorFactory.java
```diff
@@ -41,24 +37,27 @@ public abstract class AbstractConnectorFactory<T, C extends ConnectorConfigurati
       new DisabledConnectorEnvVarsConfig();
 
   public record ConnectorRuntimeConfiguration<T extends ConnectorConfiguration>(
-      T config, boolean isActive) {
-    public Optional<T> getActiveConfig() {
```
johnBgood, line 45: this was unnecessary as we can just `filter()` when using `stream()`

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableQueryService.java
```diff
@@ -66,7 +66,7 @@ public List<ActiveExecutableResponse> query(ActiveExecutableQuery query) {
    */
   public String getConnectorName(String type) {
     try {
-      return connectorFactory.getConfigurations().stream()
+      return connectorFactory.getActiveConfigurations().stream()
```
johnBgood, line 69: @copilot Create a follow up issue (don't implement anything) about updating InboundExecutableQueryService to also use `getRuntimeConfigurations()`. We will then need to adapt some more code, but this way we will see the inbound connectors state (enabled/disabled).

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/controller/OutboundConnectorsRestController.java
```diff
@@ -0,0 +1,77 @@
...
+public class OutboundConnectorsRestController {
+
+  private final InstanceForwardingRouter instanceForwardingRouter;
+  private final OutboundConnectorsService outboundConnectorsService;
+
+  @Value("${camunda.connector.hostname:${HOSTNAME:localhost}}")
+  private String hostname;
+
+  public OutboundConnectorsRestController(
+      @Autowired(required = false) InstanceForwardingRouter instanceForwardingRouter,
```
ztefanie: So the router can be null and when we call one of the methods below we will get a null pointer exception, right? As the InboundConnectorsAutoConfiguration creates a InstanceForwardingConfiguration bean, this will work in most setups, but if the inbound is disabled/excluded, the outbound will fail with NPE as the router will be null as there is no bean created for it.

ztefanie: Ok I just saw Copilot mentioned this and I see you resolved it by also importing InstanceForwardingConfiguration from OutboundConnectorsAutoConfiguration. That covers the starter-based setup, but if someone pulls in OutboundConnectorRuntimeConfiguration directly without going through the auto-config, the router bean is still absent and we'd NPE on the first request. Why not just make the dependency required here instead?

ztefanie (approved):
Nice improvment 🚀 

Looks all good to me. Just to make sure i get this right

We need the Reducer for Outbound as it is possible that different nodes handle different outbound types, whereas we do not need this for inbound, as there all nodes have the same "complete" store?

Just one minor inconsistency i found, up to you if you want to fix it:
`/inbound/nothing-known` 
returns 404 without a message
`/outbound/nothing-known`
returns 404 with message: ""Data of type 'OutboundConnectorResponse' with id 'nothing-known' not found""

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/controller/OutboundConnectorsRestController.java
johnBgood: I plan to keep it as optional (raise an exception if some related variables are used), and fallback to the local calls if `null`.

johnBgood:
@ztefanie I added some changes, and reworked the configuration part entirely, it should be cleaner now :)

Szik (approved):

johnBgood:
@copilot run `mvn spotless:apply` and commit please

PR #6606, opened 2026-03-05
refactor(agentic-ai): consolidate AI Agent job worker into Connectors SDK OutboundConnectorFunction

maff:
## Description

**Based on:** https://github.com/camunda/connectors/pull/6781

Consolidates the AI Agent job worker into the Connectors SDK `OutboundConnectorFunction` pattern using the `ConnectorResponse` sealed hierarchy from #6781.

- `AiAgentJobWorker` becomes a plain `@OutboundConnector` returning `AiAgentSubProcessResponse` (implements `AdHocSubProcessConnectorResponse`)
- Deleted custom handler infrastructure (`AiAgentJobWorkerHandlerImpl`, `AiAgentJobWorkerValueCustomizer`, `JobWorkerAgentExecutionContextFactory`, etc.)
- Replaced `AgentJobContext` with SDK's `JobContext`
- Removed `completionErrorHandler` callback — the SDK does not yet support job completion callbacks. `ConversationStore.compensateFailedJobCompletion()` is kept as a placeholder for a future follow-up

## Related issues

closes #6604

## Checklist

- [ ] Backport labels are added if these code changes should be backported.
- [x] Tests/Integration tests for the changes have been added if applicable.

PR #6994, opened 2026-04-20
feat: update Camunda Client API to 8.10 and support job completion listeners

### LINKED ISSUE #7020: Support job completion callbacks for conversation store
## What should we do?

Add job completion awareness to the conversation storage SPI so that store implementations can react to job completion outcomes.

### Prerequisites

This builds on camunda/camunda#49396, which redesigns the Camunda Client's `CommandWrapper` into a callback-aware `JobCallbackCommandWrapperFactory` with observable `CommandOutcome` results (`Completed`, `Failed`, `Ignored`). This is what makes it possible for the connector runtime to know whether a Zeebe command was accepted or rejected.

### Connector SDK changes

- Update `version.camunda` to 8.10.0 to consume the new `JobCallbackCommandWrapperFactory` / `CommandOutcome` APIs
- Migrate `SpringConnectorJobHandler` from the old `CommandWrapper`/`CommandExceptionHandlingStrategy` to the new callback-aware command APIs
- Introduce `JobCompletionListener` and `JobCompletionFailure` interfaces in the connector SDK so that connector responses can optionally receive notification after Zeebe command dispatch

### Agentic AI changes

- Add `onJobCompleted` and `onJobCompletionFailed` hooks to `ConversationStore` (default no-ops)
- AI Agent responses implement `JobCompletionListener` and delegate to the store's hooks with the relevant `AgentContext`
- Store implementations use these hooks to act on completion outcomes — e.g., orphan cleanup on failure, or pushing tracing/observability data on success

Depends on #7019 (the redesigned SPI).

## Why should we do it?

Conversation messages are persisted to external storage *before* the job completion command is sent to Zeebe. Today there is no mechanism for a store to know the outcome of that completion. This means:

- **On failure**: If Zeebe rejects the completion (job superseded, network error, etc.), the persisted state becomes orphaned — it exists in the store but is never referenced by the process. Orphaned documents/branches accumulate silently.
- **On success**: There is no hook to perform post-commit actions like pushing tracing data or confirming state to external systems.

Completion callbacks close this gap by giving store implementations a way to react to both outcomes.

maff:
## Description

Based on #6784

Adds job completion callbacks to the connector SDK and wires them through the runtime into the agentic-ai conversation store SPI.  Depends on camunda/camunda#49396 for the `JobCallbackCommandWrapperFactory` / `CommandOutcome` APIs.

**SDK + Runtime:**
- Bump `version.camunda` to `8.10.0-SNAPSHOT` for `JobCallbackCommandWrapperFactory` / `CommandOutcome` APIs
- Migrate `SpringConnectorJobHandler` from `CommandWrapper`/`CommandExceptionHandlingStrategy` to the new callback-aware command APIs
- Add `JobCompletionListener` and `JobCompletionFailure` interfaces to the connector SDK — `OutboundConnectorFunction` implementations can optionally implement `JobCompletionListener` to receive lifecycle notifications after Zeebe command dispatch
- `JobCompletionFailure` is a sealed hierarchy: `CommandFailure` (sub-sealed; Zeebe-side rejections — `CommandFailed` / `CommandIgnored`), `ExecutionFailed` (connector function exceptions, error-expression evaluation failures, runtime-synthesized rejections), `BpmnErrorThrown`, `JobErrorRaised`. The latter three carry an optional `CommandFailure` to surface the case where the failJob/throwBpmnError command sent in response was itself rejected
- `SpringConnectorJobHandler` notifies all listener callbacks asynchronously after the relevant Zeebe command future resolves (completeJob, throwBpmnError, failJob)
- Unit tests cover BPMN error, job error, IgnoreError, completeJob outcomes (success/Failed/Ignored), pre-response failures, and command-failure-on-response paths

**Agentic AI:**
- `AiAgentFunction` and `AiAgentJobWorker` implement `JobCompletionListener` (via the shared `AgentConnectorFunction` mixin) and delegate to an internal `AgentJobCompletionListener` carried by the response. The internal listener captures per-execution state (conversation store + agent context) at response build time, so callbacks fire with full context once Zeebe resolves
- Add `onJobCompleted(AgentExecutionContext, AgentContext)` and `onJobCompletionFailed(AgentExecutionContext, AgentContext, JobCompletionFailure)` default hooks to `ConversationStore` — the execution context is provided so implementations can create temporary sessions for compensation if needed
- Implement `onJobCompletionFailed` on `CamundaDocumentConversationStore` to delete the orphaned document written by `storeMessages` when job completion is rejected by Zeebe
- Unit tests for response delegation, function-level delegation, and document store orphan cleanup
- Clean up `TestConfig` (remove unused beans from custom job handler era) and unused test dependencies

**Agentic AI E2E:**
- Add e2e tests for both job worker and outbound connector flavors verifying orphaned document cleanup after job completion failure (fails the job via Zeebe API during execution, then verifies the exact document created by `storeMessages` is deleted by `onJobCompletionFailed`)

## Related issues

closes #7020

Based on #6784. Depends on camunda/camunda#49396 for the `JobCallbackCommandWrapperFactory` / `CommandOutcome` APIs.

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
@@ -37,6 +37,9 @@ import io.camunda.connector.api.outbound.ConnectorResponse.AdHocSubProcessConnec
 import io.camunda.connector.api.outbound.ConnectorResponse.AdHocSubProcessConnectorResponse.ElementActivation;
 import io.camunda.connector.api.outbound.ConnectorResponse.StandardConnectorResponse;
 import io.camunda.connector.api.outbound.JobCompletionFailure;
+import io.camunda.connector.api.outbound.JobCompletionFailure.BpmnErrorThrown;
+import io.camunda.connector.api.outbound.JobCompletionFailure.CommandFailure;
+import io.camunda.connector.api.outbound.JobCompletionFailure.JobErrorRaised;
 import io.camunda.connector.api.outbound.JobCompletionListener;
 import io.camunda.connector.api.outbound.OutboundConnectorContext;
 import io.camunda.connector.api.outbound.OutboundConnectorFunction;
@@ -61,6 +64,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -205,9 +209,7 @@ public class SpringConnectorJobHandler implements JobHandler {
           () -> handleFinalResult(client, job, context, finalResult, counterMetricsContext));
     } catch (Exception ex) {
       notifyJobCompletionFailed(
-          context,
-          connectorResponseOrNull(finalResult),
-          new JobCompletionFailure.CommandFailure.CommandFailed(ex));
+          context, connectorResponseOrNull(finalResult), new CommandFailure.CommandFailed(ex));
       failJob(
           client,
           job,
@@ -233,12 +235,11 @@ public class SpringConnectorJobHandler implements JobHandler {
           JobForLog.from(job),
           errorResult.exception().getMessage(),
           errorResult.exception());
+
       // pre-response failure path: function threw before returning a response, so notify with a
       // null response (subscribers to JobCompletionListener can still react)
       notifyJobCompletionFailed(
-          context,
-          null,
-          new JobCompletionFailure.CommandFailure.CommandFailed(errorResult.exception()));
+          context, null, new CommandFailure.CommandFailed(errorResult.exception()));
       failJob(jobClient, job, errorResult, counterMetricsContext);
     }
   }
@@ -256,23 +257,21 @@ public class SpringConnectorJobHandler implements JobHandler {
       case BpmnError bpmnError -> {
         LOGGER.debug(
             "Throwing BPMN error for job {} with code {}", job.getKey(), bpmnError.errorCode());
-        var future = throwBpmnError(client, job, bpmnError, counterMetricsContext);
-        if (call instanceof JobCompletionListener) {
-          future.whenComplete(
-              (outcome, throwable) ->
-                  notifyJobCompletionFailed(
-                      context,
-                      response,
-                      new JobCompletionFailure.BpmnErrorThrown(
-                          bpmnError.errorCode(),
-                          bpmnError.errorMessage(),
-                          bpmnError.variables(),
-                          toCommandFailure(outcome, throwable))));
-        }
+        var bpmnErrorFuture = throwBpmnError(client, job, bpmnError, counterMetricsContext);
+        notifyFailureOnCommandOutcome(
+            bpmnErrorFuture,
+            context,
+            response,
+            completionFailure ->
+                new BpmnErrorThrown(
+                    bpmnError.errorCode(),
+                    bpmnError.errorMessage(),
+                    bpmnError.variables(),
+                    completionFailure));
       }
       case JobError jobError -> {
         LOGGER.debug("Throwing incident for job {}", job.getKey());
-        var future =
+        var failJobFuture =
             failJob(
                 client,
                 job,
@@ -282,52 +281,58 @@ public class SpringConnectorJobHandler implements JobHandler {
                     jobError.retries(),
                     jobError.retryBackoff()),
                 counterMetricsContext);
-        if (call instanceof JobCompletionListener) {
-          future.whenComplete(
-              (outcome, throwable) ->
-                  notifyJobCompletionFailed(
-                      context,
-                      response,
-                      new JobCompletionFailure.JobErrorRaised(
-                          jobError.errorMessage(),
-                          jobError.variablesWithErrorMessage(),
-                          toCommandFailure(outcome, throwable))));
-        }
+        notifyFailureOnCommandOutcome(
+            failJobFuture,
+            context,
+            response,
+            completionFailure ->
+                new JobErrorRaised(
+                    jobError.errorMessage(),
+                    jobError.variablesWithErrorMessage(),
+                    completionFailure));
       }
-      case IgnoreError ignoreError -> {
-        if (finalResult instanceof ConnectorResult.SuccessResult successResult
-            && successResult.connectorResponse() instanceof AdHocSubProcessConnectorResponse) {
-          LOGGER.debug(
-              "IgnoreError not supported for AdHocSubProcessConnectorResponse, job {}",
-              job.getKey());
-          var cause =
-              new UnsupportedOperationException("IgnoreError is not supported for this connector");
-          var future =
-              failJob(
-                  client,
-                  job,
-                  new ConnectorResult.ErrorResult(ignoreError.variables(), cause, 0, null),
-                  counterMetricsContext);
-          if (call instanceof JobCompletionListener) {
-            // listener-relevant failure is the IgnoreError misuse, not the failJob outcome
-            future.whenComplete(
-                (outcome, throwable) ->
-                    notifyJobCompletionFailed(
-                        context,
-                        response,
-                        new JobCompletionFailure.CommandFailure.CommandFailed(cause)));
-          }
-        } else {
-          LOGGER.debug("Ignoring error for job {}", job.getKey());
-          completeJob(
+      case IgnoreError ignoreError ->
+          handleIgnoreError(
+              client, job, context, finalResult, response, ignoreError, counterMetricsContext);
+    }
+  }
+
+  private void handleIgnoreError(
+      JobClient client,
+      ActivatedJob job,
+      OutboundConnectorContext context,
+      ConnectorResult finalResult,
+      ConnectorResponse response,
+      IgnoreError ignoreError,
+      CounterMetricsContext counterMetricsContext) {
+    if (finalResult instanceof ConnectorResult.SuccessResult successResult
+        && successResult.connectorResponse() instanceof AdHocSubProcessConnectorResponse) {
+      LOGGER.debug(
+          "IgnoreError not supported for AdHocSubProcessConnectorResponse, job {}", job.getKey());
+      var cause =
+          new UnsupportedOperationException("IgnoreError is not supported for this connector");
+      var failJobFuture =
+          failJob(
               client,
               job,
-              context,
-              new ConnectorResult.SuccessResult(
-                  StandardConnectorResponse.of(null), ignoreError.variables()),
+              new ConnectorResult.ErrorResult(ignoreError.variables(), cause, 0, null),
               counterMetricsContext);
-        }
-      }
+
+      // listener-relevant failure is the IgnoreError misuse, not the failJob outcome
+      notifyFailureOnCommandOutcome(
+          failJobFuture,
+          context,
+          response,
+          completionFailure -> new CommandFailure.CommandFailed(cause));
+    } else {
+      LOGGER.debug("Ignoring error for job {}", job.getKey());
+      completeJob(
+          client,
+          job,
+          context,
+          new ConnectorResult.SuccessResult(
+              StandardConnectorResponse.of(null), ignoreError.variables()),
+          counterMetricsContext);
     }
   }
 
@@ -417,18 +422,14 @@ public class SpringConnectorJobHandler implements JobHandler {
               prepareAdHocSubProcessCompleteJobCommand(client, job, ahsp);
         };
 
-    var future =
+    var completeJobFuture =
         jobCallbackCommandWrapperFactory
             .create(command, job.getDeadline(), counterMetricsContext, MAX_ZEEBE_COMMAND_RETRIES)
             .executeAsync();
 
-    if (call instanceof JobCompletionListener) {
-      future.whenComplete(
-          (outcome, error) ->
-              notifyJobCompletionOutcome(context, connectorResponse, outcome, error));
-    }
+    notifyOnCommandOutcome(completeJobFuture, context, connectorResponse);
 
-    return future;
+    return completeJobFuture;
   }
 
   private static JobCallbackFinalCommandStep<CompleteJobResponse> prepareCompleteJobCommand(
@@ -477,42 +478,13 @@ public class SpringConnectorJobHandler implements JobHandler {
         : null;
   }
 
-  /**
-   * Translates an internal {@link CommandOutcome} into the SDK-level {@link
-   * JobCompletionFailure.CommandFailure}. Returns {@code null} for a successful outcome so it can
-   * be embedded directly in {@link JobCompletionFailure.BpmnErrorThrown} / {@link
-   * JobCompletionFailure.JobErrorRaised}.
-   */
-  private static JobCompletionFailure.CommandFailure toCommandFailure(
-      CommandOutcome outcome, Throwable throwable) {
-    if (throwable != null) {
-      return new JobCompletionFailure.CommandFailure.CommandFailed(throwable);
-    }
-    return switch (outcome) {
-      case CommandOutcome.Completed c -> null;
-      case CommandOutcome.Failed f ->
-          new JobCompletionFailure.CommandFailure.CommandFailed(f.cause());
-      case CommandOutcome.Ignored i ->
-          new JobCompletionFailure.CommandFailure.CommandIgnored(i.cause());
-    };
-  }
-
-  private void notifyJobCompletionOutcome(
-      OutboundConnectorContext context,
-      ConnectorResponse response,
-      CommandOutcome outcome,
-      Throwable error) {
+  private void notifyJobCompleted(OutboundConnectorContext context, ConnectorResponse response) {
     if (!(call instanceof JobCompletionListener listener)) {
       return;
     }
 
     try {
-      var commandFailure = toCommandFailure(outcome, error);
-      if (commandFailure == null) {
-        listener.onJobCompleted(context, response);
-      } else {
-        listener.onJobCompletionFailed(context, response, commandFailure);
-      }
+      listener.onJobCompleted(context, response);
     } catch (Exception e) {
       LOGGER.warn("JobCompletionListener callback failed", e);
     }
@@ -529,5 +501,64 @@ public class SpringConnectorJobHandler implements JobHandler {
     } catch (Exception e) {
       LOGGER.warn("JobCompletionListener callback failed", e);
     }
+  }
+
+  /**
+   * Dispatches the listener once a Zeebe command future resolves: success → {@link
+   * #notifyJobCompleted}, failure (rejection or future exception) → {@link
+   * #notifyJobCompletionFailed} with the {@link CommandFailure} derived from the outcome.
+   *
+   * <p>Used by the {@code completeJob} path, where success and failure are both possible.
+   */
+  private void notifyOnCommandOutcome(
+      CompletableFuture<CommandOutcome> future,
+      OutboundConnectorContext context,
+      ConnectorResponse response) {
+    future.whenComplete(
+        (outcome, throwable) -> {
+          var commandFailure = toCommandFailure(outcome, throwable);
+          if (commandFailure == null) {
+            notifyJobCompleted(context, response);
+          } else {
+            notifyJobCompletionFailed(context, response, commandFailure);
+          }
+        });
+  }
+
+  /**
+   * Dispatches a {@link #notifyJobCompletionFailed} once a Zeebe command future resolves, building
+   * the failure via {@code failureBuilder} from the resolved {@link CommandFailure} (which is
+   * {@code null} when the command was accepted).
+   *
+   * <p>Used by paths that always end in failure regardless of the command outcome (BPMN error, job
+   * error, IgnoreError misuse) — the builder produces the appropriate {@link JobCompletionFailure}
+   * subtype and embeds the command outcome where applicable.
+   */
+  private void notifyFailureOnCommandOutcome(
+      CompletableFuture<CommandOutcome> future,
+      OutboundConnectorContext context,
+      ConnectorResponse response,
+      Function<CommandFailure, JobCompletionFailure> failureBuilder) {
+    future.whenComplete(
+        (outcome, throwable) ->
+            notifyJobCompletionFailed(
+                context, response, failureBuilder.apply(toCommandFailure(outcome, throwable))));
+  }
+
+  /**
+   * Translates an internal {@link CommandOutcome} into the SDK-level {@link CommandFailure}.
+   * Returns {@code null} for a successful outcome so it can be embedded directly in {@link
+   * BpmnErrorThrown} / {@link JobErrorRaised}.
+   */
+  private static CommandFailure toCommandFailure(CommandOutcome outcome, Throwable throwable) {
+    if (throwable != null) {
+      return new CommandFailure.CommandFailed(throwable);
+    }
+
+    return switch (outcome) {
+      case CommandOutcome.Completed c -> null;
+      case CommandOutcome.Failed f -> new CommandFailure.CommandFailed(f.cause());
+      case CommandOutcome.Ignored i -> new CommandFailure.CommandIgnored(i.cause());
+    };
   }
 }
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
@@ -207,7 +207,7 @@ public class SpringConnectorJobHandler implements JobHandler {
       notifyJobCompletionFailed(
           context,
           connectorResponseOrNull(finalResult),
-          new JobCompletionFailure.CommandFailed(ex));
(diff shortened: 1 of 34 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -19,15 +19,15 @@
...
 import io.camunda.client.api.response.ActivatedJob;
 import io.camunda.client.api.response.CompleteJobResponse;
 import io.camunda.client.api.response.FailJobResponse;
 import io.camunda.client.api.response.ThrowErrorResponse;
 import io.camunda.client.api.worker.JobClient;
 import io.camunda.client.api.worker.JobHandler;
-import io.camunda.client.jobhandling.CommandExceptionHandlingStrategy;
-import io.camunda.client.jobhandling.CommandWrapper;
+import io.camunda.client.jobhandling.CommandOutcome;
+import io.camunda.client.jobhandling.JobCallbackCommandWrapperFactory;
```
maff, line 30: ℹ️ This is the changed API from https://github.com/camunda/camunda/pull/49396

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/autoconfigure/TestConfig.java
```diff
@@ -41,24 +36,4 @@ public DocumentFactory documentFactory() {
...
-
-  @Bean
-  public CommandExceptionHandlingStrategy commandExceptionHandlingStrategy() {
-    return mock(CommandExceptionHandlingStrategy.class);
-  }
-
-  @Bean
-  public MetricsRecorder metricsRecorder() {
-    return new DefaultNoopMetricsRecorder();
-  }
```
maff, line 63: ℹ️ Leftover from https://github.com/camunda/connectors/pull/6606

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/AiAgentTaskResponse.java
```diff
@@ -0,0 +1,41 @@
...
+import io.camunda.connector.api.outbound.ConnectorResponse.StandardConnectorResponse;
+import io.camunda.connector.api.outbound.JobCompletionFailure;
+import io.camunda.connector.api.outbound.JobCompletionListener;
+import org.springframework.lang.Nullable;
+
+/**
+ * Response type for the AI Agent task flavor (outbound connector on a service task). Delegates job
+ * completion callbacks to the provided listener.
+ */
+public record AiAgentTaskResponse(
```
maff: ℹ️ New wrapper type implementing `StandardConnectorResponse` instead of directly returning the response payload.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/ConversationStore.java
```diff
@@ -32,4 +33,16 @@ public interface ConversationStore {
    */
   ConversationSession createSession(
       AgentExecutionContext executionContext, AgentContext agentContext);
+
+  /**
+   * Best-effort: the job carrying the given agentContext was accepted by Zeebe. Implementations may
+   * use this to clean up previous state (old documents, orphaned branches).
+   */
+  default void onJobCompleted(AgentContext committedContext) {}
```
maff: 🔧 These event handlers should get the same data as `createSession`, so they are able to potentially open a dedicated session for clean-up/compensation.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
```diff
@@ -1687,4 +1691,61 @@ void nullElementActivationsTreatedAsEmptyList() throws Exception {
       assertThat(result.elementActivations()).isEmpty();
     }
   }
+
+  @Nested
+  class JobCompletionListenerTests {
```
maff, line 1725: 🔧 Missing tests for normal command outcome (successfully completed, ignored, failed, exception thrown).

connectors-e2e-test/connectors-e2e-test-csv/src/test/java/io/camunda/connector/e2e/csv/CsvConnectorTests.java
```diff
@@ -48,7 +46,6 @@
 @CamundaSpringProcessTest
 @SlowTest
 @ExtendWith(MockitoExtension.class)
-@DirtiesContext(classMode = ClassMode.AFTER_EACH_TEST_METHOD)
```
maff: ℹ️ `@DirtiesContext` [currently clashes with 8.10 CPT](https://github.com/camunda/camunda/issues/51444), but these tests seem to also work without it.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
```diff
@@ -1687,4 +1694,161 @@ void nullElementActivationsTreatedAsEmptyList() throws Exception {
...
+          .withErrorExpressionHeader(
+              "=if response.status = \"fail\" then bpmnError(\"ERR_001\", \"test error\") else null")
+          .executeAndCaptureResult(handler, false, true);
+
+      var captor = ArgumentCaptor.forClass(JobCompletionFailure.class);
+      verify(listener).onJobCompletionFailed(captor.capture());
+      assertThat(captor.getValue())
+          .isInstanceOfSatisfying(
+              JobCompletionFailure.BpmnErrorThrown.class,
+              failure -> assertThat(failure.errorCode()).isEqualTo("ERR_001"));
```
maff: 🔧 Let's assert all the values passed to `bpmnError` here. Can we also assert variables and code?

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
```diff
@@ -1687,4 +1694,161 @@ void nullElementActivationsTreatedAsEmptyList() throws Exception {
...
+          .withErrorExpressionHeader(
+              "=if response.status = \"fail\" then jobError(\"something went wrong\") else null")
+          .executeAndCaptureResult(handler, false, false);
+
+      var captor = ArgumentCaptor.forClass(JobCompletionFailure.class);
+      verify(listener).onJobCompletionFailed(captor.capture());
+      assertThat(captor.getValue())
+          .isInstanceOfSatisfying(
+              JobCompletionFailure.JobFailed.class,
+              failure -> assertThat(failure.errorMessage()).isEqualTo("something went wrong"));
```
maff: 🔧 Let's assert all the values passed to jobError.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
```diff
@@ -1687,4 +1694,161 @@ void nullElementActivationsTreatedAsEmptyList() throws Exception {
...
+    void noListenerDoesNotCrash() throws Exception {
+      // response that does NOT implement JobCompletionListener
+      var handler =
+          newConnectorJobHandlerWithOutcome(
+              context -> StandardConnectorResponse.of(Map.of("key", "value")),
+              CompletableFuture.completedFuture(new CommandOutcome.Completed(null, 1)));
+
+      // should not throw
+      JobBuilder.create().execute(handler);
+    }
```
maff, line 2114: 🔧 We're missing a test with a listener that does throw (on completion and on error).

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandlerTest.java
```diff
@@ -1687,4 +1694,161 @@ void nullElementActivationsTreatedAsEmptyList() throws Exception {
...
+      when(factory.create(any(), anyLong(), any(), anyInt()).executeAsync()).thenReturn(outcome);
+      return new SpringConnectorJobHandler(
+          metricsRecorder,
+          factory,
+          new SecretProviderAggregator(List.of(new FooBarSecretProvider())),
+          new DefaultValidationProvider(),
+          mock(DocumentFactory.class),
+          TestObjectMapperSupplier.INSTANCE,
+          call);
+    }
```
maff: 🔧 Can we consolidate this with `newConnectorJobHandler`?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -227,20 +238,30 @@ private void handleFinalResult(
...
       ConnectorResult finalResult,
       ConnectorError error,
       CounterMetricsContext counterMetricsContext) {
+    var listener = extractJobCompletionListener(finalResult);
+
     switch (error) {
       case BpmnError bpmnError -> {
         LOGGER.debug(
             "Throwing BPMN error for job {} with code {}", job.getKey(), bpmnError.errorCode());
+        notifyJobCompletionFailed(
```
johnBgood: ❓do we intentionally notify before sending the actual error? We don't what users can do here, maybe it's safer to change the order (same question for the other cases)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -424,4 +437,43 @@ private static String truncateErrorMessage(String message) {
         ? message.substring(0, Math.min(message.length(), MAX_ERROR_MESSAGE_LENGTH))
         : null;
   }
+
+  private static JobCompletionListener extractJobCompletionListener(ConnectorResult result) {
+    if (result instanceof ConnectorResult.SuccessResult successResult
+        && successResult.connectorResponse() instanceof JobCompletionListener listener) {
+      return listener;
+    }
+    return null;
```
johnBgood: ❓If that's intentional, should we thrown an exception here, or add a comment? It's not clear to me if this was done on purpose or not when first reading it

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -227,20 +238,30 @@ private void handleFinalResult(
...
       ConnectorError error,
       CounterMetricsContext counterMetricsContext) {
+    var listener = extractJobCompletionListener(finalResult);
+
     switch (error) {
       case BpmnError bpmnError -> {
         LOGGER.debug(
             "Throwing BPMN error for job {} with code {}", job.getKey(), bpmnError.errorCode());
+        notifyJobCompletionFailed(
+            listener,
```
johnBgood: this is always null given how the `extractJobCompletionListener` is implemented, or did I miss something?

connector-sdk/core/src/main/java/io/camunda/connector/api/outbound/JobCompletionListener.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Callbacks for job completion outcomes. Connector responses can implement this interface to
+ * receive notification after the Zeebe command is dispatched.
+ *
+ * <p>For the completeJob path, callbacks fire asynchronously after the command future resolves. For
+ * failJob and throwBpmnError paths (triggered by error expressions), callbacks fire synchronously
+ * before the command is dispatched, since the outcome is already determined.
+ */
+public interface JobCompletionListener {
```
johnBgood, line 31: ❓ Currently, this is attached to the connector's response, but I guess I would've expected it at the same level of the `execute` method in `OutboundConnectorFunction`. 
wdyt?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -424,4 +437,43 @@ private static String truncateErrorMessage(String message) {
...
+
+  private static JobCompletionListener extractJobCompletionListener(ConnectorResult result) {
+    if (result instanceof ConnectorResult.SuccessResult successResult
+        && successResult.connectorResponse() instanceof JobCompletionListener listener) {
+      return listener;
+    }
+    return null;
+  }
+
+  private static void notifyJobCompletionOutcome(
```
johnBgood: 🔧notifyJobCompletionListener ?

nikonovd: `onJobCompletion()` ?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
maff: The primary use case of the listener setup is to notify listeners if the job completion succeeded or failed with a detail (the `JobCompletionFailure`) when it didn't. As the outcome in this branch can only be that a BPMN error was thrown, the idea here is to notify listeners synchronously before the actual command is sent because they are primarily interested in the fact that the job didn't complete with the failure reason being secondary.

The nuance here is that this command can potentially fail as well, leading to another failure mode ("BPMN error should have been thrown, but the command failed"). We can try to transport this information as well (either with a dedicated `JobCompletionFailure` type or by making it part of `BpmnErrorThrown`/`JobErrorRaised` - in that case we'd subscribe to the future in the same way as in the job completion case and act on the result. WDYT?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
maff: It is intentional as currently the `ConnectorResponse` and `JobCompletionListener` are separate interfaces. In https://github.com/camunda/connectors/pull/6781, we introduced the new `ConnectorResponse` hierarchy which can be returned by a connector. Connectors can either return any object as they always did or return a `ConnectorResponse` to have more influence on the job completion. If no `ConnectorResponse` is returned, it will be automatically wrapped in `StandardConnectorResponse` earlier in this class before further processing.

With this change, the `ConnectorResponse` can additionally implement `JobCompletionListener` in order to opt in to be notified of these events (see for example `AiAgentSubProcessConnectorResponse` which implements both `AdHocSubProcessConnectorResponse` and `JobCompletionListener`).

If we want to simplify this we can also merge the `JobCompletionListener` interface into `ConnectorResponse`, making all connector responses a listener. As the implementations would be no-ops, it shouldn't make a big difference, only the short circuits here on `listener == null` would always be called. Otherwise, yes, we can definitely add a comment to explain this better.

---

Some AI fluff on why this looks as it does right now - one use case for a separate interface is already used in the AI Agent part where the listener is constructed separately and used with delegation in the response:

  The strongest reasons:

  1. **`ConnectorResponse` is sealed** (`connector-sdk/core/.../ConnectorResponse.java:26`) with two `non-sealed` permitted variants. Merging would either (a) force callback methods onto the base — meaning every `StandardConnectorResponse`/`AdHocSubProcessConnectorResponse` impl carries lifecycle concerns it doesn't need, or (b) only attach them to one variant —
   which breaks the symmetry with no real win. `StandardConnectorResponse.of(Object)` works as a lambda factory precisely because the interface stays a single-abstract-method shape; bolting lifecycle onto it makes that brittle.

  2. **Composition over inheritance — and the PR already relies on it.** `BaseAgentRequestHandler.createStoreCompletionListener(...)` builds a standalone `JobCompletionListener` and the response just delegates to it (`AiAgentTaskConnectorResponse` field `completionListener`). That only works because the listener is a free-standing abstraction. If you merged it
  into `ConnectorResponse`, every callback-bearing thing has to *be* a response, which it isn't.

  3. **Orthogonal concerns.** `ConnectorResponse#responseValue()` is "what does the runtime put back into the process?". `JobCompletionListener` is "tell me what happened after dispatch." A given response may or may not care about the latter; making it opt-in via a marker-style interface checked at the dispatch site (`SpringConnectorJobHandler.java:441-447`) is
  the conventional Java pattern for exactly this (cf. `Closeable`, `AutoCloseable`, `Flushable` on streams).

  4. **The `instanceof` check is a dispatch detail, not a coupling.** "Only checked against `ConnectorResponse`" describes *where* we look for it today, not *what it is*. If a future feature wants to register listeners outside the response object (e.g. a connector-level interceptor), the abstraction is already free to be reused.

  The case to merge would be YAGNI — but the cost of merging (sealed-hierarchy contamination, loss of standalone delegation) is higher than the cost of one extra interface. I'd keep the split.

connector-sdk/core/src/main/java/io/camunda/connector/api/outbound/JobCompletionListener.java
maff: You mean something like the following, either directly on the `OutboundConnectorFunction` or the `OutboundConnectorFunction` being able to implement `JobCompletionListener`?

```java
public interface OutboundConnectorFunction {
  Object execute(OutboundConnectorContext context) throws Exception;
  
  void onJobCompleted();
  
  void onJobCompletionFailed(JobCompletionFailure failure);
}
```

The reason why this currently lives on the response is that the implementation might need to refer to data created during the execution, making the response a good fit to a) carry information how to complete and b) to optionally listen on the completion outcome. For example, on this PR the `BaseAgentRequestHandler` wires the used store implementation and execution context into the listener, so it can delegate potential cleanups to the store implementation. Without the data from the current execution, it would need to go all the way of parsing input mappings/initializing data and then potentially being able to derive what to do in the listener.

We could maybe model this in a way having the listeners be part of the `OutboundConnectorFunction` by feeding the respective response and potentially the execution context back into the listener, e.g.:

```java
public interface OutboundConnectorFunction {
  // CAN return ConnectorResponse, is wrapped in StandardConnectorResponse otherwise
  Object execute(OutboundConnectorContext context) throws Exception;

  // receives the ConnectorResponse (or the wrapped one) returned from execute()
  void onJobCompleted(OutboundConnectorContext context, ConnectorResponse connectorResponse);

  void onJobCompletionFailed(
      OutboundConnectorContext context,
      ConnectorResponse connectorResponse,
      JobCompletionFailure failure);
}
```

Would you prefer that alternative? I think that' something we can explore, it just might to lead to a bit more error handling/potential error cases. For example, the AI Agent might only want to deal with `AiAgentTaskConnectorResponse` and `AiAgentSubProcessConnectorResponse` objects in this listener, so it needs to check the type before processing.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
johnBgood: `onJobCompletion` looks like something you would see in the Listener, not the notifier imho

connector-sdk/core/src/main/java/io/camunda/connector/api/outbound/JobCompletionListener.java
johnBgood: My initial gut feeling was exactly what you posted here, where we provide the connectorResponse. I would go for this option, mostly because of what I said about the methods being in the response vs. in the OutboundConnectorFunction.

That said, you have an actual use case, so if you think that doesn't make sense, we can keep the current implementation :)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
johnBgood: Got it, thank you for the explanation. 
Maybe an Optional would make sense then. I'll leave it to you. 

Thanks

connector-sdk/core/src/main/java/io/camunda/connector/api/outbound/JobCompletionListener.java
maff: I think the use case we have also fits this pattern well. I updated it accordingly. `OutboundConnectorFunction`s can now implement `JobCompletionListener` if they are interested in the outcome.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
maff: The changeset is bigger now, but it now implements a more consistent pattern, notifying async after the command. The logical `JobCompletionFailure` cases (such as BPMN error thrown) now contain an additional nullable `CommandFailure` field indicating that the actual Zeebe command failed.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
maff: Fixed with the latest changes to also notify async on BPMN/Job error. Now 2 methods exist to drive the listener (completion/completion failure) + one to handle the future for the cases emitting a "failure" command.

connector-sdk/core/src/main/java/io/camunda/connector/api/outbound/JobCompletionListener.java
johnBgood: Awesome, thank you!

johnBgood (approved):
Looks really nice!

maff:
@Szik @gbetances089 I assigned both of you for QA as these changes concern both the SDK and the AI Agent:

- The SDK should behave as before for all connectors as the listener support is opt-in - if not, this could indicate a regression in the 8.10 Camunda Client which we update to here or in the job handler changes. The e2e test suite passes though.
- The AI Agents should behave as before as well with the small adaption that when using the document storage and the job completion fails, the document will be attempted to be cleaned up/deleted. This is covered by an e2e test.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -196,25 +205,33 @@ private void processFinalResult(
...
+          error ->
+              handleConnectorError(client, job, context, finalResult, error, counterMetricsContext),
+          () -> handleFinalResult(client, job, context, finalResult, counterMetricsContext));
     } catch (Exception ex) {
-      failJob(
-          client,
-          job,
-          this.outboundConnectorExceptionHandler.handleFinalResultException(ex, job),
-          counterMetricsContext);
+      var failJobFuture =
```
nikonovd: ⛏️ instead of naming the variable like that, we could just rely on typing instead on inference in such places 🙂 

```java
Future<?> failedJobRequest = failJob(...);
```

nikonovd (approved):
LGTM, i left a very nitpicky comment but feel free to check yourself 🙂

gbetances089:
The AI agent on Camunda 8.10 with PR [#6994](https://github.com/camunda/connectors/pull/6994) completed successfully end-to-end. Here's the validation summary:

Validation Result: ✅ No regression

Test environment: Camunda 8.10 (chart 15.0.0-alpha0) with connector image registry.camunda.cloud/team-connectors/connectors-bundle:8.10.0-pr6994-run25310024859-a1

BPMN: ai-agent-chat-with-tools_open_ai.bpmn (OpenAI GPT-4o variant)

Execution:

AI agent job received → tool elements resolved from process definition (Agent_Tools ad-hoc subprocess)
First model call: GPT-4o decided to call GetDateAndTime tool (FEEL =now()) → job completed
Tool executed, second AI agent job triggered with tool result
Second model call: GPT-4o synthesized the final answer → job completedThe AI agent on Camunda 8.10 with PR [#6994](https://github.com/camunda/connectors/pull/6994) completed successfully end-to-end. Here's the validation summary:

Validation Result: ✅ No regression

Test environment: Camunda 8.10 (chart 15.0.0-alpha0) with connector image registry.camunda.cloud/team-connectors/connectors-bundle:8.10.0-pr6994-run25310024859-a1

BPMN: ai-agent-chat-with-tools_open_ai.bpmn (OpenAI GPT-4o variant)

Execution:

AI agent job received → tool elements resolved from process definition (Agent_Tools ad-hoc subprocess)
First model call: GPT-4o decided to call GetDateAndTime tool (FEEL =now()) → job completed
Tool executed, second AI agent job triggered with tool result
Second model call: GPT-4o synthesized the final answer → job completed

gbetances089 (approved):

Szik (approved):

PR #6999, opened 2026-04-20
feat(agentic-ai): extract documents from tool call results into user messages

### LINKED ISSUE #7005: Agentic AI: Make Camunda Documents in tool call results visible to LLMs
## Problem

Tool call results (from BPMN tool activities) can contain Camunda Documents — either at the root level or nested within maps and lists. Currently, these documents are serialized as base64-encoded strings embedded within the tool result JSON text. LLMs cannot meaningfully interpret base64 blobs in tool result text, making documents in tool call results effectively invisible to the model.

In contrast, documents provided via user messages already work correctly — they are converted to native multi-modal content blocks (images, PDFs, text) that the model can interpret. This process is handled by the `DocumentToContentConverterImpl`, which inspects the document's content type and converts it to the appropriate LangChain4J content type (`ImageContent`, `PdfFileContent`, or `TextContent`). The resulting content blocks are added to the `UserMessage` and rendered by the LLM provider as native multi-modal input. This proven path does not exist for tool call results.

## Requirements

- Documents in tool call results must be provided to the LLM in a format it can process (native content blocks, not base64 in JSON).
- The solution must work across all supported LLM providers, not just those with native multi-content tool result support.
- The model must be able to correlate which documents belong to which tool call when multiple tool calls return documents.
- Document handling must be visible and auditable in the persisted conversation history.
- The context window (message count limit) should not be unnecessarily inflated by the document delivery mechanism.
- The existing document handling infrastructure (used for user prompt documents) should be reused where possible.
- The stale base64 serialization infrastructure should be removed.

maff:
## Description

Tool call results containing Camunda Documents were serialized as base64 strings embedded in JSON via a custom `DocumentToContentSerializer`. Most LLMs cannot properly interpret this format.

This PR extracts documents from tool call results into a synthetic `UserMessage` with `DocumentContent` blocks appended after the `ToolCallResultMessage`. The tool result text retains document references (serialized by the standard `DocumentSerializer`) so the model can correlate references with actual content.

### Document message format

A single `UserMessage` (metadata: `toolCallDocuments=true`) is appended with:
- A preamble: `"Documents extracted from tool calls (<doc /> tag + content pair):"`
- Per document: a self-closing `<doc />` XML tag immediately followed by its `DocumentContent` block

Tag name and preamble are intentionally terse to minimise token usage while keeping the pair structure (reference tag → content block) explicit. The tag shape is dispatched on the document's reference type and uses attribute names that mirror the JSON fields emitted by the standard `DocumentSerializer`, so the model can correlate a reference in the tool result JSON with its content block 1:1. Attribute values are XML-escaped (via `commons-text` `StringEscapeUtils.escapeXml10`); blank attributes are omitted.

Event messages containing documents receive the same `<doc />` labels (without `toolName` / `toolCallId`), preceded by the equivalent `"Documents extracted from event data (<doc /> tag + content pair):"` preamble.

#### Worked example

User: *"Summarise the quarterly report and fetch the spec sheet by URL."*

The agent calls `download_report` (returns a Camunda doc) and `fetch_spec` (returns a nested `{summary, document}` with an external ref). The messages sent to the LLM are:

```
UserMessage:        "Summarise the quarterly report and fetch the spec sheet by URL."
AssistantMessage:   tool calls [call_1 → download_report, call_2 → fetch_spec]
ToolCallResultMessage(call_1):
    {"camunda.document.type": "camunda", "documentId": "25ece9fa-…", "storeId": "in-memory"}
ToolCallResultMessage(call_2):
    {"summary": "Q4 spec sheet",
     "document": {"camunda.document.type": "external", "url": "https://…/spec.pdf", "name": "spec-sheet"}}
UserMessage (toolCallDocuments=true):
    TextContent:     Documents extracted from tool calls (<doc /> tag + content pair):
    TextContent:     <doc toolName="download_report" toolCallId="call_1" documentId="25ece9fa-…" storeId="in-memory" contentType="application/pdf" fileName="report.pdf" />
    DocumentContent: [report.pdf bytes]
    TextContent:     <doc toolName="fetch_spec" toolCallId="call_2" url="https://…/spec.pdf" name="spec-sheet" />
    DocumentContent: [spec.pdf bytes]
```

External refs deliberately omit `contentType` / `fileName` because `Document.metadata()` on an external doc reads HTTP response headers — surfacing those fields would trigger a network fetch just to render a label. Camunda refs carry them because metadata is in-memory on resolve.

### Document extraction architecture

Document extraction is driven by `ToolCallResultDocumentExtractor`, called from `AgentMessagesHandlerImpl` after the `ToolCallResultMessage` is built. For each result it asks `GatewayToolHandlerRegistry.handlerForToolDefinition(toolName)` for the responsible handler:

- if a handler manages the tool → delegate to `GatewayToolHandler.extractDocuments(ToolCallResult)` — handlers walk their own typed content (sealed-type `switch` over `McpContent` / `A2aSendMessageResult`);
- otherwise → fall back to `ContentTreeDocumentWalker.extractDocumentsFromContent(...)`, a stateless static utility that recursively walks `Map`, `Collection`, `Object[]` and `Document` nodes (used for plain BPMN tools whose content is a raw FEEL tree).

The default `GatewayToolHandler.extractDocuments` implementation also delegates to the walker, so third-party handlers that return raw maps work without overriding. `ContentTreeDocumentWalker` is public so handlers whose typed content embeds raw user-generated subtrees can call it directly on those subtrees.

Documents are de-duplicated per tool call (same reference referenced multiple times collapses to one content block).

### Gateway tool handlers

Gateway handlers (MCP, A2A) restore typed content as `ToolCallResult.content()` (the pre-PR shape: `List<McpContent>` for MCP, `A2aSendMessageResult` for A2A) and override `extractDocuments` to walk that typed structure:

- **MCP**: collects documents from `McpDocumentContent` and `McpEmbeddedResourceContent.BlobDocumentResource` via a sealed-type switch over `McpContent`.
- **A2A**: walks `A2aMessage.contents` at the root, plus `A2aTask.artifacts` and (recursively) `A2aTask.history` to collect `DocumentContent` entries.

This removes the earlier requirement that handlers preserve raw `Map`/`List`/`Document` content trees just so the document extractor's `instanceof` walk could find nested `Document` instances.

### Message window memory

The synthetic document `UserMessage` does not count toward the `maxMessages` context window limit and is evicted together with its associated `ToolCallResultMessage`.

### Cross-provider viability test

A manual cross-provider integration test (`DocumentToolCallResultsIT`, `@Disabled` by default) validates that real LLM providers can receive and reason about PDF documents extracted from tool call results. It covers single documents, multiple documents, and nested structures. Provider configs: OpenAI, Anthropic, AWS Bedrock, and OpenAI-compatible (Docker Model Runner, Ollama). CPT judge assertions verify the model correctly extracted facts from the PDFs.

### E2E coverage

The L4J job worker and outbound connector tool-calling test suites both cover document responses from BPMN tools — a Camunda-document tool and a nested external-reference tool — exercising the full agent flow.

### Deleted

`DocumentToContentModule`, `DocumentToContentResponseModel`, `DocumentToContentSerializer`, and related tests.

### ADR

See [ADR-004: Document Handling in Tool Call Results](connectors/agentic-ai/docs/adr/004-document-handling-in-tool-call-results.md).

## Related issues

closes #7005

## Checklist


connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/BaseL4JAiAgentJobWorkerTest.java
```diff
@@ -341,5 +341,5 @@
     }
   }
 
-  protected record DownloadFileToolResult(int status, DocumentToContentResponseModel document) {}
+  protected record DownloadFileToolResult(int status, DocumentReferenceModel document) {}
```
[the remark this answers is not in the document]
maff: Fixed — removed the unused record in commit `d94b7e52d`.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/outboundconnector/BaseL4JAiAgentConnectorTest.java
```diff
@@ -339,5 +339,5 @@
     }
   }
 
-  protected record DownloadFileToolResult(int status, DocumentToContentResponseModel document) {}
+  protected record DownloadFileToolResult(int status, DocumentReferenceModel document) {}
```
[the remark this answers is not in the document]
maff: Fixed — removed the unused record in commit `d94b7e52d`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    return result;
+  }
+
+  /**
+   * Recursively extracts all {@link Document} instances from an arbitrary object tree. Handles
+   * {@link Document}, {@link Map}, {@link List}/{@link Collection}, and skips all other types.
+   */
+  public List<Document> extractDocuments(Object content) {
```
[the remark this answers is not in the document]
maff: Fixed — renamed `extractDocuments(Object)` to `extractDocumentsFromContent(Object)` in commit `d94b7e52d`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/runtime/MessageWindowRuntimeMemory.java
```diff
@@ -72,7 +73,7 @@ public void clear() {
   // original implementation see Langchain4j
   private static List<Message> filteredMessages(List<Message> messages, int maxMessages) {
     final var filtered = new ArrayList<>(messages);
-    while (filtered.size() > maxMessages) {
+    while (countMessages(filtered) > maxMessages) {
       int messageToEvictIndex = 0;
```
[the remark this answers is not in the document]
maff: Fixed — now tracks effective count as an int, decremented on eviction. Commit `7c21394f2`.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/outboundconnector/L4JAiAgentConnectorToolCallingTests.java
```diff
@@ -148,19 +131,104 @@ void supportsDocumentResponsesFromToolCalls(
                 Map.of("userPrompt", initialUserPrompt))
             .waitForProcessCompletion();
 
-    assertLastChatRequest(3, expectedConversation);
+    // Assert the conversation structure with document extraction
+    await()
+        .alias("Chat request captured")
+        .untilAsserted(() -> assertThat(chatRequestCaptor.getValue()).isNotNull());
+
+    final var lastMessages = chatRequestCaptor.getValue().messages();
```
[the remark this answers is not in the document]
maff: Fixed — restored `chatRequestCaptor.getAllValues().hasSize(3)` assertion.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerToolCallingTests.java
```diff
@@ -148,18 +132,103 @@ void supportsDocumentResponsesFromToolCalls(
                 Map.of("userPrompt", initialUserPrompt))
             .waitForProcessCompletion();
 
-    assertLastChatRequest(expectedConversation);
+    // Assert the conversation structure with document extraction
+    await()
+        .alias("Chat request captured")
+        .untilAsserted(() -> assertThat(chatRequestCaptor.getValue()).isNotNull());
```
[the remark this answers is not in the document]
maff: Fixed — restored `chatRequestCaptor.getAllValues().hasSize(3)` assertion.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/discovery/McpClientGatewayToolHandler.java
```diff
@@ -259,12 +259,29 @@ private ToolCallResult toolCallResultFromMcpToolCall(ToolCallResult toolCallResu
...
+  private Object getRawMcpContent(ToolCallResult toolCallResult) {
+    if (toolCallResult.content() instanceof Map<?, ?> map) {
+      var content = map.get("content");
+      if (content == null && !map.isEmpty()) {
+        LOGGER.warn(
+            "MCP tool call result map has no 'content' key but contains keys: {}. "
+                + "Documents may be lost if the response structure has changed.",
+            map.keySet());
+      }
+      return content;
```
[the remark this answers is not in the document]
maff: The `content` key is part of the MCP spec (`CallToolResult.content`). If absent, the response shape has changed and we log a warning (added in commit `4dfbea24c`). Falling back to the full map would expose the MCP wrapper to the model, which would be more confusing than an empty result.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandlerImpl.java
```diff
@@ -249,6 +290,41 @@ private boolean interruptToolCallsOnEventResults(AgentExecutionContext execution
...
   }
 
+  static String documentXmlTag(Document document, String toolName, String toolCallId) {
+    var sb = new StringBuilder("<document");
+    if (StringUtils.isNotBlank(toolName)) {
+      sb.append(" tool=\"").append(toolName).append("\"");
+    }
+    if (StringUtils.isNotBlank(toolCallId)) {
+      sb.append(" call-id=\"").append(toolCallId).append("\"");
+    }
```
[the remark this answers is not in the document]
maff: Fixed — all attribute values are now XML-escaped via `escapeXmlAttribute()`. Commit `318fdb5b5`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandlerImpl.java
```diff
@@ -249,6 +290,41 @@ private boolean interruptToolCallsOnEventResults(AgentExecutionContext execution
...
+    var shortId = documentShortId(document);
+    if (shortId != null) {
+      sb.append(" document-short-id=\"").append(shortId).append("\"");
+    }
+    var fileName = document.metadata() != null ? document.metadata().getFileName() : null;
+    if (StringUtils.isNotBlank(fileName)) {
+      sb.append(" filename=\"").append(fileName).append("\"");
+    }
+    sb.append(" />");
```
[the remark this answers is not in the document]
maff: Fixed — same `escapeXmlAttribute()` applied to filename. Commit `318fdb5b5`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/runtime/MessageWindowRuntimeMemory.java
```diff
@@ -82,6 +85,7 @@ private static List<Message> filteredMessages(List<Message> messages, int maxMes
 
       // remove the message at the current index
       Message evictedMessage = filtered.remove(messageToEvictIndex);
+      effectiveCount--;
```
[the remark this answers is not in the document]
maff: Fixed — effectiveCount is now only decremented when the evicted message is not a tool-call document message. Added a test for the orphaned document message edge case.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/discovery/McpClientGatewayToolHandler.java
```diff
@@ -259,12 +259,29 @@ private ToolCallResult toolCallResultFromMcpToolCall(ToolCallResult toolCallResu
         && callToolResult.content().getFirst() instanceof McpTextContent textContent) {
       toolCallResultBuilder.content(textContent.text());
     } else {
-      toolCallResultBuilder.content(callToolResult.content());
+      // use the raw content from the original tool call result (preserving document references
+      // as deserialized by the engine) rather than the typed McpContent list, which may lose
+      // document reference fidelity during re-serialization.
+      toolCallResultBuilder.content(getRawMcpContent(toolCallResult));
```
maff: ℹ️ This is not ideal. As we need to traverse the content structure to find documents, it's easy to rely on the types created by FEEL deserialization. As soon as we deserialize into the actual domain specific DTOs, the current traversation logic in the document extractor doesn't work anymore -> pass the raw map here instead of the parsed `McpClientCallToolResult`.

connectors/agentic-ai/pom.xml
```diff
@@ -297,6 +297,11 @@
     </dependency>
 
     <!-- Test dependencies -->
+    <dependency>
+      <groupId>io.camunda.connector</groupId>
+      <artifactId>jackson-datatype-document</artifactId>
+      <scope>test</scope>
+    </dependency>
     <dependency>
```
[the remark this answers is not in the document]
maff: Fixed — removed the duplicate `jackson-datatype-document` entry.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/model/message/DocumentXmlTag.java
```diff
@@ -0,0 +1,88 @@
...
+    return from(document, null, null);
+  }
+
+  /** Serializes this tag to an XML self-closing element string. */
+  public String toXml() {
+    var sb = new StringBuilder("<document");
+    appendAttribute(sb, "tool-name", toolName);
+    appendAttribute(sb, "tool-call-id", toolCallId);
+    appendAttribute(sb, "document-short-id", documentShortId);
+    appendAttribute(sb, "filename", filename);
```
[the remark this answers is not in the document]
maff: The PR description already uses the correct attribute names (`tool-name`, `tool-call-id`, `document-short-id`). This was fixed earlier — the comment was filed against an older revision.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/cpt/DocumentToolCallResultsIT.java
```diff
@@ -0,0 +1,400 @@
...
+
+  // ---------------------------------------------------------------------------
+  // Helpers
+  // ---------------------------------------------------------------------------
+
+  private io.camunda.client.api.response.ProcessInstanceEvent startProcess(
+      ProviderConfig provider,
+      String userPrompt,
+      List<String> downloadUrls,
+      WireMockRuntimeInfo wireMock) {
```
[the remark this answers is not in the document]
maff: Fixed — removed unused `wireMock` parameter from `startProcess()`.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/cpt/DocumentToolCallResultsIT.java
```diff
@@ -0,0 +1,400 @@
...
+    var model = buildModel(provider);
+
+    // deploy and wait for process definition to be available
+    var zeebeTest =
+        io.camunda.connector.e2e.ZeebeTest.with(camundaClient)
+            .awaitCompleteTopology()
+            .deploy(model);
```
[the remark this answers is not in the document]
maff: Fixed — `zeebeTest` variable removed, deploy call is now a statement.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/e2e/DocumentToolCallResultsIT.java
```diff
@@ -0,0 +1,432 @@
...
+      "camunda.process-test.judge.chat-model.credentials.access-key=${AWS_BEDROCK_ACCESS_KEY:NOT_SET}",
+      "camunda.process-test.judge.chat-model.credentials.secret-key=${AWS_BEDROCK_SECRET_KEY:NOT_SET}",
+      "camunda.process-test.judge.threshold=0.6"
+    },
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@CamundaSpringProcessTest
+@WireMockTest
+@Import(CamundaDocumentTestConfiguration.class)
+@Disabled("Manual viability test - requires real LLM API keys via environment variables")
+class DocumentToolCallResultsIT {
```
maff, line 94: ℹ️ Works well with all the listed hosted models (also older ones like GPT 4.1). Does not work with Ollama/DMR as it fails with invalid file type errors (might be generally related to multimodality support).

maff: The judge assertions are really a gamechanger here! :)

connectors/agentic-ai/docs/adr/004-document-handling-in-tool-call-results.md
```diff
@@ -0,0 +1,159 @@
+# Document handling in tool call results
+
+* Deciders: Agentic AI Team
+* Date: Apr 10, 2026
+
+## Status
+
+**Proposed**.
```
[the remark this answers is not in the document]
maff: Fixed — updated status to **Implemented** with PR reference.

connectors/agentic-ai/docs/adr/004-document-handling-in-tool-call-results.md
```diff
@@ -0,0 +1,159 @@
...
+## Context and Problem Statement
+
+BPMN tool activities (modeled as ad-hoc subprocess tasks) can return complex data structures containing Camunda
+Documents, either at the root level or nested within maps and lists. When these tool call results are passed to the LLM,
+the documents need to be represented in a way the model can actually interpret.
+
+The current implementation converts `ToolCallResult.content()` to a single JSON text string for LangChain4J's
+`ToolExecutionResultMessage`. Documents encountered during Jackson serialization are converted to a Claude-specific
+content block format (`DocumentToContentSerializer`) containing base64-encoded data embedded within the text. This
+approach does not work -- models cannot meaningfully interpret large base64 blobs embedded in tool result text.
```
nikonovd: 💭 Actually it's only one part of the story, because for instance, OpenAI also converts binary contents to inline base64 URIs, but encloses them in a data object that hints the LLM that it's an image.

Generally the current limitation is also rather around the implementation of our tool call result handling interface, not particularly a limitation on the provider side.

OpenAI supports tool call result contents as part of response API -> we could potentially migrate to the `OpenAiResponsesChatModel`. Anthropic likely also supports content as tool call results.

Our current limitation is AFAIK in `langchain4j`'s implementation.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerToolCallingTests.java
```diff
@@ -148,18 +132,94 @@ void supportsDocumentResponsesFromToolCalls(
...
+
+    // tool result: document serialized as document reference
+    assertThat(lastMessages.get(3))
+        .isInstanceOfSatisfying(
+            ToolExecutionResultMessage.class,
+            msg -> {
+              assertThat(msg.id()).isEqualTo("aaa111");
+              assertThat(msg.toolName()).isEqualTo("Download_A_File");
+              assertThat(msg.text()).contains("camunda.document.type");
+              assertThat(msg.text()).contains(mimeType);
```
maff: 🔧 Can we do a better assertion on the document structure? Maybe even look at the map and JSONAssert the parts we know? This could also help us in extracting the short ID without this regex match on the whole text. Same for the other, similar tests.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/tool/GatewayToolHandler.java
```diff
@@ -54,4 +56,21 @@ boolean allToolDiscoveryResultsPresent(
...
+   * after {@link #transformToolCallResults} so the {@code toolCallResult.content()} carries this
+   * handler's transformed shape (typically a typed domain object).
+   *
+   * <p>The default implementation delegates to {@link ContentTreeDocumentWalker}, which walks
+   * {@link java.util.Map}, {@link java.util.Collection}, {@code Object[]} and {@link Document}
+   * nodes. This is sufficient for handlers whose transformed content remains a raw tree.
+   *
+   * <p>Handlers that return typed records or POJOs as content must override this method and walk
+   * their own structure (typically via a sealed-type switch). For mixed shapes — typed wrappers
+   * around nested raw subtrees — call {@link ContentTreeDocumentWalker#INSTANCE} on the raw parts.
```
maff: 🔧 This is not `INSTANCE` anymore.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ContentTreeDocumentWalkerTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+  @BeforeEach
+  void setUp() {
+    documentStore.clear();
+  }
+
+  @Test
+  void extractsRootLevelDocument() {
+    final var doc = createDocument("hello", "text/plain", "test.txt");
+    final var result = ContentTreeDocumentWalker.extractDocumentsFromContent(doc);
```
maff: 🔧 Let's import the method in this class.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ContentTreeDocumentWalkerTest.java
```diff
@@ -0,0 +1,145 @@
...
+  void returnsEmptyForNullContent() {
+    final var result = ContentTreeDocumentWalker.extractDocumentsFromContent(null);
+    assertThat(result).isEmpty();
+  }
+
+  @Test
+  void returnsEmptyForScalarContent() {
+    assertThat(ContentTreeDocumentWalker.extractDocumentsFromContent("text")).isEmpty();
+    assertThat(ContentTreeDocumentWalker.extractDocumentsFromContent(42)).isEmpty();
+    assertThat(ContentTreeDocumentWalker.extractDocumentsFromContent(true)).isEmpty();
```
maff: 🔧 Parameterized test, also for null?

maff:
@nikonovd updated the approach as discussed - the gateway handlers now implement domain-specific extraction using their native models.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerToolCallingTests.java
```diff
@@ -148,18 +132,91 @@ void supportsDocumentResponsesFromToolCalls(
...
+              assertThat(contents.get(1))
+                  .isInstanceOfSatisfying(
+                      TextContent.class,
+                      tc ->
+                          assertThat(tc.text())
+                              .isEqualTo(
+                                  "<document tool-name=\"Download_A_File\" tool-call-id=\"aaa111\" document-short-id=\"%s\" />"
+                                      .formatted(documentReference.shortId())));
+              assertDocumentContentBlock(contents.get(2), type, mimeType);
+            });
```
nikonovd: 🔧 let's try to organize larger assertions into methods with readable names to get a better glue of the test flow instead of reading a lot of assertion specific implementation details. WDYT?

here we could put it into a `assertHasUserMessageWithDocumentReferencesAndContent(...);` or similar.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/AiAgentTestFixtures.java
```diff
@@ -89,4 +92,77 @@ public interface AiAgentTestFixtures {
...
+    }
+
+    final JsonNode docNode = findFirstCamundaDocumentNode(root);
+    assertThat(docNode)
+        .as("Camunda document reference in tool result text: %s", toolResultText)
+        .isNotNull();
+    assertThat(docNode.path("camunda.document.type").asText())
+        .as("camunda.document.type discriminator")
+        .isEqualTo("camunda");
+    assertThat(docNode.path("documentId").asText()).as("documentId").isNotBlank();
```
nikonovd: 🤔 do you think we should move the assertions to another layer? here they are a bit hidden inside something where i'd expect just a read operation. WDYT?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/e2e/DocumentToolCallResultsIT.java
```diff
@@ -0,0 +1,435 @@
...
```
nikonovd, line 94: 🤔 we could secureguard this test with sth like `@EnabledIfEnvironmentVariables` to avoid it running in misconfigured environments.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractor.java
```diff
@@ -0,0 +1,66 @@
...
+  public List<ToolCallDocuments> extractDocuments(List<ToolCallResult> toolCallResults) {
+    final var result = new ArrayList<ToolCallDocuments>();
+
+    for (ToolCallResult toolCallResult : toolCallResults) {
+      final var documents = extractFromToolCallResult(toolCallResult);
+      if (!documents.isEmpty()) {
+        result.add(
+            new ToolCallDocuments(
+                StringUtils.defaultString(toolCallResult.id()),
+                StringUtils.defaultIfBlank(toolCallResult.name(), "unknown"),
```
nikonovd: ❓ when is a toolcall name `unknown` ?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/discovery/McpClientGatewayToolHandler.java
```diff
@@ -265,6 +271,34 @@ private ToolCallResult toolCallResultFromMcpToolCall(ToolCallResult toolCallResu
     return toolCallResultBuilder.build();
   }
 
+  @Override
+  public List<Document> extractDocuments(ToolCallResult toolCallResult) {
```
nikonovd, line 275: 🔧 we should add proper debug logging on the travesal implementations

nikonovd (commented):
Some remarks from my side! But the new implementation looks very promising when delegating to the gateway handlers!

Let's check regarding my remarks and finish that one 🚀

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/AiAgentTestFixtures.java
maff: Yes, fully agree that this makes no sense here. I'll move to a dedicated assertions class related to documents in tool call results.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractor.java
maff: I think never unless explicitely modeled as such - will remove this fallback.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/e2e/DocumentToolCallResultsIT.java
maff: As briefly discussed, I would suggest we keep it disabled for the moment, but we could think of moving it to the new e2e tests.

connectors-e2e-test/connectors-e2e-test-agentic-ai/pom.xml
```diff
@@ -88,6 +88,12 @@
       <artifactId>camunda-process-test-spring</artifactId>
       <version>${version.camunda}</version>
     </dependency>
+    <dependency>
```
nikonovd: 🔧 this is already a transitive dependency of `camunda-process-test-spring`

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandler.java
```diff
@@ -231,6 +239,56 @@ private ToolCallResult toolCallResultFromA2aSendMessage(ToolCallResult toolCallR
     return toolCallResultBuilder.build();
   }
 
+  @Override
+  public List<Document> extractDocuments(ToolCallResult toolCallResult) {
+    if (!(toolCallResult.content() instanceof A2aSendMessageResult result)) {
```
nikonovd, line 244: ⛏️ similar as we discussed in a previous PR: Carrying a diamond type on the base interface would avoid the need for a type check. Feel free to ignore 😄

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
```diff
@@ -457,6 +459,136 @@ void handlesEmptyA2aClientsList() {
...
+              .contents(
+                  List.of(
+                      TextContent.textContent("description"),
+                      DocumentContent.documentContent(document)))
+              .build();
+      var toolCallResult = createToolCallResultWithContent("call1", "A2A_a2a1", message);
+
+      var documents = handler.extractDocuments(toolCallResult);
+
+      assertThat(documents).containsExactly(document);
```
nikonovd, line 524: ❓ does this test work reliably given the fact that a mock is supplied here?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
```diff
@@ -457,6 +459,136 @@ void handlesEmptyA2aClientsList() {
...
+              .status(A2aTaskStatus.builder().state(A2aTaskStatus.TaskState.COMPLETED).build())
+              .artifacts(List.of(artifact))
+              .history(List.of(historyMessage))
+              .build();
+      var toolCallResult = createToolCallResultWithContent("call1", "A2A_a2a1", task);
+
+      var documents = handler.extractDocuments(toolCallResult);
+
+      // artifacts before history
+      assertThat(documents).containsExactly(artifactDoc, historyDoc);
```
nikonovd, line 557: ❓ same question here

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractorTest.java
```diff
@@ -0,0 +1,238 @@
...
+  private ToolCallResultDocumentExtractor extractor;
+
+  @BeforeEach
+  void setUp() {
+    documentStore.clear();
+    extractor = new ToolCallResultDocumentExtractor(registry);
+  }
+
+  @Test
+  void usesContentTreeWalkerWhenNoHandlerManagesTheToolCall() {
```
nikonovd: ⛏️ the test name suggests we use the default content walker fallback, but we actually are not verifying it. Maybe we could use a spy here?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractorTest.java
```diff
@@ -0,0 +1,238 @@
...
+    when(registry.handlerForToolDefinition(null)).thenReturn(Optional.empty());
+
+    final var extracted = extractor.extractDocuments(List.of(result));
+
+    assertThat(extracted).hasSize(1);
+    assertThat(extracted.getFirst())
+        .satisfies(
+            e -> {
+              assertThat(e.toolCallId()).isNull();
+              assertThat(e.toolCallName()).isNull();
```
nikonovd: ❓ i guess that would contain event subprocess tool calls, right?

could it potentially contain other documents as well and would this design have any undesired side effects?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractorTest.java
```diff
@@ -0,0 +1,238 @@
...
+            .name("plain_tool")
+            .content(Map.of("attachment", doc))
+            .build();
+
+    final var extracted = realExtractor.extractDocuments(List.of(result));
+
+    assertThat(extracted).hasSize(1);
+    assertThat(extracted.getFirst().documents()).containsExactly(doc);
+    verify(handler, never()).extractDocuments(any());
+  }
```
nikonovd: 🔧 we should extract those into a nested test class to avoid prefixing and underscoring, WDYT?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/model/message/DocumentXmlTagTest.java
```diff
@@ -0,0 +1,93 @@
...
+      when(doc.metadata()).thenReturn(metadata);
+      when(metadata.getFileName()).thenReturn("report.pdf");
+
+      assertThat(DocumentXmlTag.from(doc, "call_abc", "search").toXml())
+          .isEqualTo(
+              "<document tool-name=\"search\" tool-call-id=\"call_abc\" document-short-id=\"25ece9fa\" filename=\"report.pdf\" />");
+    }
+
+    @Test
+    void generatesTagWithoutToolAndCallId() {
```
nikonovd: ❓ why could that potentially happen?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/model/message/DocumentXmlTagTest.java
```diff
@@ -0,0 +1,93 @@
...
+import io.camunda.connector.api.document.Document;
+import io.camunda.connector.api.document.DocumentMetadata;
+import io.camunda.connector.api.document.DocumentReference.CamundaDocumentReference;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+class DocumentXmlTagTest {
+
+  @Nested
+  class ToXml {
```
nikonovd: ❓ why do we need a nested class here if it's still a flat test structure?

connectors/agentic-ai/pom.xml
```diff
@@ -293,18 +293,18 @@
...
-      <artifactId>guava</artifactId>
+      <groupId>io.camunda.connector</groupId>
+      <artifactId>jackson-datatype-document</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.camunda.connector</groupId>
-      <artifactId>connector-object-mapper</artifactId>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
```
nikonovd, line 301: ❓ why is this library introduced?

nikonovd (commented):
Please check my last comments 🍊

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
maff: It should - the handler iterates over the content blocks + and adds the document if the block is a `DocumentContent` without consuming/checking the actual document instance. Or did I miss something here?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
maff: See above.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractorTest.java
maff: Yes, exactly - this is for the event subprocess. I can't think of any side-effects - basically we always want to extract all documents from the tool call result structure (the event result is technically a TCR without id/name).

I refactored the tests to be more use-case driven now, that should address both this and the issue with the underscoring below.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/model/message/DocumentXmlTagTest.java
maff: I updated the XML tag to account for different document reference types - replaces this unnecessary nested class with one per variations.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/model/message/DocumentXmlTagTest.java
maff: This is for the event subprocess handling - documents are added in with the same logic to the event user message as for tool call results in the synthetic user message. I updated the ADR to contain some worked examples.

connectors/agentic-ai/pom.xml
maff: It was there before - dependencies were only reordered. I think we don't use it directly, but it's there to please the maven dependency plugin.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandler.java
maff: You mean as in `ToolCallResult<T>`? Will ignore it for now to not widen the scope of this PR, but happy to discuss separately.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractorTest.java
maff: Actually I don't want the test to verify that the walker was used, just the behavior that documents were extracted from the structure. Reframed on the test name.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/ToolCallResultDocumentExtractorTest.java
maff: Should be obsolete with the refactoring of the test class.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
nikonovd: We mock the [document that you return](https://github.com/camunda/connectors/pull/6999/changes#diff-1990208f1b47b0f189012678e13e883a53b075b6153c0bf18106e3e24cd78761R502) and I am not sure if a mock is implementing a valid `hashCode`/`equals` logic - thus i was questioning if it's really a good test to assert that a mock is part of the collection.

connectors/agentic-ai/pom.xml
nikonovd: In this diff it's only being added, that's why i asked! 

BTW we also can add such libraries to the `ignoredUsedUndeclaredDependency` tag to avoid referencing each and every bit of libs

nikonovd (approved):
Awesome! 🚀 

Please check the open comments, but it's not blocking!

connectors/agentic-ai/pom.xml
maff: I reverted the diff now, so there is now only the addition of commons-text.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
maff: Aah 💡 I think it did work as expected, but I now changed it to use real document records to be safe.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/e2e/DocumentToolCallResultsIT.java
```diff
@@ -0,0 +1,435 @@
...
+      "camunda.process-test.judge.chat-model.region=eu-central-1",
+      "camunda.process-test.judge.chat-model.credentials.access-key=${AWS_BEDROCK_ACCESS_KEY:NOT_SET}",
+      "camunda.process-test.judge.chat-model.credentials.secret-key=${AWS_BEDROCK_SECRET_KEY:NOT_SET}",
+      "camunda.process-test.judge.threshold=0.6"
+    },
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@CamundaSpringProcessTest
+@WireMockTest
+@Import(CamundaDocumentTestConfiguration.class)
+@Disabled("Manual viability test - requires real LLM API keys via environment variables")
```
maff, line 93: @gbetances089 this lives in the same package you recently added, but is currently not executed on CI as it needs credentials for multiple providers and - as it iterates over multiple providers/models - it takes ~7 min to run and spends some tokens.

I kept it disabled for now (it is still useful for manual verification) - in the long term we can think of also running it automatically, at least during a release.

maff:
@gbetances089 I know this is a huge one, but on a high level how this can be tested:

**1. Set up an agent with a tool that returns a document.** Easy options:
  - a REST connector configured to store the response (returns a Camunda document)
  - a user task that lets a tester upload a file
  - any custom worker / script task that emits `{camunda.document.type: "external", url, name}` for an external reference

  Two tools — one returning a Camunda doc, one returning an external doc inside a `{summary, document}` object — cover both XML tag shapes.

  **2. Run the flow against a real LLM provider that supports documents** (OpenAI, Anthropic, Bedrock). Prompt the agent to consume each tool (e.g. "tell me what's in this document"). Verify the model can describe the actual contents of both documents — that proves both reached the model intact.

  **3. Event flow.** Use a non-interrupting event subprocess that emits a payload containing a document. Confirm the model still receives the document, and that the synthetic `UserMessage` uses the `"event data"` preamble.

  **4. Regression — non-document tools.** Run a flow whose tools never return documents. Confirm behaviour is identical to `main` (no synthetic `UserMessage`, no preamble).

  **5. Cross-provider sanity (optional, heavier).** `DocumentToolCallResultsIT` in `connectors-e2e-test-agentic-ai` is `@Disabled` by default and runs the same scenarios against OpenAI / Anthropic / Bedrock / OpenAI-compatible providers with CPT judges. Enable it locally with credentials configured for any subset of providers - it's the most thorough manual reference for cross-provider behaviour. Local models are disabled as most of them fail with document processing.

---

With the AI Agent Chat with Tools example, this prompt should work now out of the box (or any other image/PDF) - the "Fetch URL" tool will download it and store it as document:

```
{
  "inputText": "In https://docs.camunda.io/assets/images/c8-architecture-diagram-07487ada3bab23823c5d8dab57365799.png, what is part of Management and Modeling?"
}
```

To test an external document reference, for example the Fetch URL tool can be changed to just return the external document as reference instead of actually downloading it. The Connectors SDK/AI Agent will download it implicitely:

image

gbetances089 (approved):

PR #7004, opened 2026-04-21
refactor(agentic-ai): use provider registry pattern for ChatModelFactory

### LINKED ISSUE #5547: Agentic AI: add custom model implementation support
## Goal

Let users provide their own LLM provider implementation by implementing the agent's `ChatModelApiFactory` SPI and selecting it via a "Custom" entry in the AI Agent element templates. Mirrors the pattern used for custom conversation memory storage, and gives self-managed / hybrid setups full flexibility to wire in any LLM not covered by the built-in native implementations.

## Requirements

- A new `Custom` member on the `ProviderConfiguration` sealed type carrying a user-chosen `providerType` discriminator string and a `Map<String, Object>` of custom parameters (consumed by the user's factory).
- The new AI Agent element templates expose this as a "Custom Implementation (Self-Managed/Hybrid only)" provider entry, with inputs for the `providerType` discriminator and the parameter map.
- The user implements `ChatModelApiFactory<CustomProviderConfiguration>` and registers it as a Spring bean. The registry dispatches by the `providerType()` string the user's bean returns; built-in registrations stay default but can be overridden via `@ConditionalOnMissingBean` patterns.
- Users may either implement the SPI directly, or register the LangChain4j adapter factor with their own `dev.langchain4j.model.chat.ChatModel` implementation — both paths are documented.

maff:
## Description

This PR refactors the Langchain4J `ChatModelFactory` to use a provider registry pattern, improving extensibility and maintainability. The monolithic factory class has been decomposed into individual provider implementations that are registered in a `ChatModelProviderRegistry`.

### Key Changes

1. **Added Provider Interface and Registry**
   - Added `ChatModelProvider<T extends ProviderConfiguration>` interface — generic parameter binds each provider to its configuration type, removing per-provider casts and type-check boilerplate
   - Added `ChatModelProviderRegistry` which holds the `type → provider` mapping and resolves providers for a given `ProviderConfiguration`
   - Providers are resolved at runtime via `ProviderConfiguration.providerType()` (previously `providerId()`); positions a future `CustomProviderConfiguration` to drop its accessor override (the record component naturally implements the interface)
   - Added a debug log on resolution to help self-managed users diagnose cases where a custom override is not picked up

2. **Extracted Provider Implementations**
   - Created individual provider classes for each built-in LLM:
     - `AnthropicChatModelProvider`
     - `AzureOpenAiChatModelProvider`
     - `BedrockChatModelProvider`
     - `GoogleVertexAiChatModelProvider`
     - `OpenAiChatModelProvider`
     - `OpenAiCompatibleChatModelProvider`
   - Each implements `ChatModelProvider<X>` directly; shared timeout derivation lives in a static utility `ChatModelProviderSupport`

3. **Simplified `ChatModelFactory`**
   - Reduced from ~300 lines to ~20 lines
   - Now delegates to `ChatModelProviderRegistry` for provider resolution
   - Maintains backward compatibility with existing API

4. **Reorganized Spring Configuration**
   - Introduced `AgenticAiLangchain4JChatModelConfiguration` that declares the chat model provider beans, the registry, and the factory
   - The existing `AgenticAiLangchain4JFrameworkConfiguration` now `@Import`s it, preserving the conditional enablement chain
   - Each provider bean is exposed under its parameterized type (`ChatModelProvider<XxxConfiguration>`), so users can override any built-in by registering a bean of the matching generic type

5. **Tests**
   - Split the monolithic `ChatModelFactoryTest` into per-provider unit tests (`AnthropicChatModelProviderTest` etc.) with shared helpers in `ChatModelProviderTestSupport`
   - `ChatModelFactoryTest` now covers the thin delegation from factory to registry
   - Added `ChatModelProviderRegistryTest` covering lookup and duplicate handling
   - `AgenticAiConnectorsAutoConfigurationTest` exercises the new bean wiring; a `@Nested` `CustomChatModelProviderOverrides` parameterized test verifies that a user-supplied `ChatModelProvider<X>` bean replaces the matching built-in for all six provider types

### Benefits

- **Maintainability**: Each provider is isolated in its own class, easier to understand and modify
- **Testability**: Individual providers can be tested in isolation
- **Foundation**: Prepares the code for future extensibility work (tracked in #5547 as a follow-up)

## Related issues

related to #5547 (refactoring foundation; the custom-provider extension point will land as a follow-up)

## Checklist

- [x] Tests/Integration tests for the changes have been added (per-provider unit tests, `ChatModelProviderRegistryTest`, `ChatModelFactoryTest`, `AgenticAiConnectorsAutoConfigurationTest` including the parameterized override test)
- [ ] Backport labels (not applicable for refactoring)
- [ ] Documentation updates (no user-facing behavior change)

https://claude.ai/code/session_01DA2Hb5abY72R9NtuNMcJRb


connectors/agentic-ai/element-templates/hybrid/agenticai-aiagent-job-worker-hybrid.json
```diff
@@ -776,6 +779,43 @@
       "type" : "simple"
     },
     "type" : "String"
+  }, {
+    "id" : "provider.providerType",
+    "label" : "Implementation type",
+    "description" : "Identifier of the custom ChatModelProvider implementation to use. A matching Spring bean must be registered.",
```
maff: 🔧 Do not mention `ChatModelProvider` (class name) or Spring here - keep it more generic, similar to the memory.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryTest.java
```diff
@@ -108,8 +116,19 @@ class ChatModelFactoryTest {
           new ChatModelHttpProxySupport(
               proxyConfiguration, new JdkHttpClientProxyConfigurator(proxyConfiguration)));
 
+  private final AgenticAiConnectorsConfigurationProperties configurationProperties =
+      createDefaultConfigurationProperties();
+
   private final ChatModelFactory chatModelFactory =
-      new ChatModelFactoryImpl(createDefaultConfigurationProperties(), proxySupport);
+      new ChatModelFactoryImpl(
```
maff: 🔧 What about a custom implementation here?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerProviderMappingTests.java
```diff
@@ -0,0 +1,240 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockito.ArgumentCaptor;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.annotation.Import;
+import org.springframework.test.context.bean.override.mockito.MockitoSpyBean;
+
+@SlowTest
+@Import(CustomChatModelProviderTestConfiguration.class)
+public class L4JAiAgentJobWorkerProviderMappingTests extends BaseAiAgentJobWorkerTest {
```
nikonovd: 🤔 Do we really need to create a process instance to verify that the resolution of the correct provider works?

IMO it'd be enough to have a proper unit test (of broader scope) or maybe an integration test, on the edge of the outbound connector call.

What we are actually verifying here is that the connector runtime works.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/configuration/AgenticAiLangchain4JFrameworkConfiguration.java
```diff
@@ -48,13 +58,70 @@ public ChatModelHttpProxySupport langchain4JChatModelHttpProxySupport(
...
+    return new ChatModelProviderRegistryImpl(chatModelProviders);
+  }
+
+  @Bean
+  @ConditionalOnMissingBean
+  public ChatModelFactory langchain4JChatModelFactory(
+      ChatModelProviderRegistry chatModelProviderRegistry) {
+    return new ChatModelFactoryImpl(chatModelProviderRegistry);
+  }
+
```
nikonovd: ⛏️ Probably it might be worth to extract a separate `Langchain4JChatModelConfiguration` or something similar

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProvider.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Factory for a single LLM provider implementation. Implementations are registered in the {@link
+ * ChatModelProviderRegistry} by their {@link #type()}.
+ *
+ * <p>Built-in providers are registered with the same id as the corresponding {@link
+ * ProviderConfiguration} discriminator (e.g. {@code anthropic}). Custom providers can be registered
+ * under an arbitrary id and resolved through {@link
+ * io.camunda.connector.agenticai.aiagent.model.request.provider.CustomProviderConfiguration}.
+ */
+public interface ChatModelProvider {
```
nikonovd: 🔧 this could carry a generic type parameter to implicitely cast the provider configuration, i.e.

```java
interface ChatModelProvider<T extends ProviderConfiguration> {

    ChatModel createChatModel(T providerConfiguration);
}
```

where you should be able to avoid the repeating type-check + exception cases.

For custom implementations we could create a public subtype and make `ChatModelProvider` package-private, allowing us to narrow custom implementations to a specific ProviderConfiguration type, that probably just contains a map of additional properties.

```java
public interface CustomChatModelProvider extends ChatModelProvider<CustomProviderConfiguration> {}
```

and the `CustomProviderConfiguration` with your parameters 👍 as you implemented it already.

WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProviderRegistryImpl.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private final Map<String, ChatModelProvider> chatModelProviders = new HashMap<>();
+
+  public ChatModelProviderRegistryImpl() {}
+
+  public ChatModelProviderRegistryImpl(List<ChatModelProvider> chatModelProviders) {
+    chatModelProviders.forEach(this::registerChatModelProvider);
+  }
+
+  public void registerChatModelProvider(final ChatModelProvider chatModelProvider) {
```
nikonovd: 🔧 ideally, 
this should not be public

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProviderRegistry.java
```diff
@@ -0,0 +1,17 @@
...
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.aiagent.framework.langchain4j.provider;
+
+import io.camunda.connector.agenticai.aiagent.model.request.provider.ProviderConfiguration;
+
+/**
+ * Registry used to resolve a {@link ChatModelProvider} for a given {@link ProviderConfiguration}.
+ */
+public interface ChatModelProviderRegistry {
```
nikonovd: 🔧 I don't think this interface is necessary, as its single implementation is tighly coupled to the `ChatModelFactory` anyways.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProviderRegistryImpl.java
```diff
@@ -0,0 +1,51 @@
...
+          "Chat model provider with type '%s' is already registered.".formatted(type));
+    }
+
+    LOGGER.debug("Registering chat model provider of type '{}'", type);
+
+    chatModelProviders.put(type, chatModelProvider);
+  }
+
+  @Override
+  public ChatModelProvider getChatModelProvider(ProviderConfiguration providerConfiguration) {
```
nikonovd: 🔧 I'd prefer to have some debug logging around the provider resolution

nikonovd (commented):
🚀 Awesome refactoring! I had some remarks that could improve the handling a bit.

I'm still very hesitant to expose an extension point to our customers relying on a domain model that is completely not in our hand.

For instance, langchain4j had already some minor breaking changes during minor version increments, which would also then needed to be handled by our customers that implement custom chat model providers.

I would propose to discuss that with a bigger round, as i would really like to avoid a potential breaking point to users relying on the stability of that interface.

Saying that, i really like the refactoring and the codebase would benefit from that. WDYT about extracting the custom implementation part into a separate branch and merge the general improvements right away?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProvider.java
maff: As suggested, I now removed the "custom" part on this PR, making it a pure refactoring into individual model providers. Let's discuss the design approach for "custom" as we progress.

maff: Agree on the generic type, implemented.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProviderRegistryImpl.java
maff: Would you basically log which provider was resolved on every request?

maff:
@nikonovd thanks for the review, agree with most of the points. While I still think we'll need to provide a way for custom models, we can defer that for the moment and only pursue the refactoring which already makes it easier to customize supported models. Changed that accordingly + removed the e2e test.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProviderRegistryImpl.java
nikonovd: not on every request, but maybe on debug level. This could come handy for SM users when debugging why their custom model might not be referred to. WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/AbstractChatModelProvider.java
```diff
@@ -0,0 +1,47 @@
...
+import io.camunda.connector.agenticai.aiagent.model.request.provider.ProviderConfiguration;
+import io.camunda.connector.agenticai.aiagent.model.request.provider.shared.TimeoutConfiguration;
+import io.camunda.connector.agenticai.autoconfigure.AgenticAiConnectorsConfigurationProperties;
+import java.time.Duration;
+import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Base class for built-in {@link ChatModelProvider} implementations. */
+public abstract class AbstractChatModelProvider<T extends ProviderConfiguration>
```
nikonovd: ⛏️ Could this be package-private if it's only for built-in implementations?

We would shrink public surface.

nikonovd (approved):
🚀 great, lgtm!

I left one minor comment and there's also still the one with the logging, nothing blocking.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/AbstractChatModelProvider.java
maff: I now completely removed the abstract base class as it only provided the timeout helper not even used from all implementations. Provider classes now directly implement the interface and are registered as `ChatModelProvider<ConfigClass>` on the spring context. This allows model customizations by registering a bean with the same type (added tests to check this works as expected).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/provider/ChatModelProviderRegistryImpl.java
maff: Makes sense, added in the last commt

nikonovd (approved):

PR #7029, opened 2026-04-23
TMP: Agentic AI execution tracing solution design exploration

maff:
## Description

Solution design (finalized) and starting PoC implementation for **agent execution tracing** in the AI Agent — a fine-grained, append-only event stream pushed from the agent runtime to a Camunda backend for aggregation, metrics, and auditability.

The design introduces:

- An **append-only event log** of 11 event types (`MODEL_CALL_STARTED/COMPLETED/FAILED`, `TOOL_CALLS_EMITTED`, `TOOL_CALL_RESULT_RECEIVED`, `ITERATION_COMPLETED`, `LIMIT_HIT`, `CONVERSATION_SNAPSHOT`, `SYSTEM_PROMPT_CHANGED`, …) — modeled as events (not entity updates) so retried/superseded jobs, partial tool results, and concurrent no-op turns are all captured correctly.
- A **server-side deduplication strategy** based on per-event deduplication keys, making transport naturally idempotent (events are re-pushed on job completion via `flushAll()`).
- **Delta-based conversation reconstruction**: each event carries only per-turn deltas; server replays them to rebuild conversation state at any point in time.
- **Standalone API contract types** mirrored from the runtime (`Message`, `Content`, `ToolCall`, `ToolDefinition`, `Document` …) living in the tracing package — isolating the public API from agent-internal evolution.
- A **split client / session API** (`AgentExecutionTraceClient` + `AgentExecutionTraceSession` with `NO_OP` singleton) for stateless transport + per-request state.
- **Metrics coverage mapping** cross-referencing every metric in the Agent Visibility Metrics Reference against the event stream, BPMN data, or both, with server-derivation formulas.
- **Backwards compatibility** for pre-tracing agents (`schemaVersion`, `CONVERSATION_SNAPSHOT` on upgrade, nullable boundaries).

Docs:

- [`connectors/agentic-ai/docs/tracing/design.md`](connectors/agentic-ai/docs/tracing/design.md) — full design
- [`connectors/agentic-ai/docs/tracing/metrics.md`](connectors/agentic-ai/docs/tracing/metrics.md) — metric-by-metric coverage analysis

**Status**: Design finalized (refined per review, see 3c00f16). PoC implementation in progress on this branch.

## Related issues

closes #

## Checklist


connectors/agentic-ai/docs/tracing/design.md
```diff
@@ -0,0 +1,1643 @@
...
+
+The event model treats the trace as an **append-only log** rather than a mutable document. Each
+event is a fact — "at this time, this happened" — and facts don't change. The server aggregates
+events into derived views (total tokens, conversation timeline, tool call durations) but the
+events themselves are immutable.
+
+This aligns with the auditability goal: the event stream is a complete, ordered record of
+everything that happened during the agent execution, including failed attempts, retried jobs,
+and superseded activations. An entity-based model can only show the current state; an event-based
+model shows the full history of how we got there.
```
korthout, line 139: ⚠️ This makes sense but we do need to watch out to not overload the engine. Each audited event send to the engine (as a command) has to go through RAFT commit (replication and reaching quorum) and so do the processing results (events) of it. That means a lot of extra strain on the performance. We should see to limit this, perhaps by batching the events. 

I wrote a bit more about this [here](https://github.com/camunda/camunda/commit/53848c3d57313f10fc502ab825f61d8874e272bf)

maff: Fully agree - do you think it would be viable to design this in a way that each "flush" would be a single command transporting multiple events? I'm not sure if we really need the "safety net" to re-push after completion - without it that would mean 2 flushes per job execution (before/after LLM call). If it is a concern and we don't need the data as early as possible potentially we could also reduce it to only pushing after the LLM call, with idempotency still helping with retries when the HTTP command fails/times out.

sbuettner:
@maff  Is this still relevant or can we close it?

maff:
@sbuettner the content may still be relevant for reference, but we can close the PR 👍

PR #7072, opened 2026-04-28
test(agentic-ai): add event handling regression e2e test

maff:
## Description

Adds an e2e regression test suite for the agentic-AI ad-hoc sub-process (AHSP) event-handling contract, designed to catch a bug introduced in Camunda 8.9.1 where inner-instance variables (notably `toolCall` / `toolCallResult` set by the AI Agent on tool activations) leak out of the inner-instance scope into the surrounding AHSP/root scope.

`L4JAiAgentJobWorkerEventsTests` runs on a single BPMN — `agentic-ai-ahsp-connectors-event.bpmn` — that mirrors the realistic connectors tool setup (`SuperfluxProduct`, `Search_The_Web`, etc.) and the user-feedback satisfaction loop, plus a `Pending_Tool` service task whose job tests intentionally hold to keep the AHSP open while events are published. Per-test message correlation isolation via a UUID-driven `eventCorrelationKey`.

Scenarios:
- Event message published *before* AHSP activation (buffered, correlates on entry).
- Event during in-flight tool execution with `WAIT_FOR_TOOL_CALL_RESULTS` (default).
- Event during tool execution with `INTERRUPT_TOOL_CALLS` ("Cancel tool calls").
- Event with empty payload — agent inserts the synthetic placeholder UserMessage (one variant per behavior, since the synthetic message text differs).
- Multiple events in a single AHSP iteration — published order preserved in the chat request.

Each scenario asserts the chat conversation, agent metrics and response text, the user-feedback worker firing exactly once, and the leak invariant (`toolCall` must not exist at the process-instance root scope).

In addition:
- `agentic-ai-ahsp-connectors.bpmn`: `SuperfluxProduct` writes its script result to an intermediate variable, and an `<zeebe:output>` mapping projects it to `toolCallResult`. The presence of the output mapping is what triggers the regression code path on 8.9.1 — without it, the buggy parent+local merge in `BpmnVariableMappingBehavior` is never reached. Same change is mirrored in `agentic-ai-ahsp-connectors-event.bpmn`.
- `L4JAiAgentJobWorkerToolCallingTests.executesAgentWithToolCallingAndUserFeedback` already had a defense-in-depth leak check; it was previously a no-op (the connectors BPMN's tools had no output mappings), and is now a real regression detector thanks to the `SuperfluxProduct` output mapping.

Verification:

| | engine fix image | 8.9.1 |
| --- | --- | --- |
| `L4JAiAgentJobWorkerEventsTests` (6 tests) | ✅ all pass | ❌ all 6 fail (leak detected or process never completes) |
| `L4JAiAgentJobWorkerToolCallingTests.executesAgentWithToolCallingAndUserFeedback` | ✅ pass | ❌ fail (leak detected) |

### Note on 8.9.0

8.9.0 is **not** a stable verification target for this suite when run as a class. Individual tests pass, but in suite mode 3/6 tests time out due to a known CPT/gateway interaction (camunda/camunda#45177, camunda/camunda#45667): when CPT recreates the `CamundaClient` between tests, REST long-poll job-activation requests aren't cleanly cancelled — the gateway delivers next-test jobs to the dead connection, the new test's worker polls and gets nothing, and the job sits locked until the activation timeout (~60s) expires.

8.9.0 shipped a partial workaround (camunda/camunda#49836 — force gRPC + long-polling). The proper fix landed in 8.9.1 (camunda/camunda#49424 — cancel pending REST long-polls on cluster purge) along with related cleanup of HTTP connection-pool shutdown. The engine fix image (built on 8.9.1+) doesn't exhibit the lag, so it remains the meaningful verification target alongside 8.9.1 (which fails fast on the AHSP regression assertion).

## Related issues

Adds e2e coverage for the regression tracked in camunda/camunda#51939.

## Checklist


maff:
ℹ️ Note: this will fail e2e until using Camunda 8.9.2

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/BaseL4JAiAgentJobWorkerTest.java
```diff
@@ -330,6 +331,41 @@ protected void assertToolSpecifications(ChatRequest chatRequest) {
...
+  /**
+   * Asserts that the {@code toolCall} variable does not exist at the root (process instance) scope.
+   * The AI agent sets {@code toolCall} only on inner-instance scopes when activating a tool; a hit
+   * at the root scope means an inner-instance variable has leaked upward — the regression tracked
+   * by camunda/camunda#51939, introduced in 8.9.1.
+   *
+   * <p>{@code scopeKey} on the variable search filters direct (not inherited) variables, which is
+   * exactly what we want.
+   */
+  protected void assertNoToolCallVariableLeak(ZeebeTest zeebeTest) {
```
nikonovd:
```suggestion
  protected void assertNoToolCallVariableLeakToProcessScope(ZeebeTest zeebeTest) {
```

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerEventsTests.java
```diff
@@ -28,177 +29,371 @@
...
+
+    final var expectedConversation =
+        List.of(
+            SYSTEM_MESSAGE,
+            INITIAL_USER_MESSAGE,
+            aiToolCallMessage,
+            new ToolExecutionResultMessage(
+                SUPERFLUX_TOOL_CALL_ID, SUPERFLUX_TOOL_NAME, SUPERFLUX_TOOL_RESULT),
+            new ToolExecutionResultMessage(
+                PENDING_TOOL_CALL_ID, PENDING_TOOL_NAME, CANCELLED_TOOL_RESULT),
```
nikonovd, line 288: ❓ How is it ensured that the first tool call result is being processed, while only the second is cancelled?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerEventsTests.java
```diff
@@ -28,177 +29,371 @@
...
+  }
 
-    assertThat(userFeedbackJobWorkerCounter.get()).isEqualTo(1);
+  private AiMessage aiSuperfluxToolCall() {
+    return new AiMessage(
+        "I will use the superflux tool.",
+        List.of(toolCall(SUPERFLUX_TOOL_CALL_ID, SUPERFLUX_TOOL_NAME, SUPERFLUX_TOOL_ARGS)));
+  }
+
+  private AiMessage aiTwoToolCalls() {
```
nikonovd:
```suggestion
  private AiMessage aiMessageWithTwoToolCallInstructions() {
```

❓

nikonovd (approved):
Looks good and hopefully catches such regressions earlier 😆 

Some minor remarks and a question from my side, but nothing i see as blocking!

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerEventsTests.java
maff: The superflux tool is a script task which is immediately processed. The "pending" tool is a service task which needs to be manually completed (`completePendingToolJob`) to proceed. So when the message is published, the script task is already completed with a result.

TBH I don't know if it is theoretically possible for the script task to take longer than the job being created, awaited and the message published 🤔 but in all the runs this worked reliably. If this is flaky, we could probably await a superflux result being present before proceeding, but I would do that in a follow-up if we see a reason to do so.

Issue #5807, opened 2025-11-24
[Slack] Misleading Plaintext message type


chillleader:
__Describe the Bug__

When selecting message type in the Slack Outbound connector properties panel, users can choose between two types:
1. Plaintext
2. Text blocks

In the 8.8 release, selecting the Plaintext option does not truly lead to using the `text` field of the [chat.postMessage](https://docs.slack.dev/reference/methods/chat.postMessage/) API method. The connector always passes a text block, implicitly wrapping the text provided by the user into a single block.

Before 8.8, selecting Plaintext would mean that message was passed as "true" plaintext.

This leads to changed behavior in some cases, e.g. screenshot below

<img width="737" height="109" alt="Image" src="https://github.com/user-attachments/assets/9e17f8f8-00b1-4247-898b-e5325dacd027" />

<img width="1037" height="181" alt="Image" src="https://github.com/user-attachments/assets/73cfc7bd-57ee-4777-b4b9-463ae537365c" />

__Steps to Reproduce__

This issue becomes visible when using long text lines. Slack automatically wraps lines within a block body, this isn't the case for plaintext.

1. With 8.7, send a Plaintext message using Slack outbound connector. Make the message body a single line with 85+ characters. In Slack, you will see that the message takes up full screen width.
2. With 8.8, send a Plaintext message using Slack outbound connector. The message body is now passed as a block, which leads to line wrapping and potentially broken formatting for old processes.

__Expected Behavior__

__Environment__

- Camunda version: 8.8.0+

https://camunda.slack.com/archives/C02JLRNQQ05/p1763731646748289

PR #7079, opened 2026-04-29
fix(slack): Does not send plain text with documents

ztefanie:
## Description

Plain text was not send when adding documents as attachments. I fixed this. I also checked this change does not reintroduce [this bug](https://github.com/camunda/connectors/issues/5807). As this slack api is a bit unpredictable, please let me know of edge cases i might want to check.

## Related issues

None, just found it

## Checklist

### DIFF
--- a/connectors/slack/src/main/java/io/camunda/connector/slack/outbound/model/ChatPostMessageData.java
+++ b/connectors/slack/src/main/java/io/camunda/connector/slack/outbound/model/ChatPostMessageData.java
@@ -128,6 +128,7 @@ public record ChatPostMessageData(
       if (documents != null && !documents.isEmpty()) {
         requestBuilder.blocks(
             BlockBuilder.create(new FileUploader(methodsClient))
+                .text(text)
                 .documents(documents)
                 .getLayoutBlocks());
       }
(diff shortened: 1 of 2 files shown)


connectors/slack/src/main/java/io/camunda/connector/slack/outbound/model/ChatPostMessageData.java
```diff
@@ -128,6 +128,7 @@ public SlackResponse invoke(MethodsClient methodsClient) throws IOException, Sla
       if (documents != null && !documents.isEmpty()) {
         requestBuilder.blocks(
             BlockBuilder.create(new FileUploader(methodsClient))
+                .text(text)
```
johnBgood, line 131: Can you add some steps to reproduce the initial issue? 
I see that we add `requestBuilder.text(text);` above, I'm wondering how it's rendered when adding the `text` in the block.

ztefanie: Yes sure:

Steps to reproduce:
- select slack connector + "Post message"
- select "Message type" to be "Plain Text"
- add message "my text"
- add an attachment

Result:
- only the attachment is sent

Expected Result:
- the attachment and the plain text are sent

johnBgood: Ok looks good, do you think it's worth a QA round with Andreas? we had so many issues in this connector while implementing small changes, up to you

johnBgood (approved):

connectors/slack/src/main/java/io/camunda/connector/slack/outbound/model/ChatPostMessageData.java
ztefanie: Yes i think that makes sense.

Szik (approved):

Issue #4847, opened 2025-06-05
Agentic AI: Process definition not found errors when starting a process directly after deployment


maff:
__Describe the Bug__

Depending on the cluster, a process executing AI Agents can fail when the process definition can not be loaded from the Camunda API:

![Image](https://github.com/user-attachments/assets/ef71c0c8-7e15-414c-be18-3641e19bc536)

This happens when deploying at the same time as starting the process and the exporter not yet being able to export the process definition when the connector code is executed. It tries to load the definition from the Camunda API but fails to do so with the 404 shown in the screenshot above.

__Steps to Reproduce__

The easiest way is to change a process and use "Run" in web modeler:

1. Deploy a process - make sure that it has changes, so a new process definition key is created
2. Immediately start a process instance after the deployment

__Expected Behavior__

The AI Agent connector is able to load the process definition and does not fail with the error above. Ideally we can internally retry the request to a certain extent.

__Environment__

- Camunda version: 8.8.0

maff:
This will not be an issue when the AI Agent is implemented as ad-hoc sub-process job worker as there the job worker will take care of providing the tool definitions + parameters. 💭 Maybe we can somehow leverage this also for an AI Agent as service task.

PR #7142, opened 2026-05-06
feat(agenticai): resolve external parameters

tmetzke:
**NOTE: this is just an experiment**

## Description

Resolves `externalParameters` in tool elements to create tool input schemas. This supersedes `fromAi` resolution if provided.

## Related issues

related to https://github.com/camunda/camunda/issues/52186

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/JobWorkerAgentExecutionContext.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/JobWorkerAgentExecutionContext.java
@@ -7,6 +7,8 @@
 package io.camunda.connector.agenticai.aiagent.model;
 
 import io.camunda.connector.agenticai.adhoctoolsschema.model.AdHocToolElement;
+import io.camunda.connector.agenticai.adhoctoolsschema.model.AdHocToolElementParameter;
+import io.camunda.connector.agenticai.adhoctoolsschema.processdefinition.ProcessDefinitionAdHocToolElementsResolver;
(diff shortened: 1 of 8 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/JobWorkerAgentExecutionContext.java
```diff
@@ -47,7 +62,43 @@ public List<ToolCallResult> initialToolCallResults() {
...
+   */
+  private List<AdHocToolElement> enrichWithParameters(List<AdHocToolElement> zeebeElements) {
+    if (zeebeElements == null || zeebeElements.isEmpty() || toolElementsResolver == null) {
+      return zeebeElements;
+    }
+
+    try {
+      final Map<String, List<AdHocToolElementParameter>> paramsByElementId =
+          toolElementsResolver
+              .resolveToolElements(jobContext.getProcessDefinitionKey(), jobContext.getElementId())
```
maff, line 85: Very nicely integrated into the existing logic!

As a side note: if we pursue this approach, we should probably try to find a way to push the element information directly from the job activation, similar to the existing `adHocSubProcessElements`. The client/XML way can be problematic when exporting is slow (see https://github.com/camunda/connectors/issues/4847) (or secondary storage is not available at all). We addressed the linked issue by adding retries to the XML fetching part in https://github.com/camunda/connectors/pull/5390, but in theory the problem persists.

tmetzke: Yes, I agree, I will leave this comment open so we can consider it when we pick this up for real.

Issue #6843, opened 2026-03-30
Support start-or-correlate behaviour for inbound webhook Connectors on a single endpoint


crobbins215:
**Is your feature request related to a problem? Please describe.**

Developers using inbound webhook Connectors cannot model a common “start vs correlate” pattern on a single HTTP endpoint.

A typical requirement is that the same webhook should:
start a new process instance if none exists for a given business key, or
route into an event subprocess of an existing instance with that key.

Today, modelling this with two inbound webhook Connectors, one on the main start event and one on a non-interrupting message start event in an event subprocess, fails with:

Invalid input: Multiple connectors are activated for the same input

This is inconsistent with Camunda’s existing message semantics, where the same message name and correlation key can either start a new instance or correlate to an existing one depending on runtime state. It is confusing for users and forces additional modelling and configuration just to recreate behaviour they already expect from messages.

**Describe the solution you'd like**

Allow developers to model this pattern directly with inbound webhook Connectors on BPMN elements using a single endpoint and the same webhook payload.

Specifically, when multiple inbound webhook Connectors are configured on compatible message-correlating elements, such as a message start event and a message event subprocess, the runtime should not fail with “Multiple connectors are activated for the same input”. Instead, it should apply message correlation semantics:

if no instance exists for the correlation key, start a new process instance via the message start event
if an instance already exists, correlate to the matching intermediate or event subprocess element instead of starting a new instance

This would make inbound webhook Connectors behave more consistently with how message start and correlation already work in Camunda.

**Describe alternatives you've considered**

The main workaround is to use a single inbound webhook Connector as a generic entry point and then publish a Zeebe message from inside the Connector.

In BPMN, the process can then be modelled with a message start event plus a message catch or event subprocess using the same message name and correlation key, allowing Zeebe’s existing correlation logic to decide whether to start a new instance or correlate to an existing one.

While this works, it introduces an unnecessary extra hop and additional configuration. It also makes the model less direct, because developers cannot express the desired behaviour purely with inbound webhook Connectors on BPMN elements.

**Additional context**

A common use case is where one external system sends HTTP webhooks, using the same URL and same payload, for a business entity such as an orderId.

The first webhook for a given business key should start a new process instance. Subsequent webhooks with the same key should be routed into a non-interrupting event subprocess of the existing instance.

This is a critical gap because it blocks a natural and expected modelling pattern for webhook-driven processes.

[...]

PR #7154, opened 2026-05-07
feat: Inbound Connector Runtime Lifecycle Refactor

mathias-vandaele:
## Summary

This PR refactors the inbound connector runtime lifecycle to make activation safer, easier to reason about, and friendlier for connector authors.

## Motivation

The previous model leaked too much runtime/concurrency complexity into connector implementations.

In practice, `activate` could not safely block for a few seconds without risking stalling the runtime, so connector authors were pushed toward workarounds such as starting their own thread and returning immediately. That means connector implementations had to care about lifecycle orchestration when they should mostly care about establishing the connection and becoming ready.

At the same time, the runtime internals had accumulated a mix of locking, synchronization, callback-style cancellation plumbing, and fire-and-forget async behavior. That makes the activation path harder to follow and increases the risk of subtle race conditions or lifecycle bugs. The current diff removes the explicit cancellation callback from inbound context construction and removes the old `cancel(...)` path from the context implementation as part of simplifying that model.

## What This PR Changes

- Refactors the inbound runtime lifecycle/concurrency model so the runtime owns the blocking/async boundary
- Makes `activate` blocking-friendly, so a connector can spend a few seconds establishing its connection without forcing every implementation to spawn its own thread
- Simplifies the inbound context API by removing cancellation callback plumbing from the context factory and context implementations
- Reduces the amount of low-level concurrency behavior that connector authors need to reason about

## Why This Is Better

The main goal is to improve both runtime safety and the SDK contract.

**From the runtime side**, the activation flow becomes simpler and less error-prone because lifecycle control is centralized instead of being split between runtime internals and individual connector implementations.

**From the connector author side**, `activate` can now mean "activate": establish the connection, become ready, and return when activation is complete. Authors should not need to implement their own thread handoff just to avoid blocking the runtime.

Overall, this should make inbound connectors easier to implement, easier to review, and less likely to hide concurrency issues in connector-specific code.

## Compatibility

This PR does include lifecycle/API cleanup, but the intent is not change for the sake of change. The intent is to move to a safer and simpler model while staying pragmatic about existing connectors that may rely on older patterns.

Review follow-ups will help decide where to keep temporary compatibility or deprecation shims — reviewer feedback has already highlighted the need to be careful here for existing connectors.


.gitignore
```diff
@@ -71,3 +71,6 @@ THIRD-PARTY.properties
 element-template-generator/npm/node_modules
 element-template-generator/npm/src/element-templates
 element-template-generator/npm/src/connectors.js
+
+.claude
+graphify-out/
```
chillleader: can we move this (and claude.md) a separate PR as this is unrelated to the refactoring

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/exception/GlobalExceptionHandler.java
```diff
@@ -28,6 +31,22 @@ public ResponseEntity<String> handleNotFound(DataNotFoundException ex) {
     return ResponseEntity.status(HttpStatus.NOT_FOUND).body(ex.getMessage());
   }
 
+  @ExceptionHandler(InboundExecutableNotFoundException.class)
+  public ResponseEntity<String> handleExecutableNotFound(InboundExecutableNotFoundException ex) {
+    return ResponseEntity.status(HttpStatus.NOT_FOUND).body(ex.getMessage());
+  }
```
chillleader: do we need this new exception type when we already have `DataNotFoundException` handler that seemingly covers this?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
```diff
@@ -82,13 +82,6 @@ public interface InboundConnectorContext extends DocumentFactory {
    */
   CorrelationResult correlate(CorrelationRequest correlationRequest);
 
-  /**
-   * /** Signals to the Connector runtime that inbound Connector execution was interrupted. As a
-   * result of this call, the runtime may attempt to retry the execution or provide the user with an
-   * appropriate alert.
-   */
-  void cancel(Throwable exception);
-
```
chillleader, line 91: I would keep this as `@Deprecated` for at least one more minor version so that existing connectors using it don't break unexpectedly. Can be made a no-op though.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/lifecycle/LaneKey.java
```diff
@@ -0,0 +1,52 @@
...
+public record ProcessKey(String tenantId, String bpmnProcessId) {
+
+  public static ProcessKey of(ProcessStateChanged event) {
+    return new ProcessKey(event.tenantId(), event.bpmnProcessId());
+  }
+
+  public static ProcessKey of(InboundConnectorElement element) {
+    return new ProcessKey(element.tenantId(), element.element().bpmnProcessId());
+  }
+
```
chillleader, line 37: Future concern: we have evidence that we will need to support cross-process deduplication. How can the new approach adapt to this new requirement, should we decide to ship it?

I suggest a thought experiment: If tomorrow we decide to deduplicate executables across processes, how will this record look like?

On that note, probably it makes sense to build this abstraction in a way that does not bind us to Process key as a unit of work (maybe by naming it differently e.g. `LaneKey` or similar)

chillleader:
@mathias-vandaele could you please provide a description for this PR stating the motivation behind these changes (would be great to have this for easier review & to make archeological quests easier in the future)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/lifecycle/LaneKey.java
mathias-vandaele: @chillleader We could call it `laneKey`, but that will not correct the problem of having cross-process deduplication, because if any process has an inbound connector that can be shared with another random process. the problem becomes way harder. 

I believe this is even impossible, imagine you have process A coming having inbound connector [a,b], then process B with [c,d], if process C comes with inbound connector [b, c], which lane should it end up into ? 

Or the entire logic of `activateBatch` should be outside of lane, and then we can use deduplication properties as a `laneKey` but that's a lot of refactor, not sure how that would work.

Or a single lane can solve this problem because everything is then executed sequentially, but that means activate will have to be non blocking again, and I don't like this.

Is it something that is important for us to implement ? why would we introduced such a complexity ? would there be real use case or gain from doing this ?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorContext.java
mathias-vandaele: That's correct, I will fix the PR once I have more time 👍🏼

.gitignore
mathias-vandaele: 👍🏼

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/exception/GlobalExceptionHandler.java
mathias-vandaele: That could be removed without problem yes 👍🏼

mathias-vandaele:
> @mathias-vandaele could you please provide a description for this PR stating the motivation behind these changes (would be great to have this for easier review & to make archeological quests easier in the future)

@chillleader Done 👍🏼

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/lifecycle/LaneKey.java
chillleader: @sbuettner has approached me with this idea a few weeks ago as there was a customer request apparently. Can't find it documented anywhere though. It might be related to this webhook issue https://github.com/camunda/connectors/issues/6843 but I don't see this point called out explicitly anywhere. Let's ask him for more context

PR #7233, opened 2026-05-13
fix: prefer connector virtual thread executor

### LINKED ISSUE #7186: Connectors does not start with virtual thread by default
__Describe the Bug__

on SM, we seem not to start connectors with virtual thread by default. 

https://jira.camunda.com/browse/SUPPORT-32696?atlLinkOrigin=c2xhY2staW50ZWdyYXRpb258aXNzdWU%3D

__Expected Behavior__

Starting with virtual thread should be the default, we should not have to add properties to make it work

I believe the issue comes from `@ConditionalOnMissingBean`

__Environment__

- OS: [e.g. Windows 7]
- Library version: [e.g. 2.0.0]
- Camunda version: [e.g. 8.0.4 (self-managed)] --> please also use affects/8.X label

johnBgood:
## Description

Fixes the default virtual-thread executor setup for the connector runtime.

The connector runtime already defines a `CamundaClientExecutorService` backed by virtual threads when `camunda.connector.virtual-threads.enabled` is missing or set to `true`. However, Camunda client auto-configuration can create its own fallback `CamundaClientExecutorService` first. Because both beans are guarded by `@ConditionalOnMissingBean`, the connector-provided virtual-thread executor may never be created.

This PR makes the connector outbound auto-configuration run before the Camunda client executor auto-configuration, ensuring that the connector-specific virtual-thread executor wins by default.

Changes included:

- Configure `OutboundConnectorsAutoConfiguration` before:
  - `CamundaAutoConfiguration`
  - `ExecutorServiceConfiguration`
- Rename the connector executor bean factory method to `connectorCamundaClientExecutorService` so the bean name is explicit and distinguishable from Camunda’s fallback bean.
- Add regression tests that verify:
  - the connector executor bean is created by default
  - Camunda’s fallback executor bean is not created when connector virtual threads are enabled
  - disabling `camunda.connector.virtual-threads.enabled` allows Camunda’s fallback executor to be created
  - the auto-configuration ordering includes the Camunda client executor configuration

## Related issues

closes https://github.com/camunda/connectors/issues/7186

## Checklist

### DIFF
--- a/connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java
+++ b/connector-runtime/spring-boot-starter-camunda-connectors/src/main/java/io/camunda/connector/runtime/OutboundConnectorsAutoConfiguration.java
@@ -18,6 +18,8 @@ package io.camunda.connector.runtime;
 
 import io.camunda.client.jobhandling.CamundaClientExecutorService;
 import io.camunda.client.metrics.MeteredCamundaClientExecutorService;
+import io.camunda.client.spring.configuration.CamundaAutoConfiguration;
+import io.camunda.client.spring.configuration.ExecutorServiceConfiguration;
 import io.camunda.connector.runtime.instances.InstanceForwardingConfiguration;
 import io.camunda.connector.runtime.outbound.OutboundConnectorRuntimeConfiguration;
 import io.micrometer.core.instrument.MeterRegistry;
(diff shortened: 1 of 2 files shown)


johnBgood:
@copilot apply changes based on the comments in [this thread](https://github.com/camunda/connectors/pull/7233#pullrequestreview-4281863229)

mathias-vandaele (approved):
Thank you 👍🏼

PR #7240, opened 2026-05-15
fix(agentic-ai): Fix socket and connection timeout config on LLM clients being too short on default settings

nikonovd:
## Description

* Added explicit connection and socket timeout configuration to LLM clients:
  * Anthropic
  * OpenAI
  * OpenAI compatible
  * Amazon Bedrock

#### Settings 
* Connection Timeout: 15 seconds
* Socket (Read) Timeout: Configured LLM timeout on connector or default timeout (3 minutes)

#### Reasoning
Due to the nature of the synchronous LLM APIs that are currently used in our implementations, bigger context windows might delay a response from the LLM provider, especially during high load or longer thinking scenarios.

In synchronous LLM APIs the entire response is generated before it is transmitted, thus the socket (read) timeout must be large enough to wait in such cases.

## Related issues

closes #7193 

## Checklist

- [x] Backport labels are added if these code changes should be backported. No backport label is added to the latest
  release, as this branch will be rebased onto main before the next release. Example backport labels:
    - `backport stable/8.8`: for changes that should be included in the next 8.8.x release.
    - **or** `backport release-8.8.7`: for changes that should be included in the specific release 8.8.7, and this
      *release has already been created*. The release branch will be merged back into stable/8.8 later, so the change
      will be included in future 8.8.x releases as well.
- [x] Tests/Integration tests for the changes have been added if applicable.
- [ ] If the change requires a documentation update, it has been added to the appropriate section in the documentation.

PR #7242, opened 2026-05-15
fix(agentic-ai): Fix socket and connection timeout config on LLM clients being too short on default settings (#7240)

nikonovd:
## Description

Backport of #7240 to `stable/8.8`

Deviations:
* Only applying socket and connection timeouts to Amazon Bedrock chat models
* Other providers in the original PR respect the timeout settings that are already configured, as L4J implementation applies it, when no custom client is being configured.

### DIFF
--- a/connectors/agentic-ai/pom.xml
+++ b/connectors/agentic-ai/pom.xml
@@ -136,6 +136,11 @@
     </dependency>
 
     <dependency>
+      <groupId>software.amazon.awssdk</groupId>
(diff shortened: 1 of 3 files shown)


nikonovd:
ℹ️ In 8.8, there's no proxy configuration, thus, the `timeout` setting applies on all providers besides Amazon Bedrock.

Therefor, we are creating the apache http client here with the respective timeout config.

maff (approved):

PR #7269, opened 2026-05-19
feat(documents): Unified source dropdown

### LINKED ISSUE #7057: Element template generator — document source dropdown for Document and List<Document> fields
## Summary

Auto-generate a unified document source selection UX in element templates for every connector that takes a `Document` or `List<Document>` as input. Replaces the current raw FEEL expression field with a discoverable dropdown.

**Parent epic:** camunda/product-hub#3224

Image

## Pattern

For `Document` fields, the ETG generates 8 properties:
- Source dropdown: **Camunda Document** / **Inline Content** / **External Document**
- Conditional fields per source (reference / content + filename + contentType / URL + filename)
- Hidden FEEL composer field — the only `zeebe:input` binding, assembles the final document JSON

For `List<Document>` fields, an additional **Document mode** toggle switches between single (full source dropdown) and multiple (FEEL array expression) modes.

Visible fields use `zeebe:property` (stored in BPMN XML, not sent to connector). Hidden composer uses `zeebe:input`.

## Per-connector configuration

Default behavior:
- Filename optional everywhere (UUID generated if missing)
- Content type hidden by default (Tika infers from filename), exposed for power users via a marker

Connectors can override via annotation on the `Document` field (e.g., `@DocumentInput(requireFileName = true)`):
- Storage connectors (S3, GCS, Azure, Drive, Box) — require filename (used as object key/name)
- Messaging connectors (Slack, Teams, SendGrid, Email Outbound) — require filename (attachment name)
- Bedrock, Agentic AI — require filename (used in API calls)
- Textract, IDP Extraction, Embeddings VDB — filename optional (process bytes only)

Full inventory in the high-level planning doc.

Create new version of each element template that is affected.

## Files to modify

- `element-template-generator/core/.../generator/java/util/TemplatePropertiesUtil.java` — detect `Document` and `List<Document>` types, generate the property pattern
- New `@DocumentInput` annotation in `connector-sdk/core` — for per-connector field requirements
- Element templates regenerated for all 13 affected connectors

## Not a breaking change

Old templates continue to work; only newly generated templates use the new UX. No runtime behavior changes.

## Test plan

- ETG unit tests: correct property generation for `Document`, `List<Document>`, with and without `@DocumentInput` overrides
- Generated template validates against element template schema
- Integration: generated templates work end-to-end in the Modeler

## Depends on

- camunda/connectors Step 1 issue (inline document type) — this step relies on the inline type existing

ztefanie:
## Description

Added new `@TemplateDocumentProperty` annotation, that adds unified input fields to connectors with document input. 

Adjusted the annotation of the affected connectors. Three connectors that require special handling will be done in a separate PR, issue [here](https://github.com/camunda/team-connectors/issues/1251).

Regenerated templates with version bump and engine version 8.10 as inline document is only available with 8.10.

See linked issue for desired output.

This are the four options:

image

## Related issues

closes #7057 

## QA

The new element templates need testing, are all fields for the document selection shown correctly? Is their runtime behavior as expected?

Connectors need testing for inline, external, camunda doc:
1. S3
2. Google cloud storage
3.  Azure blob storage
4. Google drive
5. Box

Connectors need testing for inline, external, camunda doc and lists:
1. Slack
2. Teams
3. SendGrid
4. Email
5. Bedrock
6. Embeddings Vector DB

As i did not change the runtime older templates should still work, with latest code changes.

## Checklist


connectors/slack/element-templates/slack-outbound-connector.json
```diff
@@ -186,12 +186,246 @@
...
     "type" : "String"
   }, {
-    "id" : "data.documents",
-    "label" : "attachments",
+    "id" : "data_documents_documentMode",
+    "label" : "Number of documents",
     "description" : "<a href=\"https://docs.camunda.io/docs/apis-tools/camunda-api-rest/specifications/upload-document-alpha/\">Camunda documents</a> can be added as attachments",
-    "optional" : true,
+    "value" : "single",
+    "group" : "message",
```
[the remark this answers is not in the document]
ztefanie: @copilot a new option "none" was added to the template, so this should be fixed.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
```diff
@@ -60,6 +61,11 @@ public static List<PropertyBuilder> extractTemplatePropertiesFromParameter(
 
     var type = parameter.getType();
     var annotation = parameter.getAnnotation(TemplateProperty.class);
+    var documentAnnotation = parameter.getAnnotation(TemplateDocumentProperty.class);
```
johnBgood, line 66: ❓ Having a check so that we don't have both annotations provided might prevent users from doing that? Otherwise, we could silently fall back to the Document property.

```java
// In extractTemplatePropertiesFromParameter and the field loop
if (annotation != null && documentAnnotation != null) {
  throw new IllegalStateException(
      "@TemplateProperty and @TemplateDocumentProperty are mutually exclusive on: " + parameter.getName());
}
```

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
```diff
@@ -185,6 +198,53 @@ public static List<PropertyGroupBuilder> groupProperties(List<PropertyBuilder> p
...
+    }
+    throw new IllegalStateException(
+        "@TemplateDocumentProperty on '"
+            + declaredName
+            + "' requires type Document or List<Document>, got "
+            + declaredType.getSimpleName());
+  }
+
+  private static Class<?> getListElementType(java.lang.reflect.AnnotatedElement element) {
+    java.lang.reflect.Type genericType = null;
```
johnBgood: 🔧 we could remove the import path

johnBgood (approved):
Great addition! ty

maff (commented):
The feature and implementation is nice in general, just my 2c: it adds visual clutter (especially the "Number of documents" dropdown) and complexity (for the user & in the SDK) to work around the fact that element templates do not support adding "multiples"/lists of a data structure (similar to multiple I/O mappings, listeners, ...).

Ideally there would be a single "Documents" block with a plus icon and I can select the type of document and fill out the type-related fields for every entry. Today I get the nice, assisted way only for a single document and need to fall back to completely raw mode as soon as there is a second one.

Not objecting the change, but maybe there is a chance to solve this more holistically before this lands as an official feature? 🤔 I think the "multiples" would be beneficial for different other use cases as well.

Related modeler issue: https://github.com/camunda/camunda-modeler/issues/1762

ztefanie:
> The feature and implementation is nice in general, just my 2c: it adds visual clutter (especially the "Number of documents" dropdown) and complexity (for the user & in the SDK) to work around the fact that element templates do not support adding "multiples"/lists of a data structure (similar to multiple I/O mappings, listeners, ...).
> 
> Ideally there would be a single "Documents" block with a plus icon and I can select the type of document and fill out the type-related fields for every entry. Today I get the nice, assisted way only for a single document and need to fall back to completely raw mode as soon as there is a second one.

Thanks for your feedback and thoughts. As this epic is already in implement and approved by UI and PM like this, i would stick to the current plan and how it is implemented with this PR. I get your point that this "plus icon" feature is nice and i am happy to see it in the product in a later iteration. For this epic this is out of scope and having an easy way to use inline and external documents as inputs is already a huge improvement. 

> 
> Not objecting the change, but maybe there is a chance to solve this more holistically before this lands as an official feature? 🤔 I think the "multiples" would be beneficial for different other use cases as well.
> 
> Related modeler issue: [camunda/camunda-modeler#1762](https://github.com/camunda/camunda-modeler/issues/1762)


When this plus icon feature is supported by the modeler, we can of course use it for the multiple documents input.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
ztefanie: Good point, i added it.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/TemplatePropertiesUtil.java
ztefanie: Removed all occurrences of this pattern.

Szik (approved):

PR #7272, opened 2026-05-19
feat(agentic-ai): Create agent instance during initialization on engine side

nikonovd:
## Description

Create an agent instance via the Camunda client primitives during first-time agent initialization.
The `Agent Instance` is a new first-class primitive on the engine side to ensure that the entire AI agent connector lifecycle is traced and captured.

This PR introduces the creation of the agent instance reference on the engine side, capturing the `agentInstanceKey` and storing it on the agent context metadata alongside `processInstanceKey` and `elementInstanceKey`.

## Related issues

closes #7198 

## Checklist


connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/resources/application.yml
```diff
@@ -0,0 +1,3 @@
+camunda:
+  process-test:
+    runtime-mode: shared
```
nikonovd, line 3: This property ensures that the CPT runtime is shared accross the test context, where applicable, improving test runtime.

nikonovd:
@gbetances089 this is the first PR on our side for the https://github.com/camunda/product-hub/issues/3462 EPIC.

We can definitely sync on that topic for Todos on our side. Besides that, @RobertDorlan might also be a good point of contact for further QA input.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
```diff
@@ -51,7 +57,10 @@ public AgentInitializationResult initializeAgent(AgentExecutionContext execution
         Optional.ofNullable(executionContext.initialToolCallResults()).orElseGet(List::of);
 
     return switch (agentContext.state()) {
-      case INITIALIZING -> initiateToolDiscovery(executionContext, agentContext, toolCallResults);
+      case INITIALIZING -> {
+        agentContext = ensureAgentInstanceRegistered(executionContext, agentContext);
```
maff: 🔧 Do we want to do this also for already running instances? Otherwise I'd suggest we do it in the `orElseGet` branch above (maybe extract into something like `createAgentContext()`), making the engine-side agent instance creation part of bootstrapping a new agent context.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetadata.java
```diff
@@ -14,13 +14,15 @@
  *
  * @param processDefinitionKey The key of the process definition this agent was initialized with
  * @param processInstanceKey The key of the process instance this agent is executing in
+ * @param agentInstanceKey The key of the agent instance created on the engine, if any
  */
 @AgenticAiRecord
-public record AgentMetadata(Long processDefinitionKey, Long processInstanceKey)
+public record AgentMetadata(
+    Long processDefinitionKey, Long processInstanceKey, Long agentInstanceKey)
```
maff: 🔧 We should mark the `agentInstanceKey` as `@Nullable` as it can be null. Currently only relevant with the builder implementations, but we should go over the codebase and properly annotate it with JSpecify annotations soon. Let's already do that now for new code, at least for models.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,75 @@
...
+        this::sleep);
+  }
+
+  private AgentInstanceKey executeCreate(InitialAgentInstanceData params) {
+    var command =
+        camundaClient
+            .newCreateAgentInstanceCommand()
+            .elementInstanceKey(params.elementInstanceKey())
+            .model(params.model())
+            .provider(params.provider())
```
maff: 🔧 Nit: hierarchy-wise, we should probably use `provider - model` - can we try to keep the pair in that consistent order?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/InitialAgentInstanceData.java
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.aiagent.agentinstance;
+
+import io.camunda.connector.agenticai.aiagent.model.AgentExecutionContext;
+
+public record InitialAgentInstanceData(
```
maff: 💭 Is it really necessary & worth the effort to abstract the `AgentExecutionContext` (which is already an abstraction over the Subprocess/Task data structures and its configurations)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
```diff
@@ -0,0 +1,54 @@
...
+
+public final class AgentInstanceErrorClassifier implements ErrorClassifier {
+
+  public static final AgentInstanceErrorClassifier INSTANCE = new AgentInstanceErrorClassifier();
+
+  @Override
+  public Decision classify(Throwable t) {
+    Throwable current = t;
+    while (current != null) {
+      if (current instanceof ClientHttpException httpEx) {
```
maff, line 22: 🔧 Does this also need gRPC support (see for example the job completion handler: https://github.com/camunda/camunda/blob/main/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/jobhandling/JobCallbackCommandWrapper.java#L198) or this this REST-only?

In general, I'm asking myself if it is worth the effort building the infrastructure for error handling/retryable logic depending on the response or if something like "I tried and failed 3 times, giving up now" (like the process definition client did) isn't enough for a starter. 🤔

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
```diff
@@ -0,0 +1,54 @@
...
+      }
+      current = cause;
+    }
+
+    return Decision.PERMANENT;
+  }
+
+  private static boolean isIoException(Throwable t) {
+    return t instanceof IOException;
+  }
```
maff: 🔧 Used only once, inline?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @Override
+  public Decision classify(Throwable t) {
+    Throwable current = t;
+    while (current != null) {
+      if (current instanceof ClientHttpException httpEx) {
+        int status = httpEx.code();
+        if (status == 404) {
+          return Decision.RETRYABLE;
+        }
```
maff, line 26: 🔧 Why is a 404 retryable for the agent instance?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/ProviderModelExtractor.java
```diff
@@ -0,0 +1,58 @@
...
+  public static String extract(ProviderConfiguration provider) {
+    var model = extractModelByProviderConfiguration(provider);
+
+    return model.orElseThrow(
+        () -> new IllegalStateException("Could not extract model from provider configuration"));
+  }
+
+  private static Optional<String> extractModelByProviderConfiguration(
+      ProviderConfiguration provider) {
+    return switch (provider) {
```
maff: 🔧 WDYT of turning this responsibility around and making a model accessor part of the `ProviderConfiguration` interface? This would allow us to get rid of this extractor completely and play nice with future custom providers.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,75 @@
...
+import static io.camunda.connector.agenticai.aiagent.agent.AgentErrorCodes.ERROR_CODE_AGENT_INSTANCE_CREATION_FAILED;
+
+import io.camunda.client.CamundaClient;
+import io.camunda.connector.agenticai.autoconfigure.AgenticAiConnectorsConfigurationProperties.AiAgentProperties.AgentInstanceProperties.RetriesProperties;
+import io.camunda.connector.agenticai.util.retry.CamundaApiRetry;
+import io.camunda.connector.agenticai.util.retry.CamundaApiRetry.FailureReason;
+import io.camunda.connector.api.error.ConnectorException;
+import java.time.Duration;
+
+public class CamundaAgentInstanceClient implements AgentInstanceClient {
```
maff, line 19: 💭  Thinking of a future where there will be more than only the creation, should the scope of this client be "tracing"/"tracking" and not "agent instance" to avoid having to autowire multiple smaller clients into the execution?

Package-wise, WDYT of grouping the infrastructure implementations for this initiative under a common  `metrics`, `tracing`, ... package instead of `agentinstance`?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsConfigurationProperties.java
```diff
@@ -21,7 +21,16 @@ public record AgenticAiConnectorsConfigurationProperties(
...
-  public record AiAgentProperties(@Valid @DefaultValue ChatModelProperties chatModel) {}
+  public record AiAgentProperties(
+      @Valid @DefaultValue ChatModelProperties chatModel,
+      @Valid @DefaultValue AgentInstanceProperties agentInstance) {
+
+    public record AgentInstanceProperties(@Valid @DefaultValue RetriesProperties retries) {
+      public record RetriesProperties(
+          @DefaultValue("4") @PositiveOrZero Integer maxRetries,
+          @DefaultValue("PT1S") Duration initialRetryDelay) {}
+    }
```
maff: 🔧 Can we reuse the existing `RetriesProperties` record? Probably it wouldn't be an issue to align on the defaults on the process definition side (not sure if there is a way to set different defaults this way while reusing the class)?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  private void setupCommandChain() {
+    when(camundaClient.newCreateAgentInstanceCommand()).thenReturn(step1);
+    when(step1.elementInstanceKey(anyLong())).thenReturn(step2);
+    when(step2.model(anyString())).thenReturn(step3);
+    when(step3.provider(anyString())).thenReturn(step4);
+    when(step4.systemPrompt(anyString())).thenReturn(step5);
+    when(step5.maxModelCalls(anyInt())).thenReturn(step5);
+  }
```
maff: 💭 I don't recall where, but IIRC there was a way to stub these command chains in a nicer way, (probably with `RETURNS_DEEP_STUBS`?). A light variant exists on `AdHocToolsSchemaFunctionIntegrationTest`, but without the builder steps involved.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/util/retry/ExponentialBackoffRetryTest.java
```diff
@@ -0,0 +1,57 @@
...
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.util.retry;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.time.Duration;
+import org.junit.jupiter.api.Test;
+
+class ExponentialBackoffRetryTest {
```
maff, line 18: 🔧 Could these test cases be condensed into a parameterized test?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,75 @@
...
+  private final RetriesProperties retriesProperties;
+
+  public CamundaAgentInstanceClient(
+      CamundaClient camundaClient, RetriesProperties retriesProperties) {
+    this.camundaClient = camundaClient;
+    this.retriesProperties = retriesProperties;
+  }
+
+  @Override
+  public AgentInstanceKey create(InitialAgentInstanceData params) {
```
maff: 💭 What happens if the job creates an agent instance, but fails job completion and is triggering a retry - effectively creating a second agent instance for the same element instance key. Is this handled engine-side? With a retry, potentially the params (e.g. the system prompt) might change if it depends on process variables which (I think) may be is different on a retry - which one wins in that case?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
nikonovd: I don't think it should be done for running instances, but TBH this point of the code is immediately reached after the initialization of the agent context above.

Before the initial job is completed the state will already be either `TOOL_DISCOVERY` or `READY` so afaik it won't ever be `INITIALIZING` from an incoming job 🤔

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
nikonovd: Because the API yields HTTP 404 in the case of the associated elementInstanceKey being not existing.

My assumption is that it could potentially happen based on their implementation so that's why I decided to make it retryable.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/ProviderModelExtractor.java
nikonovd: That was my fallback, but i kept this for now. I can turn it around for sure.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: I get your point. My current rationale is to keep it close to the domain name of the engine, so we can associate to which concept the client is belonging.

As long as we don't have anything like general tracing concerns embedded into our code, i'd like to keep the semantics close.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
nikonovd: Yep i also had the same intention, but `RETURNS_DEEB_STUBS` unfortunately i was always getting also a deep stub for the response that was yield by `send()`.

That's anyways still on my list, but it did not work as expected.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: The API on the engine side is designed in a way that it would return the existing agent instance in case it was already created (by a failing job). So i don't see an issue in this case.

Regarding changing parameters, that's about the incident case with somebody adjusting variables?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetadata.java
nikonovd: The fun part here is that we don't even use the same annotation (package) in our own codebase 😄 

So before we put another type of annotation (like JSpecify) inside, let's agree (with the connectors team), which standard we want to follow, because for instance for `@Nullable` I found after a brief check:

* `javax.annotation.Nullable`
* `jakarta.annotation.Nullable`
* `org.springframework.lang.Nullable`
* `org.jspecify.annotations.Nullable`

that are somehow used.

nikonovd: For this builder i'll adjust now, but i'll also create a general issue regarding NullMarking.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/InitialAgentInstanceData.java
nikonovd: I was hesitant to rely on the complex object but now with the `model()` method in place, it's much smoother.

The only caveat was that i had to introduce a custom `TestAgentExecutionContext` as `ProviderConfiguration` is sealed and not mockable.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
nikonovd: Good question, afaik there is only an API implemented for that. We can discuss the retry handling in general, but for now i morphed it from the `ProcessDefinitionClient` and we should have a general solution in place, not change it on one spot only. WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsConfigurationProperties.java
nikonovd: My mistake, will merge the two of them.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/util/retry/ExponentialBackoffRetryTest.java
nikonovd: Done

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
maff: Technically you are right, but I think it makes it unnecessarily complex to understand when this instance creation API call is emitted. I think we should make it a prerequisite of even creating the initial agent context instead of creating it and then ensuring it is registered. A `createAgentContext()` method doing the API call first would clearly indicate this.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
maff: > Regarding changing parameters, that's about the incident case with somebody adjusting variables?

Yes, or when they depend on process variables which change between retries.

A practical example: a typo in the model name would raise an incident which can be fixed by adjusting the input mapping & retrying the job. The created instance would still carry the wrong model name from the failed job.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
maff: Ok, let's keep it this way and maybe reconsider when we have an e2e solution. WDYT of at least grouping this and similar clients for this initiative under a common package (can still have sub-packages)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetadata.java
maff: Yes I know it's pretty mixed now :D The builder only looks at the annotatio name without considering the package, so technically it does not make a difference at the moment. However I think we should converge to the JSpecify annotations (introduced as default with Spring Boot 4, the monorepo is also [going into that direction](https://github.com/camunda/camunda/pull/51757)).

Let's use JSpecify for all new changes + adapt the whole codebase in a follow-up.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: Your ⛏️ is not possible here, because the client's step definition is hardwired and i can't change the order 👎

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
nikonovd: Got a better version now. Unfortunately `RETURNS_DEEP_STUBS` only works with exact arguments no matchers like `any()`.

Should be good now.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: As discussed, currently it's not possible to change those data entries anyways. We'll raise it with the core engine team.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: Agreed 👍 i moved the relevant classes for now into the root `agentinstance` package.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/ProviderModelExtractor.java
nikonovd: Inversed the relationship as discussed

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
nikonovd: As discussed let's keep the retry handling for now 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
nikonovd: As discussed let's keep the retry handling for now 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetadata.java
nikonovd: I added the annotation to `AgentMetadat` for now 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
nikonovd: Moved into the `orElseGet` path together with AgentContext population.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/processdefinition/ProcessDefinitionClient.java
```diff
@@ -7,76 +7,48 @@
...
-import org.slf4j.LoggerFactory;
 
 public class ProcessDefinitionClient {
-  private static final Logger LOGGER = LoggerFactory.getLogger(ProcessDefinitionClient.class);
   private static final String ERROR_CODE_AD_HOC_SUB_PROCESS_XML_FETCH_ERROR =
       "AD_HOC_SUB_PROCESS_XML_FETCH_ERROR";
 
   private final CamundaClient camundaClient;
-  private final RetriesProperties retriesProperties;
+  private final AgenticAiConnectorsConfigurationProperties.RetriesProperties retriesProperties;
```
maff: Nit: can we keep `RetriesProperties` and only switch the import?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,82 @@
...
+import io.camunda.connector.agenticai.autoconfigure.AgenticAiConnectorsConfigurationProperties;
+import io.camunda.connector.agenticai.util.retry.CamundaApiRetry;
+import io.camunda.connector.agenticai.util.retry.CamundaApiRetry.FailureReason;
+import io.camunda.connector.api.error.ConnectorException;
+import java.time.Duration;
+
+public class CamundaAgentInstanceClient implements AgentInstanceClient {
+
+  private final CamundaClient camundaClient;
+  private final AgenticAiConnectorsConfigurationProperties.RetriesProperties retriesProperties;
```
maff: Nit: same here regarding the import (as far as I could see affects every usage)

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,82 @@
...
+        this::sleep);
+  }
+
+  private AgentInstanceKey executeCreate(AgentExecutionContext agentExecutionContext) {
+    var command =
+        camundaClient
+            .newCreateAgentInstanceCommand()
+            .elementInstanceKey(agentExecutionContext.jobContext().getElementInstanceKey())
+            .model(agentExecutionContext.provider().model())
+            .provider(agentExecutionContext.provider().providerType())
```
maff, line 49: Nit: I think it was on the previous version as comment: can we try to follow the hierarchical order `provider -> model` consistently?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,82 @@
...
+        switch (reason) {
+          case PERMANENT_ERROR ->
+              "Failed to create agent instance for element instance key %d: %s"
+                  .formatted(elementInstanceKey, cause.getMessage());
+          case RETRIES_EXHAUSTED ->
+              "Failed to create agent instance for element instance key %d after %d attempt(s): %s"
+                  .formatted(elementInstanceKey, attempt, cause.getMessage());
+          case INTERRUPTED ->
+              "Interrupted while waiting to retry agent instance creation for element instance key %d"
+                  .formatted(elementInstanceKey);
```
maff: 💭 Is the element instance key actually relevant in the returned exception or is it already part of the incident metadata? Fine to keep it, but I think we don't have it on the other exception thrown from the execution.

Related to the need to make common metadata like this available in logs and errors for debugging.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -0,0 +1,82 @@
...
+        };
+    return new ConnectorException(ERROR_CODE_AGENT_INSTANCE_CREATION_FAILED, message, cause);
+  }
+
+  protected void sleep(Duration delay) throws InterruptedException {
+    Thread.sleep(delay.toMillis());
+  }
```
maff: 🔧 ~Can we use `Sleeper.threadSleep()` instead of this one?~ I've seen that it is overridden in a test. WDYT of optionally passing the `Sleeper` as constructor arg instead?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerTest.java
```diff
@@ -561,6 +558,80 @@ void skipsToolUpdateWhenProcessDefinitionKeyMatches() {
     }
   }
 
+  @Nested
+  @MockitoSettings(strictness = Strictness.LENIENT)
```
maff: Feel free to ignore for this scoped nested class: I find it always a bit dangerous to apply broadly as it swallows often valid cases where Mockito raises an issue. Any chance we can scope this to the respective mocks (`lenient().when()...`)?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerTest.java
```diff
@@ -183,15 +178,17 @@ void handlesNullInitialToolCallResults() {
   class ToolDiscoveryInitiation {
 
     private static final AgentContext AGENT_CONTEXT =
-        AgentContext.empty().withProperty("hello", "world");
+        AgentContext.empty()
+            .withProperty("hello", "world")
+            .withMetadata(new AgentMetadata(null, null, 99999L));
```
maff: 🔧 This is actually an invariant which shouldn't be possible and was already lacking in the existing test setup, right? Can we set this up with non-null values to be closer to actual execution data?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerTest.java
```diff
@@ -561,6 +558,80 @@ void skipsToolUpdateWhenProcessDefinitionKeyMatches() {
...
+
+      final var result =
+          (AgentContextInitializationResult) agentInitializer.initializeAgent(executionContext);
+
+      verify(agentInstanceClient, times(1)).create(any(AgentExecutionContext.class));
+      assertThat(result.agentContext().metadata().agentInstanceKey()).isEqualTo(12345L);
+    }
+
+    @Test
+    void shouldSkipAgentInstanceCreationWhenKeyAlreadyPresent() {
```
maff, line 586: 🔧 Should we also add a specific test case that it skips creation when the agent is pre-existing? It's implicitly covered e.g. in the migration test cases, but maybe we could either add a specific case here or a comment on the migration cases.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
```diff
@@ -0,0 +1,248 @@
...
+    verify(camundaClient, times(2)).newCreateAgentInstanceCommand();
+  }
+
+  @Test
+  void shouldThrowConnectorExceptionWithAttemptCountWhenAllRetriesAreExhausted() {
+    when(step5.execute()).thenThrow(new ClientHttpException(500, "Internal Server Error"));
+
+    assertThatThrownBy(() -> client.create(TestAgentExecutionContext.withLimits()))
+        .isInstanceOf(ConnectorException.class)
+        .satisfies(
```
maff: 🔧 Nit: could be condensed into `isInstanceOfSatisfying()`

maff (approved):
The API retry facility (and the PR overall) looks great now 💪 I added a couple of nitpicks & minor remarks, but nothing blocking.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/adhoctoolsschema/processdefinition/ProcessDefinitionClient.java
nikonovd: fixed 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: fixed 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: as mentioned, it's not possible due to the hard order of the steps in `CamundaClient` implementation.

Noted down for a follow up with the engine team.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: You're right, i aligned with the existing schema. The incident is anyways fixed to an element instance.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: fixed 👍

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerTest.java
nikonovd: fixed 👍

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerTest.java
nikonovd: fixed 👍

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
nikonovd: fixed 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
```diff
@@ -173,11 +176,21 @@ public AgentToolsResolver aiAgentToolsResolver(
     return new AgentToolsResolverImpl(toolsSchemaResolver, gatewayToolHandlers);
   }
 
+  @Bean
+  @ConditionalOnMissingBean
+  public AgentInstanceClient agentInstanceClient(
+      CamundaClient camundaClient, AgenticAiConnectorsConfigurationProperties configuration) {
+    return new CamundaAgentInstanceClient(
+        camundaClient, configuration.aiagent().agentInstance().retries(), Sleeper.threadSleep());
+  }
```
maff, line 185: 🔧 Can we list this new bean in `AgenticAiConnectorsAutoConfigurationTest`? Probably we should mention this concept somewhere in agent instructions.

nikonovd: Good point! I added the interface `AgentInstanceClient` to the bean definition list.

gbetances089 (approved):

gbetances089 (approved):
LGTM

Szik (approved):

PR #7273, opened 2026-05-19
fix(aws): restore STS dependency to re-enable IRSA authentication

### LINKED ISSUE #7271: AWS IRSA not supported by Connectors AWS SDK in 8.8+: switching from static credentials fails
__Describe the Bug__

When migrating SaaS connectors from using AWS static credentials (via environment variables) to IRSA (IAM Roles for Service Accounts), the connectors fail to access secrets via AWS Secret Manager. While IRSA configuration works on a dummy aws-cli pod with the same setup, Connectors AWS SDK appears to be incompatible with IRSA web token authentication.

For additional discussion and a reference, see: https://github.com/camunda/camunda-operator/issues/3704#issuecomment-4487575320

__Steps to Reproduce__

1. Deploy Camunda Connectors (version 8.8+) in SaaS with AWS Secret Manager access.
2. Use static AWS credentials as environment variables: confirm secrets access works.
3. Switch deployment to use IRSA, removing static credentials and providing the web token for IRSA.
4. Attempt to access secrets from AWS Secret Manager.

__Expected Behavior__

Credentials obtained via IRSA web token should allow the connectors to authenticate and access AWS Secret Manager, similar to the aws-cli pod.

__Environment__

- OS: (Kubernetes cluster, e.g., EKS)
- Library version: N/A (Connectors AWS SDK)
- Camunda version: 8.8+ (SaaS)
- Affects: 8.8+

__Possible Reasons for Issue__

- The AWS SDK used by connectors may not be configured to detect & use the projected IRSA web token file.
- Environmental variables or pod permissions may differ between the aws-cli test pod and connectors pod.
- The connectors' container image might lack necessary IAM or SDK configurations for IRSA.
- There may be an incompatibility with the authentication chain in the SDK version used by connectors.

Further investigation is required to pin down the root cause, including comparing environment variables and filesystem tokens in both successful (aws-cli pod) and failing (connectors pod) scenarios.

[View original Slack conversation](https://camunda.slack.com/archives/C02JLRNQQ05/p1779194352793209?thread_ts=1779194352.793209&cid=C02JLRNQQ05)

johnBgood:
## Summary

AWS IRSA (IAM Roles for Service Accounts) stopped working in 8.8+ because `software.amazon.awssdk:sts` was accidentally removed during the AWS SDK v1→v2 migration.

### Root cause

IRSA relies on `WebIdentityTokenFileCredentialsProvider`, which AWS SDK v2's `DefaultCredentialsProvider` includes in its credential chain **only if `software.amazon.awssdk:sts` is on the classpath** — it is loaded reflectively. When `sts` is absent the provider is silently skipped, causing all IRSA-based authentication to fall through without any obvious error.

**The regression was introduced by [`1c7d6bdfb`](https://github.com/camunda/connectors/commit/1c7d6bdfb8d66091afb7a8887723ae4921af5463)** (`chore: migrate AWS connectors to AWS SDK v2`), which correctly removed the v1 `com.amazonaws:aws-java-sdk-sts` but inadvertently also dropped the v2 `software.amazon.awssdk:sts` that had been added by [`3dee3ebb2`](https://github.com/camunda/connectors/commit/3dee3ebb261db7f3cab782ffc3ac873dc590ca16) (`feat(aws): Role can now be assumed within an AWS EKS cluster using IRSA`).

Confirmed absent via `mvn dependency:tree -Dincludes=software.amazon.awssdk:sts` against the SaaS bundle (produced no output before this fix).

### Changes

- **`connectors/aws/aws-base/pom.xml`** — restore `software.amazon.awssdk:sts`; fixes IRSA for all AWS connectors using `CredentialsProviderSupportV2` with default credentials (Lambda, SQS, SNS, EventBridge, Bedrock, etc.)
- **`secret-providers/aws/pom.xml`** — add `software.amazon.awssdk:sts`; fixes IRSA for `AwsSecretProvider` (which does not depend on `aws-base`)
- **`parent/pom.xml`** — add `sts` to `dependencyManagement` and `ignoredUnusedDeclaredDependencies` (it is a runtime-only classpath dep, not directly referenced in source)
- **`CredentialsProviderSupportV2Test`** (new) — classpath guard + baseline behaviour tests for the credentials provider
- **`AwsSecretProviderTest`** — same classpath guard for the secret provider path

### Test plan

- [x] `mvn -pl connectors/aws/aws-base,secret-providers/aws verify` — all 9 tests pass, `BUILD SUCCESS`
- [x] `mvn dependency:tree -Dincludes=software.amazon.awssdk:sts` against `camunda-saas-bundle` now shows `software.amazon.awssdk:sts:jar:2.44.8:compile`
- [ ] Deploy to a cluster with IRSA configured and verify AWS Secrets Manager access works without static credentials

Closes #7271

### DIFF
--- a/connectors/aws/aws-base/pom.xml
+++ b/connectors/aws/aws-base/pom.xml
@@ -42,6 +42,10 @@
     </dependency>
     <dependency>
       <groupId>software.amazon.awssdk</groupId>
+      <artifactId>sts</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>software.amazon.awssdk</groupId>
       <artifactId>aws-core</artifactId>
       <version>${version.aws-sdk2}</version>
     </dependency>
--- a/connectors/aws/aws-base/src/test/java/io/camunda/connector/aws/CredentialsProviderSupportV2Test.java
+++ b/connectors/aws/aws-base/src/test/java/io/camunda/connector/aws/CredentialsProviderSupportV2Test.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.aws;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+import io.camunda.connector.aws.model.impl.AwsAuthentication;
+import io.camunda.connector.aws.model.impl.AwsBaseConfiguration;
+import io.camunda.connector.aws.model.impl.AwsBaseRequest;
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+
+class CredentialsProviderSupportV2Test {
+
+  @Test
+  void staticCredentials_shouldReturnStaticCredentialsProvider() {
+    var request =
+        requestWith(new AwsAuthentication.AwsStaticCredentialsAuthentication("key", "secret"));
+    var provider = CredentialsProviderSupportV2.credentialsProvider(request);
+    assertThat(provider.resolveCredentials().accessKeyId()).isEqualTo("key");
+  }
+
+  @Test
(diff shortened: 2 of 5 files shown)


mathias-vandaele (approved):
Thank you 👍

ztefanie (approved):
Looks good. Do we want to QA this?

Szik (approved):

PR #7320, opened 2026-05-21
feat: Support IP based filtering for hostnames

sbuettner:
## Description

This pull request introduces a new host/IP validation module for connectors, designed to defend against SSRF and unauthorized outbound HTTP calls by validating hostnames and their resolved IP addresses against configurable CIDR allow/deny lists. The implementation is split into two Maven modules: one for the Jakarta Bean Validation annotation and another for the core validator logic. The validator is configurable and intended for integration with dependency injection frameworks like Spring.

## Related issues

closes camunda/security-testing-findings/issues/2

## Checklist

### DIFF
--- a/connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
+++ b/connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
@@ -82,7 +82,8 @@ public @interface VerifiedHost {
    * {@code example.com}) and validated directly without any URI parsing.
    *
    * <p>Set this to {@code true} when annotating properties that hold full request URLs (see {@code
-   * HttpCommonRequest} and {@code GraphQLRequest}); leave it {@code false} for plain host/IP fields.
+   * HttpCommonRequest} and {@code GraphQLRequest}); leave it {@code false} for plain host/IP
+   * fields.
    *
    * @return {@code true} if the value is a URI from which the host must be extracted before
    *     validation; {@code false} if the value is already a bare hostname
--- a/connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
+++ b/connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
@@ -69,5 +69,23 @@ public @interface VerifiedHost {
 
   Class<? extends Payload>[] payload() default {};
 
+  /**
+   * Indicates whether the annotated string is a full URI rather than a bare hostname.
+   *
+   * <p>The validator resolves and checks an IP address, so it needs a hostname — not a URI. When
+   * this flag is {@code true}, the {@code VerifiedHostValidator} first extracts the host component
+   * from the value (via {@link java.net.URI#getHost()}) before resolving and validating its IP
+   * address. For example, {@code https://example.com/path} is reduced to {@code example.com} prior
+   * to validation. If the value cannot be parsed as a URI with a host, it is validated as-is.
+   *
+   * <p>When {@code false} (the default), the annotated value is treated as a bare hostname (e.g.
+   * {@code example.com}) and validated directly without any URI parsing.
+   *
+   * <p>Set this to {@code true} when annotating properties that hold full request URLs (see {@code
+   * HttpCommonRequest} and {@code GraphQLRequest}); leave it {@code false} for plain host/IP fields.
+   *
+   * @return {@code true} if the value is a URI from which the host must be extracted before
+   *     validation; {@code false} if the value is already a bare hostname
+   */
   boolean isUri() default false;
 }
--- a/connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
+++ b/connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
@@ -38,6 +38,24 @@ import java.lang.annotation.Target;
  * options live on the stateful VerifiedHostValidator that the {@link
  * jakarta.validation.ConstraintValidatorFactory} returns. Register a Spring bean (or otherwise wire
  * up a constraint validator factory) holding the desired VerifiedHostValidator.Config.
+ *
+ * <h2>Security caveat — TOCTOU / DNS rebinding</h2>
+ *
+ * <p><strong>Validation is not a substitute for connection-time enforcement.</strong> This
+ * annotation runs at Bean Validation time — typically when a connector job is picked up — but the
+ * outbound HTTP/SMTP/IMAP/POP3 connection that uses the host happens later, after a separate DNS
+ * lookup. An attacker who controls the authoritative DNS for an allowed hostname can serve a
+ * short-TTL record that resolves to a <em>public</em> IP for the validation lookup and to a
+ * <em>private</em> IP (e.g. {@code 169.254.169.254}, RFC 1918) for the connection lookup. This is
+ * the classic DNS-rebinding / time-of-check-to-time-of-use (TOCTOU) gap and is inherent to any
+ * annotation-based host check.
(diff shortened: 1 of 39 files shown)


connector-commons/host-ip-validator/src/main/java/io/camunda/connector/hostvalidator/HostIpValidator.java
```diff
@@ -0,0 +1,151 @@
...
+      String host,
+      List<CidrRange> allowRanges,
+      List<CidrRange> denyRanges,
+      boolean unsafeAllowPrivateRanges,
+      boolean unsafeAllowLoopback)
+      throws UnknownHostException {
+    if (host == null || host.isBlank()) {
+      throw new IllegalArgumentException("host must not be null or blank");
+    }
+    InetAddress[] resolved = InetAddress.getAllByName(host);
```
johnBgood, line 117: I assume we don't want to cache this, to prevent some kind of attacks? Under high load, this might have some impact

connector-commons/host-ip-validator/src/main/java/io/camunda/connector/hostvalidator/HostIpValidator.java
```diff
@@ -0,0 +1,151 @@
...
+ *
+ * <p>Unlike smokescreen — which only checks the first resolved IP — this validator inspects every
+ * address returned by the resolver, to defend against DNS rebinding / multi-record SSRF.
+ */
+public final class HostIpValidator {
+
+  /** Standard RFC 1918 (IPv4) and RFC 4193 (IPv6 ULA) private ranges. */
+  private static final List<CidrRange> RFC_PRIVATE_RANGES =
+      List.of(
+          CidrRange.parse("10.0.0.0/8"),
```
johnBgood, line 39: from Claude:

>   CidrRange.contains() compares raw byte-array lengths. An address like ::ffff:10.0.0.1 is represented in Java as a 16-byte Inet6Address, while
  10.0.0.0/8 is stored as 4 bytes. The length mismatch causes contains() to return false, so ::ffff:10.0.0.1 passes the private-range check and is
  classified ALLOW_DEFAULT. A user input of http://[::ffff:10.0.0.1]/ would have its host extracted by URI.getHost() and pass validation despite
  pointing at an RFC-1918 address. isGlobalUnicast() does not catch this either (not loopback, not multicast, not link-local). Fix: in isPrivate(),
  normalize Inet6Address IPv4-mapped addresses to Inet4Address before the CIDR check, or add the IPv4-mapped ranges explicitly.

connector-runtime/connector-runtime-test/src/main/java/io/camunda/connector/runtime/test/outbound/OutboundConnectorContextBuilder.java
```diff
@@ -189,6 +197,45 @@ public OutboundConnectorContextBuilder validation(ValidationProvider validationP
...
+            } catch (ReflectiveOperationException e) {
+              throw new IllegalStateException(e);
+            }
+          }
+
+          @Override
+          public void releaseInstance(ConstraintValidator<?, ?> instance) {}
+        };
+
+    // TODO extract into test utils
```
johnBgood: :fyi: still some TODOs

connector-commons/host-ip-validator/src/main/java/io/camunda/connector/hostvalidator/HostIpValidator.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  /**
+   * Resolves {@code host} and throws {@link HostDeniedException} if any resolved address is denied.
+   *
+   * @throws UnknownHostException when the hostname cannot be resolved
+   * @throws HostDeniedException when any resolved IP is disallowed
+   */
+  public static void validate(
```
johnBgood, line 107: from Claude:

> Validation happens at job-pickup time; the actual HTTP call happens later. With a 1-second TTL, an attacker can flip DNS from a public IP (passes
  validation) to a private IP (used for the actual connection). This is inherent to annotation-based checking, but it should be called out
  prominently in the class Javadoc. True defense requires validating the resolved IP at connection time (e.g., a custom HTTP client that re-checks
  before connecting).

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
```diff
@@ -33,6 +34,7 @@ public class HttpCommonRequest {
   @FEEL
   @NotBlank
   @Pattern(regexp = "^(=|(http://|https://|secrets|\\{\\{).*$)", message = "Must be a http(s) URL")
+  @VerifiedHost(isUrl = true)
```
johnBgood: we might want to annotate the GraphQL and JDBC connectors as well

johnBgood (approved):
really nice overall 👍

connector-commons/host-ip-validator/src/main/java/io/camunda/connector/hostvalidator/HostIpValidator.java
sbuettner: Hey @johnBgood with impact you mean negative performance impact?

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/HttpCommonRequest.java
sbuettner: Added it to the GraphQL Connector. JDBC requires dedicated work that can happen in a separate topic.

connector-commons/host-ip-validator/src/main/java/io/camunda/connector/hostvalidator/HostIpValidator.java
johnBgood: yes, exactly

sbuettner: @johnBgood As we will only release this for 8.10 do you think we can release, measure and decide whether we want to add caching later on? As caching for DNS is done on so many levels (Java, OS, ...)

sbuettner:
@johnBgood can you give this another look please?

connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/HttpTests.java
```diff
@@ -70,7 +70,9 @@
     properties = {
       "spring.main.allow-bean-definition-overriding=true",
       "camunda.connector.webhook.enabled=true",
-      "camunda.connector.polling.enabled=false"
+      "camunda.connector.polling.enabled=false",
+      "camunda.connector.validation.hosts.enabled=true",
+      "camunda.connector.validation.hosts.unsafe-allow-loopback=true"
```
johnBgood, line 75: ❓ we added this, but I don't see any tests

connector-commons/host-ip-validator-annotation/src/main/java/io/camunda/connector/hostvalidator/VerifiedHost.java
```diff
@@ -0,0 +1,73 @@
...
+@Constraint(validatedBy = {}) // Configured via XML in the validator
+public @interface VerifiedHost {
+
+  String message() default "Host is not allowed";
+
+  Class<?>[] groups() default {};
+
+  Class<? extends Payload>[] payload() default {};
+
+  boolean isUri() default false;
```
johnBgood, line 91: ❓ Maybe we can add some javadoc here, as a user it's not always when I should use this and why

connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/HttpTests.java
sbuettner: Added a dedicated e2e test now.

Szik (approved):

PR #7337, opened 2026-05-26
other(native-operations): Add validator for new native operations fields

### LINKED ISSUE #7338: Add new rules for native operations to validator
Add new rules to element template validator, to verify all native operations metadata we ship is valid.

ztefanie:
## Description
- Added new rules to verify all native operations metadata of all connectors are valid.
- Currently all connectors are in the ignore list, as there is also a check that connectors must have this metadata, so in the future when creating new connectors you get an error, if missing the metadata. 

## Related issues

closes https://github.com/camunda/connectors/issues/7338

## Checklist

### DIFF
--- a/element-template-generator/validator/src/main/java/io/camunda/connector/validator/core/ConditionEvaluator.java
+++ b/element-template-generator/validator/src/main/java/io/camunda/connector/validator/core/ConditionEvaluator.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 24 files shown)


element-template-generator/validator/src/main/java/io/camunda/connector/validator/core/ConditionEvaluator.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Evaluates an element-template property {@code condition} against an assignment of property
+ * values.
+ */
+public final class ConditionEvaluator {
+
+  private ConditionEvaluator() {}
+
+  public static boolean evaluate(JsonNode condition, Map<String, String> assignment) {
```
mathias-vandaele, line 30: I don't understand why this is called evaluate() 🤔 it seems to verify something, I find it hard to understand what `true` or `false` means when returned here

mathias-vandaele (approved):

element-template-generator/validator/src/main/java/io/camunda/connector/validator/core/ConditionEvaluator.java
ztefanie: Given a condition and an assigment it returns if this assigments evaluates the condition to true or false and returns this.

PR #7372, opened 2026-05-28
feat(agentic-ai): Update agent instance metrics on LLM turns

nikonovd:
## Description

  Implements real-time agent instance status and metrics reporting for AI Agent connectors. The engine exposes an agent instance resource per process instance; this PR populates it by PATCHing status (`THINKING`, `TOOL_CALLING`, `IDLE`, `TOOL_DISCOVERY`) and cumulative metrics (`modelCalls`, `tokenUsage`, `toolCalls`) at the right moments during agent execution.

  ## What's new

  **`AgentInstanceClient`:**  wraps the Zeebe `UpdateAgentInstanceCommand` API. `CamundaAgentInstanceClient` is the production implementation, with retry logic (`CamundaApiRetry`) and error classification (`AgentInstanceErrorClassifier`) to handle transient failures and superseded-job 404s gracefully.

  **Agent instance lifecycle:**
  - On first turn: `create()`: registers the agent instance with the engine (see #7272)
  - Before each LLM call: PATCH `THINKING`
  - After LLM response: PATCH `TOOL_CALLING` + metrics delta (tool calls present) or `IDLE` + metrics delta (final answer)
  - On tool discovery job completion: PATCH `TOOL_DISCOVERY`

  **Metrics source:** `toolCalls` is counted from the LLM response (`assistantMessage.toolCalls().size()`). This is symmetric with `modelCalls` (one per LLM invocation) and attributes what the agent *requested* on the current turn.

  **Hybrid deferred/immediate PATCH timing** (ADR 005, option B3): `UpdateAgentInstanceCommand` requires a live element instance, which dies when its containing BPMN element closes. The timing depends on flavor:

  | Condition | Element instance fate | PATCH timing |
  |---|---|---|
  | Outbound connector (task) | Dies on every job completion | Immediate, before complete command |
  | Job worker, final turn (`completionConditionFulfilled=true`) | AHSP completes | Immediate |
  | Job worker, intermediate turn (`completionConditionFulfilled=false`) | AHSP remains active | Deferred via `onJobCompleted()` |

  `BaseAgentRequestHandler` delegates the timing decision to `shouldUpdateAgentInstanceBeforeJobCompletion(AgentResponse)`, implemented per flavor. On the deferred path, a metrics `AgentJobCompletionListener` is composed with the existing store listener, `onJobCompletionFailed()` strips `toolCalls` from the delta (tool elements were never activated on a superseded job).

  ## Tests

  - `CamundaAgentInstanceClientTest`: full unit coverage of create/update, retry, error classification
  - `JobWorkerAgentRequestHandlerTest`: covers deferred path (intermediate turn), immediate path (final turn), failure/supersession stripping
  - `OutboundConnectorAgentRequestHandlerTest`: covers always-immediate path
  - E2E: `L4JAiAgentJobWorkerAgentInstanceTests` and `L4JAiAgentConnectorAgentInstanceTests`,  verify the full ordered sequence of `AgentInstanceClient` calls for single and multi-tool-call scenarios against an embedded Zeebe engine with WireMocked LLM responses
## Related issues

closes #7199 

## Checklist


connectors/agentic-ai/docs/adr/005-agent-instance-metrics-reporting.md
```diff
@@ -0,0 +1,141 @@
...
+
+The Camunda engine exposes an agent instance resource that tracks the runtime state of an AI Agent execution.
+Connectors can push status and metric updates via a PATCH endpoint wrapped by `AgentInstanceClient.update()`.
+Before this work, the agent instance received no updates during execution — its metrics were always zero and its
+status was always stale.
+
+The question is: **which metrics to report, when to report them, and from which source to derive the values**.
+
+Three counters are in scope: `modelCalls`, `tokenUsage`, and `toolCalls`, corresponding to the fields of
+`AgentMetrics`. Two status transitions are in scope: `THINKING` (LLM call in progress) and `TOOL_CALLING` /
```
maff, line 19: 💭 I'm honestly not a big fan of `thinking` as a technical term. It may look nice on a UI, but do you think we can find a more technically appropriate term here (processing, calling LLM, ...)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
```diff
@@ -116,9 +119,10 @@ private AgentInitializationResult initiateGatewayToolDiscovery(
 
     if (!CollectionUtils.isEmpty(initiationResult.toolDiscoveryToolCalls())) {
       // execute tool discovery tool calls before agent is ready for requests
+      final var updatedState = transitionAgentStateToToolDiscovery(executionContext, agentContext);
```
maff: 🔧 I'd prefer if we either re-assign `agentContext` for consistency or indicate in the name that this is the agent context, not only the state.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
```diff
@@ -13,7 +13,19 @@
...
 public final class AgentInstanceErrorClassifier implements ErrorClassifier {
 
-  public static final AgentInstanceErrorClassifier INSTANCE = new AgentInstanceErrorClassifier();
+  /** For create: 404 is retryable (idempotent re-create against a transient lookup race). */
+  public static final AgentInstanceErrorClassifier FOR_CREATE =
+      new AgentInstanceErrorClassifier(true);
+
+  /** For update: 404 is permanent (the instance must exist; 404 = not found = bug). */
+  public static final AgentInstanceErrorClassifier FOR_UPDATE =
+      new AgentInstanceErrorClassifier(false);
```
maff, line 22: 💭 Is any or the update calls retryable? This will lead to wrong data without some kind of idempotency, correct? Can we try to document/annotate that somewhere that this is still missing?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetrics.java
```diff
@@ -16,14 +16,17 @@
...
     implements AgentMetricsBuilder.With {
   public AgentMetrics {
     if (modelCalls < 0) {
       throw new IllegalArgumentException("Model calls must be non-negative");
     }
-
     Objects.requireNonNull(tokenUsage, "Token usage must not be null");
+    if (toolCalls < 0) {
+      throw new IllegalArgumentException("Tool calls must be non-negative");
+    }
```
maff, line 29: 🔧  Are we sure about that? What about agents directly returning a response without tool calls? How would that work with existing processes?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetrics.java
```diff
@@ -34,6 +37,18 @@ public AgentMetrics incrementTokenUsage(TokenUsage additionalTokenUsage) {
...
+    return withToolCalls(toolCalls + additionalToolCalls);
+  }
+
+  public AgentMetrics minus(AgentMetrics other) {
+    return builder()
+        .modelCalls(modelCalls - other.modelCalls())
+        .tokenUsage(tokenUsage.minus(other.tokenUsage()))
+        .toolCalls(toolCalls - other.toolCalls())
+        .build();
+  }
```
maff, line 50: 🔧 Should we add some kind of guard here to prevent subtracting from an empty agent metrics instance (leading to `-2` model calls for example). It will probably never happen when the code and data is correct, but it would probably be good to catch the invariant early.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
```diff
@@ -131,6 +135,16 @@ private AgentInitializationResult initiateGatewayToolDiscovery(
...
 
+  private AgentContext transitionAgentStateToToolDiscovery(
+      AgentExecutionContext executionContext, AgentContext agentContext) {
+    final var discoveryAgentContext = agentContext.withState(AgentState.TOOL_DISCOVERY);
+    agentInstanceClient.update(
+        executionContext,
+        discoveryAgentContext,
+        AgentInstanceUpdateRequest.statusOnly(AgentInstanceUpdateStatus.TOOL_DISCOVERY));
+    return discoveryAgentContext;
+  }
```
maff: 🔧 We discussed this already & I see lots of potential of drifting apart with these synchronous external entity updates (entity updated, but job fails to complete) - basically what is described [here](https://github.com/camunda/connectors/blob/636628cc7f3a46df14a855af3fbb63dbfaa46924/connectors/agentic-ai/docs/tracing/design.md#3-why-events-not-entity-updates).

WDYT of also moving this to the completion listener? The listener could be be transported by `AgentInitializationResult` and wired into the produced response.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -122,72 +141,131 @@ private AgentResponse processConversation(
...
         executionContext, agentContext, runtimeMemory, executionContext.systemPrompt());
 
-    // update memory with user messages/tool call responses
-    if (LOGGER.isDebugEnabled()) {
-      LOGGER.debug(
-          "Adding user messages including the following {} tool call results: {}",
-          toolCallResults.size(),
-          toolCallResults.stream().map(tcr -> Pair.of(tcr.id(), tcr.name())).toList());
-    }
+    logToolCallResults(toolCallResults);
```
maff: 💭 More of a personal preference, but with the scattered `log*` methods I find the actual logic which is happening in here pretty hard to follow. Is something like`logToolCallResults` actually necessary?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -122,72 +141,131 @@ private AgentResponse processConversation(
...
-    agentContext = frameworkChatResponse.agentContext();
+  private AgentMetrics notifyThinking(C executionContext, AgentContext agentContext) {
+    LOGGER.debug("Notifying agent instance: status=THINKING before LLM call");
+    final var snapshot = agentContext.metrics();
+    agentInstanceClient.update(
+        executionContext,
+        agentContext,
+        AgentInstanceUpdateRequest.statusOnly(AgentInstanceUpdateStatus.THINKING));
+    return snapshot;
+  }
```
maff: 🔧 I see the same issue with the synchronous external entity update here which might also lead to race conditions between 2 jobs in case of an event subprocess. TBH I would question the need for this status update in general - can an agent not always be processing/thinking while the job is running, maybe with a special case while tool discovery is running? 

While the agent is executing, in reality it's either job is executing ("thinking" - takes long on the LLM call, everything else is instant) or job is not executing ("idle"). I'm not convinced it's worth the effort of making an additional HTTP call for this case, rather I think we should question the need for this status field in general. Basically the UI can just show "thinking" while the job is executing.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -122,72 +141,131 @@ private AgentResponse processConversation(
...
+
+  private AgentContext updateAgentInstanceMetricsAndStatus(
+      C executionContext, AgentContext agentContext, AgentMetrics preChatMetrics) {
+    final var metricsDelta = agentContext.metrics().minus(preChatMetrics);
+    final var nextStatus = nextAgentInstanceState(metricsDelta.toolCalls());
+
+    logAgentInstancePostLlmUpdate(nextStatus, metricsDelta);
+
+    // Exclude toolCalls: tool calls are reported on job completion
+    // so superseded/failed jobs don't inflate the count.
```
maff: 🔧 Do we even need this update then as it's basically running within a couple of ms before the job completion callback?

maff (commented):
I didn't go over every detail, but some high-level feedback: with these synchronous status updates I think we're approaching the risk of race conditions + we introduce at least 2 additional HTTP calls (apart from the one on completion) on every turn, only for updating a transient status field which most of the time is a 1:1 mapping to "job is executing". I fear that with this approach, we head directly into what is documented [here](https://github.com/camunda/connectors/blob/636628cc7f3a46df14a855af3fbb63dbfaa46924/connectors/agentic-ai/docs/tracing/design.md#3-why-events-not-entity-updates) as major risk & I think we should re-consider the need for this status field. 

Regarding retries & idempotency: can we add a prominent comment somewhere stating that this is an intermediary state missing the idempotency part? Patching delta updates with retries, but without idempotency pones a high risk of incorrect data.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetrics.java
nikonovd: Fair point, although unlikely it should be correct and i added the validation to `AgentMetrics`

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: We can split it up, but it would be favorable to update the LLM token consumption in a best effort way asap to avoid losing those in case of a failing job.

However, after thinking about it, we could add a failure listener to just add the token usage after a failed job. WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: As discussed, we will have racing jobs only in case of timeout exceedance and we'll investigate that with a Heartbeat check PoC.

I'll create a task soon for that. For now i would like to explore the responsive solution in the alpha state and pivot if needed.

WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: I'll see what i can do, my preference is that we shouldn't scatter the method too much with logs. That's why I extracted it a bit.

Logging the tool call results here might be a bit overshooting, but I'd then prefer to add some logging to the messages handler.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializerImpl.java
nikonovd: As mentioned in https://github.com/camunda/connectors/pull/7372#discussion_r3323638821 i'll check that we add as much as possible into the completion listener 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetrics.java
nikonovd: Good question! IMO an existing deserialization should then default the missing value to `0` but i'll verify that.

connectors/agentic-ai/docs/adr/005-agent-instance-metrics-reporting.md
nikonovd: Generally it's an enum value on the engine side. We can adapt on our code logic something you'd prefer, but would need to map it anyways.

nikonovd:
@maff i updated the PR according to our discussions here and in slack.

Check the ADR and PR description for noteworth changes

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -122,72 +150,105 @@ private AgentResponse processConversation(
...
-    final var frameworkChatResponse =
-        framework.executeChatRequest(executionContext, agentContext, runtimeMemory);
-    agentContext = frameworkChatResponse.agentContext();
+  private void notifyThinking(C executionContext, AgentContext agentContext) {
+    LOGGER.debug("Notifying agent instance: status=THINKING before LLM call");
+    agentInstanceClient.update(
+        executionContext,
+        agentContext,
+        AgentInstanceUpdateRequest.statusOnly(AgentInstanceUpdateStatus.THINKING));
+  }
```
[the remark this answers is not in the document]
nikonovd: we fixed `notifyMetrics` to rethrow and not swallow the exception if the call is still inside the agent loop.

Only swallow exceptions inside the job completion listeners

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
nikonovd: Currently it's not idempotent, but for the HTTP errors that are visible in the API spec, we can't ensure that data is published and then potentially might be lost.

I'd currently have the retry in place still, as the consequences are similar, either lost or duplicate data.

connectors/agentic-ai/docs/adr/005-agent-instance-metrics-reporting.md
maff: Yes, I'm aware that this comes from the API, but maybe we can reconsider a proper technical term at this stage, also to not clash with actual thinking/reasoning. E.g. on Claude Managed Agents the agent the respective comparable status would be [running](https://platform.claude.com/docs/en/api/beta/sessions/retrieve#beta_managed_agents_session.status).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerAgentInstanceTests.java
```diff
@@ -0,0 +1,382 @@
...
+            JobWorkerAgentResponseAssert.assertThat(agentResponse)
+                .isReady()
+                .hasAgentInstanceKey()
+                .hasMetrics(new AgentMetrics(2, new AgentMetrics.TokenUsage(25, 45), 1)));
+
+    // Turn 1: THINKING is synchronous; TOOL_CALLING is deferred (AHSP stays open after
+    // completionConditionFulfilled=false → element instance survives job completion).
+    // Turn 2: both THINKING and IDLE are synchronous (AHSP closes after
+    // completionConditionFulfilled=true
+    // → element instance dies at job completion → immediate PATCH required).
```
maff: "dies" sounds like an error case - "is completed"?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: I removed this log for now.

I'll create a general followup issue to have good logging practices embedded into the connector handler generally.

WDYT?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/langchain4j/jobworker/L4JAiAgentJobWorkerAgentInstanceTests.java
```diff
@@ -0,0 +1,382 @@
...
+
+    stubFor(
+        post(urlPathEqualTo("/v1/chat/completions"))
+            .inScenario("multi-tool-turns")
+            .whenScenarioStateIs("turn-2")
+            .willReturn(
+                aResponse()
+                    .withStatus(200)
+                    .withHeader("Content-Type", "application/json")
+                    .withBody(toolCallResponseBody()))
```
maff: 🔧 Apparently it works as it is, but could we use 2 distinct tool call IDs for the 2 tool call responses to be closer to productive data? In the messages handler there is some logic to match tool call requests and results based on the ID.

connectors/agentic-ai/docs/adr/005-agent-instance-metrics-reporting.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. **Immediate (post-LLM PATCH)** - push `{modelCalls, tokenUsage, toolCalls}` and the outcome status
+   (`TOOL_CALLING` or `IDLE`) as soon as the LLM responds, before completing the job.
+2. **Deferred (job-completion callback)** - report all post-LLM metrics and status only after Zeebe has
+   resolved the job completion command, via `AgentJobCompletionListener`. Superseded or failed jobs report
+   partial metrics (no `toolCalls`) or skip reporting entirely depending on the failure type.
+
+## Decision Outcome
+
+**Option A2** (LLM output), **B3** (hybrid - deferred when safe, immediate otherwise), **C2** (interface static).
```
maff: 🔧 Maybe I'm holding it wrong, but should B3 and C2 be part of the considered options? 😅

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -53,10 +67,71 @@ private AgentInstanceKey executeCreate(AgentExecutionContext agentExecutionConte
...
+    CamundaApiRetry.execute(
+        () -> executeUpdate(executionContext, metadata.agentInstanceKey(), request),
+        AgentInstanceErrorClassifier.FOR_UPDATE,
+        retriesProperties.maxRetries(),
+        retriesProperties.initialRetryDelay(),
+        this::buildUpdateException,
+        sleeper);
+  }
+
+  private Void executeUpdate(
```
maff:
```suggestion
  private void executeUpdate(
```

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetrics.java
maff: Never mind - I misread as `toolCalls == 0`

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandlerTest.java
```diff
@@ -107,15 +111,19 @@ class JobWorkerAgentRequestHandlerTest {
 
   @BeforeEach
   void setUp() {
-    conversationStore = spy(new InProcessConversationStore());
+    ConversationStore conversationStore = spy(new InProcessConversationStore());
     doReturn(conversationStore)
         .when(conversationStoreRegistry)
         .getConversationStore(eq(agentExecutionContext), any(AgentContext.class));
+    lenient()
```
maff, line 118: 🔧 Is this mock really needed here or could we add it to the new tests/the tests where it is needed, potentially without making it lenient?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandlerTest.java
```diff
@@ -427,6 +435,214 @@ void silentlyCompletesJobWhenNoUserMessageContent() {
...
+    // completionConditionFulfilled=false → deferred PATCH is safe
+    mockSystemPrompt();
+    mockUserPrompt(USER_PROMPT_CONFIGURATION_WITH_TOOLS, List.of());
+    when(agentInitializer.initializeAgent(agentExecutionContext))
+        .thenReturn(new AgentContextInitializationResult(INITIAL_AGENT_CONTEXT, List.of()));
+    final var assistantMessage = AssistantMessage.builder().toolCalls(TOOL_CALLS).build();
+    mockFrameworkExecution(assistantMessage);
+    when(gatewayToolHandlers.transformToolCalls(any(AgentContext.class), anyList()))
+        .thenAnswer(i -> i.getArgument(1));
+    lenient()
```
maff, line 450: 🔧 Why does this one have to be lenient (also for the other tests)?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/OutboundConnectorAgentRequestHandlerTest.java
```diff
@@ -344,20 +519,24 @@ static Stream<MemoryConfiguration> memoryConfigurationsWithoutMaxMessages() {
         null, new MemoryConfiguration(new InProcessMemoryStorageConfiguration(), null));
   }
 
-  private void mockSystemPrompt(SystemPromptConfiguration systemPromptConfiguration) {
-    when(agentExecutionContext.systemPrompt()).thenReturn(systemPromptConfiguration);
+  private void mockSystemPrompt() {
+    when(agentExecutionContext.systemPrompt())
+        .thenReturn(OutboundConnectorAgentRequestHandlerTest.SYSTEM_PROMPT_CONFIGURATION);
```
maff: Doesn't need the FQCN, correct (also below)?

```suggestion
        .thenReturn(SYSTEM_PROMPT_CONFIGURATION);
```

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifierTest.java
```diff
@@ -53,4 +65,27 @@ static Stream<Throwable> permanentExceptions() {
...
+        new ClientHttpException(502, "Bad Gateway"),
+        new ClientHttpException(503, "Service Unavailable"),
+        new IOException("connection refused"),
+        new InterruptedIOException("timeout"),
+        new RuntimeException(new IOException("transport")));
+  }
+
+  static Stream<Throwable> permanentExceptionsIncluding404() {
+    return Stream.of(
+        new ClientHttpException(404, "Not Found"),
```
maff: 🔧 Nit: can we put this in order?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
```diff
@@ -156,6 +181,167 @@ void shouldThrowConnectorExceptionWithAttemptCountWhenAllRetriesAreExhausted() {
     verify(camundaClient, times(5)).newCreateAgentInstanceCommand();
   }
 
+  // --- update() tests ---
```
maff: 🔧 Nit: should we put the operations into nested classes instead of the prefixes and delimiter comment?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceUpdateRequestTest.java
```diff
@@ -0,0 +1,57 @@
...
+package io.camunda.connector.agenticai.aiagent.agentinstance;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.camunda.client.api.command.AgentInstanceUpdateStatus;
+import io.camunda.connector.agenticai.aiagent.model.AgentMetrics;
+import io.camunda.connector.agenticai.aiagent.model.AgentMetrics.TokenUsage;
+import org.junit.jupiter.api.Test;
+
+class AgentInstanceUpdateRequestTest {
```
maff: 🔧 Do we actually need these tests, given the builder is also implicitly tested in the `CamundaAgentInstanceClient` tests?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
```diff
@@ -156,6 +181,167 @@ void shouldThrowConnectorExceptionWithAttemptCountWhenAllRetriesAreExhausted() {
...
+            Duration.ofSeconds(8));
+    verify(camundaClient, times(5)).newUpdateAgentInstanceCommand(AGENT_INSTANCE_KEY);
+  }
+
+  private static AgentContext agentContextWithInstanceKey() {
+    final var metadata = new AgentMetadata(1L, 2L, AGENT_INSTANCE_KEY);
+    return AgentContext.builder().state(AgentState.READY).metadata(metadata).build();
+  }
+
+  // --- create() tests ---
```
maff: 🔧 I guess this belongs before the update tests (see comment regarding nested ones).

maff (approved):
Some nitpicks, primarily on tests, but otherwise LGTM after our discussion today and the planned changes for idempotency with the job activation count! 🚀 Approving as none of the comments is blocking.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: It was a bit trickier due to the `return null` but i fixed as suggested 👍

connectors/agentic-ai/docs/adr/005-agent-instance-metrics-reporting.md
nikonovd: C2 is orphaned so i removed it.

Added B3 as a considered option 👍

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
nikonovd: done

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClientTest.java
nikonovd: reorganized into nested test classes

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifierTest.java
nikonovd: done

nikonovd:
@gbetances089 @RobertDorlan could you please check and approve accordingly so we can merge it, thank you 🚀

connectors/agentic-ai/docs/adr/005-agent-instance-metrics-reporting.md
nikonovd: As discussed, will take that to the bigger round for discussion.

RobertDorlan (approved):
Thanks for the thorough work on this, @nikonovd 🙌
Reviewed the implementation and test coverage — everything looks solid.

gbetances089 (approved):

PR #7404, opened 2026-06-03
fix(runtime): generic variable deserialization in operation bindings

### LINKED ISSUE #7377: Generic type parameters in @Variable bindings erased — List<T> deserializes to List<LinkedHashMap>, causing late ClassCastException
__Describe the Bug__

When using the `@Operation`/`@Variable` pattern with a parameter typed as a parameterized collection (e.g. `List<MyRecord>`), the runtime erases the generic type and Jackson deserializes the JSON array as `List<LinkedHashMap>`. Accessing an element then throws `ClassCastException: LinkedHashMap cannot be cast to MyRecord` — at the first dereference inside the connector method, not at the framework boundary, so the failure is reported far from its root cause.

The cause is in [`ConnectorOperations#map`](https://github.com/camunda/connectors/blob/main/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java):

```java
return new ParameterDescriptor.Variable<>(
    getVariableName(variableAnnotation),
    parameter.getType(),               // <-- raw Class, no generic info
    variableAnnotation.required());
```

`Parameter.getType()` returns the raw `Class<?>` (`List.class`), so by the time Jackson gets the binding target, the element type is gone. Jackson falls back to `LinkedHashMap`, and `Span s : spans` blows up when it tries to read fields.

Arrays work around it (`Span[]` retains component type at runtime), but the array workaround is unidiomatic and surprising — code authors reach for `List<T>` first.

__Steps to Reproduce__

1. Define a connector with the `@Operation`/`@Variable` annotation style:

   ```java
   @OutboundConnector(
       name = "RepoConnector",
       type = "io.example:repro:1",
       inputVariables = {"spans"})
   public class ReproConnector implements OutboundConnectorProvider {

     public record Span(int start, int end) {}

     @Operation(id = "consume")
     public String consume(@Variable("spans") List<Span> spans) {
       return spans.stream().mapToInt(Span::start).sum() + ""; // CCE here
     }
   }
   ```

2. Activate the connector with a job whose `spans` variable is a JSON array, e.g. `[{"start":1,"end":3},{"start":5,"end":7}]`.

3. The job fails with:

   ```
   java.lang.ClassCastException: class java.util.LinkedHashMap cannot be cast to class ReproConnector$Span
   ```

__Expected Behavior__

The framework should use the parameter's full type (e.g. via `Parameter.getParameterizedType()` and Jackson's `TypeFactory.constructType(...)` / `ObjectMapper.convertValue(value, javaType)`) so that `List<Span>`, `Map<String, Span>`, `Optional<Span>`, etc. all deserialize to the correctly typed elements. Either that, or document the limitation prominently and reject methods with parameterized generic `@Variable` parameters at startup, instead of letting them fail at first dereference.

__Environment__

- Camunda Connectors: 8.9.x (`connector-runtime-core`, `spring-boot-starter-camunda-connectors`)
- Camunda: 8.9.5
- JDK: 24

__Workaround__

Replace `List<T>` parameters with arrays (`T[]`) — arrays preserve their component type through erasure, so Jackson resolves them correctly.

```java
@Operation(id = "consume")
public String consume(@Variable("spans") Span[] spans) { ... }
```

azan-baloch:
## Description

Fix generic deserialization for `@Variable` bindings in operation-based connectors.

Previously, parameterized types such as `List<T>` were deserialized using the raw type, which results in values being mapped as `LinkedHashMap` instead of the expected type.

Parameterized types such as List<T> now preserve their type information during variable binding and are deserialized correctly.

Added a regression test for generic collection variables

## Related issues

closes #7377

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/ConnectorOperations.java
@@ -69,7 +69,9 @@ public record ConnectorOperations(Object connector, Map<String, OperationInvoker
     if (parameter.isAnnotationPresent(Variable.class)) {
       Variable variableAnnotation = parameter.getAnnotation(Variable.class);
       return new ParameterDescriptor.Variable<>(
-          getVariableName(variableAnnotation), parameter.getType(), variableAnnotation.required());
+          getVariableName(variableAnnotation),
+          parameter.getParameterizedType(),
+          variableAnnotation.required());
     } else if (parameter.getType().equals(OutboundConnectorContext.class)) {
       return new Context();
     } else if (parameter.isAnnotationPresent(Header.class)) {
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/OperationInvoker.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/operation/OperationInvoker.java
@@ -19,6 +19,7 @@ package io.camunda.connector.runtime.core.outbound.operation;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonPointer;
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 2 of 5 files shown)


sbuettner (approved):
🥇

johnBgood (approved):

PR #7431, opened 2026-06-05
fix: allow reset of inbound connectors in FailedToActivate/ConnectorNotRegistered state

johnBgood:
### Problem

Calling `POST /inbound-instances/executables/{id}/reset` on a connector that was **not healthy** (e.g. missing secret causing `FailedToActivate`) returned a confusing **404** instead of attempting to reset the connector.

### Root cause

The chain of failures that produced the 404:

1. `InboundInstancesService.resetExecutable()` first calls `findExecutable()` — this succeeds because `query()` returns all states including `FailedToActivate`.
2. It then calls `executableRegistry.reset(id)` which internally calls `validateResettable()`.
3. `validateResettable()` only accepted `Activated` or `Cancelled` states and threw `IllegalStateException` for anything else → **Spring returned 500**.
4. In the multi-instance setup, `InstanceForwardingHttpClient` filters out all responses with status ≥ 400. When every instance returns 4xx/5xx (500 from the owning instance, 404 from others), no successful response remains.
5. The controller's `Optional.ofNullable(...).orElseThrow(DataNotFoundException)` then fires → **404**.

### Fix

- **`validateResettable()`** now also allows `FailedToActivate`. `ConnectorNotRegistered` and `InvalidDefinition` remain non-resettable: `ConnectorNotRegistered` means the connector type isn't available in this runtime — resetting would immediately produce another `ConnectorNotRegistered`, so the right action is a runtime restart with the correct JARs. `InvalidDefinition` requires fixing the process model itself.
- **New `extractConnectorElements()` helper** replaces `extractContext()` and covers all three resettable states: for `Activated`/`Cancelled` it reads from the context; for `FailedToActivate` it reads from `data.connectorElements()`.
- The existing `RESTART` state-transition pipeline already handles `FailedToActivate` correctly: `deactivateSingle()` is a no-op for non-`Activated` entries, and `activateBatch()` retries activation from scratch — so if the root cause is fixed (e.g. secret added), the reset succeeds.

### State machine after this fix

| State | Resettable? | Notes |
|---|---|---|
| `Activated` | ✅ | Was already supported |
| `Cancelled` | ✅ | Was already supported |
| `FailedToActivate` | ✅ | **New** — retries activation (e.g. after adding a missing secret) |
| `ConnectorNotRegistered` | ❌ | Runtime doesn't have the connector type; needs runtime restart |
| `InvalidDefinition` | ❌ | Broken process model config; needs redeployment |

## Related issues

closes #

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
@@ -339,7 +339,7 @@ public class InboundExecutableRegistryImpl implements InboundExecutableRegistry 
       throw new IllegalStateException(
           "Cannot reset executable '"
               + id
-              + "': not in a resettable state, but was: "
+              + "': not in a resettable state, was: "
               + current.getClass().getSimpleName());
     }
     return current;
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
@@ -30,6 +30,7 @@ import io.camunda.connector.runtime.inbound.executable.InboundExecutableStateTra
 import io.camunda.connector.runtime.inbound.executable.InboundExecutableStateTransitionService.TargetState;
 import io.camunda.connector.runtime.inbound.executable.RegisteredExecutable.Activated;
 import io.camunda.connector.runtime.inbound.executable.RegisteredExecutable.Cancelled;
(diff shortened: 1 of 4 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/InboundExecutableRegistryImpl.java
```diff
@@ -332,39 +333,42 @@ private RegisteredExecutable validateResettable(ExecutableId id) {
...
     }
-    if (!(current instanceof Activated) && !(current instanceof Cancelled)) {
+    if (!(current instanceof Activated)
+        && !(current instanceof Cancelled)
+        && !(current instanceof FailedToActivate)) {
       throw new IllegalStateException(
           "Cannot reset executable '"
               + id
-              + "': must be in Activated or Cancelled state, but was: "
+              + "': not in a resettable state, but was: "
```
chillleader:
```suggestion
              + "': not in a resettable state, was: "
```

chillleader (approved):
nice catch

PR #7444, opened 2026-06-08
refactor(agentic-ai): Refactor agent initialization around domain oriented concepts

nikonovd:
## Description

Improve code organization around the agent initialization to reflect actual logic around domain oriented concepts.

Key improvements:
* AgentInitializer result typing reflects the actual intent for the conversation, either
   * Defer the conversation
   * Proceed with the conversation
   * Dispatch a tool discovery
* Introduce `AgentConversation` domain model as the center model around the agent lifecycle
  * In this PR it begins to wrap several other objects to ensure that the data flow is clearer 

## Checklist

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
@@ -14,7 +14,7 @@ import java.util.List;
 public sealed interface AgentInitializationResult {
 
   /** Agent provisioned and tools resolved: proceed to the conversation phase. */
-  record ReadyToConverse(AgentContext agentContext, List<ToolCallResult> engineToolCallResults)
+  record ReadyToConverse(AgentContext agentContext, List<ToolCallResult> toolCallResults)
       implements AgentInitializationResult {}
 
   /** Gateway tools require discovery: dispatch these discovery tool calls, then await results. */
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
@@ -7,28 +7,23 @@
 package io.camunda.connector.agenticai.aiagent.agent;
 
 import io.camunda.connector.agenticai.aiagent.model.AgentContext;
-import io.camunda.connector.agenticai.aiagent.model.AgentResponse;
+import io.camunda.connector.agenticai.model.tool.ToolCall;
 import io.camunda.connector.agenticai.model.tool.ToolCallResult;
 import java.util.List;
-import org.jspecify.annotations.Nullable;
 
 public sealed interface AgentInitializationResult
-    permits AgentInitializationResult.AgentContextInitializationResult,
-        AgentInitializationResult.AgentDiscoveryInProgressInitializationResult,
-        AgentInitializationResult.AgentResponseInitializationResult {
+    permits AgentInitializationResult.ReadyToConverse,
+        AgentInitializationResult.DiscoverTools,
+        AgentInitializationResult.DeferConversation {
 
-  record AgentContextInitializationResult(
-      AgentContext agentContext, List<ToolCallResult> toolCallResults)
+  /** Agent provisioned and tools resolved: proceed to the conversation phase. */
+  record ReadyToConverse(AgentContext agentContext, List<ToolCallResult> engineToolCallResults)
       implements AgentInitializationResult {}
 
-  record AgentDiscoveryInProgressInitializationResult() implements AgentInitializationResult {}
+  /** Gateway tools require discovery: dispatch these discovery tool calls, then await results. */
+  record DiscoverTools(AgentContext agentContext, List<ToolCall> toolDiscoveryToolCalls)
+      implements AgentInitializationResult {}
 
-  record AgentResponseInitializationResult(
-      AgentResponse agentResponse, @Nullable AgentJobCompletionListener completionListener)
-      implements AgentInitializationResult {
-
-    AgentResponseInitializationResult(AgentResponse agentResponse) {
-      this(agentResponse, null);
-    }
-  }
+  /** Discovery already dispatched; results not all present yet — no-op this turn. */
+  record DeferConversation() implements AgentInitializationResult {}
 }
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
@@ -11,10 +11,7 @@ import io.camunda.connector.agenticai.model.tool.ToolCall;
 import io.camunda.connector.agenticai.model.tool.ToolCallResult;
 import java.util.List;
 
-public sealed interface AgentInitializationResult
-    permits AgentInitializationResult.ReadyToConverse,
-        AgentInitializationResult.DiscoverTools,
-        AgentInitializationResult.DeferConversation {
+public sealed interface AgentInitializationResult {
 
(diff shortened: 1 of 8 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
```diff
@@ -7,28 +7,23 @@
...
 import java.util.List;
-import org.jspecify.annotations.Nullable;
 
 public sealed interface AgentInitializationResult
-    permits AgentInitializationResult.AgentContextInitializationResult,
-        AgentInitializationResult.AgentDiscoveryInProgressInitializationResult,
-        AgentInitializationResult.AgentResponseInitializationResult {
+    permits AgentInitializationResult.ReadyToConverse,
+        AgentInitializationResult.DiscoverTools,
+        AgentInitializationResult.DeferConversation {
```
maff: I missed this on the initial implementation, but actually we could omit these permits completely as the implementation are nested classes of the interface.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -75,48 +77,41 @@ public BaseAgentRequestHandler(
...
   public R handleRequest(final C executionContext) {
-    final var agentInitializationResult = agentInitializer.initializeAgent(executionContext);
-    return switch (agentInitializationResult) {
-      // directly return agent response if needed (e.g. tool discovery tool calls before calling the
-      // LLM)
-      case AgentResponseInitializationResult(
-              AgentResponse agentResponse,
-              AgentJobCompletionListener initCompletionListener) -> {
+    return switch (agentInitializer.initializeAgent(executionContext)) {
+      case ReadyToConverse(var agentContext, var engineToolCallResults) -> {
```
maff: 🔧 Nit: do you think we could list the different states in their order of appearance, so it's easier to follow the workflow (discover, defer, ready)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -75,48 +77,41 @@ public BaseAgentRequestHandler(
...
         LOGGER.debug(
-            "AI Agent initialization returned direct response including {} tool calls. Completing job without further processing.",
-            agentResponse.toolCalls().size());
-        yield buildConnectorResponse(executionContext, agentResponse, initCompletionListener);
+            "Handling agent request with {} tool call results",
+            engineToolCallResults != null ? engineToolCallResults.size() : 0);
```
maff: 🔧 Do we need this null-check here or can we expect (maybe also enforce) this to never be null?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
```diff
@@ -7,28 +7,23 @@
...
 
-  record AgentContextInitializationResult(
-      AgentContext agentContext, List<ToolCallResult> toolCallResults)
+  /** Agent provisioned and tools resolved: proceed to the conversation phase. */
+  record ReadyToConverse(AgentContext agentContext, List<ToolCallResult> engineToolCallResults)
```
maff: 🔧 Do we need the naming `engineToolCallResults` outside of the initializer logic? IIRC the "initial" tool call results were introduced to distinguish between "unprocessed" (before gateway logic) and the actually handled tool call results. I'm wondering if we need an additional naming concept of "engine tool call results" or if we could have unprocessed/initial and everything else (happy to rename the former to something else than "initial" if it makes sense). Anything after initialization could stay `toolCallResults` and would not need to carry the `engine` part.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -0,0 +1,74 @@
...
+    return context.state();
+  }
+
+  public List<ToolDefinition> toolDefinitions() {
+    return context.toolDefinitions();
+  }
+
+  public RuntimeMemory messageMemory() {
+    return messageMemory;
+  }
```
maff: 🔧 I'm a bit hesitant to introduce a second name for the same concept. WDYT of naming this `runtimeMemory()`?

maff (commented):
The approach looks great & no blocking comments, assuming that e2e tests still pass. I added some minor comments, primarily on naming.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
nikonovd: So, if I understood correctly, it's the discovered tool call results for the initialization phase, and everything inside the ready loop is actually then normal tool call results, right? Did I miss some point here? Sorry, we had this discussion once already, but I forgot what we've discussed 😄 

Happy to adjust namings afterwards 👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: Good hint, fixed.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInitializationResult.java
maff: Yes exactly, to have Claude answer it in a bit more prose 😄 :

> The distinction is about the lifecycle of tool call results during the agent initialization phase.
> 
> initialToolCallResults() is the raw list of ToolCallResult objects read directly from process variables — before the AgentInitializerImpl state machine has processed them.
> 
> The key is the TOOL_DISCOVERY state. When the agent has gateway tools (MCP, A2A) that need to discover their available tools, the process is:
> 
> 1. Agent state = INITIALIZING → initializer emits discovery tool calls (e.g., "list available MCP tools"), sets state to TOOL_DISCOVERY, and completes the job
> 2. Those discovery tool calls execute, results come back in toolCallResults process variable
> 3. On the next job, initialToolCallResults() contains both:
>   - Gateway discovery results (consumed internally by handleToolDiscoveryResults())
>   - Any actual user-originated tool call results that may have arrived alongside them
> 
> handleToolDiscoveryResults() strips out the discovery results and returns remainingToolCallResults() — the subset that didn't belong to the discovery phase. Only these "remaining" results flow forward to BaseAgentRequestHandler and ultimately into the LLM conversation context.
> 
> So: "initial" = straight from process variables, unfiltered. The "normal" (non-initial) tool call results are remainingToolCallResults() — after the discovery phase has consumed what it needed. In the common case where there's no gateway tool discovery in progress, initialToolCallResults and the remaining results are identical (the full list passes through unchanged).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -0,0 +1,70 @@
...
+ * Transient domain aggregate representing the agent's conversation state for one turn.
+ *
+ * <p>{@code context} and {@code runtimeMemory} are <em>restored</em> from persisted state at the
+ * start of each turn. {@code toolCallResults} is <em>this turn's new engine input</em>, primed to
+ * be folded into the conversation by the message handler.
+ */
+public final class AgentConversation {
+
+  private AgentContext context;
+  private final RuntimeMemory messageMemory;
```
maff: 🔧 This one is still called `messageMemory`

maff (approved):
Small naming leftover, otherwise LGTM, great work! 🚀

PR #7463, opened 2026-06-09
test(agentic-ai): refactor e2e tests to framework agnostic approach

### LINKED ISSUE #7400: Update e2e tests to use HTTP mocking/Wiremock instead of Mockito mocks
## What should we do?

The e2e tests in `connectors-e2e-test/connectors-e2e-test-agentic-ai/` (the `langchain4j/` subtree) currently mock the LLM layer by replacing the `ChatModelFactory` Spring bean with a Mockito mock and returning a stubbed `CloseableChatModel`:

```java
@MockitoBean private ChatModelFactory chatModelFactory;
@Mock protected CloseableChatModel chatModel;

@BeforeEach
void setUp() {
  when(chatModelFactory.createChatModel(any())).thenReturn(chatModel);
}
```

These are LangChain4j-specific implementation details. Once the LangChain4j framework layer is replaced with native provider implementations, both `ChatModelFactory` and `CloseableChatModel` will no longer exist, and all tests built on top of this pattern will stop working.

Before any refactoring starts, replace the mock with WireMock stubs that intercept at the HTTP level. The connector should be configured to point at the WireMock server (e.g. via the `openai-compatible` provider pointing to `http://localhost:<wiremock-port>`), and the test stubs the expected HTTP endpoint and asserts on the request body instead.

## Why?

The refactoring will remove `ChatModelFactory` and the LangChain4j providers entirely. Tests that mock these classes either disappear along with the code they test or need to be rewritten anyway. Rewriting them first, before the refactoring, means:
- The new native implementations are covered by the same behavioral tests from day one.
- Reviewers can verify the refactoring produces the same wire-level behavior rather than taking it on faith.

nikonovd:
### Motivation

  The previous e2e test suite mocked the LLM at the Java level (Mockito ChatModel + ArgumentCaptor<ChatRequest>), which bypassed LangChain4j's HTTP transport entirely. The tests also carried LangChain4j-specific class names and types throughout, making them useless as regression baselines when switching AI frameworks.

#### WireMock LLM transport (was: Mockito mock)
All 20 L4JAiAgent* test classes now stub the LLM at the HTTP level via `OpenAiChatModelStubs` (`POST /v1/chat/completions`) and verify recorded requests through `RecordedLlmConversation`.
The real LangChain4j OpenAiChatModel executes the full HTTP round-trip in every test.

#### Framework-agnostic refactoring

  - `AiAgentToolSpecifications` interface with `record ExpectedTool(String name, String description)` replaces `Langchain4JAiAgentToolSpecifications` and its LangChain4j `ToolSpecification` constants
  - `AiAgentA2aIntegrationTestSupport` with `record A2aExpectedConversation(...)` replaces the `List<ChatMessage>`-returning old helper; callers now receive plain strings
  - `ToolCallResultDocumentAssertions` updated: `ExtractedDocument.contentBlockAssertion` changed from `Consumer<Content>` to `Consumer<JsonNode>`, unused `ChatMessage` overload removed, `decodeJsonContentBlock` eliminated
  - `BaseWireMockL4JAiAgentConnectorTest` + `BaseWireMockL4JAiAgentJobWorkerTest` merged into the new `BaseAiAgentConnectorTest` / `BaseAiAgentJobWorkerTest` in flattened packages with a single inheritance level
  - All test classes renamed `L4JAiAgent*` to `AiAgent*` and moved away from `langchain4j` package.{outboundconnector,jobworker,wiremock,common} to {outboundconnector,jobworker,wiremock,common}
  - Zero `dev.langchain4j.*` imports remain in test code
## Related issues

closes #7400

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -6,65 +6,280 @@
...
+   * {@code pendingInputMessages} and appends a new {@link ConversationTurn}.
+   *
+   * @throws IllegalStateException if called before {@link #applyInput}
+   */
+  public AgentConversation ingest(
+      AssistantMessage assistantMessage, AgentMetrics.TokenUsage tokenUsage) {
+    if (pendingInputMessages == null) {
+      throw new IllegalStateException("ingest() called before applyInput()");
+    }
+    int nextKey = turns.size() + 1;
```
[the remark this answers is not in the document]
nikonovd: It was the wrong base branch, this code is not part of this branch.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/RecordedLlmConversation.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Inspects the chat completion requests the connector actually sent to WireMock. This replaces the
+ * Mockito {@code ArgumentCaptor<ChatRequest>} used by the previous mock-based tests: instead of
+ * capturing the in-process LangChain4j request object, we parse the real OpenAI-compatible request
+ * bodies recorded by WireMock.
```
maff: 🔧 I know LLMs tend to do this a lot, but I'm not a fan of commenting what previous implementations did here (we can find that out from the history) - I'd suggest we focus on the status quo with the comments and not mention previous implementations (or L4J at all?).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/common/AiAgentA2aIntegrationTestSupport.java
```diff
@@ -47,12 +41,17 @@
 import org.springframework.core.io.Resource;
 import org.springframework.util.StreamUtils;
 
-public class L4JAiAgentA2aIntegrationTestSupport {
+/**
+ * Framework-agnostic test support for A2A integration tests. Provides the expected conversation
+ * data as a plain record rather than LangChain4j message objects, so it can serve as a regression
+ * baseline when switching AI frameworks.
+ */
```
maff: 🔧 I also noted it further down below: I'd suggest we don't include future plans or references to the past in these class level docblocks, but rather describe what it does at the current state (if necessary).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/AiAgentToolSpecifications.java
```diff
@@ -0,0 +1,108 @@
...
+public interface AiAgentToolSpecifications {
+
+  /** Lightweight representation of an expected tool visible to the LLM. */
+  record ExpectedTool(String name, String description) {}
+
+  // ── Standard tools ──────────────────────────────────────────────────────────
+
+  ExpectedTool GET_DATE_AND_TIME_TOOL =
+      new ExpectedTool(
+          "GetDateAndTime", "Returns the current date and time including the timezone.");
```
maff: 🔧 If I interpret this correctly, we are now only asserting tool definition names + descriptions and not on the full tool schema. We might be able to cover it with unit tests, but as this also involves the mechanics how the cluster provides the tool specs to the client I would highly prefer if we could keep asserting the whole definition.

WDYT of having a helper in the recorded conversation which is able to return the tool definitions in a unified format we can compare against (e.g. our `ToolDefinition` model or something custom to these tests)? In case we add support for other wire formats (e.g. Anthropic API), the `RecordedConversation` implementation could return the tools in this unified format.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/BaseAiAgentTest.java
```diff
@@ -49,20 +52,39 @@ public abstract class BaseAiAgentTest extends BaseAgenticAiTest {
...
       new AtomicReference<>(Collections.emptyMap());
 
+  /**
+   * Ordered user feedback consumed one entry per {@code user_feedback} job invocation. This
+   * decouples per-turn feedback from the model call, which the WireMock-based tests need (the
+   * previous Mockito mock set feedback as a side effect of each {@code chat()} call). When the
+   * queue is exhausted, the worker falls back to {@link #userFeedbackVariables}.
+   */
+  protected final ConcurrentLinkedQueue<Map<String, Object>> userFeedbackQueue =
+      new ConcurrentLinkedQueue<>();
```
maff: 🔧 TBH when I initially built this I was either not aware of CPT's job worker mocking possibilities or they were not there yet. Is there maybe a simpler way now to handle these feedback jobs?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/OpenAiChatModelStubs.java
```diff
@@ -0,0 +1,182 @@
...
+ * <p>A conversation is expressed as an ordered list of {@link Turn turns}. Each model call advances
+ * a WireMock {@link Scenario} state, so sequential calls deterministically receive the next turn's
+ * response regardless of request body. This mirrors the previous Mockito {@code
+ * mockChatInteractions} queue.
+ *
+ * <p>The real ad-hoc sub-process tools execute between turns (as before), so tool <em>results</em>
+ * are produced by the engine — only the assistant turns (text and tool-call requests) are stubbed
+ * here.
+ */
+public final class OpenAiChatModelStubs {
```
maff: 🔧 Should we call this `OpenAiCompletionsChatModelStubs`?

Do you think we'd have a use case to add different provider support in the future? In that case I'd also name `RecordedLlmConversation` accordingly and include OpenAI completions.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/OpenAiChatModelStubs.java
```diff
@@ -0,0 +1,182 @@
...
+     * calls.
+     */
+    public static Turn toolCalls(
+        String text, int promptTokens, int completionTokens, ToolCall... toolCalls) {
+      return new Turn(text, Arrays.asList(toolCalls), promptTokens, completionTokens);
+    }
+
+    private String toResponseJson() {
+      final ObjectNode root = OBJECT_MAPPER.createObjectNode();
+      root.put("id", "chatcmpl-test");
```
maff: 🔧 Can we return a unique ID per message here to reflect production? Otherwise this might bite us if we consume the message ID somewhere.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/OpenAiChatModelStubs.java
```diff
@@ -0,0 +1,182 @@
...
+      } else {
+        message.putNull("content");
+      }
+
+      if (!toolCalls.isEmpty()) {
+        final ArrayNode toolCallsNode = message.putArray("tool_calls");
+        for (final ToolCall toolCall : toolCalls) {
+          final ObjectNode toolCallNode = toolCallsNode.addObject();
+          toolCallNode.put("id", toolCall.id());
+          toolCallNode.put("type", "function");
```
maff: 🔧 Is there a reason we use the lower level Jackson APIs vs. serializing some record structure? Asking primarily for readability reasons as it's not super straightforward to follow the produced JSON.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/BaseAiAgentTest.java
nikonovd: That's valid, because for user tasks we mostly added the good solution in 8.9 with conditional behaviors. I'll apply those.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/RecordedLlmConversation.java
nikonovd: Valid point 👍

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/common/AiAgentA2aIntegrationTestSupport.java
nikonovd: Valid point 👍

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/OpenAiChatModelStubs.java
nikonovd: Potentially we should cover all of those in a single test case about that specific topic. Agree about the renaming.

nikonovd: Generally i would like to not create conversation abstractions here for various providers now, but rather when we'd introduce new providers in e2e tests then.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/AiAgentToolSpecifications.java
nikonovd: Good point and i did some fixed around that and now all tool specs inside that class also contain inputSchema again.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/jobworker/AiAgentJobWorkerMcpIntegrationTests.java
```diff
@@ -166,24 +157,19 @@ void executesMcpToolDiscovery() throws Exception {
 
     CamundaAssert.assertThat(zeebeTest.getProcessInstanceEvent())
         .hasCompletedElements(
-            // entering the AI Agent
```
maff, line 169: 💭 Is there a reason to remove these comments as they might still be helpful to understand the processed elements?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/jobworker/AiAgentJobWorkerMcpIntegrationTests.java
```diff
@@ -65,17 +57,20 @@
...
 import org.mockito.Captor;
 import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.core.io.Resource;
 import org.springframework.test.context.TestPropertySource;
 import org.springframework.test.context.bean.override.mockito.MockitoBean;
 
 @SlowTest
+@ExtendWith(MockitoExtension.class)
```
maff, line 71: 💭 Why is this necessary now?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/jobworker/AiAgentJobWorkerA2aIntegrationTests.java
```diff
@@ -190,17 +151,34 @@ void handlesA2aToolCalls(WireMockRuntimeInfo wireMock) throws IOException {
...
+        recorded.lastRequest(),
+        ExpectedMessage.system(expectedSystemPrompt()),
+        ExpectedMessage.user(testSupport.initialUserPrompt),
+        ExpectedMessage.assistantWithToolCalls(
+            conversation.aiToolCallMessageText(),
+            "SuperfluxProduct",
+            "A2A_Travel_Agent",
+            "A2A_Weather_Agent",
+            "A2A_Exchange_Rate_Agent"),
+        ExpectedMessage.toolResult("aaa111", conversation.superfluxToolResult()),
```
maff: 💭 Until now we tried to consistently call these "tool call" & "tool call result" - do you think we could keep the `toolCallResult` naming across these new changes to keep a single naming concept?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/AiAgentToolSpecifications.java
```diff
@@ -0,0 +1,173 @@
...
+import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.springframework.core.io.ClassPathResource;
+
+/**
+ * Framework-agnostic tool specification constants for AI agent e2e tests. Uses {@link
```
maff: 🔧 Is framework-agnostic still relevant here?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/ToolCallResultDocumentAssertions.java
```diff
@@ -97,39 +96,41 @@ private static <T> T parseFirstDocumentNode(
...
+   * stays in sync with the production format. The content-block {@link JsonNode} is passed directly
+   * to each {@link ExtractedDocument#contentBlockAssertion()}.
    */
   public static void assertExtractedDocumentsUserMessage(
-      ChatMessage message, ExtractedDocument... expectedDocuments) {
-    assertThat(message).isInstanceOf(UserMessage.class);
-    final var contents = ((UserMessage) message).contents();
-
-    assertThat(contents)
+      RecordedMessage message, ExtractedDocument... expectedDocuments) {
```
maff, line 108: 🔧 I'm wondering if we could make these assertions agnostic of the wire format at this stage, similar as we did for the tool definitions, so we'd be able to use these assertions for other provider API tests as well. If not can we maybe indicate in the class or method name that this is specific to OpenAI completions?

maff (approved):
Some minor non-blocking remarks but otherwise looks great and will be much more helpful in the future! 🚀

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/ToolCallResultDocumentAssertions.java
nikonovd: Definitely, I had that in mind as well, but that should be an easy extract in a follow-up, once we want to test more provider configurations here. I'll leave it as is for now.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/jobworker/AiAgentJobWorkerMcpIntegrationTests.java
nikonovd: This extension annotation was previously present on the base test classes in the l4j package for all classes.

Now that we removed the mockito mocks for the LLM calls, we still need to keep the mockito mocks for MCP client interaction for now.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/ToolCallResultDocumentAssertions.java
nikonovd: @copilot create a follow-up issue:

* Make the `OpenAiCompletionsRecordedConversation` class of this PR wire-format agnostic, so that we can reuse a RecoredConversation together with RecordedMessage messages independent of the configured LLM provider, when testing against WireMock in these E2E tests
* Expand the test suite by smoke tests with the reference test bpmn process covering all AI agent connector supported providers (OpenAI, Azure OpenAI, OpenAiCompatible, Anthropic, Amazon Bedrock)
   * For the JobWorker implementation
   * For the outbound connector implementation

PR #7471, opened 2026-06-11
fix(webhook): exclude response expression from inbound deduplication

### LINKED ISSUE #6684: Webhook response expression should not be considered for deduplication
__What should we do?__

Webhook response expression allows the users to return an arbitrary response from the webhook endpoint in case of a successful correlation. As we now support deduplicating multiple elements into the same executable, even across process versions, we are encountering issues when changing the response expression.

As discussed in a team round, we think that webhook response expression should NOT be considered for deduplication, i.e. the logic is as follows:
- If two or more elements match (same/compatible activation conditions), response expression must be the same (otherwise we return a runtime error)
- If only one element matches, we should have no problem resolving the right response expression.

This task is somewhat complicated because the response expression is considered a user-space property, i.e. included in the request model of the webhook connector. We generally always include these properties in the deduplication scope. Our options are:
- Make it a runtime-level property
- Keep it in user space but add a keyword to `Keywords` in the runtime-core. This duplicates the ID so we need to link them somehow ideally, maybe by referencing this keyword from runtime-core in the template generator annotation to make sure property ID is fetched from the same source everywhere

__Why should we do it?__

It is currently very easy to break the webhook by changing the response expression, it's counterintuitive.

chillleader:
## Description

The webhook `responseExpression` was being included in the inbound connector **deduplication** key. As a result, changing only the response expression across deployments produced a different dedup ID and broke webhook correlation for the affected element. The response expression is a pure output concern and should not influence deduplication.

This PR removes `responseExpression` from the bound webhook model and from the deduplication scope, and resolves it **per request from the activated element** instead of from a shared connector instance. Consequently, several elements that share one webhook executable may legitimately declare different response expressions.

### Key changes
- **Deduplication** — `responseExpression` is excluded from the dedup scope (`Keywords.PROPERTIES_EXCLUDED_FROM_DEDUPLICATION`); compatibility checks compare only dedup-scoped properties. Adds SDK `WebhookPropertyNames`.
- **Per-request resolution** — the controller resolves the response expression from the element that was actually activated at request time (`InboundConnectorManagementContext.bindProperties(Class, rawProperties)`, `WebhookResponseExpressionProperties`), falling back to `WebhookResult.response()` for custom executables. `ActivationConditionEvaluator` guards divergent multi-match cases.
- **Template-only property** — `responseExpression` is no longer part of the bound model. It is declared via a new uninstantiable `TemplateOnly` marker type (element-template-generator), as a `@JsonIgnore`'d record component kept in its original position. `@JsonIgnore` keeps Jackson from binding it; the marker type makes "no runtime value" type-enforced (only `null` is assignable) and forces the template type to be owned by the annotation. The generator **fails the build** if a `TemplateOnly` field is not excluded from binding (`static` or `@JsonIgnore`) or omits an explicit `type`.
- **Templates unchanged** — because the property keeps its original declaration position, the generated element templates are **byte-identical to `main`** (no display-order churn).

### Scope notes
- `verificationExpression` is unchanged: it is consumed pre-correlation, so it stays bound and within the dedup scope.
- The legacy `responseBodyExpression` is removed from the bound model.

## Related issues

closes #6684

## Checklist

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/Keywords.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/Keywords.java
@@ -16,6 +16,7 @@
  */
 package io.camunda.connector.runtime.core;
 
+import io.camunda.connector.api.inbound.webhook.WebhookPropertyNames;
 import java.util.Set;
 
 public class Keywords {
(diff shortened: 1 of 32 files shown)


chillleader:
Superseded by two stacked PRs that split this change for cleaner review:

- #7487 — `feat(element-template-generator)`: the `TemplateOnly` template-only-property concept (→ `main`).
- #7488 — `fix(webhook)`: exclude the response expression from inbound deduplication and resolve it per request (→ #7487).

The combined change is identical, minus the intermediate static-field detour from this branch's history. Closing in favor of those.

PR #7504, opened 2026-06-15
feat(inbound): element-scoped properties — resolve webhook response per activated element (#6684)

chillleader:
## What & why

Closes the design behind #6684 with a reusable concept.

**Element-scoped properties.** Connector properties can now be split into two classes:
- **connector-scoped** (the existing `inputDataClass`) — bound once at `activate()`, participates in deduplication;
- **element-scoped** (a new `@ElementTemplate.elementInputDataClass`) — merged into the *same* element template, but bound **per activated element at correlation time** and **excluded from deduplication**.

Applied to the webhook `responseExpression`: webhook elements that differ *only* in their response expression now deduplicate into a single executable while each still produces its own response — resolved from the element that actually matched the request, instead of an arbitrary one captured at activation.

This supersedes the earlier `TemplateOnly`-marker approach (old stacked PRs), which left a null-by-design field on the model.

## Change set
| Layer | Change |
|---|---|
| ETG annotations | `@ElementTemplate.elementInputDataClass()` |
| ETG core | one-line merge in `ClassBasedTemplateGenerator` (reuses existing group-merge + duplicate-id guard) |
| SDK | `ProcessElement.properties()` default — exposes the activated element's raw properties |
| runtime-core | `InboundConnectorManagementContext.bindProperties(Class, Map)` (+ impl/delegate/test stub); dedup-exclusion keywords |
| webhook | new element-scoped `DynamicWebhookProperties` (`inbound`-wrapped → binding keys unchanged); response fields removed from `WebhookConnectorProperties`; executable defers the response |
| spring | `WebhookResponseExpressionProperties` + controller resolves the response from the activated element per request |

## Template impact
Regenerated templates change in exactly one way: `responseExpression` and `verificationExpression` swap order within the `webhookResponse` group. **All binding keys (e.g. `inbound.responseExpression`) are unchanged**, so deduplication ids and deployed models are unaffected.

## Verified
- All touched modules compile; spotless applied.
- Webhook module tests green.
- Template regenerated; diff is the reorder above only.

## Not done yet (follow-ups in this PR)
- [ ] **Walker**: derive the dedup exclusion from `elementInputDataClass` and remove the hardcoded `Keywords` entries (blocked on a module seam — `connector-runtime-core` doesn't depend on the ETG annotations module).
- [ ] **End-to-end test**: two deduplicated elements with different response expressions each get their own response (no `InboundWebhookRestController` test harness exists yet).
- [ ] **Template version bump** 14 → 15 (content changed).
- [ ] Run `connector-runtime-core` / `connector-runtime-spring` suites; write/replace ADR.

Draft — opening for early review of the approach.


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseExpressionProperties.java
```diff
@@ -0,0 +1,41 @@
...
+ * <p>This is a structural twin of the webhook connector's element-scoped template class: the
+ * connector module owns the template definition, while the runtime (which builds the HTTP response)
+ * owns this binding target, because {@code connector-runtime-spring} does not depend on the webhook
+ * connector module.
+ */
+public record WebhookResponseExpressionProperties(WebhookResponseExpressions inbound) {
+
+  public record WebhookResponseExpressions(
+      Function<WebhookResultContext, WebhookHttpResponse> responseExpression,
+      Function<WebhookResultContext, Object> responseBodyExpression) {}
```
chillleader: needs to be marked as deprecated / legacy, replaced with the former

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -309,6 +309,35 @@ public <T> T bindProperties(Class<T> cls) {
     }
   }
 
+  @Override
+  public <T> T bindProperties(Class<T> cls, Map<String, String> rawProperties) {
```
chillleader: let's try to move this API a level or two downward, into CorrelationResult.Success probably. This would imply splitting the class between SDK and runtime but that's an acceptable trade-off. Another option is the element class (one more level down)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorManagementContext.java
```diff
@@ -20,9 +20,22 @@
...
+   * Binds the given raw element properties to a typed object using the same secret-replacement and
+   * FEEL-evaluation pipeline as {@link #bindProperties(Class)}, but against an explicitly provided
+   * property map rather than the executable's connector-level properties.
+   *
+   * <p>Intended for resolving element-scoped properties from a specific activated element after
+   * correlation (for example, a webhook response expression), where several elements may have been
+   * deduplicated into a single executable. The raw map is typically obtained from {@link
+   * io.camunda.connector.api.inbound.ProcessElement#properties()} of the activated element.
+   */
+  <T> T bindProperties(Class<T> cls, Map<String, String> rawProperties);
```
chillleader: let's have a clean API where we don't need to pass rawProperties manually (reduce boilerplate)

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestController.java
```diff
@@ -255,20 +254,17 @@ protected ResponseEntity<?> verify(
...
+    }
+    if (correlationResult instanceof CorrelationResult.Failure failure) {
+      // Ignore strategy: treat as a successful (no-op) correlation; otherwise forward the error.
+      if (failure.handlingStrategy() instanceof Ignore) {
+        return buildSuccessfulResponse(webhookResult, documents, null);
       }
+      return buildErrorResponse(failure);
     }
-    return response;
+    throw new IllegalStateException("Illegal correlation result : " + correlationResult);
```
chillleader: why did we change this block? seems functionally equivalent, if so, let's roll back for cleaner diff

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/CorrelationResult.java
```diff
@@ -26,6 +26,18 @@ sealed interface Success extends CorrelationResult {
...
     ProcessElement activatedElement();
 
+    /**
+     * Binds the raw properties of the {@link #activatedElement()} to a typed object using the
+     * runtime's secret-replacement and FEEL-evaluation pipeline.
+     *
+     * <p>Use this to resolve element-scoped properties (for example, a webhook response expression)
+     * from the element that actually matched this correlation, even when several elements were
+     * deduplicated into a single executable.
+     */
```
chillleader, line 42: add a warning here that this happens past transaction boundary so careful exception handling is critical

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessElement.java
```diff
@@ -35,4 +37,30 @@ default String processName() {
...
+
+  /**
+   * Binds this element's raw properties to a typed object using the runtime's secret-replacement
+   * and FEEL-evaluation pipeline.
+   *
+   * <p>Use this to resolve element-scoped properties (for example, a webhook response expression)
+   * from the specific element that matched a request, even when several elements were deduplicated
+   * into a single connector executable. Only elements supplied by the runtime support binding; the
+   * default implementation throws.
+   */
```
chillleader, line 64: same here (warning)

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessElement.java
```diff
@@ -35,4 +37,35 @@ default String processName() {
...
+  /**
+   * Raw properties of this element as defined in the process model. FEEL expressions are not
+   * evaluated and secret placeholders are not resolved.
+   *
+   * <p>This allows the runtime to resolve element-scoped properties (for example, a webhook
+   * response expression) from the specific activated element even when several elements were
+   * deduplicated into a single connector executable.
+   */
+  default Map<String, String> properties() {
+    return Map.of();
```
sbuettner: Should we also throw an exception here for the default implementation?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookResponseExpressionProperties.java
```diff
@@ -0,0 +1,43 @@
...
+ * <p>The raw properties of the activated element are bound to this record per request (see {@link
+ * InboundWebhookRestController}). The {@code inbound} wrapper mirrors the {@code
+ * inbound.responseExpression} / {@code inbound.responseBodyExpression} property bindings used in
+ * the element template, so the same raw model keys deserialize here.
+ *
+ * <p>This is a structural twin of the webhook connector's element-scoped template class: the
+ * connector module owns the template definition, while the runtime (which builds the HTTP response)
+ * owns this binding target, because {@code connector-runtime-spring} does not depend on the webhook
+ * connector module.
+ */
```
sbuettner: Should we think about a way where we dont have to duplicate this part?

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/ProcessElement.java
chillleader: made it abstract

johnBgood (approved):
LGTM, just a bit sad that we use ET-level annotation to solve a runtime issue. 
Did you consider a simple annotation that could be used on, say, the `responseExpression` property. It would:
- exclude it from the deduplicationId 
- allow lazy resolution (as you did)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/Keywords.java
```diff
@@ -151,6 +151,19 @@ public enum DeduplicationMode {
...
+   * between BPMN elements that are deduplicated into a single webhook executable, so they must be
+   * excluded from deduplication and resolved per activated element at correlation time.
+   *
+   * <p>TODO(#6684): derive this exclusion from the connector's element-scoped input class (see
+   * {@code ElementTemplate#elementInputDataClass()}) instead of hardcoding webhook specifics here.
+   */
+  public static final String WEBHOOK_RESPONSE_EXPRESSION_KEYWORD = "inbound.responseExpression";
+
+  public static final String WEBHOOK_RESPONSE_BODY_EXPRESSION_KEYWORD =
+      "inbound.responseBodyExpression";
```
sbuettner: Agree with @johnBgood, could we find another way to exclude it from the deduplication calculation by leveraging the deduplication properties for example?

chillleader: @johnBgood @sbuettner makes sense, what if we make the ET annotation generic (e.g. it accepts any number of data classes that define the ET model and the ETG merges them all without caring about runtime aspects):

```
 @ElementTemplate(inputDataClasses = { A.class, B.class, … })
```
and repurpose `deduplicationProperties` to accept a class reference as well to avoid having to list individual properties 

```
@InboundConnector(deduplicationProperties = A.class)
```
(if not specified, everything is included in deduplication properties by default)

Then we:
* have a clear separation of concerns (ET annotations only care about ET stuff, runtime annotations declare runtime concepts)
* fix the awkward current syntax of `@InboundConnector(deduplicationProperties)` where you have to list all the properties explicitly (very error-prone especially for AI)
* keep the lazy resolution class-level not property-level (more convenient to reason about, because we bind to classes not individual values)

sbuettner: Sounds like a plan @chillleader!

Can we somehow remodel the existing `inputDataClass` to an Array type. Did a quick test and might not require a code change in the Connectors (possibly a recompile but no source change).

What about a `deduplicationClass` property that accepts an array of type class? Would be more consistent with the data class properties.

johnBgood: I love it! 👍

chillleader: done, please take a final look at the latest 3 commits 🙂

sbuettner (approved):
Great work!

Szik (approved):

PR #7518, opened 2026-06-16
fix: remove activity log entries on permanent connector deactivation

johnBgood:
## Problem

`ActivityLogRegistry` stores activity logs per `ExecutableId` in an internal `Map` but never removes entries. Each time a connector is deployed then undeployed, the map grows by one entry and those entries are never reclaimed. Over time this causes unbounded memory growth in long-running runtimes with frequent redeployments.

## Solution

Added `ActivityLogRegistry.remove(ExecutableId)` — a synchronized map removal, consistent with the existing `getLogs` and `log` methods.

The call site is `InboundExecutableRegistryImpl.executeStateTransition`, immediately after the permanent **DEACTIVATE** batch completes. This is the only action that permanently discards an executable:

- **DEACTIVATE** → remove logs ✅ (connector is gone for good)
- **RESTART** → keep logs (same `ExecutableId` re-activated moments later)
- **REPLACE_WITH_INVALID** → keep logs (executable stays visible via the query API as `InvalidDefinition`)

## Tests

- **`ActivityLogRegistryTest`** (new unit test): verifies `remove` clears all logs for the target id, leaves other ids untouched, and is a no-op on an unknown id.
- **`activityLog_permanentDeactivation_logsShouldBePurged`**: integration test — confirms the registry is empty after a permanent deactivation event.
- **`activityLog_deactivation_deactivationLifecycleEntryIsEmitted`**: regression test using a spy — confirms the deactivation lifecycle entry is still written to the log before being removed.
- **`activityLog_afterRestart_priorLogsShouldSurvive`**: regression test — confirms that a `reset()` (RESTART path) does not remove logs for the reactivated connector.

## Checklist

- [x] Memory leak fixed
- [x] RESTART and REPLACE_WITH_INVALID paths unaffected
- [x] All existing tests pass
- [x] New tests cover the fix and guard against regressions

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/activitylog/ActivityLogRegistry.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/activitylog/ActivityLogRegistry.java
@@ -48,6 +48,12 @@ public class ActivityLogRegistry implements ActivityLogWriter {
     }
   }
 
+  public void remove(ExecutableId executableId) {
+    synchronized (activityLogs) {
+      activityLogs.remove(executableId);
+    }
+  }
+
   @Override
   public void log(ActivityLogEntry logEntry) {
     String message = logEntry.activity().toString();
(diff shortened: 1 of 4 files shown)


mathias-vandaele (approved):
Good improvement @johnBgood Thank you

PR #7474, opened 2026-06-11
test(e2e): stabilize flaky agentic-ai e2e tests

nikonovd:
## Description

  The agentic-ai e2e suite (mainly the A2A integration tests) flaked intermittently: webhook context collisions, non-deterministic feedback handling, cross-test state bleed, and ad-hoc timeouts. This PR addresses each:

  - Unique webhook context per A2A test fixes the Context: test-webhook-id already in use collision that parked the webhook executable DOWN across consecutive tests/reruns (new customizeModel hook plus TestUtil.setWebhookContext).
  - Per-class CPT runtime: drop runtime-mode: shared, whose cross-class carryover reintroduced the same collision.
  - Per-test WireMock reset: clear scenarios/stubs in `@BeforeEach` to stop stub bleed.
  - Deterministic user-feedback: replace the timing-sensitive when/then conditional (gated behind a costly condition-await timeout) with a queued mock job worker.
  - Centralized timeouts: Duration-based waitForProcessCompletion/waitForActiveIncidents (default 30s) instead of per-test timeout setup.
  - PT2S A2A polling: tighten task/process polling in the A2A BPMNs so polled completions surface promptly.
  - Split E2E test execution into matrix based workflow
     - Each branch (on PR it's only main) builds connectors and e2e test base modules first
     - For each e2e test module a separate matrix branch is created
     - Each matrix branch pulls the pre-built artifacts from cache
  - Retain CI logs: upload surefire *-output.txt (redirected stdout) alongside the XML reports.

## Checklist

- [ ] Backport labels are added if these code changes should be backported. No backport label is added to the latest
  release, as this branch will be rebased onto main before the next release. Example backport labels:
    - `backport stable/8.8`: for changes that should be included in the next 8.8.x release.
    - **or** `backport release-8.8.7`: for changes that should be included in the specific release 8.8.7, and this
      *release has already been created*. The release branch will be merged back into stable/8.8 later, so the change
      will be included in future 8.8.x releases as well.
- [ ] Tests/Integration tests for the changes have been added if applicable.
- [ ] If the change requires a documentation update, it has been added to the appropriate section in the documentation.

PR #7527, opened 2026-06-17
refactor(agentic-ai): introduce domain-oriented code flow for agent conversation

nikonovd:
## Description
`AgentConversation` previously held a flat message list and delegated all  turn-level reasoning to the handler. The handler reconstructed turn  boundaries, tracked per-call metrics, partitioned tool results from event  messages, and decided whether to proceed or wait, all interleaved with the  LLM invocation logic in `BaseAgentRequestHandler`.

This branch promotes `AgentConversation` to a rich domain model. A  `ConversationTurn` groups the input messages that drove an LLM call, the  assistant response, and per-turn metrics. `AgentConversation` manages the  turn lifecycle (pending → complete via `ingest`) and exposes derived state  (`expectingToolCallResults`, `hasToolCalls`, `metricsDelta`, limits).
`TurnReconstructor` re-inflates the turn structure from the flat message  list in `AgentContext` on load, so the storage format is unchanged and no  data migration is needed.

Message assembly and the proceed/wait/cancel decision move into  `ConversationTurnComposer` (the former `AgentMessagesHandlerImpl`), returning a sealed `AgentInput` type instead of a boolean flag + separate  message list. `AiFrameworkAdapter` receives a `ConversationSnapshot` (the  windowed transient view) instead of a mutable `RuntimeMemory`.

## Related issues

relates to #7450 (as a precondition)

## Checklist


nikonovd:
I'll activate e2e tests after #7474 is merged

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -6,65 +6,193 @@
...
+  public ConversationTurn currentTurn() {
+    return currentTurn;
   }
 
-  public void updateContext(AgentContext newContext) {
-    this.context = newContext;
+  /** Returns the agent instance key of that conversation, or null if it is not existing (yet). */
+  public AgentInstanceKey agentInstanceKey() {
+    return AgentInstanceKey.from(currentContext.metadata());
   }
```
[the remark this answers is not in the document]
nikonovd: Fixed: `agentInstanceKey()` now returns `@Nullable AgentInstanceKey`, matching the `@Nullable` contract of `AgentInstanceKey.from(...)` it delegates to. The only call site (`notifyThinking`) forwards the value straight into `agentInstanceClient.update(...)`, which already accepts a nullable key.

---

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -290,39 +286,51 @@ protected abstract boolean modelCallPrerequisitesFulfilled(
...
           metricsDelta.modelCalls(),
           metricsDelta.tokenUsage().inputTokenCount(),
           metricsDelta.tokenUsage().outputTokenCount(),
           metricsDelta.toolCalls());
       agentInstanceClient.update(
           executionContext,
-          agentContext,
-          AgentInstanceUpdateRequest.builder().status(nextStatus).delta(metricsDelta).build());
+          AgentInstanceKey.from(context.metadata()),
+          AgentInstanceUpdateRequest.builder().status(nextState).delta(metricsDelta).build());
```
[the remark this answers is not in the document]
nikonovd: Fixed: the `nextState` parameter is now marked `@Nullable`, and the update request is built without calling `.status(...)` when it is null (instead of passing null into the builder). This keeps the superseded-job intent explicit and avoids the nullness warning.

---

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
```diff
@@ -0,0 +1,182 @@
...
+## Decision Drivers
+
+* **Explicit turn structure**: a "turn" (one LLM call with its input messages, assistant response,
+  and per-turn metrics) should be a first-class domain concept, not an implicit slice of a flat list.
+* **Single decision point**: the proceed / wait / cancel outcome should be expressed as a typed
+  value returned from a single component, not inferred by the caller from an empty list + boolean.
+* **No mutable shared state**: the working representation of a conversation should be immutable;
+  each transformation returns a new instance.
+* **Backward compatibility**: existing persisted conversations (flat message lists in `agentContext`)
+  must rehydrate without a data migration.
```
maff, line 50: 💭 This is what we could do next as implicit data migration on the next save, correct? Do we already plan to do this in combination with a schema version?

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
```diff
@@ -0,0 +1,182 @@
...
+
+**`ConversationTurn` record**: the unit of one LLM call:
+
+```
+ConversationTurn(int iterationKey, List<Message> inputMessages,
+                 @Nullable AssistantMessage assistantMessage, AgentMetrics metrics)
+```
+
+A turn is *pending* (`assistantMessage == null`) when created by `rehydrate`, and *complete* after
+`ingest`. `iterationKey` is 1-based and counts LLM calls across the agent's lifetime.
```
maff, line 83: 💭 Assuming we'd do multiple LLM interactions within in the same job execution, would we be able to model this with the conversation turn data model? Is it relevant that the `iterationKey` might be reused between retried jobs (in combination with the agent instance API)?

Examples:
- compacting the message history (1 turn compaction, a second turn the actual model call or vice versa)
- calling "internal tools" which do not route through the AHSP (loading a skill, potentially executing code in a sandbox)

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
```diff
@@ -0,0 +1,182 @@
...
+The composed system message is passed in by the handler and may be `null` when the composed system
+prompt is blank (in which case no system message is sent to the LLM or persisted).
+
+**`TurnReconstructor`**: rebuilds the `ConversationTurn` list from the persisted flat message
+list. Groups messages by scanning for `AssistantMessage` boundaries; assigns `iterationKey` by
+position (1-based count of completed assistant messages). All reconstructed turns carry
+`AgentMetrics.empty()`: per-invocation metrics are computed live from the current turn, not read
+from historical turns.
+
+This provides backward compatibility with all existing conversations without a data migration.
```
maff: 💭 By actually storing the turns in an updated storage format we can avoid this step + only apply it on existing processes for data migration, correct?

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
```diff
@@ -0,0 +1,182 @@
...
+- Per-turn historical metrics are not persisted in the message list: reconstructed previous turns
+  always carry `AgentMetrics.empty()`. Cumulative metrics (model calls, tokens, tool calls) are
+  therefore read from the durable `AgentContext.metrics` — `totalMetrics()` returns
+  `baseAgentContext().metrics()` and `toAgentContext()` increments it by the current turn's delta.
+  The model-call limit check must use this cumulative counter, **not** a sum over reconstructed
+  turns (which is always zero). Per-turn historical cost reporting would need the metrics to be
+  persisted alongside messages (see Future improvements).
+- Two "conversation" vocabularies coexist: `ConversationStore` / `ConversationSession` /
+  `ConversationContext` (message persistence layer) vs. `AgentConversation` / `ConversationTurn`
+  (turn aggregate layer). The distinction must be kept clear in code reviews and documentation.
```
maff: 🤔 That is a good point. Maybe we can address this via consistent naming? `Conversation*` related to persistence, `AgentConversation*` (or something else) related to runtime data?

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Future improvements
+
+- If per-turn historical metrics become a requirement, store them alongside messages (e.g. as a
+  message metadata field or a parallel structure), and update `TurnReconstructor` to read them.
```
maff, line 198: 💭 We already store them as part of the metadata and could potentially use them to feed the reconstructed messages. Cached tokens vary by provider, but at least I/O/total should be there:

Image

nikonovd: Nice, I wasn't really aware of that 😄 i'll check.

connectors/agentic-ai/docs/reference/ai-agent.md
```diff
@@ -677,41 +718,50 @@ t4                                      Job #2 picked up
...
 
 ```java
-// BaseAgentRequestHandler.handleRequest()
-var addedUserMessages = messagesHandler.addUserMessages(...);
-if (!modelCallPrerequisitesFulfilled(addedUserMessages)) {
-    return buildConnectorResponse(executionContext, null);
-}
+return switch (input) {
+    case AgentInput.None ignored ->
+        handleNoOp(executionContext);                              // wait for more results
```
maff: 🔧 Should we align this with the initializer result which uses "deferred" when handling it as no-op (e.g. `AgentInput.Deferred`)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInput.java
```diff
@@ -0,0 +1,30 @@
...
+package io.camunda.connector.agenticai.aiagent.agent;
+
+import io.camunda.connector.agenticai.model.message.Message;
+import java.util.List;
+
+/**
+ * Decision produced by ConversationTurnComposer: whether and how to proceed with the next LLM call.
+ */
+public sealed interface AgentInput
+    permits AgentInput.None, AgentInput.Cancellation, AgentInput.NextTurn {
```
maff: 🔧 Nit: permits can be omitted if the records are inner classes of the interface.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentInput.java
```diff
@@ -0,0 +1,30 @@
...
+
+  /** No messages ready yet — wait for more tool results before proceeding. */
+  record None() implements AgentInput {}
```
maff: 💭  I already commented above regarding naming this `Deferred` - actually we always comment that it is waiting for more tool results, would it also be viable to call that respectively?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentResponseHandlerImpl.java
```diff
@@ -37,21 +37,28 @@ public class AgentResponseHandlerImpl implements AgentResponseHandler {
...
 
   @Override
-  public AgentResponse createResponse(
-      AgentExecutionContext executionContext,
-      AgentContext agentContext,
-      AssistantMessage assistantMessage,
-      List<ToolCallProcessVariable> toolCalls) {
+  public AgentResponse createResponse(AgentConversation conversation) {
+    final var agentContext = conversation.toAgentContext();
+    final var assistantMessage = conversation.lastTurn().orElseThrow().assistantMessage();
```
maff: 🔧 I'm aware that this should never happen, but can we throw with an actual error message here?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -45,39 +48,40 @@ public abstract class BaseAgentRequestHandler<
...
 
   @Override
   public R handleRequest(final C executionContext) {
     return switch (agentInitializer.initializeAgent(executionContext)) {
+      case ReadyToConverse(var agentContext, var engineToolCallResults) -> {
+        LOGGER.debug(
+            "Handling agent request with {} tool call results",
+            engineToolCallResults != null ? engineToolCallResults.size() : 0);
+        yield converse(executionContext, agentContext, engineToolCallResults);
+      }
```
maff: 🔧 Didn't we switch the order here on the agent initializer PR to match the actual execution order? 😅 I still think the code would be easier to follow if the steps would come in the order they are processed during the lifetime of an agent.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -87,125 +91,137 @@ case DiscoverTools(var agentContext, var toolDiscoveryToolCalls) -> {
...
       }
     };
   }
 
   private R converse(
       final C executionContext,
-      AgentContext agentContext,
-      final List<ToolCallResult> toolCallResults) {
+      final AgentContext agentContext,
+      final List<ToolCallResult> engineToolCallResults) {
```
maff: 🔧 Is this a leftover from the other PR? Can we drop `engine` basically for every tool call results variable at this layer?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -87,125 +91,137 @@ case DiscoverTools(var agentContext, var toolDiscoveryToolCalls) -> {
...
+              executionContext.userPrompt(),
+              engineToolCallResults != null ? engineToolCallResults : List.of());
+
+      LOGGER.trace("Loading previous conversation (if any) for rehydration");
+      final var loadedMessages = session.loadMessages(agentContext).messages();
+      final var history = TurnReconstructor.reconstruct(loadedMessages);
+
+      LOGGER.trace("Composing turn input from history and invocation state");
+      final var input =
+          agentInputComposer.compose(history, invocationInput, agentContext, configuration);
```
maff: 🔧 Given that the `TurnReconstructor` is only a helper concept to support existing process data, I assume `compose()` should accept a data structure which is decoupled from the reconstructor instead of `TurnReconstructor.Result`. WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -87,125 +91,137 @@ case DiscoverTools(var agentContext, var toolDiscoveryToolCalls) -> {
...
+      return switch (input) {
+        case AgentInput.None ignored -> {
```
maff: 💭 Conceptually it's rather hard to distinguish `AgentInvocationInput` and `AgentInput`. Do you think we can find a naming scheme which makes them less likely to be mixed up?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposer.java
```diff
@@ -0,0 +1,21 @@
...
+import io.camunda.connector.agenticai.aiagent.model.AgentInvocationInput;
+import io.camunda.connector.agenticai.aiagent.model.TurnReconstructor;
+
+/** Composes the next conversation turn input from raw history and the current invocation state. */
+public interface ConversationTurnComposer {
+  AgentInput compose(
+      TurnReconstructor.Result history,
+      AgentInvocationInput invocationInput,
+      AgentContext agentContext,
+      AgentConfiguration configuration);
```
maff: 🔧 Until now, we tried to preserve a consistent order or parameters from generic to more specific, mostly with the pair `(agentExecutionContext, agentContext)` + additional parameters. Can we try to find a consistent scheme for the new approach as well? In this case I'd probably order by `configuration, context, history, invocationInput`.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentMessagesHandler.java
maff, line 1: Awesome to see this getting transformed into something more manageable! ❤️

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
```diff
@@ -63,97 +59,86 @@ public class AgentMessagesHandlerImpl implements AgentMessagesHandler {
...
+          "No user message content available to start the conversation.");
+    }
+
+    return new AgentInput.NextTurn(messages);
   }
 
   private UserMessage createUserPromptMessage(UserPromptConfiguration userPrompt) {
+    if (userPrompt == null) {
+      return null;
+    }
```
maff: 🔧 The configuration itself is marked as non-null on the request object. Maybe we find a way to ensure all these top-level configuration objects are not nullable, so we can avoid guards deep down here in the execution (can be a follow-up, but I guess we could avoid this change here).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
```diff
@@ -63,97 +59,86 @@ public class AgentMessagesHandlerImpl implements AgentMessagesHandler {
...
 
-    List<Message> messages = new ArrayList<>();
-    if (lastChatMessage instanceof AssistantMessage assistantMessage
-        && assistantMessage.hasToolCalls()) {
+    if (expectingToolCallResults) {
       boolean interruptMissingToolCalls =
           interruptToolCallsOnEventResults && !eventMessages.isEmpty();
 
-      ToolCallResultMessage toolCallResultMessage =
+      final var toolCalls = history.turns().getLast().assistantMessage().toolCalls();
```
maff: 🔧 IntelliJ flags this as potential for an NPE as we never check if `assistantMessage()` does not return `null`, however we implicitly test this with `hasToolCalls()` above. WDYT of either add an additional `toolCalls()` method to the turn or add some kind check here?

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
nikonovd: Exactly, now we have back-and-forth mapping from and to `AgentContext`.

With a one-time migration for running processes we could read AgentContext and store the relevant AgentConversation data finally so we can also avoid the mapping layer.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
```diff
@@ -230,7 +216,7 @@ private UserMessage createDocumentMessageForToolResults(List<ToolCallResult> res
     content.add(textContent(TOOL_CALL_DOCUMENTS_PREAMBLE));
     content.addAll(createDocumentPairs(toolCallDocuments));
 
-    final var metadata = new HashMap<String, Object>(defaultMessageMetadata());
+    final var metadata = new java.util.HashMap<String, Object>(defaultMessageMetadata());
```
maff: 🔧 Nit: remove FQCN

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
```diff
@@ -37,62 +34,41 @@ public class JobWorkerAgentRequestHandler
...
+        systemPromptComposer,
         responseHandler,
         agentInstanceClient);
   }
 
   @Override
-  protected boolean shouldUpdateAgentInstanceBeforeJobCompletion(AgentResponse agentResponse) {
-    return agentResponse.toolCalls().isEmpty();
+  protected boolean shouldUpdateAgentInstanceBeforeJobCompletion(AgentConversation conversation) {
+    return !conversation.lastTurn().map(ConversationTurn::hasToolCalls).orElse(false);
```
maff: 🔧 This statement with the `orElse` branch and the negation is rather hard to follow, can we make that more readable, maybe paired with a comment?

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
nikonovd: The current definition of a turn would be input -> assistantmessage.
I think we should then also reflect that and have multiple turns in a single execution then.

* Compaction would then be a turn pre-actual chat request
* caling internal tools could also be `Turn1{input -> assistant message with internal tool calls}` -> `Turn2{internal tool call results -> assistant message}`

Would then be transparent to the agent instance whether we do it in one or multiple jobs. What's your vision on this?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
```diff
@@ -114,16 +90,19 @@ public AiAgentSubProcessConnectorResponse buildConnectorResponse(
...
   private AiAgentSubProcessConnectorResponse buildResponse(
       JobWorkerAgentExecutionContext executionContext,
+      AgentConversation conversation,
       AgentResponse agentResponse,
       AgentJobCompletionListener completionListener) {
     boolean completionConditionFulfilled = agentResponse.toolCalls().isEmpty();
-    boolean cancelRemainingInstances = executionContext.cancelRemainingInstances();
+    // cancel remaining instances if any tool call in this turn's input was interrupted
+    boolean cancelRemainingInstances =
+        conversation != null && conversation.currentTurn().hasInterruptedToolCallResults();
```
maff, line 106: Great to get rid of this mutable property! 😊

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
```diff
@@ -63,97 +59,86 @@ public class AgentMessagesHandlerImpl implements AgentMessagesHandler {
...
 
     messages = messages.stream().filter(Objects::nonNull).toList();
-    messages.forEach(memory::addMessage);
 
-    return messages;
+    if (messages.isEmpty()) {
+      LOGGER.debug("Not proceeding as no user content was found to add.");
+      return new AgentInput.Cancellation(
+          ERROR_CODE_NO_USER_MESSAGE_CONTENT,
+          "No user message content available to start the conversation.");
```
maff: 🔧 Is this now only to start the conversation or does it also apply mid-conversation? What about re-entry?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
```diff
@@ -37,62 +34,41 @@ public class JobWorkerAgentRequestHandler
...
-                    Boolean.TRUE.equals(
-                        result
-                            .properties()
-                            .getOrDefault(ToolCallResult.PROPERTY_INTERRUPTED, false)));
-
-    // cancel remaining instances if any tool call was interrupted
-    if (hasInterruptedToolCalls) {
-      executionContext.setCancelRemainingInstances(true);
-    }
+  protected AiAgentSubProcessConnectorResponse handleInputCancel(
```
maff: 💭 Maybe I didn't fully understand when this is raised now, but I'm wondering if "input cancel" is the right framing here and if it makes sense to generalize the error code/message in the composer as we currently do. If I understand correctly, in the job worker variant we receive an error code/message for a valid state (waiting on further results) + log this debug warning, but do not consume the error code and message. In the outbound connector we use it to throw an exception.

WDYT of only transporting the information "we can't continue" and letting the respective implementation decide on the error code & message & what to log?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+import io.camunda.connector.agenticai.aiagent.model.request.ResponseConfiguration;
+import io.camunda.connector.agenticai.aiagent.model.request.provider.ProviderConfiguration;
+import java.util.Optional;
+import org.springframework.lang.Nullable;
+
+/**
+ * Static per-invocation configuration. Built once from AgentExecutionContext at the start of each
+ * handler invocation; does not change mid-conversation.
+ */
+public record AgentConfiguration(
```
maff, line 25: 💭 Until now, the execution context was this abstraction over different configurations, now we additionally have this record which we build from the execution context, but it's not 100% consistent when to read from the context or from the configuration. Do you think we can consolidate this? I see 2 approaches:

- Do not introduce this new configuration structure & continue passing the execution context as "configuration abstraction with additional data". I assume this is what we want to avoid by introducing this record.
- Make this configuration the authoritative configuration source from the start. The execution context could be stripped of all the configuration-related methods and only expose a non-nullable `configuration()`. 

WDYT?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+        executionContext.response());
+  }
+
+  public int contextWindowSize() {
+    return Optional.ofNullable(memory).map(MemoryConfiguration::contextWindowSize).orElse(20);
+  }
+
+  public int maxModelCalls() {
+    return Optional.ofNullable(limits).map(LimitsConfiguration::maxModelCalls).orElse(10);
+  }
```
maff, line 47: 🔧 Can we extract the defaults into constants as they have been before?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.lang.Nullable;
+
+/**
+ * Static per-invocation configuration. Built once from AgentExecutionContext at the start of each
+ * handler invocation; does not change mid-conversation.
+ */
+public record AgentConfiguration(
+    ProviderConfiguration provider,
+    @Nullable SystemPromptConfiguration systemPrompt,
+    @Nullable MemoryConfiguration memory,
```
maff: 🔧 I guess this should also contain the user prompt config for completeness?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+    @Nullable LimitsConfiguration limits,
+    @Nullable EventHandlingConfiguration events,
+    @Nullable ResponseConfiguration response) {
```
maff, line 32: 💭 Could be handled in a follow-up, but we should double-check what really should be nullable. Maybe we can use this configuration class as the fallback layer mentioned above to initialize all the configs and to avoid having to keep null-guards throughout the codebase.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -6,65 +6,193 @@
...
 public final class AgentConversation {
 
-  private AgentContext context;
-  private final RuntimeMemory runtimeMemory;
-  private final List<ToolCallResult> toolCallResults;
+  private final @Nullable SystemMessage systemMessage;
+  private final List<ConversationTurn> previousTurns;
+  private final ConversationTurn currentTurn;
+  private final AgentContext currentContext;
+  private final AgentConfiguration configuration;
```
maff: 🔧 Can we try to keep some kind of hierarchic structure here? E.g. I'd probably put at least the configuration first (generic to the whole execution), maybe also the context.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -6,65 +6,193 @@
...
+  /** Returns the current turn. Always non-null; pending until {@link #ingest} completes it. */
+  public ConversationTurn currentTurn() {
+    return currentTurn;
   }
 
-  public void updateContext(AgentContext newContext) {
-    this.context = newContext;
+  /** Returns the agent instance key of that conversation, or null if it is not existing (yet). */
+  public @Nullable AgentInstanceKey agentInstanceKey() {
+    return AgentInstanceKey.from(currentContext.metadata());
```
maff: 🔧 Should this be assigned once at construction time rather than lazily on each invocation?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -6,65 +6,193 @@
...
+    var all = allTurns();
+    return all.isEmpty() ? Optional.empty() : Optional.of(all.getLast());
   }
 
-  public RuntimeMemory runtimeMemory() {
-    return runtimeMemory;
+  private List<ConversationTurn> allTurns() {
+    if (currentTurn.assistantMessage() == null) {
+      return previousTurns;
+    }
```
maff, line 189: 💭 This basically means all turns which have a completed LLM message? Would it then maybe be `allCompletedTurns`?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfiguration.java
```diff
@@ -253,19 +245,17 @@ public ToolCallResultDocumentExtractor toolCallResultDocumentExtractor(
 
   @Bean
   @ConditionalOnMissingBean
-  public AgentMessagesHandler aiAgentMessagesHandler(
-      GatewayToolHandlerRegistry gatewayToolHandlers,
-      SystemPromptComposer systemPromptComposer,
-      ToolCallResultDocumentExtractor documentExtractor) {
-    return new AgentMessagesHandlerImpl(
-        gatewayToolHandlers, systemPromptComposer, documentExtractor);
+  public ConversationTurnComposer aiConversationTurnComposer(
```
maff: 🔧 `aiAgentConversationTurnComposer`

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/autoconfigure/AgenticAiConnectorsAutoConfigurationTest.java
```diff
@@ -99,8 +97,6 @@ class AgenticAiConnectorsAutoConfigurationTest {
           AwsAgentCoreConversationMapper.class,
           AwsAgentCoreConversationStore.class,
           ConversationStoreRegistry.class,
-          AgentLimitsValidator.class,
-          AgentMessagesHandler.class,
```
maff, line 103: 🔧 We should add the new conversation turn composer bean here.

connectors/agentic-ai/docs/adr/007-agent-conversation-turn-aggregate.md
maff: Yes, I would see it the same. It's always a turn, independent if executed all within a job or by looping through the process. Just wanted to make sure our architecture & the metrics handling is able to cope with that.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
nikonovd: TBH i'd like to see an Issue under #7537 for introducing nullability contracts on the codebase 👍 we can create an issue there to do it once to achieve a solid baseline.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: Yep and this PR was based on the oild one and overrode it after rebasing 😓 i'll fix that back

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
nikonovd: The user prompt config is currently really rather the new `AgentInput` and wouldn't serve any meaning here in the current design, so i'd leave it as is, if you don't have strong objections.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
nikonovd: Fine, generally i wanted to decouple the agent conversation from the a bit heavy bloated execution contexts to what we need -> keeps tests a bit simpler.

I followed your second approach now.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
nikonovd: No that also appies mid cancellation, if we don't have anything to add.

Basically no messages at all. In combination with your 💭 i renamed it to NoInput() without an errorcode and mesage and both implementations of the handler actually decide what to do, as it was beforehand.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
nikonovd: see above

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
nikonovd: Agree, let's have a followup task under #7537 - I think AI agents are going to benefit from nullability annotations as well.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: Good point. I extracted an independent class `PreviousConversation` - once we have a solid path to avoid reconstruction, we could drop it and just rely entirely on the conversation.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: Good point! 

`AgentInput` became simply `CompositionResult` to allow space for the different outcomes.
`AgentInvocationInput` became `AgentInput` after all.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposer.java
nikonovd: Done!

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
nikonovd: Done!

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/JobWorkerAgentRequestHandler.java
nikonovd: Changed it a bit and added a comment. Let me know if that suits you better.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -87,125 +86,142 @@ case DiscoverTools(var agentContext, var toolDiscoveryToolCalls) -> {
...
         LOGGER.debug(
             "AI Agent initialization tool discovery is still in progress. Completing job without further processing.");
-        yield buildConnectorResponse(executionContext, null, null);
+        yield handleNoOp(executionContext);
       }
       case ReadyToConverse(var agentContext, var toolCallResults) -> {
-        LOGGER.debug("Handling agent request with {} tool call results", toolCallResults.size());
+        LOGGER.debug(
+            "Handling agent request with {} tool call results",
+            toolCallResults != null ? toolCallResults.size() : 0);
```
maff: 🔧 I think toolCallResults can never be null, here, correct? Should be done in the initializer & the record should enforce this is never null, so we can avoid all subsequent null-checks.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -87,125 +86,142 @@ case DiscoverTools(var agentContext, var toolDiscoveryToolCalls) -> {
...
-        if (shouldUpdateAgentInstanceBeforeJobCompletion(agentResponse)) {
-          notifyMetrics(executionContext, agentResponse.context(), metricsDelta, nextStatus, true);
-        } else {
-          metricsListener =
-              createMetricsCompletionListener(
-                  executionContext, agentResponse.context(), metricsDelta, nextStatus);
+      final var configuration = executionContext.configuration();
+      final var agentInput =
+          AgentInput.from(
+              executionContext.userPrompt(), toolCallResults != null ? toolCallResults : List.of());
```
maff: 🔧 Also here, we should be able to assume toolCallResults is never null and avoid the null check.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -87,125 +86,142 @@ case DiscoverTools(var agentContext, var toolDiscoveryToolCalls) -> {
...
-    handleAddedUserMessages(executionContext, conversation.context(), addedUserMessages);
-
+  private R proceed(
+      final C executionContext,
+      final PreviousConversation previousConversation,
+      final List<Message> inputMessages,
+      final AgentContext agentContext,
+      final AgentConfiguration configuration,
+      final ConversationSession session,
+      final ConversationStore store) {
```
maff, line 138: 🔧 Given configuration is now part of the execution context, do we still need to pass it as dedicated parameter here? Could be something like (reordered a bit):


```suggestion
  private R proceed(
      final C executionContext,
      final AgentContext agentContext,
      final PreviousConversation previousConversation,
      final List<Message> inputMessages,
      final ConversationStore store,
      final ConversationSession session
) {
```

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
maff: I think for this PR can assume this already now and mark both system and user prompt config as not nullable (only the wrapper config object, not the individual fields) as they are both marked as `@Valid @NotNull` from the request object inwards. If we remove/not introdcue unnecessary null-checks here it will probably be easier to follow back and properly annotate nullable paths.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -49,21 +49,22 @@ public AgentInstanceKey create(AgentExecutionContext agentExecutionContext) {
...
     var command =
         camundaClient
             .newCreateAgentInstanceCommand()
             .elementInstanceKey(elementInstanceKey)
-            .model(agentExecutionContext.provider().model())
-            .provider(agentExecutionContext.provider().providerType())
-            .systemPrompt(agentExecutionContext.systemPrompt().prompt());
+            .model(configuration.provider().model())
+            .provider(configuration.provider().providerType())
+            .systemPrompt(configuration.systemPrompt().prompt());
```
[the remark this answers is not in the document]
maff: See above - we should be able to remove the `@Nullable` on the system prompt config.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
maff: Hmm I'd still prefer this configuration as "all configuration passed to this execution when the input mappings were applied". The user prompt can change across AHSP instances, but so can the response format or the limits (at least for the moment, also if they're pushed to the cluster agent instance). As a consumer I'd not understand why this does not carry also the user prompt config as the naming implies a complete view.

maff: As commented above, I think we should be able to remove the `@Nullable` on the 2 prompt fields.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -6,65 +6,196 @@
...
 
-  private AgentContext context;
-  private final RuntimeMemory runtimeMemory;
-  private final List<ToolCallResult> toolCallResults;
+  private final AgentConfiguration configuration;
+  private final AgentContext currentContext;
+  private final @Nullable SystemMessage systemMessage;
+  private final List<ConversationTurn> previousTurns;
+  private final ConversationTurn currentTurn;
+  private final @Nullable AgentInstanceKey agentInstanceKey;
```
maff, line 34: 🔧 As this is the primary identity of the conversation, should we elevate this field to come before tthe turn specific-ones (e.g. after the current context)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/JobWorkerAgentExecutionContext.java
```diff
@@ -50,46 +45,32 @@ public List<AdHocToolElement> toolElements() {
...
   @Override
-  public MemoryConfiguration memory() {
-    return request.data().memory();
-  }
-
-  @Override
-  public LimitsConfiguration limits() {
-    return request.data().limits();
+  public AgentConfiguration configuration() {
+    if (configuration == null) {
```
maff: 🔧 As this is cheap and we can be sure this is needed subsequently in the request, is it necessary to lazily build the object vs initializing it in the constructor (for both contexts)?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/systemprompt/SystemPromptContributor.java
```diff
@@ -22,11 +22,11 @@ public interface SystemPromptContributor {
...
    * Returns the additional system prompt content to be appended.
    *
-   * @param executionContext The current agent execution context
-   * @param agentContext The current agent context
+   * @param agentContext the current durable agent context
+   * @param configuration the current agent configuration
    * @return The system prompt contribution, or null/empty if no contribution needed
    */
-  String contributeSystemPrompt(AgentExecutionContext executionContext, AgentContext agentContext);
+  String contribute(AgentContext agentContext, AgentConfiguration configuration);
```
maff: 🔧 I would prefer if we could design most of the interfaces (not only this contributor here) to receive the execution context + agent context. The execution context can carry relevant job metadata + configuration, the agent context data specific to the agent. I'm aware this is not needed for this A2A contributor, but I would see it as important for these moving parts to be able to access job metadata (let's say, to use the element instance key or element ID as a cache key). In this case I'd suggest keeping the `(executioncontext, agentcontext)` tuple.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/systemprompt/SystemPromptComposerImpl.java
```diff
@@ -42,22 +42,19 @@ public SystemPromptComposerImpl(List<SystemPromptContributor> contributors) {
   }
 
   @Override
-  public String composeSystemPrompt(
-      AgentExecutionContext executionContext,
-      AgentContext agentContext,
-      SystemPromptConfiguration baseSystemPrompt) {
-
+  public String compose(AgentContext agentContext, AgentConfiguration configuration) {
```
maff: 🔧 Same comment as on the contributor below, but we should be able to omit the baseSystemPrompt as it was already changed, trimming it down to the 2 contexts tuple.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/ConversationTurn.java
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * One LLM invocation: the messages that drove it, the response, and per-turn metrics. iterationKey
+ * is 1-based and counts LLM calls across the entire agent lifecycle.
+ *
+ * <p>A turn is <em>pending</em> when {@code assistantMessage} is {@code null} — created by {@link
+ * AgentConversation#rehydrate} before the LLM call — and <em>complete</em> after {@link
+ * AgentConversation#ingest}.
+ */
+public record ConversationTurn(
```
maff: 🔧 As copilot commented on the ADR, maybe it would make sense to keep different "Conversation" concepts (turn, persistence) distinct via naming. As this is part of an `AgentConversation`, this could be an `AgentConversationTurn`? 

There is also the `ConversationTurnComposer` which actually does not compose or produce turns, but the decision (and when it decides to proceed the inputs/messages to the turn). Do you think we can & should decouple these 2 concepts naming-wise (e.g. `AgentConversationTurnInputComposer`) or is it intentional?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
maff: Thanks! Does "is left to add" from the previous variant still apply? Is this also relevant on re-entry (not starting conversation). IIRC the case covered here was an AI Agent Task (re-)entered without tool call results and without a user message (or UM being null).

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
nikonovd: > Hmm I'd still prefer this configuration as "all configuration passed to this execution when the input mappings were applied". The user prompt can change across AHSP instances, but so can the response format or the limits (at least for the moment, also if they're pushed to the cluster agent instance). As a consumer I'd not understand why this does not carry also the user prompt config as the naming implies a complete view.

Agreed, i'll add this. 

> As commented above, I think we should be able to remove the @Nullable on the 2 prompt fields.

👍

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/ConversationTurnComposerImpl.java
nikonovd: Fixed 👍

maff (approved):
Awesome work! 🚀

PR #7532, opened 2026-06-17
fix(github-connector): qualify FEEL variable references in authentication.token expression

### LINKED ISSUE #7533: bug(github-connector): authentication.token always empty — FEEL expression uses unqualified variable names
### Connector
GitHub Connector (element-template only, no Java module)

### Affected versions
Template versions 9, 10, 11 (current)

### Symptom
Every execution of the GitHub connector fails with:

```
jakarta.validation.ValidationException: Found constraints violated while validating input:
 - Property: authentication.token: Validation failed. Original message: must not be empty
```

This happens regardless of whether **Personal Access Token** or **GitHub App** authentication is selected.

### Root cause
The Hidden field that computes `authentication.token` contains this FEEL expression:

```
= if authType = "github_app"
  then {"camunda.function.type":"createGithubAppInstallationToken","params":[githubAppPrivateKey,githubAppAppId,githubAppInstallationId]}
  else pat
```

All sibling form fields bind to **dotted** `zeebe:input` names (`authentication.authType`, `authentication.pat`, etc.). Zeebe evaluates input mappings in source order and materialises dotted names into a nested object, so the FEEL scope contains `authentication.authType` — not a top-level `authType`. All bare references resolve to `null`, the expression collapses to `null`, and `authentication.token` is empty.

The underlying HTTP REST connector then fails the `@NotEmpty` constraint on `BearerAuthentication.token`.

### Fix
Qualify every variable reference with the `authentication.` prefix:

```
= if authentication.authType = "github_app"
  then {"camunda.function.type":"createGithubAppInstallationToken","params":[authentication.githubAppPrivateKey,authentication.githubAppAppId,authentication.githubAppInstallationId]}
  else authentication.pat
```

Fixed in PR #7532.

### Note for users
Existing BPMN diagrams with the old embedded expression must be **re-saved in Camunda Modeler** (open the task, click save) so the corrected expression is re-emitted. Diagrams created after the fix is deployed will work automatically.

gbetances089:
## Description

The GitHub connector element template has a Hidden field that computes the `authentication.token` input for the underlying HTTP REST connector. Its FEEL expression referenced bare variable names (`authType`, `pat`, `githubAppPrivateKey`, `githubAppAppId`, `githubAppInstallationId`), but every sibling form field binds to a dotted `zeebe:input` name under the `authentication.*` prefix.

Per the [Camunda variables docs](https://docs.camunda.io/docs/components/concepts/variables/#inputoutput-variable-mappings), Zeebe evaluates input mappings in source order and materialises dotted names into a nested object. This means the FEEL scope at evaluation time contained `authentication.authType` (nested), not a top-level `authType`. All bare references resolved to `null`, the `if/else` expression evaluated to `null`, and `authentication.token` was set to empty — triggering:

```
jakarta.validation.ValidationException: Found constraints violated while validating input:
 - Property: authentication.token: Validation failed. Original message: must not be empty
```

This affected both the Personal Access Token path (token always null) and the GitHub App path (function call descriptor never constructed).

**Fix:** qualify every variable reference with the `authentication.` prefix in the single Hidden field's `value`:

```diff
- "= if authType = \"github_app\" then {\"camunda.function.type\":\"createGithubAppInstallationToken\",\"params\":[githubAppPrivateKey,githubAppAppId,githubAppInstallationId]} else pat"
+ "= if authentication.authType = \"github_app\" then {\"camunda.function.type\":\"createGithubAppInstallationToken\",\"params\":[authentication.githubAppPrivateKey,authentication.githubAppAppId,authentication.githubAppInstallationId]} else authentication.pat"
```

**Note for users:** existing BPMN diagrams that already embedded the old expression must be re-saved in Camunda Modeler (open the task, click save) to regenerate the `zeebe:input` source with the corrected expression. Diagrams saved after this template is deployed will work automatically.

**Scope:** only `connectors/github/element-templates/github-connector.json` (the unversioned/latest template). The frozen versioned copies (`versioned/github-connector-9.json`, `versioned/github-connector-10.json`) are left untouched per the versioning convention — users on those older template versions should upgrade to the latest.

## Related issues

closes #

## Checklist

- [x] Tests/Integration tests for the changes have been added if applicable. *(The GitHub connector has no Java module — it is element-template only. The fix was verified locally using `LocalFeelExpressionEvaluator` against both the PAT and GitHub App branches, plus a negative-control test confirming the old broken expression returned `null`.)*

### DIFF
--- a/connectors/github/element-templates/github-connector.json
+++ b/connectors/github/element-templates/github-connector.json
@@ -589,7 +589,7 @@
     {
       "group": "authentication",
       "type": "Hidden",
-      "value": "= if authType = \"github_app\" then {\"camunda.function.type\":\"createGithubAppInstallationToken\",\"params\":[githubAppPrivateKey,githubAppAppId,githubAppInstallationId]} else pat",
+      "value": "= if authentication.authType = \"github_app\" then {\"camunda.function.type\":\"createGithubAppInstallationToken\",\"params\":[authentication.githubAppPrivateKey,authentication.githubAppAppId,authentication.githubAppInstallationId]} else authentication.pat",
       "binding": {
         "type": "zeebe:input",
         "name": "authentication.token"


gbetances089:
Manual test confirmed

Screenshot 2026-06-17 at 15 11 52

johnBgood:
## Root Cause Analysis: Why this fix is required

### What the element template does

The GitHub connector element template has a `Hidden` field that computes `authentication.token`:

```
// Before fix (broken):
= if authType = "github_app" then {"camunda.function.type":"createGithubAppInstallationToken","params":[githubAppPrivateKey,githubAppAppId,githubAppInstallationId]} else pat

// After fix (this PR):
= if authentication.authType = "github_app" then {...params:[authentication.githubAppPrivateKey,...]} else authentication.pat
```

All sibling `zeebe:input` bindings use dotted names: `authentication.authType`, `authentication.pat`, `authentication.githubAppPrivateKey`, etc. The Hidden field's FEEL expression was referencing bare names (`authType`, `pat`) that don't exist in the scope.

---

### Why it used to work — and what broke it in 8.9

**Zeebe [#52737](https://github.com/camunda/camunda/pull/52737)** (landed in 8.9+gen7 / 8.8+gen24 / 8.7+gen29) changed how `VariableMappingTransformer.buildIncrementalMappingExpression()` generates the FEEL context for input mappings.

Before that PR, the implementation exposed leaf segments of dotted names as top-level local variables in the incremental context — so `authentication.authType` would also make `authType` available as a top-level binding. The bare-name references worked as a side effect of this internal implementation detail, never by design.

After #52737:
- The context generation was restructured
- Dotted names are materialized **only** as nested objects (`authentication.authType` = nested path), not as top-level flat variables
- Bare names like `authType` resolve to `null`
- This produces: `authentication.token = null` → `ValidationException: must not be empty`

**This same PR also introduced a separate but related critical bug** ([Zeebe issue #54946](https://github.com/camunda/camunda/issues/54946), L1 incident `#inc-rest-connector-no-longer-building-request-body-saas-likely-feel-mapping`): when a later input mapping uses `=a.b` as its source (referencing a variable created by an earlier mapping with a nested target), the parent variable `a` is stale and the traversal returns `null`. This is what broke the REST connector's presales demo. Affected versions were pulled from update paths; fixed in 8.9+gen8, 8.8+gen25, 8.7+gen30.

---

### Why it works in `main` but not `stable/8.9`

This PR was merged into **main** — it updates the element template to use `authentication.authType`, `authentication.pat`, etc. The connector runtime built from `main` ships the corrected template.

**stable/8.9 still ships the old element template** with bare names. When deployed against a Zeebe 8.9 cluster (where #52737's behavior is in effect), the Hidden field FEEL expression evaluates to `null`, triggering the validation failure.

The fix needs to be backported to `stable/8.9`.

johnBgood:
@gustavo-camunda they reverted their "fix" [2 days ago](https://github.com/camunda/camunda/pull/55301).

I tried your fix with 8.9.9 (the version we just released, release train almost done), and it doesn't work. 
It works currently because you tested it on 8.9.8 I assume, where the "fix" I mentioned is present.

Using your fix on 8.9.9 would break the connector, bringing the same issue you tried to avoid here (token=null).

I think we should close this PR once 8.9.9 is out and we can confirm this hypothesis.

gbetances089:
Screenshot 2026-06-18 at 11 54 33

I tested now with 8.9.9 cluster using the v11 version, and it seems to be working, think we can close this and will ask the mono release manager to upgrade the cluster where the train is running.

PR #7549, opened 2026-06-18
fix(deps): pin jaxb-core to resolve SOAP connector build failure

sbuettner:
## Root Cause

The `connector-soap` build failed because `org.glassfish.jaxb:jaxb-core:4.0.9` could not be resolved from the Camunda internal Nexus proxy. The dependency chain:

1. Spring Boot 4.1.0 BOM sets `glassfish-jaxb.version=4.0.9`
2. `spring-ws-core:5.0.2` (managed by Spring Boot 4.1.0) hardcodes `jaxb-runtime:4.0.9` as a runtime dependency
3. `jaxb-runtime:4.0.9` requires `jaxb-core:4.0.9`
4. The Camunda Nexus timed out when proxying `jaxb-core:4.0.9` — this version was never cached there (the GitHub Actions Maven cache previously served it, but the `actions/cache@v6` upgrade caused a cache invalidation)

## Fix

Added an explicit `jaxb-core:4.0.5` override in `parent/pom.xml`'s `dependencyManagement` section, following the same pattern already used for the postgresql version override. Version `4.0.5` is used by `wss4j-ws-security-common:4.0.1` (already a direct dependency of the SOAP connector) and is guaranteed to be in the Nexus cache from prior builds. The `4.0.x` series maintains API compatibility, so mixing `jaxb-runtime:4.0.9` with `jaxb-core:4.0.5` is safe at runtime.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -247,6 +247,17 @@ limitations under the License.</license.inlineheader>
         <version>42.7.11</version>
       </dependency>
 
+      <!--
+        Spring Boot 4.1.0 BOM bumped glassfish-jaxb to 4.0.9 which is not yet cached in the
+        Camunda internal Nexus proxy. Pin jaxb-core to 4.0.5 (the version used by wss4j) so
+        that the SOAP connector can be built without requiring a fresh Nexus proxy fetch.
+      -->
+      <dependency>
+        <groupId>org.glassfish.jaxb</groupId>
+        <artifactId>jaxb-core</artifactId>
+        <version>4.0.5</version>
+      </dependency>
+
       <!-- gRPC BOM - the version has to be equal or greater than the one from the Camunda client -->
       <dependency>
         <groupId>io.grpc</groupId>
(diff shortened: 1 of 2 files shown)


johnBgood (approved):

parent/pom.xml
```diff
@@ -247,6 +247,17 @@ limitations under the License.</license.inlineheader>
         <version>42.7.11</version>
       </dependency>
 
+      <!--
+        Spring Boot 4.1.0 BOM bumped glassfish-jaxb to 4.0.9 which is not yet cached in the
+        Camunda internal Nexus proxy. Pin jaxb-core to 4.0.5 (the version used by wss4j) so
+        that the SOAP connector can be built without requiring a fresh Nexus proxy fetch.
+      -->
+      <dependency>
+        <groupId>org.glassfish.jaxb</groupId>
```
johnBgood, line 256: @sbuettner should we add a Maven enforcer rule to remind us to delete this at some point?

sbuettner: @johnBgood Where did you do this again? You can also push to this branch directly.

johnBgood: I don't have all the context, but[ here's the link](https://github.com/camunda/connectors/pull/7469/changes/fea1aca07b987147557271408d4bf0ade5cac2b7#diff-b5a06276719e759fe07dfe6f75d781be5f83d2215179d82bdb195ad035348214L1024)

PR #7553, opened 2026-06-18
feat(document-handling): Unified response dropdown 

ztefanie:
## Description

This is part 1/3 of this issue: https://github.com/camunda/connectors/issues/7058
Part two is adding the same functionality to the rest connector https://github.com/camunda/connectors/pull/7554
Part three is adding the same functionality to the CSV connector (tbd)

This also includes the fix for this bug https://github.com/camunda/connectors/issues/7049 for 8.10/main. A separate PR to fix this bug on older versions and the old flow is here: https://github.com/camunda/connectors/pull/7608

The new behavior is only for the new versions of the element templates, the old templates keeps working with the new runtime.

Solution proposal document is [here](https://docs.google.com/document/d/1kJsAGoXr1WLF71n7ad3vjViKIqJJQ_ma8WB-BIvhYwc/edit?tab=t.0#bookmark=id.kn8zx66g3dgt)

## QA

I already tested manually all 5 connectors, for backwards compatibility and new feature as well as the file size bug.

## Related issues

partially solves https://github.com/camunda/connectors/issues/7049
closes: https://github.com/camunda/connectors/issues/7058

## Checklist

- [ ] Backport labels are added if these code changes should be backported. No backport label is added to the latest
  release, as this branch will be rebased onto main before the next release. Example backport labels:
    - `backport stable/8.8`: for changes that should be included in the next 8.8.x release.
    - **or** `backport release-8.8.7`: for changes that should be included in the specific release 8.8.7, and this
      *release has already been created*. The release branch will be merged back into stable/8.8 later, so the change
      will be included in future 8.8.x releases as well.
- [ ] Tests/Integration tests for the changes have been added if applicable.
- [ ] If the change requires a documentation update, it has been added to the appropriate section in the documentation.

Issue #7058, opened 2026-04-28
Unified download return format — replace asFile / asDocument / storeResponse booleans


ztefanie:
## Summary

TBD: Should connectors developer be able to specify the default of this dropdown per connector?

Replace the inconsistent boolean toggles (`asFile`, `asDocument`, `storeResponse`) across storage and HTTP connectors with a unified 3-option return format dropdown:

- **Document reference** — saves to Document Store
- **As text** — `new String(bytes, charset)` — optional encoding sub-field, default UTF-8
- **As JSON** — `objectMapper.readTree(bytes)`, Zeebe incident if invalid JSON

<img width="1358" height="532" alt="Image" src="https://github.com/user-attachments/assets/b29f6d25-edfd-4be8-a425-2623ac03c866" />

**Parent epic:** camunda/product-hub#3224
**Replaces:** camunda/connectors#5486 (closed in favor of this technical-scope issue)

For base64 encoding, users apply FEEL's `base64encode()` on the text result.

## Scope

| Connector | Current toggle | After |
|---|---|---|
| AWS S3 | `asFile` boolean | DocumentResponse dropdown |
| Google Cloud Storage | `asDocument` boolean | DocumentResponse dropdown |
| Azure Blob Storage | `asFile` boolean | DocumentResponse dropdown |
| Box | none (always doc store) | DocumentResponse dropdown |
| Google Drive | none (always doc store) | DocumentResponse dropdown |
| HTTP REST / GraphQL | `storeResponse` boolean | DocumentResponse dropdown |
| CSV (Write) | `createDocument` boolean | 2-option dropdown (Document reference / As text only — JSON makes no sense for CSV) |

## Implementation

- New `DocumentResponse` SDK type in `connector-sdk/core` with builder API:
  ```java
  DocumentResponse.from(bytes)
    .format(userChoice)
    .encoding(charset)
    .fileName(name)
    .contentType(contentType)
    .documentCreator(context::create)
    .build();
  ```
- `Format` enum: `DOCUMENT`, `TEXT`, `JSON`
- ETG auto-detects `DocumentResponse` typed fields, generates the dropdown + conditional encoding field

## Backward compatibility — NOT a breaking change at runtime

Old boolean fields (`asFile`, `asDocument`, `storeResponse`) kept on each connector's request model with `@Deprecated`, with fallback logic in the executor:

```java
private Format effectiveFormat() {
  if (returnFormat != null) return returnFormat;
[...]

PR #7554, opened 2026-06-18
feat(document-handling): Unifed response dropdown for rest connector

ztefanie:
## Description

Extends the unified `Response format` dropdown from #7553 to the **HTTP REST** and **GraphQL** connectors,
replacing the old `storeResponse` boolean with `Document reference` / `as text` / `as JSON` (default `as JSON`).

- Streams the response body to the document store / decoder instead of buffering it fully in heap
  (`CustomApacheHttpClient.executeStreaming`).
- Back-compat: `storeResponse` stays but is `@Deprecated` and hidden; old templates keep the legacy path unchanged.
- Templates regenerated — REST **v17**, GraphQL **v11**, with `versioned/` snapshots.

**Behavior note:** on new templates, the `as JSON` default fails the job on non-JSON bodies (was previously
lenient, returning them as text). Intentional — matches the 5 connectors from #7553 and follows least-surprise;
use `as text` for non-JSON. Existing deployments are unaffected.

## Related issues

Part of #7058 · Stacks on #7553

## Checklist

### DIFF
--- a/connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/CustomApacheHttpClient.java
+++ b/connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/CustomApacheHttpClient.java
@@ -26,10 +26,19 @@ import io.camunda.connector.http.client.mapper.HttpResponse;
 import io.camunda.connector.http.client.mapper.ResponseMapper;
 import io.camunda.connector.http.client.mapper.StreamingHttpResponse;
 import io.camunda.connector.http.client.model.HttpClientRequest;
+import java.io.FilterInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.SocketTimeoutException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 import javax.net.ssl.SSLException;
 import org.apache.hc.client5.http.ClientProtocolException;
+import org.apache.hc.core5.http.ClassicHttpRequest;
+import org.apache.hc.core5.http.ClassicHttpResponse;
+import org.apache.hc.core5.http.Header;
 import org.apache.hc.core5.http.HttpStatus;
 
 public class CustomApacheHttpClient implements HttpClient {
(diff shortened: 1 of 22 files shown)


connectors/http/graphql/src/main/java/io/camunda/connector/http/graphql/GraphQLFunction.java
```diff
@@ -79,7 +80,20 @@ public Object execute(OutboundConnectorContext context) {
...
     LOGGER.debug("Executing graphql connector with request {}", commonRequest);
-    var rawResult = httpService.executeConnectorRequest(commonRequest, context);
+    var responseChoice = context.readDocumentReturnFormat().map(f -> f.choice()).orElse(null);
+    var rawResult = httpService.executeConnectorRequest(commonRequest, context, responseChoice);
+    if (rawResult instanceof DocumentReturn<?> documentReturn) {
+      // The body is only materialized inside the wrap lambda, so run the error check there too.
+      return new DocumentReturn<>(
+          documentReturn.payload(),
+          (converted, choice) ->
+              failOnGraphQLErrors(documentReturn.wrap().apply(converted, choice)));
```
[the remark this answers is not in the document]
ztefanie: @copilot I pushed i fixed, is this resolved now?

Szik (approved):

connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/CustomApacheHttpClient.java
```diff
@@ -116,4 +95,142 @@ private static String rootMessage(Throwable t) {
...
+          "TLS handshake failed: "
+              + rootMessage(e)
+              + ". The server certificate may not be trusted — provide the server's CA "
+              + "certificate via 'clientTls.trustedCertificate', or check the client certificate "
+              + "configuration.",
+          e);
+    } catch (IOException e) {
+      closeQuietly(client);
+      throw new ConnectorException(
+          String.valueOf(HttpStatus.SC_REQUEST_TIMEOUT),
```
chillleader, line 137: Is there a more suitable status code / reason phrase for this generic exception?

chillleader (approved):
LGTM, looks clean + nice adjustments 🙌 
just one minor comment

connector-commons/http-client/src/main/java/io/camunda/connector/http/client/client/apache/CustomApacheHttpClient.java
ztefanie: Yes fair point. I just copied this from the existing `execute` method, that handles this the same way. As i do not want to change anything for users especially if not using document handling at all, i think we should keep it as it is. But if you have an idea how this in general could/should be handled better, i am happy to follow up on it in another PR.

PR #7557, opened 2026-06-18
feat(secret-util): add env var to disable legacy unbraced secrets. pattern

vringar:
## Summary

- Adds `CAMUNDA_CONNECTOR_SECRETS_LEGACY_PATTERN_ENABLED=false` env var to disable the unbraced `secrets.KEY` secret pattern in `SecretUtil`
- When disabled, only `{{secrets.KEY}}` is accepted; the bare `secrets.KEY` form is silently ignored
- Both `replaceSecrets` and `retrieveSecretKeysInInput` respect the flag; the check is per-call so the env var is always fresh
- Tests use `system-stubs-jupiter` (`@SystemStub EnvironmentVariables`) to pin the env var to a known value for the whole class, ensuring the suite is deterministic regardless of the JVM environment

## Motivation

The unbraced `secrets.KEY` pattern matches any text that contains the substring `secrets.` — including third-party documentation, Helm chart READMEs, and other untrusted content that agents routinely process. For example, a connector that fetches a Helm chart README (such as the one for [external-secrets](https://github.com/camunda/helm-charts-mirror/blob/main/charts/charts.external-secrets.io/external-secrets/2.6.0/external-secrets/README.md)) will have `secrets.` references in that content silently resolved (or silently left as-is, masking failures), making connectors brittle in the face of untrusted input data.

The `{{secrets.KEY}}` form is unambiguous and intentional; the bare form is a historical convenience that now does more harm than good in environments that process arbitrary text.

## Test plan

- [ ] `SecretUtilTests` passes (26 tests: two new for `replaceSecrets` with legacy disabled, one new for `retrieveSecretKeysInInput` with legacy disabled)
- [ ] Verify default behaviour is unchanged (env var absent → legacy pattern still active)
- [ ] Set `CAMUNDA_CONNECTOR_SECRETS_LEGACY_PATTERN_ENABLED=false` and confirm `secrets.KEY` in a connector input is no longer resolved

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretUtil.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretUtil.java
@@ -37,13 +37,17 @@ public class SecretUtil {
   private static final Pattern SECRET_PATTERN_PARENTHESES =
       Pattern.compile("\\{\\{\\s*secrets\\.(?<secret>\\S+?\\s*)}}");
 
+  private SecretUtil() {}
+
   public static String replaceSecrets(
-      String input, SecretContext context, SecretReplacer secretReplacer) {
+      String input, SecretContext context, SecretReplacer secretReplacer, SecretResolverMode mode) {
     if (input == null) {
       throw new IllegalStateException("input cant be null.");
     }
     input = replaceSecretsWithParentheses(input, context, secretReplacer);
-    input = replaceSecretsWithoutParentheses(input, context, secretReplacer);
+    if (mode == SecretResolverMode.ALL) {
+      input = replaceSecretsWithoutParentheses(input, context, secretReplacer);
+    }
     return input;
   }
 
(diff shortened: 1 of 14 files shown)


connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretUtil.java
```diff
@@ -42,10 +42,19 @@ public static String replaceSecrets(
...
+    if (isLegacyPatternEnabled()) {
+      input = replaceSecretsWithoutParentheses(input, context, secretReplacer);
+    }
     return input;
   }
 
+  // Set CAMUNDA_CONNECTOR_SECRETS_LEGACY_PATTERN_ENABLED=false to accept only {{secrets.KEY}}
+  // and disable the unbraced secrets.KEY pattern that can accidentally match arbitrary text.
+  private static boolean isLegacyPatternEnabled() {
+    return !"false"
```
johnBgood: 🔧 can we avoid double negation for readability?

johnBgood: also fine if you want to keep it  like this, it's a matter of pref

johnBgood (approved):

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretUtil.java
vringar: I don't know if people actually set random env vars, but this way `CAMUNDA_CONNECTOR_SECRETS_LEGACY_PATTERN_ENABLED=YES_PLEASE` still stays active
and only `CAMUNDA_CONNECTOR_SECRETS_LEGACY_PATTERN_ENABLED=FALSE` actually disables something?

jonathanlukas: Why would this even be an env var. Why not enhance the existing SecretUtil.resolveSecrets method to take another parameter `SecretResolverMode` that can be `ALL` or `LEGACY_DISABLED`. In that way, we can inject it in a proper spring way rather than building around it.

vringar: Because these are currently all static methods, so setting via spring runtime wiring would require a static volatile which I really don't want

jonathanlukas: Ok, how about a method parameter? My assumption is these methods are not public API, so we can break their method signature. And then, we can hand a property from the caller to the static method.

A bit more effort, but one less cringe env-parsing.

PR #7308, opened 2026-05-21
feat: add compatibility with scope key tenant evaluation

mathias-vandaele:
This pull request introduces significant improvements to FEEL expression evaluation in inbound connector contexts by integrating a new builder pattern for configuring evaluators, supporting tenant and scope scoping, and updating how properties are bound and evaluated. The changes enhance flexibility, allow for more precise context-aware evaluation (including tenant and element instance scoping), and improve the maintainability of the codebase by removing legacy mechanisms.

Key changes include:

### FEEL Expression Evaluator Enhancements

* Introduced the `FeelExpressionEvaluatorBuilder` class, providing a step-builder pattern to configure FEEL evaluators for both local and Camunda client-based evaluation, with support for tenant ID, scope key, and custom `ObjectMapper` instances. This enables more flexible and explicit evaluator creation.
* Updated `CamundaClientFeelExpressionEvaluator` to support tenant and scope key scoping, with new constructors and logic to apply these options during evaluation. [[1]](diffhunk://#diff-69560fad66ffd01fc9d723c7815a37a7c0e6fcee33bca69cea6723f578c68000R35-R36) [[2]](diffhunk://#diff-69560fad66ffd01fc9d723c7815a37a7c0e6fcee33bca69cea6723f578c68000R46-R74) [[3]](diffhunk://#diff-69560fad66ffd01fc9d723c7815a37a7c0e6fcee33bca69cea6723f578c68000R124-R130)

### Inbound Connector Context Refactoring

* Refactored `InboundConnectorContextImpl` and `DefaultProcessInstanceContext` to use the new FEEL evaluator builder for property binding, replacing the previous supplier-based approach. This allows for FEEL evaluation with tenant and scope context, and ensures properties are bound using the correct evaluation context. [[1]](diffhunk://#diff-3b621a7444b3de88ba434fa0251b7bda0c4b6d3e1d25f44a737011c4b9661d37L278-R301) [[2]](diffhunk://#diff-a60cbba070109d310c3ee61d96356885ed63326d5ccefa5c6cefc7e67c590d8bR69-R77)
[...]

PR #7568, opened 2026-06-19
fix(runtime): restrict outbound connector secret resolution to declared secrets

### LINKED ISSUE #7629: feat(runtime): restrict outbound connector secret resolution to declared secrets (best-effort)
## Problem

Outbound connectors currently resolve any secret referenced in their input, regardless of whether that secret is declared in the connector's BPMN input mappings. A misconfigured or malicious connector template could therefore cause secrets it has no legitimate need for to be resolved and included in the job payload — leaking credentials to unintended recipients.

## Goal

Restrict secret resolution for outbound connectors to only the secret names that appear in the connector element's Zeebe input mappings, as declared in the BPMN process definition. Secrets not explicitly referenced should be left as-is (unresolved) rather than silently injected into the connector input.

## Approach

- `ProcessDefinitionInspector` parses outbound connector elements (ServiceTask / SendTask / ScriptTask / BusinessRuleTask) at job activation time and extracts the set of secret names referenced via Zeebe input mappings.
- `SecretHandler` consults a per-job allow list (`SecretFilter`) so only declared secrets are resolved; undeclared references are left as placeholder strings.
- The filter is resolved lazily (`LazyLoadingSecretFilter`) so lookup failures propagate through the normal connector error path instead of being swallowed silently.

Three configurable modes via `camunda.connector.secret-resolver.allow-list.mode`:

| Mode | Behaviour | Security guarantee |
|------|-----------|-------------------|
| `DISABLED` (default) | All secrets resolved freely — existing behaviour unchanged. | None — feature is off. |
| `LAX` | Enforces allow list when the process definition is available; falls back to allow-all with a warning on any lookup failure (API unavailable, eventual-consistency delay). | Best-effort. A lookup failure silently bypasses the restriction. |
| `STRICT` | Enforces allow list unconditionally; fails the Zeebe job (triggering retries) when the process definition cannot be retrieved. | Strong — no secret resolves outside the declared set, ever. |

The default mode is `DISABLED` so the change is fully backwards-compatible and operators can opt in at their own pace.

## Out of scope

Inbound connectors are unaffected by this change.

jonathanlukas:
## Summary

Prevents outbound connectors from resolving secrets that are not declared in their BPMN input mappings.

### New abstractions (connector-runtime-core)

- **`SecretFilter`** — predicate that controls whether a named secret may be resolved. Two factory methods: `allowAll()` (existing behaviour) and `allowOnly(List<String>)` (restricts to a declared set; empty list denies all).
- **`SecretFilterFactory`** — per-job factory that creates a `SecretFilter` from a `SecretFilterContext` (process definition key + element id).
- **`AbstractConnectorContext`** — now accepts a `SecretFilter` at construction; the `SecretHandler` it builds passes filtered names through before resolving. Unallowed secret references are left as-is in the connector input (logged at DEBUG).
- **Inbound** connectors hardcode `SecretFilter.allowAll()` for now (extending filtering to inbound is tracked in #7730).

### Outbound implementation (connector-runtime-spring)

| Class | Role |
|---|---|
| `SecretKeyCache` | Interface: given a process definition key + element id, returns the list of declared secret names. |
| `ProcessDefinitionSecretKeyCache` | Implementation: fetches BPMN XML via `CamundaClient.newProcessDefinitionGetXmlRequest`, parses Zeebe input mappings for ServiceTask/SendTask/ScriptTask/BusinessRuleTask elements that carry an element template, and extracts secret names via `SecretUtil.retrieveSecretKeysInInput`. Results are cached by process definition key. |
| `ConfigurableSecretFilterFactory` | `SecretFilterFactory` implementation wired via `OutboundConnectorRuntimeConfiguration`. Delegates to `ProcessDefinitionSecretKeyCache` and wraps the result in a `LazyLoadingSecretFilter`. |
| `LazyLoadingSecretFilter` | Defers the BPMN lookup to the first `isAllowed()` call (which happens inside the job handler's try/catch), so lookup failures propagate as a proper Zeebe `failJob` rather than escaping error handling. The supplier is invoked exactly once per filter instance via double-checked locking; a `null` result (LAX fallback) is cached. |

### Configuration

Property: `camunda.connector.secret-resolver.secret-filter.mode` (default: `DISABLED`)

| Mode | Behaviour | Security guarantee |
|------|-----------|-------------------|
| `DISABLED` | All secrets resolved freely — existing behaviour. | None — feature is off. |
| `LAX` | Enforces allow-list when the process definition is available. Falls back to **allow-all** on any lookup failure (API unavailable, eventual-consistency delay, element not yet indexed). | Best-effort. Any lookup failure silently bypasses the restriction. Choose this when uninterrupted job processing matters more than strict secret isolation. |
| `STRICT` | Enforces allow-list unconditionally. Fails the Zeebe job (triggering retries) when the process definition cannot be retrieved. | Strong — no secret resolves outside the declared set. |

Secret-key results are cached separately from the inbound process-definition cache, controlled by:
- `camunda.connector.secret-resolver.secret-filter.cache.enabled` (default: `true`)
- `camunda.connector.secret-resolver.secret-filter.cache.max-size` (default: `1000`)

### Design note — lazy loading

`ConfigurableSecretFilterFactory.create()` is called before the `try/catch` block in `SpringConnectorJobHandler.getConnectorResult`. The actual BPMN lookup is deferred to the first `isAllowed()` invocation inside the guarded path, so exceptions are routed through the standard connector error handling. Jobs with no secret references never trigger a lookup at all.

## Test plan

- [x] `LazyLoadingSecretFilterTest` — single-invocation guarantee, null-fallback caching, allow/deny behaviour
- [x] `ConfigurableSecretFilterFactoryTest` — all three modes including LAX fallback and STRICT propagation
- [x] `ProcessDefinitionSecretKeyCacheTest` — BPMN fixture-based: bracket and bare notation, multi-task isolation, no-template exclusion, no-secrets, unknown element id
- [x] `OutboundConnectorRuntimeConfigurationTest` — `secretKeyCacheManager` enabled/disabled/size-clamping
- [x] `SecretUtilTests.shouldOnlyReplaceAllowListedSecrets` — unreferenced secrets left as-is
- [x] Existing `SpringConnectorJobHandlerTest`, `AnnotatedOperationTests`, `JobHandlerContextTest` — all pass with new API

---

Closes #7629

### DIFF
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
@@ -50,6 +50,7 @@ import io.camunda.connector.runtime.core.inbound.activitylog.ActivityLogWriter;
 import io.camunda.connector.runtime.core.inbound.activitylog.ActivitySource;
 import io.camunda.connector.runtime.core.inbound.correlation.InboundCorrelationHandler;
 import io.camunda.connector.runtime.core.inbound.details.InboundConnectorDetails.ValidInboundConnectorDetails;
+import io.camunda.connector.runtime.core.secret.SecretFilter;
 import java.io.IOException;
 import java.util.List;
 import java.util.Map;
@@ -87,7 +88,7 @@ public class InboundConnectorContextImpl extends AbstractConnectorContext
       ObjectMapper objectMapper,
       ActivityLogWriter activityLogWriter,
       CamundaClient camundaClient) {
-    super(secretProvider, validationProvider);
+    super(secretProvider, SecretFilter.allowAll(), validationProvider);
     this.documentFactory = documentFactory;
     this.correlationHandler = correlationHandler;
     this.connectorDetails = connectorDetails;
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
@@ -33,6 +33,7 @@ import io.camunda.connector.api.secret.SecretContext;
 import io.camunda.connector.api.secret.SecretProvider;
 import io.camunda.connector.api.validation.ValidationProvider;
 import io.camunda.connector.runtime.core.AbstractConnectorContext;
+import io.camunda.connector.runtime.core.secret.SecretFilter;
 import java.util.Objects;
 import org.jspecify.annotations.Nullable;
 import org.slf4j.Logger;
@@ -59,8 +60,9 @@ public class JobHandlerContext extends AbstractConnectorContext
       final SecretProvider secretProvider,
       final ValidationProvider validationProvider,
       final DocumentFactory documentFactory,
-      final ObjectMapper objectMapper) {
-    super(secretProvider, validationProvider);
+      final ObjectMapper objectMapper,
+      final SecretFilter secretFilter) {
+    super(secretProvider, secretFilter, validationProvider);
     this.documentFactory = documentFactory;
     this.job = job;
     this.objectMapper = objectMapper;
--- a/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretHandler.java
+++ b/connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretHandler.java
@@ -20,22 +20,31 @@ import io.camunda.connector.api.error.ConnectorInputException;
 import io.camunda.connector.api.secret.SecretContext;
 import io.camunda.connector.api.secret.SecretProvider;
 import java.util.Optional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class SecretHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(SecretHandler.class);
 
   protected final SecretProvider secretProvider;
 
   protected SecretReplacer secretReplacer;
 
-  public SecretHandler(final SecretProvider secretProvider) {
+  public SecretHandler(final SecretProvider secretProvider, SecretFilter secretFilter) {
     this.secretProvider = secretProvider;
     secretReplacer =
-        (name, context) ->
-            Optional.ofNullable(secretProvider.getSecret(name, context))
+        (name, context) -> {
+          if (secretFilter.isAllowed(name)) {
+            return Optional.ofNullable(secretProvider.getSecret(name, context))
                 .orElseThrow(
                     () ->
                         new ConnectorInputException(
                             String.format("Secret with name '%s' is not available", name)));
+          }
+          LOG.debug("Secret '{}' not in allow-list — placeholder left unreplaced", name);
+          return null;
+        };
   }
 
   public String replaceSecrets(String input, SecretContext context) {
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
@@ -17,6 +17,7 @@
 package io.camunda.connector.runtime.outbound;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import io.camunda.client.CamundaClient;
 import io.camunda.client.jobhandling.JobCallbackCommandWrapperFactory;
 import io.camunda.client.jobhandling.JobWorkerManager;
@@ -32,21 +33,30 @@ import io.camunda.connector.runtime.core.document.store.CamundaDocumentStore;
 import io.camunda.connector.runtime.core.document.store.CamundaDocumentStoreImpl;
 import io.camunda.connector.runtime.core.outbound.DefaultOutboundConnectorFactory;
 import io.camunda.connector.runtime.core.outbound.OutboundConnectorFactory;
+import io.camunda.connector.runtime.core.secret.SecretFilterFactory;
 import io.camunda.connector.runtime.core.secret.SecretProviderAggregator;
 import io.camunda.connector.runtime.core.validation.ValidationUtil;
 import io.camunda.connector.runtime.instances.InstanceForwardingConfiguration;
 import io.camunda.connector.runtime.instances.service.OutboundConnectorsService;
 import io.camunda.connector.runtime.outbound.controller.OutboundConnectorsRestController;
+import io.camunda.connector.runtime.outbound.job.ConfigurableSecretFilterFactory;
+import io.camunda.connector.runtime.outbound.job.ConfigurableSecretFilterFactory.SecretFilterMode;
 import io.camunda.connector.runtime.outbound.jobstream.BrokerJobStreamClient;
 import io.camunda.connector.runtime.outbound.lifecycle.OutboundConnectorManager;
+import io.camunda.connector.runtime.outbound.secret.ProcessDefinitionSecretKeyCache;
+import io.camunda.connector.runtime.outbound.secret.SecretKeyCache;
 import java.net.URI;
 import java.util.Arrays;
 import java.util.List;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.cache.CacheManager;
+import org.springframework.cache.caffeine.CaffeineCacheManager;
+import org.springframework.cache.support.NoOpCacheManager;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
@@ -134,6 +144,37 @@ public class OutboundConnectorRuntimeConfiguration {
   }
 
   @Bean
+  public CacheManager secretKeyCacheManager(
+      @Value("${camunda.connector.secret-resolver.secret-filter.cache.enabled:true}")
+          boolean cacheEnabled,
+      @Value("${camunda.connector.secret-resolver.secret-filter.cache.max-size:1000}")
+          int cacheMaxSize) {
+    if (!cacheEnabled) {
+      return new NoOpCacheManager();
+    }
+    int boundedMaxSize = cacheMaxSize > 0 ? cacheMaxSize : 1000;
+    CaffeineCacheManager cacheManager =
+        new CaffeineCacheManager(SecretKeyCache.SECRET_KEY_CACHE_NAME);
+    cacheManager.setCaffeine(Caffeine.newBuilder().maximumSize(boundedMaxSize));
+    return cacheManager;
+  }
+
+  @Bean
+  public SecretKeyCache secretKeyCache(
+      CamundaClient camundaClient, @Qualifier("secretKeyCacheManager") CacheManager cacheManager) {
+    return new ProcessDefinitionSecretKeyCache(
+        camundaClient, cacheManager.getCache(SecretKeyCache.SECRET_KEY_CACHE_NAME));
+  }
+
+  @Bean
+  public SecretFilterFactory secretFilterFactory(
+      @Value("${camunda.connector.secret-resolver.secret-filter.mode:DISABLED}")
+          SecretFilterMode secretFilterMode,
+      SecretKeyCache secretKeyCache) {
+    return new ConfigurableSecretFilterFactory(secretFilterMode, secretKeyCache);
+  }
+
+  @Bean
   public OutboundConnectorManager outboundConnectorManager(
       JobWorkerManager jobWorkerManager,
       OutboundConnectorFactory connectorFactory,
@@ -142,7 +183,8 @@ public class OutboundConnectorRuntimeConfiguration {
       ValidationProvider validationProvider,
       MetricsRecorder metricsRecorder,
       DocumentFactory documentFactory,
-      @OutboundConnectorObjectMapper ObjectMapper objectMapper) {
+      @OutboundConnectorObjectMapper ObjectMapper objectMapper,
+      SecretFilterFactory secretFilterFactory) {
     return new OutboundConnectorManager(
         jobWorkerManager,
         connectorFactory,
@@ -151,6 +193,7 @@ public class OutboundConnectorRuntimeConfiguration {
         validationProvider,
         documentFactory,
         objectMapper,
-        metricsRecorder);
+        metricsRecorder,
+        secretFilterFactory);
   }
 }
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.outbound.job;
+
+import io.camunda.connector.runtime.core.secret.SecretFilter;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Supplier;
+
+/**
+ * A {@link SecretFilter} that resolves the allowed secret names on the first {@link
+ * #isAllowed(String)} call rather than at construction time. This defers the BPMN process
+ * definition lookup to the point where secrets are actually being replaced, so that any exception
+ * thrown by the supplier propagates through the normal connector error-handling path and results in
+ * a proper Zeebe {@code failJob} command.
+ *
+ * <p>The supplier is called exactly once per filter instance regardless of outcome. When the
+ * supplier returns {@code null} (LAX mode: lookup failed, fall back to allow-all), all subsequent
+ * calls to {@link #isAllowed(String)} return {@code true} without re-invoking the supplier.
+ */
+public class LazyLoadingSecretFilter implements SecretFilter {
+  private final Supplier<List<String>> secretNamesSupplier;
+
+  private volatile boolean initialized = false;
+  private Set<String> secretNames;
+
+  public LazyLoadingSecretFilter(Supplier<List<String>> secretNamesSupplier) {
+    this.secretNamesSupplier = secretNamesSupplier;
+  }
+
+  @Override
+  public boolean isAllowed(String secretName) {
+    if (!initialized) {
+      synchronized (this) {
+        if (!initialized) {
+          List<String> names = secretNamesSupplier.get();
+          secretNames = names != null ? Set.copyOf(names) : null;
+          initialized = true;
+        }
+      }
+    }
+    if (secretNames != null) {
+      return secretNames.contains(secretName);
+    } else {
+      return true;
+    }
+  }
+}
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
@@ -58,6 +58,9 @@ import io.camunda.connector.runtime.core.error.JobError;
 import io.camunda.connector.runtime.core.outbound.ConnectorResult;
 import io.camunda.connector.runtime.core.outbound.ErrorExpressionJobContext;
 import io.camunda.connector.runtime.core.outbound.JobHandlerContext;
+import io.camunda.connector.runtime.core.secret.SecretFilter;
+import io.camunda.connector.runtime.core.secret.SecretFilterFactory;
+import io.camunda.connector.runtime.core.secret.SecretFilterFactory.SecretFilterContext;
 import io.camunda.connector.runtime.core.secret.SecretProviderAggregator;
 import io.camunda.connector.runtime.core.secret.SecretProviderDiscovery;
 import io.camunda.connector.runtime.metrics.ConnectorMetrics;
@@ -90,6 +93,7 @@ public class SpringConnectorJobHandler implements JobHandler {
   private final ValidationProvider validationProvider;
   private final DocumentFactory documentFactory;
   private final ObjectMapper objectMapper;
+  private final SecretFilterFactory secretFilterFactory;
 
   public SpringConnectorJobHandler(
       MetricsRecorder outboundMetrics,
@@ -98,12 +102,14 @@ public class SpringConnectorJobHandler implements JobHandler {
       ValidationProvider validationProvider,
       DocumentFactory documentFactory,
       ObjectMapper objectMapper,
-      OutboundConnectorFunction connectorFunction) {
+      OutboundConnectorFunction connectorFunction,
+      SecretFilterFactory secretFilterFactory) {
     this.call = connectorFunction;
     this.secretProvider = secretProviderAggregator;
     this.validationProvider = validationProvider;
     this.documentFactory = documentFactory;
     this.objectMapper = objectMapper;
+    this.secretFilterFactory = secretFilterFactory;
     this.outboundConnectorExceptionHandler =
         new OutboundConnectorExceptionHandler(getSecretProvider());
     this.connectorResultHandler = new ConnectorResultHandler(objectMapper);
@@ -152,14 +158,23 @@ public class SpringConnectorJobHandler implements JobHandler {
         job.getKey(),
         job.getType(),
         job.getTenantId());
+    var secretFilter =
+        secretFilterFactory.create(
+            new SecretFilterContext(job.getProcessDefinitionKey(), job.getElementId()));
     var context =
         new JobHandlerContext(
-            job, getSecretProvider(), validationProvider, documentFactory, objectMapper);
-    ConnectorResult result = getConnectorResult(job, context);
-    processFinalResult(client, job, context, result, counterMetricsContext);
+            job,
+            getSecretProvider(),
+            validationProvider,
+            documentFactory,
+            objectMapper,
+            secretFilter);
+    ConnectorResult result = getConnectorResult(job, context, secretFilter);
+    processFinalResult(client, job, context, result, counterMetricsContext, secretFilter);
   }
 
-  private ConnectorResult getConnectorResult(ActivatedJob job, OutboundConnectorContext context) {
+  private ConnectorResult getConnectorResult(
+      ActivatedJob job, OutboundConnectorContext context, SecretFilter secretFilter) {
     Duration retryBackoff = null;
     try {
       retryBackoff = getBackoffDuration(job);
@@ -183,7 +198,7 @@ public class SpringConnectorJobHandler implements JobHandler {
       return new ConnectorResult.SuccessResult(connectorResponse, responseVariables);
     } catch (Exception e) {
       return outboundConnectorExceptionHandler.manageConnectorJobHandlerException(
-          e, job, retryBackoff);
+          e, job, retryBackoff, secretFilter);
     }
   }
 
@@ -202,7 +217,8 @@ public class SpringConnectorJobHandler implements JobHandler {
       ActivatedJob job,
       OutboundConnectorContext context,
       ConnectorResult finalResult,
-      CounterMetricsContext counterMetricsContext) {
+      CounterMetricsContext counterMetricsContext,
+      SecretFilter secretFilter) {
     try {
       Optional<ConnectorError> optionalConnectorError =
           connectorResultHandler.examineErrorExpression(
@@ -219,7 +235,8 @@ public class SpringConnectorJobHandler implements JobHandler {
           failJob(
               client,
               job,
-              this.outboundConnectorExceptionHandler.handleFinalResultException(ex, job),
+              this.outboundConnectorExceptionHandler.handleFinalResultException(
+                  ex, job, secretFilter),
               counterMetricsContext);
       notifyFailureOnCommandOutcome(
           failJobRequest,
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.outbound.secret;
+
+import io.camunda.client.CamundaClient;
+import io.camunda.connector.api.inbound.ElementTemplateDetails;
+import io.camunda.connector.runtime.core.secret.SecretUtil;
+import io.camunda.zeebe.model.bpmn.Bpmn;
+import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
+import io.camunda.zeebe.model.bpmn.instance.BaseElement;
+import io.camunda.zeebe.model.bpmn.instance.BusinessRuleTask;
+import io.camunda.zeebe.model.bpmn.instance.FlowElement;
+import io.camunda.zeebe.model.bpmn.instance.Process;
+import io.camunda.zeebe.model.bpmn.instance.ScriptTask;
+import io.camunda.zeebe.model.bpmn.instance.SendTask;
+import io.camunda.zeebe.model.bpmn.instance.ServiceTask;
+import io.camunda.zeebe.model.bpmn.instance.SubProcess;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeInput;
+import io.camunda.zeebe.model.bpmn.instance.zeebe.ZeebeIoMapping;
+import java.io.ByteArrayInputStream;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.cache.Cache;
+
+public class ProcessDefinitionSecretKeyCache implements SecretKeyCache {
+  private static final Logger LOG = LoggerFactory.getLogger(ProcessDefinitionSecretKeyCache.class);
+  private static final List<Class<? extends BaseElement>> OUTBOUND_ELIGIBLE_TYPES =
+      new ArrayList<>();
+
+  static {
+    OUTBOUND_ELIGIBLE_TYPES.add(ServiceTask.class);
+    OUTBOUND_ELIGIBLE_TYPES.add(SendTask.class);
+    OUTBOUND_ELIGIBLE_TYPES.add(ScriptTask.class);
+    OUTBOUND_ELIGIBLE_TYPES.add(BusinessRuleTask.class);
+  }
+
+  private final CamundaClient camundaClient;
+  private final Cache cache;
+
+  public ProcessDefinitionSecretKeyCache(CamundaClient camundaClient, Cache cache) {
+    this.camundaClient = camundaClient;
+    this.cache = cache;
+  }
+
+  @Override
+  public List<String> getSecretKeys(SecretKeyContext secretKeyContext) {
+    return cache
+        .get(
+            secretKeyContext.processDefinitionKey(),
+            () -> fetchSecretKeysByElementIds(secretKeyContext.processDefinitionKey()))
+        .getOrDefault(secretKeyContext.elementId(), Collections.emptyList());
+  }
+
+  private Map<String, List<String>> fetchSecretKeysByElementIds(long processDefinitionKey) {
+    String bpmnXml =
+        camundaClient.newProcessDefinitionGetXmlRequest(processDefinitionKey).execute();
+
+    BpmnModelInstance modelInstance =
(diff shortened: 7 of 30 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -101,6 +105,15 @@ public OutboundConnectorsService outboundConnectorsService(
...
   }
 
+  @Bean
+  public SecretFilterFactory secretFilterFactory(
+      ProcessDefinitionInspector processDefinitionInspector,
+      @Value("${camunda.connector.secret-resolver.allow-list.mode:disabled}")
+          AllowListMode allowListMode) {
+    return new ProcessDefinitionInspectorBoundSecretFilterFactory(
+        processDefinitionInspector, allowListMode);
+  }
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionInspector.java
```diff
@@ -120,20 +154,23 @@ private List<InboundConnectorElement> fetchAndParseConnectors(
...
+        retrieveInboundEligibleElementsFromProcess(process);
+    Collection<BaseElement> outboundEligibleElements =
+        retrieveOutboundEligibleElementsFromProcess(process);
+    if (inboundEligibleElements.isEmpty() && outboundEligibleElements.isEmpty()) {
       LOG.debug(
           "No inbound eligible elements found in process {} (key {})",
           identifier.bpmnProcessId(),
           processDefinitionKey);
-      return Collections.emptyList();
+      return new ConnectorElements(Collections.emptyList(), Collections.emptyList());
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionInspector.java
```diff
@@ -175,10 +212,33 @@ private List<InboundConnectorElement> inspectBpmnProcess(
...
+          new OutboundConnectorElement(element.getId(), definedSecrets));
+    }
+    return new ConnectorElements(discoveredInboundConnectors, discoveredOutboundConnectors);
+  }
+
+  private List<String> extractSecrets(List<ZeebeInput> inputs) {
+    return inputs.stream()
+        .flatMap(input -> SecretUtil.retrieveSecretKeysInInput(input.getSource()).stream())
+        .distinct()
+        .toList();
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
```diff
@@ -0,0 +1,40 @@
+package io.camunda.connector.runtime.outbound.job;
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
```diff
@@ -0,0 +1,40 @@
...
+    if (!initialized) {
+      secretNames = secretNamesSupplier.get();
+      initialized = true;
+    }
+    if (secretNames != null) {
+      return secretNames.contains(secretName);
+    } else {
+      return true;
+    }
+  }
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspectorOutboundTests.java
```diff
@@ -0,0 +1,120 @@
...
+          new ConcurrentMapCacheManager(ProcessDefinitionInspector.PROCESS_DEFINITION_CACHE_NAME);
+      var inspector =
+          new ProcessDefinitionInspector(
+              searchQueryClientMock,
+              cacheManager.getCache(ProcessDefinitionInspector.PROCESS_DEFINITION_CACHE_NAME));
+      var modelFile = ResourceUtils.getFile("classpath:bpmn/" + fileName);
+      var model = Bpmn.readModelFromStream(new FileInputStream(modelFile));
+      var processDefinitionRef = new ProcessDefinitionRef(processId, "tenant1");
+      var processDefinitionKey = 1L;
+      when(searchQueryClientMock.getProcessModel(processDefinitionKey)).thenReturn(model);
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspectorOutboundTests.java
```diff
@@ -0,0 +1,120 @@
...
+    try {
+      var searchQueryClientMock = mock(SearchQueryClient.class);
+      var cacheManager =
+          new ConcurrentMapCacheManager(ProcessDefinitionInspector.PROCESS_DEFINITION_CACHE_NAME);
+      var inspector =
+          new ProcessDefinitionInspector(
+              searchQueryClientMock,
+              cacheManager.getCache(ProcessDefinitionInspector.PROCESS_DEFINITION_CACHE_NAME));
+      var modelFile = ResourceUtils.getFile("classpath:bpmn/" + fileName);
+      var model = Bpmn.readModelFromStream(new FileInputStream(modelFile));
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ae0051fa6.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
```diff
@@ -58,8 +59,9 @@ public JobHandlerContext(
       final SecretProvider secretProvider,
       final ValidationProvider validationProvider,
       final DocumentFactory documentFactory,
-      final ObjectMapper objectMapper) {
-    super(secretProvider, validationProvider);
+      final ObjectMapper objectMapper,
+      SecretFilter secretFilter) {
```
jonathanlukas:
```suggestion
      final SecretFilter secretFilter) {
```

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretUtil.java
```diff
@@ -108,7 +108,7 @@ public static List<String> retrieveSecretKeysInInput(String input) {
         : Stream.of(SECRET_PATTERN_PARENTHESES, SECRET_PATTERN_SECRETS)
             .map(pattern -> pattern.matcher(input))
             .flatMap(Matcher::results)
-            .map(matchResult -> matchResult.group("secret"))
+            .map(matchResult -> matchResult.group("secret").trim())
```
jonathanlukas:
```suggestion
            .map(matchResult -> matchResult.group("secret"))
```

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/outbound/JobHandlerContext.java
jonathanlukas: Applied in 3d95ae163.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretUtil.java
jonathanlukas: Applied in 3d95ae163.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
```diff
@@ -0,0 +1,56 @@
...
+}
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in cb9de9235 — double-checked locking (volatile + synchronized) guarantees the supplier is called exactly once even under concurrent access.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -101,6 +105,15 @@ public OutboundConnectorsService outboundConnectorsService(
...
   }
 
+  @Bean
+  public SecretFilterFactory secretFilterFactory(
+      ProcessDefinitionInspector processDefinitionInspector,
+      @Value("${camunda.connector.secret-resolver.allow-list.mode:DISABLED}")
+          AllowListMode allowListMode) {
+    return new ProcessDefinitionInspectorBoundSecretFilterFactory(
+        processDefinitionInspector, allowListMode);
+  }
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in cb9de9235 — now injected as String and normalized with toUpperCase() before valueOf(), so disabled/lax/strict and DISABLED/LAX/STRICT all work.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionInspector.java
```diff
@@ -175,10 +212,33 @@ private List<InboundConnectorElement> inspectBpmnProcess(
...
   }
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 98f9bb91d.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
```diff
@@ -0,0 +1,60 @@
...
+      synchronized (this) {
+        if (!initialized) {
+          secretNames = secretNamesSupplier.get();
+          initialized = true;
+        }
+      }
+    }
+    if (secretNames != null) {
+      return secretNames.contains(secretName);
+    } else {
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 98f9bb91d.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionInspectorOutboundTests.java
```diff
@@ -0,0 +1,124 @@
...
+
+public class ProcessDefinitionInspectorOutboundTests {
+
+  @Test
+  void shouldDetectSecretInBracketNotation() {
+    var outbound = fromModel("outbound-with-secrets.bpmn", "outbound-with-secrets");
+    assertThat(outbound).hasSize(1);
+    assertThat(outbound.getFirst().definedSecrets()).contains("API_KEY");
+  }
+
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 98f9bb91d.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/ProcessDefinitionInspectorBoundSecretFilterFactory.java
```diff
@@ -0,0 +1,89 @@
...
+                            "Did not find element with id "
+                                + context.elementId()
+                                + " in process definition "
+                                + context.processDefinitionKey()));
+          } catch (Exception e) {
+            if (strict) {
+              throw new IllegalArgumentException(
+                  "Error filtering secrets for process definition "
+                      + context.processDefinitionKey(),
+                  e);
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 98f9bb91d.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -147,6 +151,15 @@ public OutboundConnectorsService outboundConnectorsService(
         outboundConnectorConfigurationRegistry, gatewayJobStreamClient, brokerJobStreamClient);
   }
 
+  @Bean
+  public SecretFilterFactory secretFilterFactory(
+      ProcessDefinitionInspector processDefinitionInspector,
+      @Value("${camunda.connector.secret-resolver.allow-list.mode:disabled}")
+          String allowListMode) {
+    return new ProcessDefinitionInspectorBoundSecretFilterFactory(
+        processDefinitionInspector, AllowListMode.valueOf(allowListMode.toUpperCase()));
```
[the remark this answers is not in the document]
jonathanlukas: Instead, we should inject the enum directly and let spring do the parsing.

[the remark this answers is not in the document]
jonathanlukas: Fixed in b08608c9b — injecting `AllowListMode` directly via `@Value` and letting Spring handle the conversion (Spring Boot's `ApplicationConversionService` registers a case-insensitive enum converter, so `disabled`, `lax`, and `strict` all work).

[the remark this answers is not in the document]
jonathanlukas: Applied in b08608c9b.

sbuettner:
@jonathanlukas this should be solved via https://camunda.slack.com/archives/C099MRHA8U8

jonathanlukas:
@sbuettner we know, but this is blocking us in production right now. This is why we designed this whole implementation to be backported and shipped with a patch release asap.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -54,7 +59,11 @@
 import org.springframework.core.env.Environment;
 
 @Configuration
-@Import({OutboundConnectorsRestController.class, InstanceForwardingConfiguration.class})
+@Import({
+  OutboundConnectorsRestController.class,
+  InstanceForwardingConfiguration.class,
+  ProcessDefinitionManagementConfiguration.class
+})
```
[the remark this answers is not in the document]
jonathanlukas: False positive — `ProcessDefinitionManagementConfiguration` exists at `io.camunda.connector.runtime.search.ProcessDefinitionManagementConfiguration` and is tracked in HEAD.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionInspector.java
```diff
@@ -120,20 +154,23 @@ private List<InboundConnectorElement> fetchAndParseConnectors(
...
+    Collection<BaseElement> outboundEligibleElements =
+        retrieveOutboundEligibleElementsFromProcess(process);
+    if (inboundEligibleElements.isEmpty() && outboundEligibleElements.isEmpty()) {
       LOG.debug(
-          "No inbound eligible elements found in process {} (key {})",
+          "No connector elements found in process {} (key {})",
           identifier.bpmnProcessId(),
           processDefinitionKey);
-      return Collections.emptyList();
+      return new ConnectorElements(Collections.emptyList(), Collections.emptyList());
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 44c7e841d — `getProcessDefinition` is now only called when `inboundEligibleElements` is non-empty.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/ProcessDefinitionInspectorBoundSecretFilterFactory.java
```diff
@@ -0,0 +1,96 @@
...
+                  context.elementId(),
+                  context.bpmnProcessId(),
+                  context.processDefinitionKey(),
+                  context.tenantId(),
+                  e);
+              return null;
+            }
+          }
+        });
+  }
```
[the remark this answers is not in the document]
jonathanlukas: The "LAX" mode is intended to function as a "best-effort" way of restricting secret access, for example to enabling demos where the setup usually is not perfectly clean.

Strict mode is the only way production should work, and this covers everything.

sbuettner:
@jonathanlukas Would it be sufficient for your use-case to limit secret resolving to only certain activities?

If so what about having a header flag that disables secrets (either a predefined list or at all) for certain jobs?

jonathanlukas:
@sbuettner that would not work. We are using Agents that are going to incident state because a call to the camunda docs mcp returns markdown snippets that contain secret references (because we query the docs for them).

We cannot configure an Agent without using secrets.


To go into details, my anticipated approach is:

* we backport this to 8.9
* this stays backwards-compatible, so you can still decide whether you want to roll out a breaking change in 8.10
* Saas runtime does not change by default
* we could modify our saas runtimes to perform strict secret resolving
* we modify our hybrid connector to perform strict secret resolving

What I see here for you:

* no breaking change
* no actual change to current behaviour
* an opt-in to have a solution in place in case a customer asks for it
* potential to enforce better security by default

What I see for us:

* we cry less because of secrets

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/ProcessDefinitionInspectorBoundSecretFilterFactory.java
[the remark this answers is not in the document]
jonathanlukas: LAX mode is intentionally insecure by design — it is a best-effort implementation where uninterrupted job processing takes priority over strict secret isolation. Any lookup failure (API unavailable, eventual-consistency delay, element not yet indexed, or element genuinely not found) falls back to allow-all. Updated the PR description to document this tradeoff explicitly for all three modes.

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/InboundConnectorRuntimeConfigurationTest.java
```diff
@@ -28,8 +29,8 @@
 
 class InboundConnectorRuntimeConfigurationTest {
 
-  private final InboundConnectorRuntimeConfiguration configuration =
-      new InboundConnectorRuntimeConfiguration();
+  private final ProcessDefinitionManagementConfiguration configuration =
+      new ProcessDefinitionManagementConfiguration();
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 59c5be13a — renamed to `ProcessDefinitionManagementConfigurationTest` and moved to the `search` package to match the configuration class.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
```diff
@@ -0,0 +1,62 @@
...
+      return true;
+    }
+  }
+}
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in 0592c6d26 — added `LazyLoadingSecretFilterTest` covering: (1) allow-listed secret permitted, (2) non-listed secret denied, (3) null supplier result → allow-all, (4) supplier called exactly once on multiple invocations, (5) null result cached (supplier not re-invoked).

sbuettner:
@jonathanlukas 

> We cannot configure an Agent without using secrets.

My point was that certain jobs should not be able to resolve secrets OR have dedicated list of secrets based on headers as an allow-list. The latter should solve your issue no?

We could also think of having a list of properties that allow to accept secrets so other user-data properties are not considered and excluded (also via headers)

jonathanlukas:
>My point was that certain jobs should not be able to resolve secrets OR have dedicated list of secrets based on >headers as an allow-list. The latter should solve your issue no?
>
>We could also think of having a list of properties that allow to accept secrets so other user-data properties are not >considered and excluded (also via headers)

All of this sounds like we would have to implement this as well. What's your point?

sbuettner:
Yes, but way less invasive and complex.

jonathanlukas:
Understood. 

That means you fear the complexity of downloading and parsing a model for outbound connectors.

As the only difference would be that the allow list of secrets would be sourced from the header, not from the process model.

I can adjust my PR to have this included, but I would actually like to leave the feature as being implemented in here.

Any objections?

sbuettner:
I would have to take a closer look but if everything is done in a way it can easily be disabled I dont see a big issue if you are relying on it for production go-live. If we keep the code the additional header logic should AFAIK not be implemented.

Did you investigate whether adding the element- + bpmn-definition-information to the secret provider context would work? This way we could keep the logic in the secret provider without changing the runime too much. 

You could have a proxy like provider that wraps the existing ones that basically implements your check.

jonathanlukas:
>I would have to take a closer look but if everything is done in a way it can easily be disabled I dont see a big issue if you are relying on it for production go-live. If we keep the code the additional header logic should AFAIK not be implemented.

The whole solution is implemented to not affect anything by default. The feature has to enabled explicitly using a property.

This is documented in the PR description. Should I also set up an issue describing the anticipated solution?

>Did you investigate whether adding the element- + bpmn-definition-information to the secret provider context would work? This way we could keep the logic in the secret provider without changing the runime too much. You could have a proxy like provider that wraps the existing ones that basically implements your check.

That was my first approach. The problem is that as soon as the secret provider is hit, I cannot propagate "I don't want to replace the secret" as the secret resolver throws as soon as the resolved secret is `null`.

This is why I came up with the "clean" approach to make the secret resolver aware of the filter.

sbuettner:
> This is documented in the PR description. Should I also set up an issue describing the anticipated solution?

This is the preferred approach in general so we can discuss this with the team before opening a PR.

> That was my first approach. The problem is that as soon as the secret provider is hit, I cannot propagate "I don't want to replace the secret" as the secret resolver throws as soon as the resolved secret is null.

I guess if you dont want to exchange a secret you could just return the key again?

jonathanlukas:
>I guess if you dont want to exchange a secret you could just return the key again?

Not really, as the replacement would make `{{secret.MY_SUPER_SECRET}}` become `MY_SUPER_SECRET` if I just return the key. 

Now, I could add the whole declaration back.

This would then not be sensitive to the declaration way of the secret, so it could alter the content, depending on whether the old `secrets.MY_SUPER_SECRET` or the new `{{secret.MY_SUPER_SECRET}}` pattern is used. 

While this might not be relevant in other cases, we quote from our docs and here, the correctness of the quoted section is relevant.

>This is the preferred approach in general so we can discuss this with the team before opening a PR.

I will create an issue for this.

sbuettner:
@jonathanlukas As this is a temporary fix anyway I dont see your case as breaking issue. Especially since the old secret format is considered deprecated so it would never be a good idea to return it. 

Please open the PR so the team can take a look. Thx!

@johnBgood @chillleader would love to get your perspective on this as well.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionInspector.java
```diff
@@ -120,65 +154,94 @@ private List<InboundConnectorElement> fetchAndParseConnectors(
...
 
-  private List<InboundConnectorElement> inspectBpmnProcess(
+  private ConnectorElements inspectBpmnProcess(
       Process process, ProcessDefinitionRef identifier, long processDefinitionKey) {
-    Collection<BaseElement> inboundEligibleElements = retrieveEligibleElementsFromProcess(process);
-    if (inboundEligibleElements.isEmpty()) {
+    Collection<BaseElement> inboundEligibleElements =
+        retrieveInboundEligibleElementsFromProcess(process);
+    Collection<BaseElement> outboundEligibleElements =
+        retrieveOutboundEligibleElementsFromProcess(process);
```
johnBgood: Now even when I'm interested in Inbound connectors only, I will deal with outbound connectors, go through a lot of for loops when it's not necessary.

jonathanlukas:
Tracked in #7629.

jonathanlukas:
@sbuettner I opened #7629

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -96,19 +99,118 @@ public SpringConnectorJobHandler(
...
+      SecretFilterFactory secretFilterFactory) {
     this.call = connectorFunction;
     this.secretProvider = secretProviderAggregator;
     this.validationProvider = validationProvider;
     this.documentFactory = documentFactory;
     this.objectMapper = objectMapper;
+    this.secretFilterFactory = secretFilterFactory;
     this.outboundConnectorExceptionHandler =
         new OutboundConnectorExceptionHandler(getSecretProvider());
     this.connectorResultHandler = new ConnectorResultHandler(objectMapper);
```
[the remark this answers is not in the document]
jonathanlukas: Fixed in ffca28de1 — filter secret keys through the job's `SecretFilter` before `fetchAll` in both `manageConnectorJobHandlerException` and `handleFinalResultException`, so only allow-listed secrets are ever fetched even during error handling.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretHandler.java
```diff
@@ -27,15 +27,19 @@ public class SecretHandler {
...
-            Optional.ofNullable(secretProvider.getSecret(name, context))
+        (name, context) -> {
+          if (secretFilter.isAllowed(name)) {
+            return Optional.ofNullable(secretProvider.getSecret(name, context))
                 .orElseThrow(
                     () ->
                         new ConnectorInputException(
                             String.format("Secret with name '%s' is not available", name), null));
+          }
+          return null;
```
johnBgood, line 46: maybe we can log something here

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
```diff
@@ -0,0 +1,62 @@
...
+  private Set<String> secretNames;
+
+  public LazyLoadingSecretFilter(Supplier<List<String>> secretNamesSupplier) {
+    this.secretNamesSupplier = secretNamesSupplier;
+  }
+
+  @Override
+  public boolean isAllowed(String secretName) {
+    if (!initialized) {
+      synchronized (this) {
```
johnBgood, line 48: do we need this? as far as I can see we create a secret filter instance during each `handle(...)` in `SpringConnectorJobHandler`

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
```diff
@@ -96,19 +100,118 @@ public SpringConnectorJobHandler(
...
     this.jobCallbackCommandWrapperFactory = jobCallbackCommandWrapperFactory;
     this.connectorsOutboundMetrics = outboundMetrics;
   }
 
+  private static JobCallbackFinalCommandStep<FailJobResponse> prepareFailJobCommand(
+      JobClient client, ActivatedJob job, ConnectorResult.ErrorResult result) {
```
johnBgood: I guess you moved some code, can you revert if that's the case?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -147,6 +156,15 @@ public OutboundConnectorsService outboundConnectorsService(
         outboundConnectorConfigurationRegistry, gatewayJobStreamClient, brokerJobStreamClient);
   }
 
+  @Bean
+  public SecretFilterFactory secretFilterFactory(
+      ProcessDefinitionInspector processDefinitionInspector,
+      @Value("${camunda.connector.secret-resolver.allow-list.mode:DISABLED}")
+          AllowListMode allowListMode) {
```
johnBgood: maybe `secretFilterMode` would be clearer at first sight

sbuettner:
@jonathanlukas Some feedback from the team:

Implement the BPMN fetching, caching in a separate module/class without touching existing functionality of the `ProcessDefinitionInspector`. Keep the functionality outside of the `inbound` packages. Use this new capability in the `SecretFilter` implementation. 

Regarding the cache, only cache the secret names or even only their computes hashes:
`(process definition key, element id) -> List of allowed secret names (or hashes)`

We like the idea of a `camunda.connector.secret-resolver.allow-list.mode` and the current implementation is fine.

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/secret/SecretHandler.java
jonathanlukas: Fixed in c30e53f80 — added `LOG.debug` in the `secretReplacer` lambda when `isAllowed` returns false.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessDefinitionInspector.java
jonathanlukas: The `inspectBpmnProcess` result is stored in a shared cache keyed by process definition. Because both the inbound and outbound subsystems draw from that same cache entry, a single BPMN parse must yield both sets of elements — splitting the paths would require two separate cache entries (or two model fetches) per process definition, which is worse overall. The outbound element iteration is O(n) over BPMN elements and negligible compared to the `getProcessModel` network call that has already happened by this point.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/LazyLoadingSecretFilter.java
jonathanlukas: The laziness isn't about thread safety — it's about error propagation. `secretFilterFactory.create()` is called in `internalHandle` *before* the try/catch in `getConnectorResult`. If the BPMN supplier were invoked eagerly in the constructor, any lookup exception would escape that handler and bypass `manageConnectorJobHandlerException`, producing an unhandled error instead of a proper `failJob` command. Deferring the supplier call to the first `isAllowed()` invocation (which happens inside secret replacement, itself inside the guarded code path) ensures exceptions route through the standard connector error handling. The secondary benefit is that jobs whose inputs contain no secret placeholders never trigger the BPMN lookup at all.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/job/SpringConnectorJobHandler.java
jonathanlukas: Fixed in 3d74d8a90 — moved the seven static helper methods back to the bottom of the class, restoring their original position.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
jonathanlukas: Fixed in f9bda9949 — renamed to `secretFilterMode`.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
```diff
@@ -0,0 +1,167 @@
...
+  public ProcessDefinitionSecretKeyCache(CamundaClient camundaClient, Cache cache) {
+    this.camundaClient = camundaClient;
+    this.cache = cache;
+  }
+
+  @Override
+  public List<String> getSecretKeys(SecretKeyContext secretKeyContext) {
+    return cache
+        .get(
+            secretKeyContext.processDefinitionKey(),
```
johnBgood, line 71: Can"t the key be `processDefKey__ElementId` (we need to add the `elementId` in the `SecretContext` if so). That would give direct access to the corresponding secrets

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
```diff
@@ -86,7 +87,7 @@ public InboundConnectorContextImpl(
       ObjectMapper objectMapper,
       ActivityLogWriter activityLogWriter,
       CamundaClient camundaClient) {
-    super(secretProvider, validationProvider);
+    super(secretProvider, SecretFilter.allowAll(), validationProvider);
```
johnBgood, line 91: Can't we have this enabled depending on the config param, similar to how it's done for Outbound?
In this case, `ProcessDefinitionSecretKeyCache` shouldn't be limited to outbound

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
jonathanlukas: Why should we not keep the element id as key of a map?

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
jonathanlukas: As inbound connectors use properties as their input, there is nothing processed on the zeebe side, so technically there is no way to declare a secret dynamically. Or am I mistaken?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
johnBgood: but for now the key is the process definition key (?), or do you mean the value Map?

johnBgood: up to you, both options are fine to me

jonathanlukas: Yes, I mean the value map. The assumption is that the amount of element ids for a process definition is finite.

And I am not a fan of composed keys :)

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImpl.java
johnBgood: The HTTP polling connector [can now fetch process variables](https://github.com/camunda/connectors/pull/7308), for instance

jonathanlukas: Ah non! 
That would mean what I create should be done for outbound AND inbound.... That changes things a bit :)

johnBgood: sorry :D

johnBgood: we can also do it in a follow-up PR, what you've implemented is already a great improvement. 

As long as you don't have to break too much work in this follow-up PR :/

jonathanlukas: I think that would work. I see your point in the distinction between inbound and outbound connector processing, so I would not change the work I already did rather than extending the inbound connector processing side.

jonathanlukas: Tracked in #7730.

sbuettner (approved):

johnBgood:
Update on the backports: opened #8496 for \`stable/8.8\` (manual, chained off the #7740 8.9 backport since that version already threads the secret filter through the AI Agent job-worker path).

**\`stable/8.7\` and \`end-of-life/8.6\`: not feasible as a backport.** This isn't a conflict-resolution problem — the feature has no data source on those branches. \`ProcessDefinitionSecretKeyCache\` fetches the process definition's BPMN XML via \`CamundaClient.newProcessDefinitionGetXmlRequest\`, part of the newer client "search" API. Neither branch has it: both are still on the pre-rename \`ZeebeClient\` and have no \`inbound/search\` package at all (that only landed with 8.8). Backporting there would mean building a new BPMN-fetch mechanism against an older client — new feature work, not a backport.

\`end-of-life/8.6\` additionally hasn't had a commit since 2026-04-16 and has no working release pipeline, so it's doubly out of scope.

I'd suggest removing the \`backport stable/8.7\` and \`backport end-of-life/8.6\` labels here to stop the bot from retrying them. If 8.7 needs this fix, it'd need to be scoped as its own change against that branch's older client API rather than a backport.

johnBgood:
Correction to my last comment: \`stable/8.7\` and \`end-of-life/8.6\` **are** backportable — I was wrong to say otherwise.

I anchored on the specific API (\`CamundaClient.newProcessDefinitionGetXmlRequest\`) instead of the underlying capability. Both branches already have an equivalent: \`CamundaOperateClient.getProcessDefinitionModel(long processDefinitionKey)\`, which is already wired as a Spring bean and used by the existing inbound \`ProcessDefinitionInspector\` on both branches. It returns a parsed \`BpmnModelInstance\` directly, so \`ProcessDefinitionSecretKeyCache\` can use it as a drop-in data source instead of fetching+parsing raw XML.

Please disregard the "not feasible" comment and keep the \`backport stable/8.7\` / \`backport end-of-life/8.6\` labels — I'm implementing both now, chained off #8496 (the \`stable/8.8\` backport).

One caveat I'll call out explicitly in each PR: on 8.7/8.6 the lookup goes through Operate rather than the Zeebe/Camunda client's REST search API, so enabling \`LAX\`/\`STRICT\` mode on those branches will require Operate to be reachable, which isn't a requirement for \`DISABLED\` (default) mode.

johnBgood:
Found via automated review on #8496 (the 8.8 backport), but this bug is already live in the merged code here too, not something the backport introduced:

**\`JobWorkerAgentExecutionContextFactoryImpl\` and \`AiAgentJobWorkerHandlerImpl\` each create their own separate \`SecretFilter\` instance for the same job** — one for input resolution (\`createExecutionContext\`), one for error masking (\`executeJob\`). In \`LAX\` mode, if the first lookup transiently fails (falls back to allow-all) and the second succeeds (or vice versa), the two instances can disagree on which secrets are allowed. The masking-side instance's disagreement means a secret that's actually not meant to be resolvable can still get its real value included in the redaction step, which functions correctly for masking (it hides that value from the message), but it exposes an inconsistency: the same job effectively toggles between the currently declared/allowed set across the two independent lookups, undermining the guarantee that a job's allow-list is fixed for that job execution.

I fixed this on #8496 (\`stable/8.8\`) by having \`createExecutionContext\` take the already-created \`SecretFilter\` as a parameter instead of building its own — see d4578729a9 for the diff. This same fix should go to \`stable/8.9\` (#7740) and \`release-8.10.0-alpha3\` (#7749), both already merged and shipped. Happy to open that follow-up PR if wanted — let me know.

PR #7592, opened 2026-06-22
fix(ci): correct nightly e2e notify jq syntax and spurious matrix jobs

nikonovd:
## Description

Bug 1 — gh api --jq does not forward --arg to jq; passing both caused "accepts 1 arg(s), received 4". Fix: pipe --paginate output through jq -rs --arg so the filter receives the prefix variable correctly. Also switch GITHUB_OUTPUT write to the multiline heredoc form since the value may now contain real newlines.

Bug 2 — matrix.include with a module name not present in the dynamic discovery list unconditionally adds a new job (GitHub Actions footgun). This caused connectors-e2e-test-agentic-ai to run on stable/8.7 where the module doesn't exist. Fix: embed runner into the discovery output as {"module": "...", "runner": "..."} objects resolved from the new e2e-runner-overrides.json config file, eliminating matrix.include entirely. To assign a custom runner to a new module, add one line to the JSON file — no YAML changes required.

## Checklist

### DIFF
--- a/.github/workflows/E2E_BRANCH_RUN.yml
+++ b/.github/workflows/E2E_BRANCH_RUN.yml
@@ -76,8 +76,9 @@ jobs:
 
       # Discover the e2e-test modules to run from this branch's checkout so the test matrix is
       # branch-accurate. Leaf poms with `jar` packaging are test modules; aggregators (`pom`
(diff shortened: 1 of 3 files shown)


maff (approved):
Thanks for fixing! 💪

sbuettner (approved):

PR #7607, opened 2026-06-23
fix: bump netty to 4.2.15.Final

ztefanie:
## Summary

Pins `io.netty:netty-bom` to `4.2.15.Final` explicitly (Spring Boot 3.5.14 BOM shipped 4.1.132.Final). Also removes the now-redundant individual `netty-codec-dns` pin introduced for a prior CVE — it is covered by the netty BOM at 4.2.15.Final.

Security fix. Details in the internal tracking issue: camunda/team-connectors#1233


ztefanie:
This version was already bumped

Issue #7640, opened 2026-06-25
AI Agent: Available tools are not provided when updating agent instances


christoph-fricke:
__Describe the Bug__

The [agent instance update command](https://docs.camunda.io/docs/next/apis-tools/orchestration-cluster-api-rest/specifications/update-agent-instance/) accepts `tools` that are available for the agent instance to call. The information is used in Operate to display available tools in the agent details. However, the agentai connector current does not provide tool information when updating an agent instance.

__Steps to Reproduce__

1. Start a process with an Agent AHSP and tools
2. Observe the agent instance response in the network tab. It's `tool` property remains empty even though the agent starts to call tools.

__Expected Behavior__

If an agent instance has tools available (and calls these tools) the `tools` property on an agent instance should not be empty. The connector should provide the information when updating the agent instance.

__Environment__

- OS: MacOS 26.5
- Library version: 8.10-SNAPSHOT
- Camunda version: 8.10-SNAPSHOT (c8Run)

PR #7683, opened 2026-06-28
test(e2e): accept empty tools list on agent-instance metrics update

nikonovd:
## Description

#7662 changed the metrics-update matcher in `AgentInstanceClientVerifier`
to require `request.tools() == null`. However `AgentInstanceUpdateRequest` is
an `@AgenticAiRecord` built with `useImmutableCollections = true`, so an unset
tools list is normalized to a non-null empty list rather than null. The
metrics update therefore carries an empty collection, and the matcher could
never match, failing every toolCallTurn/finalAnswerTurn verification.

## Related issues

relates to #7640 

## Checklist

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/assertj/AgentInstanceClientVerifier.java
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/assertj/AgentInstanceClientVerifier.java
@@ -119,9 +119,6 @@ public class AgentInstanceClientVerifier {
                     request != null
                         && request.status() == endStatus
                         && Objects.equals(request.delta(), delta)
-                        // No tools on the metrics update (only THINKING carries them).
-                        // AgentInstanceUpdateRequest uses immutable collections, so an unset tools
-                        // list is normalized to an empty list rather than null.
                         && (request.tools() == null || request.tools().isEmpty())));
 
     final var expectedIterationKey = ++turnCount;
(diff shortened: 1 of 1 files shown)


connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/assertj/AgentInstanceClientVerifier.java
```diff
@@ -119,7 +119,7 @@ private AgentInstanceClientVerifier chatTurn(
                     request != null
                         && request.status() == endStatus
                         && Objects.equals(request.delta(), delta)
-                        && request.tools() == null));
+                        && (request.tools() == null || request.tools().isEmpty())));
```
maff, line 122: 🔧 As the PR description states this can never be null due to the record builder config, why do we need both variants? Can we expect this always to be an empty list?

nikonovd: A test should be independent of the implementation behind it, to avoid regressions.

Furthrmore, the client code is also implemented equally, to avoid regressions (e.g. if we change the implementation of the annotation).

maff: Agree on the the test being independent of the implementation, however if null/empty list does have a different semantic meaning (e.g. "do not update the tools part with this update call" vs "agent has no tools"), I'd prefer expecting only the actually expected variant instead of widening the list of variations. I guess we can revisit this also as part of the nullability checks you were working on 👍

maff (approved):

PR #7684, opened 2026-06-29
refactor(agentic-ai): reorganize agentic-ai connector module structure

maff:
## Description

Converts the `agentic-ai` module to a multi-module Maven layout, aligning it with other multi-module connectors (e.g. `http`) and laying the groundwork for a future split of the agentic AI codebase into separate submodules.

For now, all code moves unchanged into a single `connector-agentic-ai` submodule under a new `connector-agentic-ai-parent` aggregator. Future submodule splits (e.g. extracting `mcp`, `a2a`, `aiagent`) can then be done incrementally without restructuring the Maven hierarchy again.

Changes:
- New aggregator POM (`connector-agentic-ai-parent`) at the `agentic-ai/` level
- All source code, build scripts, and element templates moved into `connector-agentic-ai/` submodule; docs, examples, `AGENTS.md`, and `CLAUDE.md` remain at the parent level
- Symlink at `agentic-ai/element-templates` → `connector-agentic-ai/element-templates` for backwards compatibility with any external links
- Updated path references in `AGENTS.md`, `CLAUDE.md`, and `docs/reference/ai-agent.md`

The published artifact ID (`connector-agentic-ai`) is unchanged, so no downstream consumer changes are required.

## Checklist

https://claude.ai/code/session_013JShjsz5jAom7FP2VqXnzg


connectors/agentic-ai/CHANGELOG.md
maff, line 1: ℹ️ This one was stale

connectors/agentic-ai/element-templates
```diff
@@ -0,0 +1 @@
+connector-agentic-ai/element-templates
```
nikonovd, line 1: 💭 After our discussion regarding the marketplace I quickly checked:

- The connector templates are referenced in the repository, but IMO they should also be updated eventually after merging, as the `connector-templates.json` should get updated.
- The blueprints around the AI agent are linking to the [Camunda 8 Tutorials](https://github.com/camunda/camunda-8-tutorials/tree/main/solutions) repo, so we should be good on that front

connectors/agentic-ai/docs/reference/ai-agent.md
nikonovd, line 1: ❓ Is it intentionally kept at the aggregator module now?

I'm fine with that, we just need to consider splitting up the documentation as well when modularizing the new `connector-agentic-ai` module (i.e. add it as a note to the existing GH issue)

nikonovd (approved):
Great work 🚀 thanks and lgtm. Two comments, non-blocking.

Did you consider changing the `artifactId`s later when we split the modules? We should consider then having connector-agentic-ai as a bundle jar as a legacy-artifact that contains all modules we have (which could then also be injected into the default connectors bundles.

connectors/agentic-ai/docs/reference/ai-agent.md
maff: Yes, I kept it intentionally here for now as the docs & ADRs reference the whole codebase (e.g. the decision on the MCP SDK). In follow-ups, we can decide if we want to keep this shared documentation or if it makes sense to move it into the respective modules.

maff:
> Did you consider changing the artifactIds later when we split the modules? We should consider then having connector-agentic-ai as a bundle jar as a legacy-artifact that contains all modules we have (which could then also be injected into the default connectors bundles.

That's a good point. Right now it's the same artifact ID, so the change should be transparent. We'd probably want to split the individual connectors into modules and could keep `connector-agentic-ai` as fallback bundle module as you suggested.

nikonovd:
Sounds like something we could note for the proceeding modularization. 👍

PR #7697, opened 2026-06-29
feat(agentic-ai): apply @NullMarked to aiagent package

nikonovd:
## Summary

Applies JSpecify null-safety to the `aiagent` package (79 files changed). Requires #7696 to be merged first.

- Adds `package-info.java` with `@NullMarked` to all `aiagent` subpackages.
- Fixes NullAway violations surfaced by enabling the check: explicit `@Nullable` annotations on fields and return types that may be absent, null-checks and `Objects.requireNonNull` guards where warranted, `@NullUnmarked` used sparingly as a named deferral with an explanatory comment.
- Deletes `aiagent/.gitignore` (the placeholder from #7696 that suppressed untracked `package-info.java` files in this package).
- Includes corresponding unit test and e2e test updates.

After this PR merges, NullAway enforces non-null-by-default for the entire `aiagent` package at compile time.

## Merge order

Merge #7696 first. This PR and the other three feature PRs (#7698, #7699, #7700) can be merged in any order after that.

PR #7698, opened 2026-06-29
feat(agentic-ai): apply @NullMarked to mcp package

nikonovd:
## Summary

Applies JSpecify null-safety to the `mcp` package. Requires #7696 to be merged first.

- Adds `package-info.java` with `@NullMarked` to all `mcp` subpackages.
- Fixes NullAway violations: `@Nullable` on fields/return types that may be absent, null guards at appropriate call sites.
- Deletes `mcp/.gitignore` (the placeholder from #7696).

After this PR merges, NullAway enforces non-null-by-default for the entire `mcp` package at compile time.

## Merge order

Merge #7696 first. This PR and the other three feature PRs (#7697, #7699, #7700) can be merged in any order after that.

### DIFF
--- a/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/McpRemoteClientRegistry.java
+++ b/connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/McpRemoteClientRegistry.java
@@ -28,7 +28,6 @@ import io.camunda.connector.agenticai.mcp.client.model.auth.NoAuthentication;
 import io.camunda.connector.agenticai.mcp.client.model.auth.OAuthAuthentication;
 import io.camunda.connector.api.outbound.OutboundConnectorContext;
 import org.jspecify.annotations.NonNull;
-import org.jspecify.annotations.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -55,7 +54,9 @@ public class McpRemoteClientRegistry implements AutoCloseable {
             (key, value, cause) -> {
               LOGGER.info(
                   "MCP({}): Removing cached remote HTTP client (removal cause: {})", key, cause);
-              this.closeClient(key, value);
+              if (key != null && value != null) {
+                this.closeClient(key, value);
+              }
             })
         .build();
   }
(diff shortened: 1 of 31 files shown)


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/framework/bootstrap/auth/OAuthHeadersSupplier.java
```diff
@@ -64,14 +65,16 @@ public OAuthHeadersSupplier(
 
   @Override
   public synchronized Map<String, String> get() {
-    if (tokenResponse == null || tokenResponse.isExpired(clock)) {
+    var current = tokenResponse;
+    if (current == null || current.isExpired(clock)) {
```
maff: 💭 Why do we need such constructs when introducing the nullability constraints? Is the compiler smart enough to figure out that current is non-null, but can't do so for `tokenResponse`?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/McpRemoteClientRegistry.java
```diff
@@ -153,7 +154,7 @@ public void close() {
     this.cache.asMap().forEach(this::closeClient);
   }
 
-  public void closeClient(Object clientId, Object client) {
+  public void closeClient(@Nullable Object clientId, @Nullable Object client) {
```
maff: 🔧 Should we make these non-nullable? Internal usage should definitely be non-nullable & we could adapt the caller accordingly.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/mcp/client/framework/bootstrap/auth/OAuthHeadersSupplier.java
nikonovd: In this case we don't 👍 removed it. 

AI insisted that it's relevant (as the variable can be nullified at anytime by another thread. However i couldn't find any evidence with manual research on that topic and i'm going to revisit the rest of the code.

maff (approved):

PR #7702, opened 2026-06-29
chore(connector-runtime): add jspecify/NullAway null-safety checking

sbuettner:
## Description

Introduces compile-time null-safety analysis for the `connector-runtime` modules using **NullAway + ErrorProne**, and annotates the genuinely-nullable APIs in `connector-runtime-core` with `org.jspecify.annotations.Nullable`.

### What changed

- **Tooling.** Wire NullAway (`WARN`) + ErrorProne into the `connector-runtime` compiler plugin (`AnnotatedPackages=io.camunda.connector`, test sources excluded). Versions for error-prone/nullaway and the `maven-compiler-plugin` are managed in `parent/pom.xml`, plus the javac `--add-exports`/`--add-opens` ErrorProne needs.
- **Dependency hygiene.** jspecify is declared only in the modules that actually use it (`connector-runtime-core`, `spring-boot-starter-camunda-connectors`) rather than the runtime parent, and added to the dependency-analyzer ignore list (it's a CLASS-retention annotation, invisible in bytecode). Removed the now-unused `jakarta.annotation-api` from `connector-runtime-core` after migrating its remaining `@Nullable` usages to jspecify.
- **Annotations.** Added `@Nullable` to genuinely-nullable fields, record components, parameters and returns across `connector-runtime-core`, plus three small local refactors where NullAway couldn't track non-nullness (`ConnectorResultHandler` guard, `InboundPropertyHandler` branch split, `ActivationConditionEvaluator` `requireNonNull`).

### Impact

- NullAway warnings in `connector-runtime-core`: **55 → 21**, with no new warnings introduced.
- `NullAway:WARN` (not `ERROR`) — informational for now, does not fail the build.
- Remaining warnings are not fixable within `connector-runtime-core`: they'd require annotating `connector-api` / `connector-feel`, which is intentionally out of scope here.

## Verification

- `mvn clean compile -pl connector-runtime/connector-runtime-core -am` builds clean (21 NullAway warnings, down from 55).
- `mvn install -pl connector-runtime/connector-runtime-core -am` passes, including the dependency-analysis goal.
- `ConnectorResultHandlerTest` and related unit tests pass.

### DIFF
--- a/.mvn/jvm.config
+++ b/.mvn/jvm.config
@@ -1,8 +1,13 @@
 --add-exports jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
 --add-exports jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED
 --add-exports jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED
+--add-exports jdk.compiler/com.sun.tools.javac.processing=ALL-UNNAMED
 --add-exports jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED
 --add-exports jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED
+--add-opens jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED
+--add-opens jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED
 --add-opens=java.base/java.util=ALL-UNNAMED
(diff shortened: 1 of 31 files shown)


maff:
@nikonovd FYI with this merged, we should be able to remove the dependency management & potentially the JVM flags from the build config.

sbuettner:
I found an [issue](https://github.com/jspecify/jspecify/issues/365) which requires us to use the `jakarta.annotation.Nullable` annotation for intrinsic functions.

mathias-vandaele (approved):
Thank you @sbuettner, solid PR. 💪

A little nitpeak, the PR description state: 

"Removed the now-unused jakarta.annotation-api from connector-runtime-core"

This is not the case, it was kept with a comment: `<!-- runtime-retained @Nullable read reflectively by IntrinsicFunctionParameterBinder -->`

PR #7722, opened 2026-06-30
ci(e2e): notify Slack on build failures and route pings to the owning team via CODEOWNERS

nikonovd:
## Description

Improves the nightly E2E Slack notifications (`E2E_BRANCH_RUN.yml`) the following way:

**Failures ping the owning team, not just the medic.** Previously every failure
pinged a single hardcoded medic group. The message now also pings the team that owns the
failing module, resolved from `CODEOWNERS` (last-match-wins, exactly like PR review
assignment). The medic is always included as a catch-all, so unmapped owners and
build-phase failures (no module) are never silent.

The CODEOWNERS→Slack derivation is packaged as a reusable, dependency-free composite
action — `.github/actions/codeowners-slack-mentions` — that takes paths + a team→ID map
and returns deduped Slack mentions. 
The glob matcher is a faithful port of GitHub's [CODEOWNERS](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file) semantics, and its tests are derived
from GitHub's official reference example (default `*`, `*.js` precedence, email/team
owners, `/build/logs/` anchoring, `docs/*` one-level-only, `apps/` anywhere, `**/logs`,
multi-owner, owner-less removal, literal character ranges).

Team Slack user-group IDs live in a committed map (`.github/codeowners-slack-map.json`),
not secrets, matching the `connectors-streak-detector.yml` convention. Empty/absent
values and the default owner fall back to the always-pinged medic, so routing is never
silent and never emits a broken mention; filling in a team's ID activates direct routing
with no code change.

### Changes
- `E2E_BRANCH_RUN.yml`: resolves owners and pings the owning team + medic.
- New reusable action `.github/actions/codeowners-slack-mentions` (`action.yml`,
  `index.mjs`, `index.test.mjs`, `README.md`).
- New `.github/codeowners-slack-map.json` (team handle → bare Slack group ID).
- `SLACK_MENTIONS_SETUP.md`: documents the map and routing/fallback behaviour.
- Parser fix: strip inline `#` comments in CODEOWNERS (surfaced by the GitHub doc example).

## Checklist

### DIFF
--- a/.github/actions/codeowners-slack-mentions/action.yml
+++ b/.github/actions/codeowners-slack-mentions/action.yml
@@ -32,6 +32,10 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Set up Node
+      uses: actions/setup-node@v6
+      with:
+        node-version: '24'
     - name: Install dependencies
       shell: bash
       working-directory: ${{ github.action_path }}
--- a/.github/actions/codeowners-slack-mentions/action.yml
+++ b/.github/actions/codeowners-slack-mentions/action.yml
@@ -32,6 +32,10 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Install dependencies
+      shell: bash
+      working-directory: ${{ github.action_path }}
+      run: npm ci --omit=dev
     - name: Derive mentions
       id: derive
       shell: bash
--- a/.github/actions/codeowners-slack-mentions/action.yml
+++ b/.github/actions/codeowners-slack-mentions/action.yml
@@ -1,0 +1,43 @@
+name: Derive Slack mentions from CODEOWNERS
+description: >
+  Resolve the CODEOWNERS owners of the given repo paths and map them to Slack
+  mentions, with optional always-include mentions, deduped. Pure derivation —
+  no GitHub API calls and no Slack posting. Generic: paths + map in, mentions out.
+
+inputs:
+  paths:
+    description: "Newline- or space-separated repo-relative paths to resolve owners for. Empty is allowed (then only always-include is returned)."
+    required: false
+    default: ""
+  codeowners-file:
+    description: "Path to the CODEOWNERS file."
+    required: false
+    default: CODEOWNERS
+  team-slack-map:
+    description: 'JSON mapping "@org/team" -> Slack mention (<!subteam^…> or <@…>), inserted verbatim. Either inline JSON or a path to a .json file; inline lets a caller assemble it from secrets without writing them to disk.'
+    required: true
+  always-include:
+    description: "Mention(s) always included regardless of owners (e.g. a medic group). Space-separated."
+    required: false
+    default: ""
+
+outputs:
+  mentions:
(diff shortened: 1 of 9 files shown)


.github/codeowners-slack-map.json
```diff
@@ -0,0 +1,6 @@
+{
+  "@camunda/connectors-core": "S05715X5N21",
```
johnBgood: why not use GitHub secrets? as it's done for the current medic handle IIRC

nikonovd: As discussed, there was no strong opposing against that. Rewrote it to use secrets (and configured them for IDP and agentic ai medics).

See also the test run in the support channel.

.github/actions/codeowners-slack-mentions/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Derive Slack mentions from CODEOWNERS
```
sbuettner, line 1: Nice work!

WDYT about making this more generic by just returning the codeowners? Should we give this a more generic name? Something like `get-codeowners`?

We could also re-use prior art like this: https://www.npmjs.com/package/codeowners (also happy to keep your implementation but maybe consider splitting or renaming the slack mapping part)

We could then have a separate action that does the mapping.

nikonovd: Definitely I thought about reusing existing libraries - this was rather a prototype to reduce dependencies in the supply chain.

@johnBgood WDYT about reusing existing packages here?

johnBgood: It would probably simplify the workflow and help with reading/maintaining it, so I'd agree

johnBgood (approved):

.github/actions/codeowners-slack-mentions/action.yml
nikonovd: Went with [`minimatch`](https://www.npmjs.com/package/minimatch) instead of the `codeowners` package — that one (and `codeowners-utils`) pulls in 4–9 deps for filesystem-walking/CLI machinery we don't need, since we already have the paths and CODEOWNERS location. `minimatch` adds exactly one dependency (`brace-expansion`, itself dependency-free).

I also tried a generic gitignore library (`ignore`, zero deps) first, but it recurses into matched directories even for bare wildcards like `docs/*` — which contradicts GitHub's documented CODEOWNERS behavior and broke that exact existing test. `minimatch` doesn't have that cascade, so our own (small, already-tested) anchoring rules stay in charge of that.

Pushed in 58b40cb — all existing tests in `index.test.mjs` pass unchanged, plus README now documents the "why minimatch" reasoning.

---

sbuettner: @nikonovd Nice and thanks for considering.

sbuettner (approved):
Nice!

PR #7735, opened 2026-07-01
fix(email): handle document upload failures gracefully to prevent infinite retry loop

### LINKED ISSUE #7733: bug(email): unhandled document upload exception in PollingManager causes infinite retry loop on large attachments
## Summary

When the Jakarta inbound email connector processes an email with an attachment that exceeds the SaaS document storage limit (10 MB), `connectorContext.create()` throws an exception that is not caught at the point of upload. This causes the email to never be marked as processed, and the single-threaded poller retries it on every cycle — blocking all subsequent emails from being correlated until the offending email is manually removed from the mailbox.

Confirmed via production logs (cluster DFB-INTG, `8.9.5`) and cross-referenced with the source code.

## Root cause

`PollingManager.createDocumentList()` calls `connectorContext.create()` for each attachment with no exception handling:

```java
// PollingManager.java:120-130
private List<Document> createDocumentList(Email email) {
  return email.body().attachments().stream()
      .map(
          document ->
              this.connectorContext.create(   // ← no try/catch
                  DocumentCreationRequest.from(document.inputStream())
                      .contentType(document.contentType())
                      .fileName(document.name())
                      .build()))
      .toList();
}
```

If `create()` throws (e.g. document too large, service unavailable), the exception propagates up through `correlate()` → `process()` → `processMail()` → `pollAllAndProcess()` / `pollUnseenAndProcess()` until it is caught by the top-level handler in `poll()` (line 177):

```java
} catch (Exception e) {
  // All exception are caught at highest level, ensuring the scheduler never stops
  this.connectorContext.log(...);
  this.connectorContext.reportHealth(Health.down());
}
```

This catch-all reports health as DOWN and logs the error, but **does not mark the email as processed**. Since `JakartaEmailListener` uses `newSingleThreadScheduledExecutor()`, the next poll picks up the same email, hits the same failure, and loops forever.

## Observed behaviour

- Connector gets stuck processing the same email on every poll cycle
- Health reported as DOWN (`RuntimeException: 1 of N connectors are down`)
- All emails arriving after the offending one are never processed — they effectively queue behind it
- No clear error surfaced to the process instance; the BPMN process simply stalls
- Workaround: manually delete or move the email from the monitored mailbox folder

## Expected behaviour

A document upload failure should:
1. Surface a clear error to the activity log (with the exception message and attachment name)
2. Either raise a BPMN error / incident on the correlated process instance, or skip the attachment and continue with the remaining ones
3. Mark the email as processed (or move/delete it per the configured handling strategy) so the poller is not blocked

## Affected code

- `connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java` — `createDocumentList()` (line 120) and `correlate()` (line 132)
- The same pattern exists in the outbound executor: `JakartaEmailActionExecutor.createDocumentList()` — document upload has no exception handling there either

## Steps to reproduce

1. Configure a Jakarta inbound email connector (IMAP) with attachment handling enabled
2. Send an email with an attachment > 10 MB to the monitored mailbox on a SaaS cluster
3. Observe connector health going DOWN
4. Send a second email (small, no attachment) — observe it is never processed until the large-attachment email is removed

## Environment

- Connector version: `8.9.5`
- Deployment: Camunda SaaS (document storage limit: 10 MB)
- Also affects self-managed if the document store rejects the upload for any reason

mathias-vandaele:
## Summary

- Inbound (`PollingManager`): wrap each `connectorContext.create()` call in try/catch; on failure log an `ERROR` to the activity log (visible in Operate/Connectors UI) and skip the attachment. Correlation still proceeds with any successfully uploaded attachments, and the email is marked as processed — eliminating the infinite retry loop that blocks all subsequent emails.
- Outbound (`JakartaEmailActionExecutor`): rethrow with a message including the attachment name so Zeebe creates an incident with actionable context instead of a generic error.

Closes #7733

## Root cause (short)

`createDocumentList()` in both classes called `connectorContext.create()` with no exception handling. In the inbound connector, an upload failure (e.g. attachment > 10 MB SaaS limit) propagated to the top-level `poll()` catch-all which logged the error and reported health DOWN — but **did not mark the email as processed**. The single-threaded `ScheduledExecutorService` then retried the same email on every poll cycle, permanently blocking all subsequent emails.

## Test plan

- [ ] Existing unit + integration tests pass (`mvnw -pl connectors/email test`) — 60 tests, all green
- [ ] Manual: send an email with attachment > 10 MB to a SaaS cluster monitored mailbox, confirm the connector logs a `document-upload` ERROR per attachment and continues correlating subsequent emails normally
- [ ] Manual: confirm the outbound read-email action produces a clear incident message in Operate when an attachment upload fails

### DIFF
--- a/connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
+++ b/connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
@@ -17,6 +17,7 @@ import io.camunda.connector.email.inbound.model.*;
 import io.camunda.connector.email.response.ReadEmailResponse;
 import jakarta.mail.*;
 import jakarta.mail.search.FlagTerm;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
@@ -50,7 +51,10 @@ public class PollingManager {
     this.store = store;
   }
 
-  private static ReadEmailResponse createResponse(Email email, List<Document> documents) {
+  private record DocumentUploadResult(List<Document> documents, List<String> errors) {}
+
+  private static ReadEmailResponse createResponse(
+      Email email, List<Document> documents, List<String> errors) {
     return new ReadEmailResponse(
         email.messageId(),
         email.from(),
@@ -60,7 +64,8 @@ public class PollingManager {
         email.body().bodyAsPlainText(),
         email.body().bodyAsHtml(),
         documents,
-        email.receivedAt());
+        email.receivedAt(),
+        errors);
   }
 
   public static PollingManager create(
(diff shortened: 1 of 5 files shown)


connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
```diff
@@ -120,12 +120,27 @@ public static PollingManager create(
...
-            document ->
-                this.connectorContext.create(
-                    DocumentCreationRequest.from(document.inputStream())
-                        .contentType(document.contentType())
-                        .fileName(document.name())
-                        .build()))
+            attachment -> {
+              try {
+                return this.connectorContext.create(
+                    DocumentCreationRequest.from(attachment.inputStream())
```
sbuettner: Shouldnt this behaviour be configurable? Here we are logging an error but the mails will be marked as processed.

mathias-vandaele: We have to behave this way, otherwise, emails will be processed in a loop, and we can't completely drop an email because its attachment was too big. 

We might want to add an error to the payload, so that user can handle it in the process diagram

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
```diff
@@ -117,23 +122,37 @@ public static PollingManager create(
...
+                    .fileName(attachment.name())
+                    .build()));
+      } catch (Exception e) {
+        String errorMessage =
+            "Failed to upload attachment '%s': %s".formatted(attachment.name(), e.getMessage());
+        errors.add(errorMessage);
+        this.connectorContext.log(
+            activity ->
+                activity
+                    .withSeverity(Severity.ERROR)
```
johnBgood: should be a warning as we correlated the message anyway

johnBgood (approved):
Left one minor comment, nice improvement

connectors/email/src/main/java/io/camunda/connector/email/client/jakarta/inbound/PollingManager.java
mathias-vandaele: Correct

Issue #7470, opened 2026-06-11
Replace descriptions in element templates with tooltips


chillleader:
## Background

Follow up on an experiment showing that moving property descriptions in connector element templates to tooltips significantly improves the visual clarity and usability of the templates. The experiment was shared in [#connector-sdk-development](https://camunda.slack.com/archives/C02JLRNQQ05/p1781107263423369) and received positive feedback.

- **Experiment preview**: https://jubilant-bassoon-2qgwnjw.pages.github.io/p/copilot-element-templates-again/
- **Experiment source (modeler-prototypes)**: https://github.com/camunda/modeler-prototypes/pull/8

## Goal

Replace `description` fields in all connector element templates with `tooltip` fields so that descriptive text is shown on hover rather than always-visible below the input.

## What needs to change

1. **Element template generator** (`element-template-generator`) — update the generator to emit `tooltip` instead of `description` for property-level help text (Nico's note: _"needs obviously adjustment in the template generator on your end — but seems to be rather straight forward"_).
2. **Versioned/generated templates** — regenerate all connector templates after the generator change so the produced JSON uses `tooltip`.
3. **Hand-authored templates** — audit any templates not produced by the generator and migrate `description` → `tooltip` there as well.

## Acceptance criteria

- [ ] Element template generator emits `tooltip` instead of `description` for property help text
- [ ] All generated connector templates use `tooltip` (no leftover `description` on properties)
- [ ] Hand-authored templates migrated
- [ ] Visually verified in Modeler that tooltips appear correctly on hover

crobbins215:
@barinali will pick up.

I will close down. https://github.com/camunda/connectors/issues/2328

crobbins215:
### PR map for this issue

The original umbrella PR was split into review-sized, vendor-family slices (per review feedback), plus a dedicated generator PR.

**Generator (issue item 1)**
- #7753 — `feat(element-template-generator): emit property description as tooltip`. Auto-maps property `description=` → `tooltip` at a single chokepoint; explicit `tooltip=` wins; template-level description unchanged. Generator + tests only (196 core tests green); connector regeneration is a follow-up.

**Migration slices (issue items 2 & 3 — generated + hand-authored templates)**
| Slice | PR |
|-------|----|
| AWS | #7651 |
| Google | #7742 |
| Microsoft | #7743 |
| HTTP (REST / GraphQL / polling) | #7744 |
| Messaging (Kafka / RabbitMQ / Slack) | #7745 |
[...]

PR #7753, opened 2026-07-01
feat(element-template-generator): emit property description as tooltip

crobbins215:
Part of #7470 — the **generator** follow-up (issue item 1), addressing @ztefanie's review point 6 ("do the generator first / do it in the generator so all fields are consistent").

## What this does
The generator now emits property-level help text as a **`tooltip`** (hover) instead of an always-visible **`description`**. The change is at a single chokepoint — [`PropertyBuilder.description()`](element-template-generator/core/src/main/java/io/camunda/connector/generator/dsl/PropertyBuilder.java) — so **every** generation path is covered:
- the Java `@TemplateProperty` annotation processor (all annotated connectors),
- the HTTP DSL (`FactoryUtils` / `PropertyUtil`),
- hard-coded DSL property builders.

### Rules
- An explicit `tooltip=` **always wins** over `description=`.
- Connector/template-level description (the `ElementTemplate` description) is **unchanged** — only property help text moves.
- To keep an always-visible `description` for a specific property, edit the generated template manually (per the agreed policy: auto-map by default, manual opt-out).

## Why this approach
Connector authors keep writing `description=` as they always have and get tooltip-first templates for free on regeneration — no per-field annotation churn. This is the mechanical/consistent counterpart to the hand-curated wording in the split migration PRs (#7651 + siblings).

## Scope / follow-up
- This PR is **generator + tests only**. Regenerating all connector templates so the committed JSON reflects the new output is deliberately left as a follow-up to keep this reviewable.
- Tests updated to assert the new behavior (`OutboundClassBasedTemplateGeneratorTest`, `ElementTemplateSerializationTest` + resource). `element-template-generator/core` suite: **196 passing**.

## Note on the annotation
`@TemplateProperty#description()` is kept (Javadoc updated to document the mapping) rather than removed, to avoid breaking every existing annotation. It now behaves as a tooltip source.

### DIFF
--- a/element-template-generator/annotations/src/main/java/io/camunda/connector/generator/java/annotation/TemplateProperty.java
+++ b/element-template-generator/annotations/src/main/java/io/camunda/connector/generator/java/annotation/TemplateProperty.java
@@ -48,7 +48,15 @@ public @interface TemplateProperty {
   /** Custom binding name of the property */
   PropertyBinding binding() default @PropertyBinding(name = "");
 
-  /** Description of the property */
+  /**
+   * Help text for the property.
+   *
+   * <p>As of <a href="https://github.com/camunda/connectors/issues/7470">#7470</a> the generator
+   * emits this value as a {@link #tooltip()} (shown on hover) rather than an always-visible
+   * description. Set {@link #tooltip()} explicitly to override; an explicit tooltip always wins. To
+   * keep an always-visible description for a specific property, edit the generated template
+   * manually.
+   */
   String description() default "";
 
   /** Whether the property should be marked as optional in the element template */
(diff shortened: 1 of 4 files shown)


crobbins215:
Closing — decision is **not** to auto-map in the generator.

Auto-mapping `description=` → `tooltip` would remove the ability to author an always-visible `description` for generated connectors (value-constraints, `Note:`/`Warning:`, genuinely critical guidance). That conflicts with the agreed priority — `tooltip` > `placeholder` > `description`, deliberately **keeping a curated set of descriptions** (e.g. the google-maps `Units` hint restored in #7748, and the REST auth/validation guidance).

Instead the migration stays **per-field** in the family PRs (#7651 + siblings), where each hint is placed intentionally and always-visible descriptions are preserved where they help. Generator behavior is left unchanged.

PR #7755, opened 2026-07-01
feat(runtime): add structured metrics API to /outbound and /inbound controllers

johnBgood:
## Description

Introduces a structured metrics query API on the outbound and inbound connector controllers. The API returns pre-aggregated, strongly-typed objects rather than raw metric series, making it directly consumable by dashboards and monitoring tools.

### Endpoints

```
GET /outbound/metrics                    → aggregated totals across all connector types
GET /outbound/metrics/{connectorType}   → metrics scoped to a specific connector type

GET /inbound/metrics
GET /inbound/metrics/{connectorType}
```

### Outbound response shape

```json
{
  "runtime": {
    "uptimeSeconds": 7320
  },
  "job": {
    "completed": 42,
    "failed": 1,
    "bpmnError": 0,
    "executionTime": {
      "meanMs": 123.4,
      "maxMs": 890.0
    },
    "lastCompleted": "2026-07-02T10:00:00Z",
    "lastFailed": "2026-07-01T08:30:00Z"
  },
  "worker": {
    "jobsActivated": 43,
    "jobsHandled": 43,
    "streamRecreations": 0
  }
}
```

> `job.executionTime`, `job.lastCompleted`, `job.lastFailed` are omitted if no data is available yet.

### Inbound response shape

```json
{
  "runtime": {
    "uptimeSeconds": 7320
  },
  "activation": {
    "activated": 5,
    "deactivated": 1,
    "activationFailed": 0,
    "lastActivated": "2026-07-02T10:00:00Z"
  },
  "trigger": {
    "triggered": 100,
    "correlated": 98,
    "correlationFailed": 2,
    "activationConditionFailed": 3,
    "lastTriggered": "2026-07-02T09:55:00Z"
  }
}
```

> `activation.lastActivated` and `trigger.lastTriggered` are omitted if no activation/trigger has occurred yet.

### Key design decisions

- **Single object response** (not an array): when no `{connectorType}` segment is provided, metrics are aggregated across all types (sums for counters, weighted mean + overall max for execution time, most-recent timestamp for last activity).
- **No `connectorType` in the response**: the caller already knows the type they requested; including it would be redundant.
- **`runtime.uptimeSeconds`** sourced from `process.uptime` (auto-registered by Spring Boot Actuator), present in all responses.
- **Timestamps nested inside their domain group**: `job.lastCompleted`/`job.lastFailed` sit inside the `job` sub-object; `activation.lastActivated`/`trigger.lastTriggered` sit inside their respective sub-objects. They are also exposed as Micrometer `Gauge`s so they appear in Prometheus scrapes.
- **Grouped response structure**: both `OutboundConnectorMetrics` and `InboundConnectorMetrics` are modelled as three nested records (`runtime` and two domain-specific groups) rather than flat records, improving readability and keeping related fields co-located.

### Bug fix: `bpmn-error` action tag

The `ACTION_BPMN_ERROR` constant was `"bpmnError"` but the Camunda client's `JobHandlerMetrics.Action.BPMN_ERROR.asString()` returns `"bpmn-error"` (with hyphen). This caused `job.bpmnError` to always read `0`. Fixed by aligning the constant.

### Bug fix: `activationConditionFailed` always zero

`MeteredInboundCorrelationHandler` previously tried to intercept `isActivationConditionMet()`, but that method is never called in the actual correlation flow — making the counter permanently zero. The fix went through several iterations:
1. Override `canActivate()` to detect `NoMatchingElement` results.
2. Move the increment into `correlate()` (instead of `canActivate()`) to avoid double-counting when callers invoke `canActivate()` directly.
3. Switch to inspecting `CorrelationResult` type in `correlate()` to also catch activation-condition failures that don't propagate as exceptions.

The unused `isActivationConditionMet()` override was removed from the base class as part of this cleanup.

### New component: `ConnectorOutboundMetrics`

Replaces the separate `MetricsRecorder` + `OutboundConnectorTimestampTracker` fields in `SpringConnectorJobHandler` with a single `ConnectorOutboundMetrics` object — mirroring the existing `ConnectorsInboundMetrics` pattern on the inbound side. It delegates counter/timer recording to the underlying `MetricsRecorder` and owns the timestamp gauges via `MeterRegistry`.

### `MeterRegistry` is optional

`MeterRegistry` (provided by Spring Boot Actuator) is now an optional dependency throughout the runtime. Controllers and the outbound connector manager use `Optional<MeterRegistry>` injection so the runtime starts cleanly without Actuator on the classpath. Metrics endpoints return empty results and log a one-time `WARN` when no registry is available.

## Related issues

closes #

## Checklist


sbuettner:
Overall very solid and straightforward.

Some food for thought:

@johnBgood What do you think of having a more concrete API response schema similar to what we offer for https://docs.camunda.io/docs/apis-tools/orchestration-cluster-api-rest/specifications/get-global-job-statistics/ without leaking the micrometer registry internals to clients by having a concrete schema when returning global data (/metrics) and Connector specific data (/metrics/$connector)?

We dont need the same, complex filtering structure but could think of whether we want to return the data in a similar way in case of a high number of connectors by wrapping it into a page for example?

We could define proper keys in the schema without relying too much on how the metrics are named internally.

sbuettner:
@johnBgood nice work!

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/exception/InvalidTagFormatException.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.camunda.connector.runtime.inbound.controller.exception;
+
+public class InvalidTagFormatException extends RuntimeException {
```
sbuettner: Still required?

sbuettner (approved):

PR #7779, opened 2026-07-02
fix(agentic-ai): treat 404 on agent-instance writes as non-retryable

### LINKED ISSUE #7598: AI Agent: Verify whether eventual-consistency retry on agent instance update is still needed
## Problem

`AgentInstanceErrorClassifier` retries 404 responses on `newUpdateAgentInstanceCommand` under the assumption that the just-created agent instance may not yet be visible (eventual consistency window). However this assumption is unverified against the documented API contract.

From the PR review (@maff on `AgentInstanceErrorClassifier.java:23`):

> Are we 100% sure about the eventual consistency part as this would deviate from other concepts on the engine? The API docs list the create agent instance endpoint as strongly consistent: https://docs.camunda.io/docs/next/apis-tools/orchestration-cluster-api-rest/specifications/create-agent-instance/

Current status (@nikonovd):

> Well, I misread the `x-eventually-consistent` tag there, but in CPT I observed repeated errors after the creation of the agent instance, that it was not existing, which disappeared on retry. I'll check again, but generally it's fairly acceptable to have that in place for now.

## Goal

Confirm with the engine/API team whether the `x-eventually-consistent` tag applies to the agent instance update endpoint, or whether the observed 404-after-create in CPT was a transient environment issue. If the endpoint is strongly consistent, remove the 404 retry from `AgentInstanceErrorClassifier` to avoid silently masking real 404 errors.

## Parent

#7595

nikonovd:
## Description

The create/update/history endpoints for agent instances are all `x-eventually-consistent: false` and validate against primary processing state (not an ES/OS projection), and Zeebe key-based partition routing guarantees a created record is visible before its key is returned to the caller. A 404 from these endpoints therefore reflects a genuinely missing instance, not an eventual-consistency lag, so retrying it only delays surfacing a real error.

Collapses AgentInstanceErrorClassifier into a single stateless singleton (404 now falls under the existing generic 4xx-is-permanent branch), removing the notFoundIsRetryable toggle and the three now-identical `FOR_*` variants.
## Related issues

closes #7598 

## Checklist

### DIFF
--- a/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
+++ b/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
@@ -10,32 +10,19 @@ import io.camunda.client.api.command.ClientHttpException;
 import io.camunda.connector.agenticai.common.util.retry.ErrorClassifier;
 import java.io.IOException;
 
-public final class AgentInstanceErrorClassifier implements ErrorClassifier {
-
-  /** For create: 404 is retryable (idempotent re-create against a transient lookup race). */
-  public static final AgentInstanceErrorClassifier FOR_CREATE =
-      new AgentInstanceErrorClassifier(true);
-
-  /**
-   * For update: 404 is retryable. A just-created agent instance may not yet be visible to follow-up
-   * API calls due to eventual consistency, so a transient 404 should be retried rather than failing
-   * the job.
-   */
-  public static final AgentInstanceErrorClassifier FOR_UPDATE =
-      new AgentInstanceErrorClassifier(true);
-
-  /**
-   * For history item creation: 404 is retryable for the same eventual-consistency reason as {@link
-   * #FOR_UPDATE}.
-   */
-  public static final AgentInstanceErrorClassifier FOR_HISTORY_ITEM =
-      new AgentInstanceErrorClassifier(true);
-
-  private final boolean notFoundIsRetryable;
-
-  private AgentInstanceErrorClassifier(boolean notFoundIsRetryable) {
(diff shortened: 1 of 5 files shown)


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
```diff
@@ -10,42 +10,25 @@
...
     Throwable current = t;
     while (current != null) {
       if (current instanceof ClientHttpException httpEx) {
         int status = httpEx.code();
-        if (status == 404) {
-          return notFoundIsRetryable ? Decision.RETRYABLE : Decision.PERMANENT;
-        }
         if (status >= 400 && status < 500) {
           return Decision.PERMANENT;
         }
```
maff: 💭 Do we have any need to retry temporary error codes like a 429, similar as we do [when handling jobs](https://github.com/camunda/camunda/blob/main/clients/camunda-spring-boot-starter/src/main/java/io/camunda/client/jobhandling/JobCallbackCommandWrapper.java#L41)?

maff (approved):

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceErrorClassifier.java
nikonovd: Good catch, and you're right — the `JobCallbackCommandWrapper` precedent explicitly treats 429 as retryable (along with 502/503/504), separate from the 404 "ignorable" bucket. Our classifier was lumping 429 into the generic "4xx is permanent" rule, which is wrong: rate limiting is transient, unlike 400/401/403/404/409. Added a `429 → RETRYABLE` case (grouped with 5xx) and covered it in `AgentInstanceErrorClassifierTest`.

---

Issue #7211, opened 2026-05-13
Agentic AI: native LLM provider integrations (v2 connector)


maff:
## Value proposition

Give the AI Agent first-class, native access to each LLM provider through the provider's own SDK and wire format, exposed via a clean v2 connector. This unlocks capabilities the common-denominator LangChain4j abstraction cannot reach — extended reasoning/thinking, prompt caching, native multimodal, and lossless provider-specific routing across the hyperscalers (AWS, Azure, Google) — configured through a redesigned provider/connection/model UX. Providers are delivered one at a time, platform by platform, each shipping with reasoning and prompt caching from the start.

LangChain4j remains the supported **v1** implementation and the officially-supported way to register custom providers; it is not being removed. Native v2 providers are added alongside v1. Whether and where existing v1 implementations later adopt the native SDKs is evaluated separately.

## Design decisions

- Provider config is a three-layer model: **presentation preset** (the Modeler `Provider` dropdown entry, e.g. *Anthropic*, *OpenAI*, *Mistral AI*) → **connection/backend** (endpoint + auth, e.g. *Anthropic API*, *Amazon Bedrock*, *Microsoft Foundry*) → **wire mapper** (the shared internal converter, e.g. *Anthropic Messages*, *OpenAI Responses / Chat Completions*, *AWS Converse*). The dropdown entry is decoupled from the mapper: for example a *Mistral AI* preset can reuse the *OpenAI Chat Completions* mapper with no Mistral-specific wire code.
- `ProviderConfiguration` is a sealed union keyed by preset `type`; each provider carries a backend sub-union and, where the provider exposes more than one API, an `api` family selector (e.g. OpenAI *Responses* vs *Chat Completions*).
- **HTTP proxy support is mandatory on every provider/backend.**

## Scope

Tracked as linked sub-issues:

- **Foundation:** ChatModel provider SPI · `ProviderConfiguration` sealed type · v2 Task/Sub-process connector types + element templates · custom model provider.
- **Providers** (each a provider issue with per-backend sub-issues; `direct` + `compatible` first, then per hyperscaler/platform): Anthropic · OpenAI · Google Gemini · Amazon Bedrock (Converse API) · Mistral AI.
- **Cross-cutting** (delivered per provider): prompt caching · reasoning configuration.
- **Migration & lifecycle:** deprecate old connector types · migration tooling · (optional) evaluate moving v1 implementations to the native SDKs.

PR #7786, opened 2026-07-02
test(agentic-ai): provider wire-format e2e coverage (Anthropic, Bedrock, Azure OpenAI)

### LINKED ISSUE #7785: AI Agent: Extend WireMock wire-format e2e coverage to Anthropic, Bedrock, and Azure OpenAI
## What should we do?

Extend the WireMock wire-format e2e approach from #7400 (currently only `openaiCompatible`) to Anthropic (Messages API), AWS Bedrock (Converse API), and Azure OpenAI (Chat Completions) — as a regression baseline for #7211's native-provider rewrite. Wire-format drift during that rewrite should be caught, not silent.

One parameterized JUnit test class, one row per provider/API, 4 scenarios each: single-turn text, tool calls, document in user prompt, JSON response schema.

**In scope:** `anthropic`, `bedrock`, `azureOpenAi`, plus `openaiCompatible` folded into the same structure.

**Out of scope:** `openai` (direct) — identical wire format to `openaiCompatible`. `google-vertex-ai` — LangChain4j's client hardcodes gRPC with no proxy/endpoint-override hook; revisit if a REST-based SDK is adopted. Only job-worker flavor (shared orchestration core with outbound-connector makes duplicating this redundant).

## Why?

Continues the WireMock migration from #7400. Same "verified, not assumed" wire-level coverage, extended to all production LLM providers ahead of #7211.

Sub-issue of #7211.

maff:
Closes #7785.

Extends the WireMock wire-format e2e approach from #7400 (previously `openaiCompatible`-only) to Anthropic, Bedrock, and Azure OpenAI, as a regression baseline for the native-provider rewrite in #7211.

One parameterized test class (`ProviderWireFormatSmokeTests`), one row per provider behind a shared `ProviderWireFormatFixture` SPI, 4 scenarios per row: single-turn text, tool calls, document in user prompt, JSON response schema.

Real findings surfaced while building this (all confirmed against source and/or live APIs, not assumptions):
- Anthropic's JSON-schema output field drops the schema name; Bedrock keeps it but sends the schema as a string, with a lowercase wire enum (`json_schema`) that doesn't match the SDK's constant name.
- Azure's SDK hard-requires HTTPS for API-key auth, and `langchain4j-azure-openai` doesn't support PDF documents in user prompts (only images).

Anthropic, Bedrock, and Azure rows were all calibrated against real APIs.


connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/azureopenai/AzureOpenAiCompletionsChatModelStubs.java
```diff
@@ -0,0 +1,187 @@
...
+          usage);
+    }
+
+    private record ChatCompletionResponse(
+        String id,
+        String object,
+        long created,
+        String model,
+        List<Choice> choices,
+        UsageInfo usage) {}
```
nikonovd: 💭 Should we check to extract common OpenAI artifacts into a common class, as we mostly differentiate by wire type, not API category, currently?

nikonovd (approved):
LGTM! One thought for food and the stuff we discussed per chat.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/azureopenai/AzureOpenAiCompletionsChatModelStubs.java
maff: Good point, thanks! I change the Azure OpenAI stubs to use the OpenAI ones directly and merged both into the same package.

PR #7815, opened 2026-07-06
fix(slack-inbound): fix HMAC failure for form-urlencoded webhooks (Slack Interactivity)

### LINKED ISSUE #7814: fix(slack-inbound): HMAC verification fails for Slack Interactivity (block_actions) payloads
## Summary

The Slack inbound connector fails HMAC signature verification for **Interactivity** payloads (Block Kit button clicks), returning HTTP 500 with `HMAC signature did not match`.

Events API (JSON body) and Slash Commands are not affected.

## Root cause

`InboundWebhookRestController.inbound()` declared `@RequestParam(required = false) Map<String, String> params` as a method parameter. Spring MVC resolves `@RequestParam` **before** the method body executes by calling `request.getParameterMap()`. For `application/x-www-form-urlencoded` POST requests, Tomcat parses and consumes the request body stream at that point.

When the controller then calls `httpServletRequest.getInputStream().readAllBytes()` to capture `rawBody`, it receives an empty byte array. HMAC is computed on `""` instead of the actual body, causing a guaranteed mismatch with Slack's signature.

- **Events API** (Content-Type: `application/json`): Spring's `@RequestParam` resolution never touches JSON bodies → body intact → HMAC passes ✓  
- **Interactivity / Slash Commands** (Content-Type: `application/x-www-form-urlencoded`): body consumed by `getParameterMap()` → `rawBody` = `[]` → HMAC fails ✗

Unit tests were not catching this because they mock `rawBody()` directly and call the controller method bypassing Spring's parameter resolution.

## Secondary bug

In `SlackInboundWebhookExecutable.bodyAsMap`, form values are split with `param.split("=")` (no limit). Once the body is URL-decoded, JSON values containing `=` characters (e.g. base64 data) get truncated: only the part before the second `=` is kept.

## Fix

- Remove `@RequestParam` from `InboundWebhookRestController.inbound()` and read the raw body first.
- Extract query params from `httpServletRequest.getQueryString()` only (not form body), preserving query-string access for FEEL expressions in other connectors.
- Change `param.split("=")` → `param.split("=", 2)` in `SlackInboundWebhookExecutable.bodyAsMap`.

## Reproduction

Configure a Slack app with Interactivity enabled, point the Request URL to the connector webhook endpoint, post a message with Block Kit buttons, and click one. The connector returns HTTP 500.

Events API `app_mention` events work fine on the same connector instance with the same signing secret.

mathias-vandaele:
Closes #7814

## Summary

- **Primary fix**: Remove `@RequestParam Map<String, String> params` from `InboundWebhookRestController.inbound()`. Spring was calling `getParameterMap()` to resolve this parameter before the method body ran, which consumed the servlet input stream for `application/x-www-form-urlencoded` requests. `getInputStream().readAllBytes()` then returned `[]`, so HMAC was computed on `""` instead of the actual body — a guaranteed failure for Slack Interactivity (block_actions) payloads.
- Query params are now extracted from `httpServletRequest.getQueryString()` only, so FEEL expressions referencing `params` in other connectors continue to work.
- **Secondary fix**: `param.split("=")` → `param.split("=", 2)` in `SlackInboundWebhookExecutable.bodyAsMap` to prevent truncation of form values whose decoded content contains `=` (e.g. base64 data).

## Why Events API was unaffected

`application/json` bodies are never parsed by Spring's `@RequestParam` resolution, so the input stream remained intact for JSON requests.

## Test plan

- [ ] All existing unit and integration tests pass (updated to match new 3-arg controller signature)
- [ ] Manually verify: Slack Interactivity button click reaches the connector and HMAC validates correctly
- [ ] Manually verify: Slack Events API (`app_mention`) still works
- [ ] Manually verify: Slash Commands still work
- [ ] Manually verify: webhooks using query-string `params` in FEEL expressions still receive correct values

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestControllerTest.java
+++ b/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestControllerTest.java
@@ -65,14 +65,11 @@ class InboundWebhookRestControllerTest {
     request.setServerName("example.com");
     request.setServerPort(443);
     request.setRequestURI("/inbound/myPath");
+    request.setQueryString("token=secret-token&q=visible");
     request.setMethod("POST");
     request.setContent("a".repeat(1005).getBytes(StandardCharsets.UTF_8));
 
-    controller.inbound(
-        "myPath",
-        Map.of("authorization", "******", "x-test", "visible"),
-        Map.of("token", "secret-token", "q", "visible"),
-        request);
+    controller.inbound("myPath", Map.of("authorization", "******", "x-test", "visible"), request);
 
     var latestActivity = latestActivity(activityLogRegistry, connector.id());
     assertThat(latestActivity.message())
@@ -105,7 +102,7 @@ class InboundWebhookRestControllerTest {
             "file", "test.txt", "text/plain", "top secret file contents".getBytes()));
     request.setContent("top secret file contents".getBytes(StandardCharsets.UTF_8));
 
-    controller.inbound("myPath", new HashMap<>(), new HashMap<>(), request);
+    controller.inbound("myPath", new HashMap<>(), request);
 
     var latestActivity = latestActivity(activityLogRegistry, connector.id());
     assertThat(latestActivity.message())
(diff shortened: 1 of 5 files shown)


connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/webhook/InboundWebhookRestControllerTest.java
```diff
@@ -65,14 +65,11 @@ void shouldLogRequestDetailsWithRedactionAndTruncation() throws Exception {
     request.setServerName("example.com");
     request.setServerPort(443);
     request.setRequestURI("/inbound/myPath");
+    request.setQueryString("token=secret-token&q=visible");
```
johnBgood, line 73: this is not tested or did I miss something?

mathias-vandaele: Correct ! I added another test

johnBgood (approved):
Pre approving, only [this](https://github.com/camunda/connectors/pull/7815/changes#r3530545241) Copilot's comment might be worth fixing

PR #7588, opened 2026-06-22
feat: App integrations outbound connector

shtishiv:
## Description

Adds a new **App Integrations** outbound connector for sending notifications and managing channels
via the Camunda App Integrations backend (Microsoft Teams).

The connector exposes two operations:

- **Send Message** — post a plain message, an [Adaptive Card](https://adaptivecards.io/), or a
  linked form to a user (by email) or a channel. Exactly one of `message`, `adaptiveCardJson`, or a
  linked form (supplied via the `linkedResources` custom header, resource type `form`) must be
  provided.
- **Create Channel** — create a Microsoft Teams channel in a given team. `teamId` accepts a raw
  `groupId` or a full Teams URL (the `groupId` is extracted automatically).

### Authentication

Two mechanisms are selectable per element-template instance:

- **API key** — for Self-Managed; sent in the `X-API-KEY` header.
- **OAuth 2.0 client credentials** — for SaaS; the token is fetched, cached, and attached by the
  connector SDK's `HttpClient`. On a `401` the cached token is invalidated and the request is
  retried once with a fresh token.

When running in SaaS, the SaaS context is read from the Operator environment variables and the
`X-Org-Id` / `X-Cluster-Id` headers are added automatically so the backend can attribute the call to
the originating org/cluster.

### Implementation notes

- All HTTP traffic (backend POST + OAuth token fetch) goes through the SDK's
  `CustomApacheHttpClient` and shared `OAuthTokenCacheHolder`, so the connector reuses the SDK's
  transport (timeouts, proxy, TLS) and token cache rather than maintaining a second HTTP stack.
- The element template is generated from the connector annotations by the
  `element-template-generator-maven-plugin`; both the standard and hybrid (Self-Managed) variants
  are committed.
- `app-integrations` is added to the element-template operations-metadata ignore list (it has
  multiple operations but does not yet ship the `steps`/`presets` metadata contract).
- The connector is wired into the default bundle.

## Related issues

closes [#3542](https://github.com/camunda/product-hub/issues/3542)

## Checklist

- [ ] Backport labels are added if these code changes should be backported. No backport label is added to the latest
  release, as this branch will be rebased onto main before the next release. Example backport labels:
    - `backport stable/8.8`: for changes that should be included in the next 8.8.x release.
    - **or** `backport release-8.8.7`: for changes that should be included in the specific release 8.8.7, and this
      *release has already been created*. The release branch will be merged back into stable/8.8 later, so the change
      will be included in future 8.8.x releases as well.
[...]

PR #7835, opened 2026-07-07
feat(saas): report App Integrations connector availability per cluster

### LINKED ISSUE #7832: Report App Integrations connector availability per cluster from connectors SaaS runtime
## Summary

Report each SaaS cluster's **App Integrations connector availability** to the App Integrations
backend from the connectors SaaS runtime. When the runtime starts, it should tell the backend
whether this org/cluster has the App Integrations connector enabled, so the backend knows which
clusters can receive App Integrations traffic.

## Background

- The App Integrations outbound connector is introduced in #7588. In SaaS it is configured through
  environment variables: `APP_INTEGRATIONS_BASE_URL`, `APP_INTEGRATIONS_OAUTH_TOKEN_ENDPOINT`,
  `APP_INTEGRATIONS_OAUTH_CLIENT_ID`, `APP_INTEGRATIONS_OAUTH_CLIENT_SECRET`,
  `APP_INTEGRATIONS_OAUTH_AUDIENCE`.
- The connector already sends `X-Org-Id` / `X-Cluster-Id` headers so the backend can attribute
  calls to the originating cluster.
- The App Integrations backend needs to know **up front** which clusters have the connector
  enabled (not only when a job runs). There is currently no such signal.
- After discussing with Simon Büttner, the connectors SaaS runtime (`bundle/camunda-saas-bundle`)
  is currently the best place to emit this signal: it is the component that knows both the cluster
  identity and whether the App Integrations connector settings are provisioned.
- This depends on #7588 landing first (defines the connector and its SaaS env configuration). The
  runtime code added here does **not** depend on the connector module at compile time — it only
  reads environment configuration and makes an HTTP call.

## Proposed solution

Add a dedicated class to the connectors SaaS runtime (`bundle/camunda-saas-bundle`) that runs when
the runtime starts (on `ApplicationReadyEvent`) and reports the cluster's App Integrations
availability to the App Integrations backend:

- If the App Integrations connector settings are **present** for the cluster, call
  `PUT {APP_INTEGRATIONS_BASE_URL}/api/connector/{orgId}/{clusterId}`.
- Otherwise, call `DELETE {APP_INTEGRATIONS_BASE_URL}/api/connector/{orgId}/{clusterId}`.
- `orgId` / `clusterId` come from `camunda.connector.cloud.organization.id` and
  `camunda.client.cloud.clusterId`.
- Both requests are authenticated with a shared, always-available API key sent in the `X-API-KEY`
  header, provisioned as the `APP_INTEGRATIONS_CONNECTOR_SECRET` environment variable (same header mechanism
  the connector uses in #7588). The SaaS runtime does **not** build an OAuth client for this call.
- Reporting is skipped (with a log statement) when the base URL, API key, org id, or cluster id is
  not available.
- Any failure while reporting is caught and logged; it must never prevent the runtime from
  starting.

### Determining "settings present"

`APP_INTEGRATIONS_BASE_URL` and `APP_INTEGRATIONS_CONNECTOR_SECRET` are fleet-level values that are always
injected into the connectors SaaS runtime (they address and authenticate the reporting endpoint,
including the `DELETE`/disable case). "Settings present" therefore refers to the **per-cluster**
App Integrations connector configuration — detected by the presence of the connector's OAuth
client credentials (`APP_INTEGRATIONS_OAUTH_CLIENT_ID` / `APP_INTEGRATIONS_OAUTH_CLIENT_SECRET`).

## Acceptance criteria

- [ ] A dedicated class in `bundle/camunda-saas-bundle` runs on runtime startup and performs the
      report.
- [ ] Settings present → `PUT /api/connector/{orgId}/{clusterId}`; settings absent →
      `DELETE /api/connector/{orgId}/{clusterId}`.
- [ ] `orgId` and `clusterId` are passed as path segments.
- [ ] Both requests send the API key in the `X-API-KEY` header; no OAuth client is constructed in
      the SaaS runtime.
- [ ] Reporting is skipped (and logged) when base URL, API key, org id, or cluster id is missing.
- [ ] Any error during reporting is caught and logged; runtime startup is unaffected.
- [ ] Unit tests cover: `PUT` when settings present, `DELETE` when absent, the `X-API-KEY` header
      is set, skip when required config is missing, and startup resilience when the HTTP call fails.

## Out of scope

- Provisioning `APP_INTEGRATIONS_BASE_URL` / `APP_INTEGRATIONS_CONNECTOR_SECRET` into the SaaS deployment
  (handled by the deployment/infra configuration).
- The App Integrations backend implementation of the `PUT`/`DELETE /api/connector/{orgId}/{clusterId}`
  endpoint.

yhaskell:
## Description

Reports each SaaS cluster's **App Integrations connector availability** to the App Integrations
backend when the connectors SaaS runtime starts.

A new component (`AppIntegrationsClusterRegistration`) in `bundle/camunda-saas-bundle` listens for
`ApplicationReadyEvent` and calls the App Integrations backend:

- **Settings present** (the connector is provisioned for the cluster) → `PUT {APP_INTEGRATIONS_BASE_URL}/api/connector/{orgId}/{clusterId}`
- **Settings absent** → `DELETE {APP_INTEGRATIONS_BASE_URL}/api/connector/{orgId}/{clusterId}`

`orgId` / `clusterId` come from `camunda.connector.cloud.organization.id` and
`camunda.client.cloud.clusterId`. Both requests are authenticated with a shared, always-available
API key sent in the `X-API-KEY` header (`APP_INTEGRATIONS_CONNECTOR_API_KEY`) — no OAuth client is built in
the SaaS runtime. The call uses the JDK `HttpClient`.

The report runs off the startup thread as a one-time task on the injected `TaskScheduler`. On a
transient failure (I/O error or `5xx`) it reschedules itself with a capped exponential backoff
(5s → 5min), so the runtime keeps retrying until the report succeeds. A client error (`4xx`) is
treated as permanent and is not retried. The request (including URI) is built up front inside a
guard, so a malformed base URL is skipped with a log rather than escaping startup or looping
forever. Nothing here can prevent the runtime from starting.

### Design notes

- `APP_INTEGRATIONS_BASE_URL` and `APP_INTEGRATIONS_CONNECTOR_API_KEY` are fleet-level values that address and
  authenticate the reporting endpoint (including the disable/`DELETE` case), so they are treated as
  always available.
- "Settings present" refers to the **per-cluster** connector configuration and is detected by the
  presence of the connector's OAuth client credentials (`APP_INTEGRATIONS_OAUTH_CLIENT_ID` /
  `APP_INTEGRATIONS_OAUTH_CLIENT_SECRET`).
- Reporting is skipped (and logged) when the base URL, API key, org id, or cluster id is missing.
- Any failure while reporting is caught and logged; it never prevents the runtime from starting.
- The component is excluded from the `test` profile, mirroring the other SaaS runtime configuration
  classes.

This runtime code does not depend on the App Integrations connector module (#7588) at compile time
— it only reads environment configuration and makes an HTTP call. The backend endpoint and the
`APP_INTEGRATIONS_BASE_URL` / `APP_INTEGRATIONS_CONNECTOR_API_KEY` provisioning are out of scope for this PR.

## Related issues

closes #7832

## Checklist

### DIFF
--- a/bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistration.java
+++ b/bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistration.java
@@ -16,12 +16,7 @@
  */
 package io.camunda.connector.runtime.saas;
 
-import java.net.URI;
-import java.net.http.HttpClient;
-import java.net.http.HttpRequest;
-import java.net.http.HttpRequest.BodyPublishers;
-import java.net.http.HttpResponse;
-import java.net.http.HttpResponse.BodyHandlers;
+import io.camunda.connector.runtime.saas.AppIntegrationsClient.RegistrationOutcome;
 import java.time.Duration;
 import java.time.Instant;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistration.java
```diff
@@ -0,0 +1,206 @@
...
+@Component
+@Profile("!test")
+public class AppIntegrationsClusterRegistration {
+
+  static final String API_KEY_HEADER = "X-API-KEY";
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(AppIntegrationsClusterRegistration.class);
+  private static final Duration TIMEOUT = Duration.ofSeconds(10);
+  private static final int MAX_RETRIES = 3;
```
sbuettner: @yhaskell what happens after the retries are exhausted?

I would recommend to schedule a new task if the first one fails with a certain delay until it finished successfully.

sbuettner: @yhaskell for example using [ScheduledTaskRegistrar.addOneTimeTask](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/scheduling/config/ScheduledTaskRegistrar.html#addOneTimeTask(org.springframework.scheduling.config.DelayedTask)). You should be able to simply inject ScheduledTaskRegistrar into your application event listener class.

yhaskell: Good point — replaced the bounded Failsafe burst with a one-time task scheduled off the startup thread (injected `TaskScheduler`) that reschedules itself with a capped exponential backoff (5s → 5min) on transient failures until the report succeeds. A `4xx` is treated as permanent and stops the loop. I used `TaskScheduler.schedule(...)` rather than `ScheduledTaskRegistrar.addOneTimeTask` because the registrar has already scheduled its tasks by the time `ApplicationReadyEvent` fires, whereas `TaskScheduler` schedules immediately from the listener — same one-time-task-until-success behaviour. Let me know if you prefer the registrar API.

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistration.java
```diff
@@ -0,0 +1,156 @@
...
+      LOGGER.info(
+          "Skipping App Integrations cluster registration: base URL, API key, organization id or "
+              + "cluster id is not configured");
+      return;
+    }
+
+    var uri =
+        URI.create(
+            baseUrl.replaceAll("/+$", "") + "/api/connector/" + organizationId + "/" + clusterId);
+    var method = settingsPresent ? "PUT" : "DELETE";
```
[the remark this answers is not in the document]
yhaskell: Fixed — URI/request construction now happens inside a guarded block in `reportAvailability` (on the startup thread). A malformed base URL is logged and the report is skipped, so it can neither escape startup nor loop forever on the scheduler.

bundle/camunda-saas-bundle/src/test/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistrationTest.java
```diff
@@ -0,0 +1,129 @@
...
+            BASE_URL + "///", API_KEY, true, ORG_ID, CLUSTER_ID, httpClient)
+        .reportAvailability();
+    assertThat(capturedRequest().uri()).hasToString(EXPECTED_URI);
+  }
+
+  @Test
+  void doesNotThrowWhenHttpCallFails() throws Exception {
+    doThrow(new IOException("boom")).when(httpClient).send(any(HttpRequest.class), any());
+    assertThatCode(() -> registration(true).reportAvailability()).doesNotThrowAnyException();
+  }
```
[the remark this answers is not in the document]
yhaskell: Added `skipsWhenBaseUrlIsMalformed`, which asserts a malformed base URL results in no scheduling and no HTTP call (and no exception).

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistration.java
sbuettner: All good, thanks for considering!

bundle/camunda-saas-bundle/src/main/java/io/camunda/connector/runtime/saas/AppIntegrationsClusterRegistration.java
```diff
@@ -0,0 +1,231 @@
...
+    this.settingsPresent = settingsPresent;
+    this.organizationId = organizationId;
+    this.clusterId = clusterId;
+    this.httpClient = httpClient;
+    this.taskScheduler = taskScheduler;
+    this.initialRetryDelay = initialRetryDelay;
+    this.maxRetryDelay = maxRetryDelay;
+  }
+
+  @EventListener(ApplicationReadyEvent.class)
```
sbuettner, line 116: For easier testability I recommend to decouple the actual registration from the event listening and scheduling.

Have a `AppIntegrationsClient` that offers a single method like `registerCluster()` that is called by a `AppIntegrationsClusterRegistration` which implements the `EventListener(ApplicationReadyEvent.class)` as well as the scheduling part. This way you decouple the scheduling part from the actual http client. They can live both in the same package.

yhaskell: Done in 7566f05. Split into `AppIntegrationsClient` (single `registerCluster(...)` method doing the HTTP call, returning `DONE`/`RETRY`) and `AppIntegrationsClusterRegistration` (the `@EventListener(ApplicationReadyEvent.class)` + scheduling, delegating to the client). Both live in the same package. Tests are split too: `AppIntegrationsClientTest` covers the HTTP behaviour with a mocked `HttpClient`, and `AppIntegrationsClusterRegistrationTest` covers scheduling/retry with a mocked `AppIntegrationsClient`.

johnBgood (approved):

Szik (approved):

PR #7838, opened 2026-07-07
chore(validator): auto-detect inbound connectors by elementType

ztefanie:
## Summary

- `OperationMetadataIgnoreList.isIgnored` now takes a `JsonNode` parameter and automatically exempts inbound connectors based on `elementType.value` (`bpmn:StartEvent`, `bpmn:IntermediateCatchEvent`, `bpmn:BoundaryEvent`, `bpmn:ReceiveTask`)
- Removed 4 inbound-only entries from the manual list (`email-inbound`, `email-message-start-event`, `slack-inbound`, `webhook`); the remaining entries cover outbound connectors that haven't added operations-metadata yet
- All 10 rule call sites updated from `isIgnored(file)` to `isIgnored(file, template)`
- New test verifies inbound auto-detection for all four BPMN element types

## Test plan

- [ ] `./mvnw -pl element-template-generator/validator test` passes
- [ ] Verify no false positives on connectors with both inbound and outbound templates (e.g. kafka, rabbitmq, github)

### DIFF
--- a/element-template-generator/validator/src/main/java/io/camunda/connector/validator/core/OperationMetadataIgnoreList.java
+++ b/element-template-generator/validator/src/main/java/io/camunda/connector/validator/core/OperationMetadataIgnoreList.java
@@ -16,21 +16,34 @@
  */
 package io.camunda.connector.validator.core;
 
+import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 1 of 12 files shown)


mathias-vandaele (approved):
Thank you 👍🏼

PR #7846, opened 2026-07-07
fix(csv): validate headers for object-based CSV write input

### LINKED ISSUE #7529: bug(csv): NPE in CsvUtils.createCsv when writing List<Map> without configured headers
## Describe the bug

Writing a CSV with `WriteCsv` throws a `NullPointerException` when the input `data` is a `List<Map>` and `format.headers` is not configured in the process variables.

## Steps to reproduce

1. Use the CSV outbound connector with operation `Write CSV`.
2. Pass a `data` value that is a list of maps, e.g. `[{"id": 1, "name": "a"}, {"id": 2, "name": "b"}]`.
3. Mark "Skip Header Record" checkbox and leave `format.headers` unset (or `null`).
4. Run the process — job fails with the NPE above; no actionable error.

sbuettner:
## Description

Writing a CSV with `WriteCsv` threw a `NullPointerException` in `CsvUtils.createCsv` when the input `data` was a `List<Map>` and `format.headers` was not configured (`null`). Setting `headers` to `[]` didn't throw but silently produced empty rows.

Headers are required for object-based (Map) input, so this now throws a meaningful `ConnectorInputException` for both `null` and empty headers instead of an NPE / silent bad output.

Fixes #7529

## Changes
- `CsvUtils.createCsv`: guard Map records with `headersDefined(format)`, throwing `ConnectorInputException` when headers are missing.
- Added test covering null and empty headers.

### DIFF
--- a/connectors/csv/src/main/java/io/camunda/connector/csv/CsvUtils.java
+++ b/connectors/csv/src/main/java/io/camunda/connector/csv/CsvUtils.java
@@ -6,6 +6,7 @@
  */
 package io.camunda.connector.csv;
 
+import io.camunda.connector.api.error.ConnectorInputException;
 import io.camunda.connector.csv.model.CsvFormat;
 import io.camunda.connector.csv.model.ReadCsvRequest;
 import io.camunda.connector.csv.model.ReadCsvResult;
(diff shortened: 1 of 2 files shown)


ztefanie (approved):
looks good

PR #7986, opened 2026-07-17
fix(agentic-ai): reuse existing agent instance on 409 conflict

### LINKED ISSUE #7932: Handle 429 conflict response on agent instance creation
__What should we do?__

When an agent instance was created by an earlier job execution (e.g. on a retry), the API will reject the call with a 429 Conflict response with the conflicting agent instance ID being part of the `detail` message. The AI Agent should handle this constellation and fall back to use the already existing instance.

See https://github.com/camunda/camunda/issues/57575

maff:
## Description

Creating an agent instance for an element instance that already has one now returns `409 ALREADY_EXISTS` (camunda/camunda#57575) instead of an idempotent `200`. This handles that response by parsing the existing agent instance key out of the `detail` message of the `ProblemException` and reusing it, so the AI Agent falls back to the already-existing instance instead of failing.

The whole `detail` message is matched against the engine's documented contract (not just a fragment), so an unexpectedly worded or unparseable detail raises an error instead of risking extraction of the wrong key. All other errors (400, 404, 429, 5xx, etc.) keep their existing classification and retry behavior — unchanged by this PR.

No e2e test is included: the real `409` behavior isn't available until camunda/camunda#57956 (the engine-side change) merges. Unit tests cover the parsing contract (well-formed, unrelated wording, partial match, non-numeric key, null/empty detail) and the classifier/retry interaction (no retries on conflict, single attempt).

## Related issues

closes #7932

## Checklist


johnBgood (approved):

maff:
@ce-dmelnych FYI this is the adaption for  the 409 conflict which now relies on the defined `detail` message.

Issue #6903, opened 2026-04-07
Connector Multi-Engine Support - Breakdown


johnBgood:
**Context:**
We met today to refine the implementation plan for multi-Engine support in Connectors and Clients.
[Discussion here](https://camunda.slack.com/archives/C0AF7LM7CGK/p1775561125440889)

**Key considerations for supporting multi-Engine in Connectors:**
- *Secrets management* (might depend on parallel epic #prj-pdp-3040-centralized-secret-resolution-via-zeebe)
- *Outbound Connectors*: Can be supported with (likely) minimal effort.
- *Inbound Interactions* requires significant changes:
    - Need to track process definitions per Engine (new concept)
    - Require a distinct webhook path per Engine (already a concern for multi-tenancy)
    - Potentially higher impact from noisy neighbours (e.g., for supporting 10 Engines, 10 Kafka consumers—one per Engine—would run on every replica; scaling the connectors runtime would not mitigate this, since all 10 consumers would still be active on each instance).

## ⚠️ Prerequistes

ActivatedJob needs to provide the physicalTenantId in addition to the [existing properties](https://github.com/camunda/camunda/blob/a2764245de80f7cf05cfbfb5205c4627d8c1787b/clients/java/src/main/java/io/camunda/client/api/response/ActivatedJob.java#L25).

## ⏲️ Estimate
It should be pretty straightforward, provided we didn't miss any important changes. I'd say 2-3 weeks, tops, would be okay. That includes a security margin.

## 🔒 Secrets
**Prerequisites from `OC`:**
- `ActivatedJob` needs to provide the `physicalTenantId` in addition to the [existing properties](https://github.com/camunda/camunda/blob/a2764245de80f7cf05cfbfb5205c4627d8c1787b/clients/java/src/main/java/io/camunda/client/api/response/ActivatedJob.java#L25).

We need to update the [SecretContext](https://github.com/camunda/connectors/blob/2c2d7e3e21fa7953f021c74a2e9e069333150be8/connector-sdk/core/src/main/java/io/camunda/connector/api/secret/SecretContext.java#L19) to include the `physicalTenantId` there.

## 📎 Document store
**Prerequisites from `OC`:**
- `ActivatedJob` needs to provide the `physicalTenantId` in addition to the [existing properties](https://github.com/camunda/camunda/blob/a2764245de80f7cf05cfbfb5205c4627d8c1787b/clients/java/src/main/java/io/camunda/client/api/response/ActivatedJob.java#L25).
 
[...]

Issue #6960, opened 2026-04-14
Document Store Updates for Multi-Engine Support


johnBgood:
**Context:**
Enable the document store to support multi-Engine by propagating the `physicalTenantId` throughout document-related requests and contexts.

**Tasks:**
- Ensure `ActivatedJob` provides the `physicalTenantId` [reference](https://github.com/camunda/camunda/blob/a2764245de80f7cf05cfbfb5205c4627d8c1787b/clients/java/src/main/java/io/camunda/client/api/response/ActivatedJob.java#L25).
- Update the [DocumentCreationRequest](https://github.com/camunda/connectors) to include the new property `physicalTenantId`.
- Update document readers and writers to use the `physicalTenantId`, for example:
    - In the `bindVariables(Class)` method, inject `physicalTenantId` via Jackson context:
      ```java
      objectMapper.reader().withAttribute("PHYSICAL_TENANT_ID", job.getPhysicalTenantId())
      ```
    - Use this mapper attribute in the `DocumentDeserializer`
    - Update the `DocumentFactory` so that resolving/creating a document uses the `physicalTenantId`
- Cover both READ and WRITE use cases by making sure context is always propagated to the right place.

PR #6738, opened 2026-03-20
feat(runtime): add reset executable endpoint

johnBgood:
## Description

This pull request introduces several improvements and refactorings to the inbound connector runtime, focusing on making the querying and management of connector executables more flexible and robust. The changes include replacing the immutable `ActiveExecutableQuery` record with a mutable, builder-style class, updating the query API to use this new approach, and adding the ability to reset (deactivate and reactivate) connector executables via the REST API. Additionally, the handling of connector responses and activity logs has been streamlined and improved.

**Key changes:**

### Query API and Filtering Improvements

- Replaced the immutable `ActiveExecutableQuery` record with a mutable class that uses builder-style setters, allowing for more flexible and readable query construction. The query API now accepts a function to configure the query object, enabling more expressive and composable queries. (`ActiveExecutableQuery.java`, `InboundConnectorRestController.java`, `InboundExecutableQueryService.java`) [[1]](diffhunk://#diff-f6ba30f8fd3a5ea131db4c6c7ac326cf57b03eca794aac3d6cc34f05d9ac61cfL19-R95) [[2]](diffhunk://#diff-308fdcb86c7df9aad10cf1811708f7d79e7bbe8ff4283dba8381bda89f506fdeL86-R86) [[3]](diffhunk://#diff-308fdcb86c7df9aad10cf1811708f7d79e7bbe8ff4283dba8381bda89f506fdeL109-R112) [[4]](diffhunk://#diff-b3641951876bc7d433c5dc818f1aaf647c1359f65bcc475b40c90b0c2d2ffcf9R97-L101)

- Added support for filtering by `executableId` in queries, improving the precision of executable lookups. (`ActiveExecutableQuery.java`, `InboundExecutableQueryService.java`) [[1]](diffhunk://#diff-f6ba30f8fd3a5ea131db4c6c7ac326cf57b03eca794aac3d6cc34f05d9ac61cfL19-R95) [[2]](diffhunk://#diff-b3641951876bc7d433c5dc818f1aaf647c1359f65bcc475b40c90b0c2d2ffcf9R97-L101)

### Executable Lifecycle Management

- Added a new REST endpoint to reset (deactivate and reactivate) connector instance executables, enabling runtime recovery and reconfiguration without manual intervention. (`InboundInstancesRestController.java`)

[...]

PR #8020, opened 2026-07-21
feat(connectors): multi-engine support for inbound connectors

### LINKED ISSUE #6962: Inbound Connector Changes for Multi-Engine Support
**Context:**
Inbound connectors, especially webhook connectors and importers, require significant changes for multi-Engine support. In particular, tracking process definitions per engine and distinct webhook paths is needed.

**Tasks:**
- Ensure process definitions are tracked on a per-engine basis in inbound connectors.
- Address webhook path collision: if a process is deployed on multiple engines with the same path, only the first works. Optionally, add a flag (e.g., `APPEND_ENGINE_AND_TENANT_TO_WEBHOOK_PATH`) to append identifiers and resolve conflicts.
    - Implementation idea: webhook becomes `/inbound/<PHYSICAL_TENANT_ID>/<TENANT_ID>/thePath` if the flag is enabled.
    - Default: keep first-come behavior if flag disabled.
- For importers:
    - Update importers to use the REST API per engine and support:
        - [Active process definitions](https://github.com/camunda/connectors/blob/7856111dbd08b2836ee058a34393b4ceef544e46/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/search/SearchQueryClientImpl.java#L55)
        - [Latest process definitions](https://github.com/camunda/connectors/blob/7856111dbd08b2836ee058a34393b4ceef544e46/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/search/SearchQueryClientImpl.java#L43)
    - Consider creating an `ImportSchedulers` per client—see [ImportSchedulers.java#L36](https://github.com/camunda/connectors/blob/4bc4463b4004b6897bd7278bc33b5eb93dd331b9/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ImportSchedulers.java#L36) and [ProcessDefinitionImportConfiguration.java#L48](https://github.com/camunda/connectors/blob/4bc4463b4004b6897bd7278bc33b5eb93dd331b9/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ProcessDefinitionImportConfiguration.java#L48).

**Notes:**
- Some issues, like webhook path collision, exist for multi-tenancy in general, not just multi-Engine.
- The first release may keep existing behavior and address collisions with a runtime parameter in the future.

**Goal:**
Enable reliable inbound interaction for any number of engines without process definition or webhook conflict.

nicpuppa:
## Description

Adds multi-engine ("physical tenant") support to inbound connectors, the first concrete task of the multi-engine epic (#6903). Previously the inbound runtime assumed exactly one Zeebe cluster; this lets it operate against every `CamundaClient` configured via `camunda.clients.*`, each identified by its physical tenant ID.

## Related issues

closes #6962 

## Checklist


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/InboundConnectorRuntimeConfiguration.java
```diff
@@ -85,32 +91,179 @@ public static InboundConnectorBeanDefinitionProcessor inboundConnectorBeanDefini
...
+                    name -> {
+                      var physicalTenantId =
+                          resolvePhysicalTenantId(registry, name, legacyCamundaClient);
+                      return new DefaultInboundConnectorContextFactory(
+                          mapper,
+                          correlationHandlersByPhysicalTenantId.get(physicalTenantId),
+                          secretProviderAggregator,
+                          validationProvider,
+                          processInstanceClientsByPhysicalTenantId.get(physicalTenantId),
+                          documentFactory,
```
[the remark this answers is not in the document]
nicpuppa: this should be resolved by https://github.com/camunda/connectors/issues/6960

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/ProcessElementWithRuntimeData.java
```diff
@@ -32,10 +32,14 @@ public record ProcessElementWithRuntimeData(
     @Nullable String elementName,
     @Nullable String elementType,
     String tenantId,
+    String physicalTenantId,
     ElementTemplateDetails elementTemplateDetails,
     Map<String, String> properties)
     implements ProcessElement {
 
+  /** Physical tenant ID used for single-cluster deployments and by test fixtures. */
+  public static final String DEFAULT_PHYSICAL_TENANT_ID = "default";
```
johnBgood, line 41: 🔧 this should be provided by the Camunda client directly? Consumers shouldn't need to know this

connector-sdk/core/src/main/java/io/camunda/connector/api/inbound/InboundConnectorDefinition.java
```diff
@@ -22,4 +22,19 @@ public record InboundConnectorDefinition(
...
+   * Legacy constructor retained for source/binary compatibility with existing connector code built
+   * against the pre-multi-tenant four-argument constructor. Defaults {@code physicalTenantId} to
+   * {@code "default"}.
+   */
+  public InboundConnectorDefinition(
+      String type,
+      String tenantId,
+      String deduplicationId,
+      List<? extends ProcessElement> elements) {
+    this(type, tenantId, deduplicationId, elements, "default");
```
johnBgood, line 38: :wrench: same here, this default should come from the client

connectors/webhook/src/test/java/io/camunda/connector/inbound/HttpWebhookExecutableTest.java
```diff
@@ -158,7 +158,7 @@ private static ProcessElement elementWithRawProperties(Map<String, String> rawPr
   private static InboundConnectorContext contextWithElements(ProcessElement... elements) {
     var definition =
         new InboundConnectorDefinition(
-            "io.camunda:webhook:1", "<default>", "dedup-id", List.of(elements));
+            "io.camunda:webhook:1", "<default>", "dedup-id", List.of(elements), "<default>");
```
johnBgood: :question: this should be `default`. Emphasizing the need for this const to be in the client :)

connector-runtime/spring-boot-starter-camunda-connectors/src/main/resources/application.properties
```diff
@@ -4,6 +4,12 @@
 #camunda.connector.polling.enabled=false
 #camunda.connector.webhook.enabled=false
 
+# When multiple physical tenants are configured (camunda.clients.*), enable this to scope
+# webhook paths by physical tenant and tenant (/inbound/<physicalTenantId>/<tenantId>/<path>)
+# instead of the legacy /inbound/<path>, avoiding path collisions across physical tenants/tenants.
+# Default is disabled to preserve existing webhook URLs.
+#camunda.connector.webhook.append-physical-tenant-and-tenant-to-path=false
```
johnBgood, line 14: :question: is it something we could deduce? As you said if we have PT configured, we want the path to include PT ids. That would avoid misconfiguration

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ImportSchedulers.java
```diff
@@ -27,31 +35,28 @@ public class ImportSchedulers {
...
   public void scheduleLatestVersionImport() {
-    try {
-      var result = importers.importLatestVersions();
-      stateStore.update(result);
-      ready = true;
-    } catch (Exception e) {
-      LOG.error("Failed to import LATEST process versions", e);
-      ready = false;
-    }
+    ready = pollAllPhysicalTenants("LATEST", Importers::importLatestVersions);
```
johnBgood: :wrench: use a const if possible

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ImportSchedulers.java
```diff
@@ -62,17 +67,50 @@ public void scheduleActiveVersionImport() {
       LOG.debug("Skipping active versions polling.");
       return;
     }
+    ready = pollAllPhysicalTenants("ACTIVE", Importers::importActiveVersions);
```
johnBgood: :wrench: use a const if possible

connector-runtime/connector-runtime-test/src/main/java/io/camunda/connector/runtime/test/inbound/InboundConnectorDefinitionBuilder.java
```diff
@@ -29,6 +29,8 @@ public class InboundConnectorDefinitionBuilder {
 
   private String deduplicationId = "test-deduplication-id";
 
+  private String physicalTenantId = "default";
```
johnBgood, line 32: same here

connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/inbound/webhook/WebhookTestsBase.java
```diff
@@ -89,25 +104,46 @@ private static InboundConnectorContextImpl buildContext(
...
   }
 
   private static InboundConnectorElement webhookElement(
-      long processDefinitionKey, String bpmnProcessId, int version, String path) {
+      long processDefinitionKey,
+      String bpmnProcessId,
+      int version,
+      String path,
+      String tenantId,
+      String physicalTenantId) {
```
johnBgood, line 129: :bulb: we have some E2E tests (see `WebhookActivatedDocumentTests` as an example). We could test the real flow with PT enabled

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/InboundConnectorRuntimeConfiguration.java
```diff
@@ -85,32 +91,179 @@ public static InboundConnectorBeanDefinitionProcessor inboundConnectorBeanDefini
     return new InboundConnectorBeanDefinitionProcessor(environment);
   }
 
+  /**
```
johnBgood: :wrench: this configuration class is quite big now, not sure if we can break it a bit, worth checking

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookConnectorRegistry.java
```diff
@@ -27,61 +27,81 @@ public class WebhookConnectorRegistry {
   private final Logger LOG = LoggerFactory.getLogger(WebhookConnectorRegistry.class);
 
   private final Map<String, WebhookExecutables> executablesByContext = new HashMap<>();
+  private final boolean appendPhysicalTenantAndTenantToPath;
```
johnBgood, line 30: :wrench: note that the web-modeler calls the [/inbound endpoint](https://github.com/camunda/connectors/blob/220cfabebdbc4d20ba005fb5d49e5b0945f492e6/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundConnectorRestController.java#L55) to build the webhook's URL. 
It uses the connector's properties to do that, especially `context`. 

This property doesn't change, so depending on whether PT is enabled or not, we need to append the `/{PTid}/{tenantId}` to this context property. 

We manipulated this `context` in the past, you can check the history of the [ConnectorDataMapper](https://github.com/camunda/connectors/blob/3a63db65d369b020cdeb219c8065b3650a0d7851/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ConnectorDataMapper.java#L23) class for more information.

They also call the /logs endpoint.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/model/ProcessDefinitionRef.java
```diff
@@ -16,5 +16,18 @@
...
+import io.camunda.connector.runtime.core.inbound.ProcessElementWithRuntimeData;
+
+/**
+ * A composite key for identifying a process definition by its physical tenant ID, BPMN process ID
+ * and tenant ID.
+ */
+public record ProcessDefinitionRef(String physicalTenantId, String bpmnProcessId, String tenantId) {
+
+  public static final String DEFAULT_PHYSICAL_TENANT_ID =
+      ProcessElementWithRuntimeData.DEFAULT_PHYSICAL_TENANT_ID;
```
johnBgood, line 28: same here

connector-runtime/connector-runtime-core/src/main/java/io/camunda/connector/runtime/core/inbound/InboundConnectorElement.java
```diff
@@ -87,7 +91,13 @@ public String deduplicationId(List<String> deduplicationProperties) {
     if (deduplicationMode == null) {
       // legacy deployment, return a deterministic unique id
       LOG.debug("Missing deduplicationMode property, using legacy deduplicationId computation");
-      return element.tenantId() + "-" + element.processDefinitionKey() + "-" + element.elementId();
+      return physicalTenantId()
```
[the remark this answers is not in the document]
johnBgood: did you solve this?

johnBgood: ⚠️ Another issue about this:
We now prefix the ID with physicalTenantId + "-". For every existing single-cluster deployment (physicalTenantId = "default"), this changes the string from e.g. tenant-42-myElement to default-tenant-42-myElement.
This might be an issue for some connectors, such as Kafka, for instance. Kafka brokers persist committed offsets per consumer group ID. Since the auto-generated group ID changes for every Kafka inbound connector that doesn't set dedupId explicitly, upgrading to this PR silently creates a new consumer group on the broker — losing the old group's committed offset position. Depending on auto.offset.reset, this means either reprocessing the entire retained topic history or skipping straight to the tail on first restart after upgrade.

I guess that it is acceptable, but it should be **clearly documented in the release notes** as a breaking change. Many connectors are probably affected. 

cc @chillleader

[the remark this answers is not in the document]
chillleader: are dashes even allowed in tenant names?

chillleader: Correct - Kafka is definitely affected. We already have a warning in the element template about relying on runtime default and not specifying your own stable consumer group ID, so with a docs/release notes update this change should be fine.

I can't think of other connectors where this is critical tbh. The polling connector uses dedup ID as well but it's not an issue since it's stateless.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ImportSchedulers.java
```diff
@@ -62,17 +67,50 @@ public void scheduleActiveVersionImport() {
...
+  }
+
+  /**
+   * Polls every configured physical tenant concurrently, so that one tenant stalling (e.g. a
+   * connection attempt that hangs until timeout) does not delay the others from starting and
+   * completing within the same scheduled tick.
+   */
+  private boolean pollAllPhysicalTenants(
```
chillleader, line 85: 🔧 I have a gut feeling that consolidating this tenant-fanout logic in one place (e.g. either in `Importers` or `ImportSchedulers`) is easier to think about than the current version that introduces the concept of physical tenants into both. Also having N `Importers` beans for each physical tenant ID could be exchanged for just one bean that can take a parameter in its method call. I might be wrong though / this might not be feasible, in that case feel free to disregard this.

nicpuppa:
Thanks for the thorough review, @chillleader and @johnBgood — addressed the outstanding points:

- **Web-modeler webhook URL** (johnBgood): `ConnectorDataMapper` now rewrites the exposed `inbound.context` with the physical-tenant/tenant-scoped path when `append-physical-tenant-and-tenant-to-path` is enabled, restoring the pattern that existed before #6738 removed it — so the URL shown in web-modeler matches the real active route.
- **Auto-deduce the scoping flag** (johnBgood): when the property is left unset, it's now inferred from configured client count (`>1` client ⇒ scoping on). An explicit `true`/`false` always wins, so nobody's existing override changes behavior.
- **Consolidate `Importers`/`ImportSchedulers`** (chillleader): `Importers` is now a single stateless bean called with `(physicalTenantId, client)` per invocation instead of one instance per physical tenant.
- **Split `InboundConnectorRuntimeConfiguration`** (johnBgood): extracted `PhysicalTenantIds` (client/tenant resolution helpers) and `InboundCorrelationConfiguration` (correlation-handler beans) into their own classes.
- **New E2E coverage** (johnBgood): added a Docker-free `WebhookPhysicalTenantScopingE2ETest` that deploys the same raw webhook path on two physical tenants through the real Spring wiring and asserts they register as distinct, independently-routable connectors.

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/PhysicalTenantIds.java
```diff
@@ -0,0 +1,149 @@
...
+ * InboundConnectorRuntimeConfiguration}, {@link InboundCorrelationConfiguration}, {@code
+ * ProcessDefinitionImportConfiguration}, {@code ProcessInstanceClientConfiguration}). Plain static
+ * methods, deliberately not {@code @Bean}-produced: none of these consumers may declare a {@code
+ * Map<String, X>}-typed {@code @Bean} parameter, since Spring's dependency resolution special-cases
+ * any such parameter by collecting *all* beans of type {@code X} by name — including scalar
+ * override beans (e.g. a test's {@code @MockitoBean SearchQueryClient}) — instead of using the bean
+ * whose own declared type is the map. That would silently produce a map keyed by the scalar bean's
+ * name rather than the real per-physical-tenant map.
+ */
+public final class PhysicalTenantIds {
```
johnBgood, line 40: :wrench: maybe something we can move to the SDK, customers might want to use it

connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/WebhookPhysicalTenantScopingE2ETest.java
```diff
@@ -0,0 +1,139 @@
...
+      "camunda.clients.engine-a.primary=true",
+      "camunda.clients.engine-b.mode=self-managed",
+      "camunda.clients.engine-b.grpc-address=http://engine-b.internal:26500",
+      "camunda.clients.engine-b.physical-tenant-id=tenantb",
+      "camunda.connector.polling.enabled=false",
+      "camunda.connector.webhook.enabled=true",
+      "camunda.connector.webhook.append-physical-tenant-and-tenant-to-path=true"
+    })
+@AutoConfigureMockMvc
+class WebhookPhysicalTenantScopingE2ETest {
```
johnBgood, line 75: :wrench: this is not a real E2E tests, see `WebhookNotActivatedDocumentTests` for example.
I'd like us to start a connectors runtime and call the different endpoints (legacy + new PT aware endpoints) etc

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/PhysicalTenantIds.java
nicpuppa: Thanks for the suggestion! I'd lean against moving this one, for a couple of reasons:

- `PhysicalTenantIds` depends on `io.camunda.client.spring.bean.CamundaClientRegistry` and on `SearchQueryClient` (our internal process-definition search abstraction) — both Spring/runtime-internal. `connector-sdk/core` deliberately has zero Spring dependency today, so moving this as-is would either drag Spring into the SDK or require splitting out a much thinner, more generic piece than what's actually here.
- The underlying need — a connector knowing which physical tenant it's running under — is already covered at the SDK level: `InboundConnectorDefinition` already exposes `physicalTenantId()` directly, so connector authors don't need these registry-resolution helpers to get that.

Given that, I think this is best left as internal runtime wiring. Happy to reconsider if there's a concrete customer use case that isn't already served by `InboundConnectorDefinition#physicalTenantId()`.

johnBgood: I missed the SearchQueryClient dep indeed, all good then

connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/WebhookPhysicalTenantScopingE2ETest.java
nicpuppa: done

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/executable/ConnectorDataMapper.java
```diff
@@ -18,12 +18,47 @@
...
 
-  private static Map<String, String> allPropertiesMapper(ActiveExecutableResponse response) {
-    return response.elements().getFirst().connectorLevelProperties();
+  private static final String INBOUND_CONTEXT_PROPERTY = "inbound.context";
+
+  private final boolean appendPhysicalTenantAndTenantToPath;
+
+  public ConnectorDataMapper() {
+    this(false);
+  }
```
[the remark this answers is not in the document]
johnBgood: @nicpuppa I think this should be solved differently. 
In the `ConnectorDataMapper`, we could add a boolean `appendPTAndTenantIds`.
This would be  `true` when called from the `InboundConnectorRestController`, and it would be a query param when used in the `InboundInstancesService`. I'll update the frontend accordingly later; let's default to false.

Only [this](https://github.com/camunda/connectors/blob/220cfabebdbc4d20ba005fb5d49e5b0945f492e6/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/controller/InboundInstancesRestController.java#L99) method will use a query param.

The existing `appendPhysicalTenantAndTenantToPath` boolean would still gate all this.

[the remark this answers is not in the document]
nicpuppa: on it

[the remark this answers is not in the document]
nicpuppa: addressed with [64ab5fc](https://github.com/camunda/connectors/pull/8020/commits/64ab5fc71526981807f1f7f5fc5d26390e387a94)

connectors-e2e-test/connectors-e2e-test-http/src/test/java/io/camunda/connector/e2e/WebhookPhysicalTenantScopingE2ETest.java
```diff
@@ -0,0 +1,154 @@
...
+    properties = {
+      "spring.main.allow-bean-definition-overriding=true",
+      "camunda.connector.webhook.enabled=true",
+      "camunda.connector.polling.enabled=false",
+      "camunda.connector.webhook.append-physical-tenant-and-tenant-to-path=true"
+    },
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@CamundaSpringProcessTest
+@AutoConfigureMockMvc
+class WebhookPhysicalTenantScopingE2ETest {
```
johnBgood, line 75: :wrench: this class needs to be annotated with @SlowTest so that it runs only during the nightly E2E tests

nicpuppa: done with [c31ad0a](https://github.com/camunda/connectors/pull/8020/commits/c31ad0a83be28165920120d47ff072b4e83684c5)

johnBgood (approved):
👍 Awesome work! ty

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/webhook/WebhookContextKeys.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public final class WebhookContextKeys {
+
+  private WebhookContextKeys() {}
+
+  public static String compose(String physicalTenantId, String tenantId, String path) {
+    Objects.requireNonNull(
+        physicalTenantId, "physicalTenantId must not be null when path scoping is enabled");
+    Objects.requireNonNull(tenantId, "tenantId must not be null when path scoping is enabled");
+    return physicalTenantId + "/" + tenantId + "/" + path;
```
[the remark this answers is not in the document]
johnBgood: @nicpuppa to be considered

Szik (approved):

PR #8023, opened 2026-07-21
feat(agentic-ai): LLM Layer: Introduce generic ChatModel provider SPI

### LINKED ISSUE #8022: Agentic AI: ChatModel provider SPI
## What should we do?

Own the LLM invocation layer behind a module SPI instead of the single LangChain4J-bound `AiFrameworkAdapter`, with LangChain4J as the first implementation behind it. Behavior for the existing LangChain4J providers stays identical.

- **`ChatModel` provider SPI** — `AutoCloseable` with `ChatResult execute(ChatRequest)`, resolved and closed per request. A `ChatModelRegistry` routes by `ChatModelFactory.supports(configuration)` and fails loud when zero or more than one factory matches, dispatching on a neutral `ChatModelConfiguration` (exposing `provider()`/`model()`) so a provider can supply its own configuration through the SPI.
- **Turn-based continuation** — `ChatResult` is a sealed `Completed | Continuation`; the request handler loops while the result is a `Continuation`, persisting each round as a turn. LangChain4J always returns `Completed`, so the loop runs exactly once for it.
- **Normalized stop reasons** — a sealed `StopReason` maps provider finish reasons to a neutral vocabulary; the content-filter guard is enforced once in the handler off `CONTENT_FILTERED`.
- **Structured content model** — tool-call results persist as a structured `List<Content>` (`ToolCallResultContent`) instead of an untyped payload, and the content model gains `ReasoningContent` (opaque provider reasoning payload) and `ProviderContent` (provider-native block preserved verbatim). Pre-existing (Camunda 8.9) state is migrated on read via an explicit persisted schema version at each conversation-state root plus a shared upcaster, rather than inferring the format from the shape of `content`.
- **Token metrics** — `AgentMetrics.TokenUsage` records prompt-cache (read/creation) and reasoning token counts. `inputTokenCount` is the new, non-cached input; cache-read/creation are disjoint auxiliary buckets (never also counted in input); `reasoningTokenCount` is a breakdown within `outputTokenCount`.

The decision and its drivers are captured in ADR 009. Backward compatibility of persisted state is covered by golden-state tests.

## Why should we do it?

The single LangChain4J adapter couples every provider to the framework's feature set and release cadence, models an invocation as a single request/response (no notion of a provider continuing a turn), and reduces a tool-call result to an untyped payload — losing the structure needed for documents, images, reasoning, and provider-native content. Owning the invocation layer behind an SPI lets us add providers, richer content, and turn continuation on our own terms, without changing today's behavior.

maff:
## Description

Replaces the single LangChain4J-bound `AiFrameworkAdapter` with a module-owned `ChatModel` provider SPI, with LangChain4J as the first implementation behind it. Behavior for the existing LangChain4J providers is unchanged.

- **`ChatModel` provider SPI** — `AutoCloseable` with `ChatResult execute(ChatRequest)`, resolved and closed per request. A `ChatModelRegistry` routes by `ChatModelFactory.supports(configuration)` and fails loud when zero or more than one factory matches. It dispatches on a neutral `ChatModelConfiguration` — the type carried through `AgentConfiguration` (`chatModel()`), exposing just `provider()` and `model()` — so a provider can supply its own configuration through the SPI rather than being confined to the built-in provider union. The LangChain4J providers become per-provider factories behind the SPI (abstract `LangChain4JChatModelFactory` + one concrete factory per backend), with their model-building logic unchanged.
- **Turn-based continuation** — `ChatResult` is a sealed `Completed | Continuation`; the request handler loops while the result is a `Continuation`, persisting each round as a turn. LangChain4J always returns `Completed`, so the loop runs exactly once for it.
- **Normalized stop reasons** — a sealed `StopReason` maps provider finish reasons to a neutral vocabulary. The content-filter guard is enforced once, generically, in the request handler off `CONTENT_FILTERED`.
- **Structured content model** — tool-call results are persisted as a structured `List<Content>` (`ToolCallResultContent`) instead of an untyped payload, and the sealed content model gains `ReasoningContent` (opaque provider reasoning payload) and `ProviderContent` (provider-native block preserved verbatim). The synthetic tool-call-documents user message is still built from the raw results, so document handling is unchanged.
- **Token metrics** — `AgentMetrics.TokenUsage` records prompt-cache (read/creation) and reasoning token counts. `inputTokenCount` is the new, non-cached input; cache-read/creation are disjoint auxiliary buckets, never also counted in input; `reasoningTokenCount` is a breakdown within `outputTokenCount`. The new fields are omitted from JSON when zero.

The LangChain4J integration also moves from `aiagent.framework` to `aiagent.chatmodel`, separating the generic chat-model execution layer from the provider implementation.

**Compatibility:** behavior-identical for the existing LangChain4J providers. The persisted tool-call-result format changes to the structured shape. Pre-existing (Camunda 8.9) state is migrated on read via an explicit persisted schema version at each conversation-state root — `schemaVersion` on `AgentContext` for the process-variable payload, `schemaVersion` on the Camunda-document conversation payload — plus a shared upcaster. Each root is self-describing and migrated independently: the pointer-based store (Camunda document) reads the version from the externally-stored payload, since the `AgentContext` version only describes the pointer. Covered by golden-state tests.

**Design:** recorded in [ADR 009](connectors/agentic-ai/docs/adr/009-chat-model-provider-spi.md); the agent reference docs (`ai-agent.md`) are updated for the new SPI, content model, and metrics semantics.

## Related issues

closes #8022

## Checklist


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/document/CamundaDocumentConversationSerializer.java
```diff
@@ -22,8 +25,20 @@ public CamundaDocumentConversationSerializer(ObjectMapper objectMapper) {
...
-    return objectMapper.readValue(document.asInputStream(), DocumentContent.class);
+  /**
+   * Reads the conversation document content, migrating its {@code messages} in place when {@code
+   * schemaVersion} predates {@link AgentContext#CURRENT_SCHEMA_VERSION} (Camunda 8.9 tool-call
+   * results persisted with a flat {@code content} field).
+   */
+  public DocumentContent readDocumentContent(Document document, int schemaVersion)
+      throws IOException {
+    if (schemaVersion >= AgentContext.CURRENT_SCHEMA_VERSION) {
+      return objectMapper.readValue(document.asInputStream(), DocumentContent.class);
```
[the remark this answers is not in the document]
maff: Fixed in 38b095d863 (+ docs in 4641902f4c). You're right — for the Camunda document store the `AgentContext` is only a pointer, and a legacy pointer-style context binds `schemaVersion` to the current version via the record-builder default, so the external payload was bound with the old flat shape un-migrated.

The document payload now carries its own `schemaVersion` (mirroring the AWS AgentCore blob envelope), and `readDocumentContent(Document)` reads the version off the document itself (the `schemaVersion` parameter is gone), so each persisted conversation root is self-describing and migrated independently of the pointer. Added a regression test (`loadsLegacyDocumentPayloadRegardlessOfPointerAgentContextSchemaVersion`) that loads a legacy payload behind a current-version pointer, and reworked the BC fixture test to drive the real `readDocumentContent` gate instead of migrating by hand.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -158,18 +161,63 @@ private R proceed(
...
+      boolean continued;
+      do {
+        LOGGER.debug(
+            "Sending turn (iterationKey={}) to the model",
+            workingConversation.currentTurn().iterationKey());
+
+        final var windowedSnapshot =
+            workingConversation.window(agentConfiguration.contextWindowSize());
+        final var chatResult =
+            chatModel.execute(new ChatRequest(executionContext, windowedSnapshot));
```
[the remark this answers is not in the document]
maff: Real, but latent in this PR: LangChain4J always returns `Completed`, so the continuation loop runs exactly once and the window is never re-applied across continuation rounds — this cannot manifest with the v1 provider shipped here. It first bites once a native provider returns `Continuation` (e.g. Anthropic `pause_turn`), which lands in a later PR where it is end-to-end testable.

Tracked as a known challenge to fix there: a continuation round must retain the complete current logical turn (freeze a base snapshot, or exempt the in-progress turn from windowing) so a resuming provider does not lose the input it is resuming.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -158,18 +161,63 @@ private R proceed(
...
+          // previousTurns and would otherwise never reach the instance-level counters (only the
+          // final round's delta is pushed after the loop). No status change — still mid-turn.
+          notifyMetrics(
+              executionContext,
+              workingConversation.toAgentContext(),
+              workingConversation.currentTurnMetrics(),
+              null,
+              false);
+
+          throwIfLimitsReached(workingConversation, agentConfiguration);
```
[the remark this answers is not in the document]
maff: Same scope as the windowing comment: this branch is only reached on a continuation round, and LangChain4J never returns `Continuation`, so it cannot manifest in this behavior-identical PR. The pre-loop limit check (before any model call or history emission) is the only one reachable in v1, and it fails cleanly.

Tracked as a known challenge for the native-provider PR that first emits `Continuation`: either persist a resumable partial state before the limit check, or enforce the limit without leaving the non-durable assistant-history / metrics-delta side effects, so a retry cannot double-emit.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/configuration/AgenticAiLangChain4JFrameworkConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+@Configuration
+@ConditionalOnProperty(
+    value = "camunda.connector.agenticai.framework",
+    havingValue = "langchain4j",
+    matchIfMissing = true)
```
maff: 🔧 This conditional is no longer relevant, let's remove it completely.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/message/StopReason.java
```diff
@@ -0,0 +1,81 @@
...
+package io.camunda.connector.agenticai.aiagent.model.message;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * Provider-neutral, normalized finish reason. Diagnostics + a thin predicate surface only — never
+ * load-bearing for control flow (that keys off {@link AssistantMessage#hasToolCalls()}). The raw
+ * vendor value is always preserved in {@link AssistantMessage#metadata()} in addition to living on
+ * this field for a genuinely unrecognised value (see {@link UnknownStopReason} below).
```
maff: 🔧 Remove the not load bearing framing as we raise errors on content filtered now.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentContext.java
```diff
@@ -22,6 +22,7 @@
 @AgenticAiRecord
 @JsonDeserialize(builder = AgentContext.AgentContextJacksonProxyBuilder.class)
 public record AgentContext(
+    @RecordBuilder.Initializer("CURRENT_SCHEMA_VERSION") int schemaVersion,
```
maff, line 30: 🔧 Does that actually make sense? When a record with a missing value is deserialized, it would initialize the current schema version, no?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentContext.java
```diff
@@ -30,6 +31,19 @@ public record AgentContext(
...
+   * whenever a change to the persisted shape requires migrating previously-persisted state on read
+   * (see {@code ConversationSchemaMigration}).
+   */
+  public static final int CURRENT_SCHEMA_VERSION = 1;
+
+  /**
+   * Marker for state persisted before {@code schemaVersion} was introduced (Camunda 8.9). Such
+   * state is missing the field entirely; it is treated as this version and migrated on read.
+   */
+  public static final int LEGACY_SCHEMA_VERSION = 0;
```
maff: 🔧 Better name for LEGACY? Missing?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/VersionedAgentContextDeserializer.java
```diff
@@ -0,0 +1,45 @@
...
+ * mapper.treeToValue(..., AgentContext.class)}, which uses {@code AgentContext}'s own builder
+ * deserializer, not this field deserializer.
+ */
+public class VersionedAgentContextDeserializer extends JsonDeserializer<AgentContext> {
+
+  @Override
+  public @Nullable AgentContext deserialize(JsonParser p, DeserializationContext ctxt)
+      throws IOException {
+    return ConversationSchemaMigration.migrateAndBindAgentContext(
+        p.readValueAsTree(), (ObjectMapper) p.getCodec());
```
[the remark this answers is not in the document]
maff: 🤖 _Reply from Claude Code (acting for the author)_

Fixed in c259e6ab01. `migrateAndBindAgentContext` now stamps the tree's `schemaVersion` to `CURRENT_SCHEMA_VERSION` after upcasting and before binding, so a migrated context is re-persisted as current and a later read no longer re-runs the legacy upcaster (which would otherwise wrap the already-structured `List<Content>` into a single `ObjectContent`). Added an idempotency test and a field-preservation test that asserts sibling fields and message metadata survive the upcast.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/ProviderConfiguration.java
```diff
@@ -34,7 +35,7 @@
     name = "type",
     description = "Specify the LLM provider to use.",
     defaultValue = ANTHROPIC_ID)
-public sealed interface ProviderConfiguration
+public sealed interface ProviderConfiguration extends ChatModelConfiguration
```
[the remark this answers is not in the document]
maff: 🤖 _Reply from Claude Code (acting for the author)_

Addressed by scoping the documentation rather than adding a request hook in this PR. You're right that no request path here feeds a custom `ChatModelConfiguration`: both v1 requests bind the sealed `ProviderConfiguration`, which is the only value copied into `AgentConfiguration.chatModel()`. The SPI *type* is the extension seam; the request-side wiring that supplies custom/native configurations lands with the v2 request types in a later PR. `ChatModelConfiguration`'s javadoc, ADR-009, and ai-agent.md now say this explicitly instead of implying the extension is wired end-to-end today (2ddf4947fd).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/tool/ToolCallResultContent.java
```diff
@@ -0,0 +1,128 @@
...
+   * already-deserialized Java objects). The mirror-image lift for raw JSON nodes read from legacy
+   * (Camunda 8.9) data lives in {@code ConversationSchemaMigration#liftLegacyContent}. Each branch
+   * produces a singleton list (or an empty one for null/blank) — a multi-element list only ever
+   * arises later once provider-specific chat model implementations emit structured content
+   * directly.
+   */
+  static List<Content> contentFromObject(@Nullable Object content) {
+    return switch (content) {
+      case null -> List.of();
+      case String s -> s.isBlank() ? List.of() : List.of(TextContent.textContent(s));
```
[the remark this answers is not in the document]
maff: **Claude Code:** No behavior change vs `main`. `ToolCallConverterImpl.asToolExecutionResultMessage` guards `if (StringUtils.isBlank(content)) content = CONTENT_NO_RESULT;`, and this guard predates the PR — it is identical on `main` (ToolCallConverterImpl:88-90) and fires on any blank *joined* content, whether it came from an empty list or a blank `TextContent`. On `main`, `contentAsString` returned a blank tool-result string verbatim and then hit that same guard, producing `CONTENT_NO_RESULT`. So blank/whitespace tool output maps to the synthetic no-result text identically before and after this change; model input for existing LangChain4J providers is unchanged. Resolving as no regression.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/AgentContextSchemaMigration.java
```diff
@@ -0,0 +1,207 @@
...
+   * </ul>
+   */
+  static List<Content> liftLegacyContent(@Nullable JsonNode node, ObjectMapper mapper) {
+    if (node == null || node.isMissingNode() || node.isNull()) {
+      return List.of();
+    }
+
+    if (node.isTextual()) {
+      String text = node.textValue();
+      return (text == null || text.isBlank()) ? List.of() : List.of(TextContent.textContent(text));
```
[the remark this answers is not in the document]
maff: **Claude Code:** Same as the sibling comment on `ToolCallResultContent` — no behavior change. `liftLegacyContent` maps a blank textual node to `List.of()`, and downstream `ToolCallConverterImpl` turns any blank/empty content into `CONTENT_NO_RESULT` via a guard that already exists on `main` (ToolCallConverterImpl:88-90). On `main` a blank 8.9 tool-result string was forwarded verbatim and hit that same guard, also yielding `CONTENT_NO_RESULT`. Empty-list and blank-text therefore produce identical model input, matching pre-change behavior. Resolving as no regression.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/mapping/BlobEnvelope.java
```diff
@@ -38,8 +38,17 @@
  */
 public record BlobEnvelope(String blobType, int version, JsonNode data) {
 
+  /**
+   * The AWS AgentCore conversation store shipped on the 8.10 track (never in Camunda 8.9), so
+   * version 1 is its first and current shape for every blob type, including {@code
+   * TOOL_CALL_RESULTS}'s structured {@code List<Content>} content — there is no legacy flat-{@code
+   * content} shape to migrate from, unlike the in-process and Camunda-document conversation stores.
+   */
   public static final int CURRENT_VERSION = 1;
```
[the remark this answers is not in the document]
maff: **Claude Code:** The AWS AgentCore conversation store never shipped in Camunda 8.9 (introduced on the 8.10 track), so there is no real legacy version-1 flat-content state to migrate — `agentcore-list-events.json` was leftover scaffolding from the AgentCore BC test that was removed earlier in this PR, not persisted production state. Removed the unreferenced fixture (and the version-history note on `CURRENT_VERSION`) in 0eb5465d75; `CURRENT_VERSION = 1` stands as the first and only shape, structured `List<Content>` from the start. Resolving.

connectors/agentic-ai/docs/reference/ai-agent.md
```diff
@@ -912,7 +1013,7 @@ Langchain4JAiFrameworkAdapter
 - **`JsonSchemaElementModule`**: Custom Jackson module needed because LangChain4J doesn't expose standard polymorphic annotations on `JsonSchemaElement`. Serializer/deserializer handle all concrete types (`JsonObjectSchema`, `JsonEnumSchema`, `JsonStringSchema`, `JsonArraySchema`, `JsonAnyOfSchema`, `JsonReferenceSchema`, etc.).
 - **`DocumentToContentModule`**: Jackson module registering `DocumentToContentSerializer` for Camunda `Document` objects in tool call result content — serializes to `{type, media_type, data}` structure.
```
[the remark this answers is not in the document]
maff: Fixed in 5f7f29f7f9. Both classes were removed in the converter refactor — only `DocumentToContentConverter` remains, and `ContentConverterImpl` now takes a plain injected `ObjectMapper` (no custom Jackson module). Dropped the stale `DocumentToContentModule` bullet and the trailing clause on the `ContentConverter` line.

maff:
Part of #8022 (Agentic AI: ChatModel provider SPI), under the #7211 native-provider restructure.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -117,6 +117,29 @@ public AgentConversation ingest(AssistantMessage assistantMessage, AgentMetrics
...
+    }
+    var updatedPreviousTurns = new ArrayList<>(previousTurns);
+    updatedPreviousTurns.add(currentTurn);
+    var nextTurn =
+        new AgentConversationTurn(
+            currentTurn.iterationKey() + 1, List.of(), null, AgentMetrics.empty());
+    return new AgentConversation(
+        configuration, currentContext, systemMessage, updatedPreviousTurns, nextTurn);
+  }
+
```
nikonovd, line 142: 🔧 IMO this concept is currently Anthropic-only.

I'm not very leaned towards pouring a provider-specific concept into our domain model.
Furthermore, logically this is not a new turn, but a continuation of the old one (see https://platform.claude.com/docs/en/agents-and-tools/tool-use/server-tools).

WDYT of letting the Anthropic chat model handle it internally, returning a single turn? I can imagine that process experts wouldn't expect a new turn due to a technical constraint on the server side.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -158,18 +161,53 @@ private R proceed(
...
-    final var chatResponse =
-        framework.executeMeasuringTime(
-            executionContext, conversation.window(agentConfiguration.contextWindowSize()));
-    final var updatedConversation =
-        conversation.ingest(chatResponse.assistantMessage(), chatResponse.metrics());
+    var workingConversation = conversation;
+    try (final var chatModel = chatModelRegistry.resolve(agentConfiguration.chatModel())) {
+      // Continuation loop (ADR-009): a provider may pause mid-turn (e.g. Anthropic pause_turn) and
+      // return a Continuation rather than a Completed result. Each Continuation becomes its own
+      // persisted turn and the model is re-called until a Completed ends the loop.
```
nikonovd: ⛏️ Likely I did it myself as well, but could we agree on skipping ADR references in code comments and inverse the references into the ADR (as it's already done)?

If you want to keep a comment here, I'd prefer shrinking it to essential explanations.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -158,18 +161,53 @@ private R proceed(
...
+          LOGGER.debug(
+              "Provider requested continuation (iterationKey={}); resuming with another round",
+              workingConversation.currentTurn().iterationKey());
+
+          throwIfLimitsReached(workingConversation, agentConfiguration);
+          workingConversation = workingConversation.nextContinuationRound();
+        }
+      } while (continued);
+    }
+    final var updatedConversation = workingConversation;
```
nikonovd: 🔧 As mentioned in AgentConversation, let's agree on semantics here. If we keep it as-is, i'd prefer to have it extracted into a separate method and optimize the code readability a bit.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/document/DocumentToContentConverterImpl.java
```diff
@@ -22,7 +22,7 @@
 import org.apache.hc.core5.http.ContentType;
 
 /**
- * Converts a Camunda {@link Document} to a Langchain4j {@link Content} object to be used in user
+ * Converts a Camunda {@link Document} to a LangChain4J {@link Content} object to be used in user
```
nikonovd: ⛏️ nit: actually it's being called `LangChain4j` with a lowercase letter 😆

nikonovd: see https://docs.langchain4j.dev/

nikonovd (commented):
I was not able to finish all low-level review, but here's my first high level feedback with a specific remark. Overall it looks very solid and I like the new SPI. Great work 🚀 

General remarks:
* The ADRs we started to write are pretty low-level and might get obsolete fast if we don't keep track. We have already some ADRs and I started to observe that we might need to shift some levels higher, so that the content is up-to-date while code is still agile. WDYT?
* Some comments are very explanatory and likely leaning towards specific examples. WDYT of shrinking them. Maybe we also are able to find some prompt directives for our `CLAUDE.md`🤔 
* I like the high-level abstraction that you introduced on our new domain model, especially that we have a unified terminology over the providers (e.g. `CONTENT_FILTERED` while it might differ per provider).
* The SPI overall looks solid  🚀 

* My current main concern is about the `pause_turn` looping with continuing turns - please see my other comments

---

Otherwise I'll look into the changes more deeply tomorrow 👍

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
maff: I experimented with modeling that loop internally in the provider specific part, but I also wanted it to be able to push metrics & check limits (e.g. max model calls) across each iteration as every continuation will still produce new cost for the input. Modeling the loop as an abstract concept (only emitted by Anthropic, I agree!) fits natively into the conversation concept as a dedicated loop iteration.

If we handle this inside the chat model, we'd need to surface multiple assistant messages from the call + probably put the metrics tracking + limits checking inside the chat model.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/document/DocumentToContentConverterImpl.java
maff: 1 typo fixed, 1 new typo introduced 🤦‍♂️

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
maff: For now, I kept the general logic for now, but extracted it into 2 smaller methods.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/AgentConversationTurnInputComposerImpl.java
```diff
@@ -152,7 +158,7 @@ public CompositionResult compose(
     return UserMessage.builder().content(content).metadata(defaultMessageMetadata()).build();
   }
 
-  private Optional<ToolCallResultMessage> createToolCallResultMessage(
+  private Optional<List<ToolCallResult>> resolveOrderedToolCallResults(
```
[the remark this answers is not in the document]
maff: 🤖 _Reply from Claude Code (acting for the author)_

Fixed in be71cfa488. Updated the walkthrough and code-paths reference to `resolveOrderedToolCallResults()`, and described the split: the method now only resolves/orders the results, while `compose()` builds the `ToolCallResultMessage` from them.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
nikonovd: The current definition of a loop iteration is the scope around (user) input message and its assistant reply.

With your implementation the pause continuation is basically starting with the last assistant message. I can image this could look weird, especially in the UI (though I didn't check yet).

However I agree that checking limits is important here, so it's fine to have this loop inside the base handler.
WDYT about aggregating the entire loop into a single turn, but having it as you implemented to be able to check the configured limits (max model calls, etc)?

Semantically it's still a single turn and we should fold only the complete assistant message into the history with all token metrics produced until that.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -199,6 +198,59 @@ private R proceed(
...
+      workingConversation =
+          workingConversation.ingest(chatResult.assistantMessage(), chatResult.metrics());
+
+      agentInstanceClient.createHistoryForAssistantMessage(
+          executionContext,
+          agentInstanceKey,
+          workingConversation.currentTurn(),
+          OffsetDateTime.now());
+
+      continued = chatResult instanceof ChatResult.Continuation;
```
nikonovd, line 233: 💭 WDYT of introducing a continuation limit to have a break condition for the potential infinte loop?
Anthropic pushes this practice also in their [examples](https://platform.claude.com/docs/en/build-with-claude/handling-stop-reasons#implement-retry-logic-for-pause-turn)

nikonovd (approved):
As discusssed, great work 🚀

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: As discussed, we have maxModelCalls as a global limit which would also apply here - it's fine 👍

Issue #7225, opened 2026-05-13
Agentic AI: new v2 AI Agent Task / Sub-process connector types and element templates


maff:
__Is your feature request related to a problem? Please describe.__

The current AI Agent connector type identifiers (`io.camunda.agenticai:aiagent:1`, `io.camunda.agenticai:aiagent-job-worker:1`) and element template structure carry historical baggage that blocks the clean introduction of new providers, backends, and API families.

__Describe the solution you'd like__

Introduce new connector types and element templates for the AI Agent:

| What | Task | Sub-process |
|------|------|-------------|
| Connector type / Zeebe job type | `io.camunda.agenticai:aiagent:task:2` | `io.camunda.agenticai:aiagent:subprocess:2` |
| Element template ID | `io.camunda.connectors.agenticai.ai-agent-task.v2` | `io.camunda.connectors.agenticai.ai-agent-subprocess.v2` |
| Element template filename | `agenticai-ai-agent-task.v2.json` | `agenticai-ai-agent-subprocess.v2.json` |
| Display name | AI Agent Task | AI Agent Sub-process |

*(Updated to the identifiers actually shipped in #8043, colon-delimited to match the existing v1 convention `io.camunda.agenticai:aiagent:1`, rather than this issue's original hyphenated proposal.)*

The new templates expose the new `ProviderConfiguration` shape: one provider entry per wire format, with backend and apiFamily dropdowns and conditional auth fields. Update the `element-templates/README.md` index and move the previous templates into `element-templates/versioned/` per the existing pattern.

Old connector types stay available but are handled by a separate input-rewrite shim issue.

__Describe alternatives you've considered__

Bumping the existing templates again carries forward the legacy name ("job-worker") and the legacy provider structure; we'd accumulate template versions instead of getting the structural reset.

__Additional context__

PR #8048, opened 2026-07-23
refactor(agentic-ai): normalize v1 connector naming, prepare for v2

maff:
## Summary

Preparation work for the upcoming v2 AI Agent connectors (stacked on #8023), containing no v2 code:

- Drops the "Ai" prefix and the "JobWorker"/"OutboundConnector" naming across the AI Agent Task and Sub-process connectors, settling on "Agent" and "Task"/"SubProcess" to match the flavor names already used in element templates and docs (e.g. `AiAgentFunction` -> `AgentTaskV1Function`, `AiAgentJobWorker` -> `AgentSubProcessV1Function`). The "V1" suffix is kept only on the top-level Function and Request classes, where the version is part of the public contract; it's dropped everywhere else (execution contexts, request handlers, responses, provider configuration).
- Moves the legacy sealed provider configuration into its own `v1` package (`model/request/v1`), freeing up the simple name `ProviderConfiguration` for the new wire-format-first v2 model landing in the next PR.
- Extracts the request-scoped fields shared between the two connector flavors' request records into standalone `AgentTaskRequestData` / `AgentSubProcessRequestData` classes, so the upcoming v2 request types can reuse the same data shape instead of duplicating it.
- Mirrors all of the above in the e2e test module (package/class renames).

Updates `AGENTS.md`, the AI Agent reference doc, and the affected ADRs to match.

Preparation for #7225 

## Test plan

- [x] `mvn -pl connectors/agentic-ai/connector-agentic-ai test` green
- [x] `connectors-e2e-test-agentic-ai` test-compiles cleanly


maff:
Part of #7225 (v2 connector types + element templates) / #7211 (native LLM provider integrations).

nikonovd (approved):

PR #8108, opened 2026-07-28
feat: propagate physicalTenantId for secret resolution

### LINKED ISSUE #6959: Update Secrets Management for Multi-Engine Support
**Context:**
Support for multi-Engine in Connectors requires updates to secrets management mechanisms. The primary change involves propagating the `physicalTenantId` so that secrets can be properly resolved per engine.

**Tasks:**
- Update the [SecretContext](https://github.com/camunda/connectors/blob/2c2d7e3e21fa7953f021c74a2e9e069333150be8/connector-sdk/core/src/main/java/io/camunda/connector/api/secret/SecretContext.java#L19) to include a `physicalTenantId` field.
- Ensure all places where secrets are resolved have access to the correct `physicalTenantId` (from `ActivatedJob`).
- Coordinate with the OC team to confirm that `ActivatedJob` provides `physicalTenantId` in addition to existing properties.

**References:**
- [ActivatedJob properties](https://github.com/camunda/camunda/blob/a2764245de80f7cf05cfbfb5205c4627d8c1787b/clients/java/src/main/java/io/camunda/client/api/response/ActivatedJob.java#L25)

nicpuppa:
## Description

Multi-engine deployments need secrets resolved per orchestration cluster, but `SecretProvider` implementations currently have no way to tell which engine a process is running on — `SecretContext` only carries the logical tenant and the process definition ID.

This adds a `physicalTenantId` component to `SecretContext` and populates it at every place secrets are resolved:

- **Outbound** — sourced from `ActivatedJob.getPhysicalTenantId()`, both when substituting secrets into job variables (`JobHandlerContext`) and when masking them out of error output (`OutboundConnectorExceptionHandler`). These two have to agree: masking works by re-resolving the secret, so if the substituting path scoped to engine A and the masking path didn't, the value would be re-resolved as "not found" and leak into the error payload verbatim.
- **Inbound** — sourced from the connector element, via a `getPhysicalTenantId()` helper extracted from the lookup that already existed inline in `getDefinition()`.

Two compatibility details worth calling out for review:

- The record keeps its previous 2-arg constructor, so third-party `SecretProvider` implementations and existing call sites compile unchanged. `physicalTenantId` is `@Nullable`.
- Clients report an unset physical tenant as an **empty string** (`ActivatedJobImpl` uses `getOrEmpty`), so the compact constructor normalizes blank to `null`. Without this, every provider would have to null-check *and* blank-check.

No behavior changes for single-cluster deployments — the new component is purely additive and providers that ignore it resolve exactly as before.

## Related issues

closes #6959

## Checklist

### DIFF
--- a/connector-sdk/core/src/main/java/io/camunda/connector/api/secret/SecretContext.java
+++ b/connector-sdk/core/src/main/java/io/camunda/connector/api/secret/SecretContext.java
@@ -16,4 +16,33 @@
  */
 package io.camunda.connector.api.secret;
 
-public record SecretContext(String tenantId, String processDefinitionId) {}
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Scope in which a secret is resolved.
+ *
+ * @param tenantId the logical (multi-tenancy) tenant the process belongs to
+ * @param processDefinitionId the BPMN process definition ID the secret is resolved for
+ * @param physicalTenantId identifies the orchestration cluster (engine) the process runs on, so
+ *     that secrets can be resolved per engine in multi-engine deployments. {@code null} when the
+ *     runtime could not determine it, e.g. against a cluster that predates multi-engine support.
+ */
+public record SecretContext(
+    String tenantId, String processDefinitionId, @Nullable String physicalTenantId) {
+
+  public SecretContext {
+    // clients report an unset physical tenant as an empty string (protobuf/REST default);
+    // normalize it so providers only have to check for null
+    if (physicalTenantId != null && physicalTenantId.isBlank()) {
+      physicalTenantId = null;
+    }
+  }
+
+  /**
+   * Retains the pre-multi-engine constructor signature so code compiled against the previous record
+   * shape keeps working. Leaves {@link #physicalTenantId()} unset.
+   */
+  public SecretContext(String tenantId, String processDefinitionId) {
+    this(tenantId, processDefinitionId, null);
+  }
+}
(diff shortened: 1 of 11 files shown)


johnBgood (approved):

nicpuppa:
@johnBgood does this PR need QA as well ?

connector-sdk/core/src/main/java/io/camunda/connector/api/secret/SecretContext.java
```diff
@@ -16,4 +16,33 @@
...
+ * Scope in which a secret is resolved.
+ *
+ * @param tenantId the logical (multi-tenancy) tenant the process belongs to
+ * @param processDefinitionId the BPMN process definition ID the secret is resolved for
+ * @param physicalTenantId identifies the orchestration cluster (engine) the process runs on, so
+ *     that secrets can be resolved per engine in multi-engine deployments. {@code null} when the
+ *     runtime could not determine it, e.g. against a cluster that predates multi-engine support.
+ */
+public record SecretContext(
+    String tenantId, String processDefinitionId, @Nullable String physicalTenantId) {
```
johnBgood, line 31: 🔧 Maybe we should use this new attirbute in `EnvironmentSecretProvider`

johnBgood:
> @johnBgood does this PR need QA as well ?

We can, see my last comment about `EnvironmentSecretProvider`. When it's done, we can test this PR. Without it we can't really. 

We would need to test:
- EnvironmentSecretProvider using the physicalTenantId
- non regression tests

connector-sdk/core/src/main/java/io/camunda/connector/api/secret/SecretContext.java
nicpuppa: solved

nicpuppa:
@johnBgood do you want to take another look ?

johnBgood (approved):
LGTM, you can add the `qa:required` label 🟢

johnBgood (approved):

Szik (approved):

PR #8146, opened 2026-07-29
feat(ms-email-inbound): expose attachment metadata before activation

polenz:
Resolve lightweight attachment metadata (name, content type, size, isInline) at poll time so inbound activation conditions can filter on attachment properties such as file type, e.g.:

  some a in attachmentMetadata satisfies a.contentType = "application/pdf"

Metadata is fetched with a byte-free $select (no contentBytes) and only for messages where hasAttachments is true; the actual attachment documents are still downloaded lazily after correlation, preserving the late-download behaviour. Metadata lookup failures are split by cause: permanent failures degrade gracefully to an empty list (email is processed as if it had no resolvable attachment metadata), while transient failures (429/5xx) leave the email untouched so it is re-evaluated on the next poll, instead of risking a false negative on the activation condition.

## Known trade-offs

- Metadata is fetched for every attachment-bearing message regardless of whether the configured activation condition actually references `attachmentMetadata`. Gating on that would require introspecting the FEEL condition text to determine whether it touches the field, which seemed like disproportionate complexity for the Graph-call savings involved, so we accept the extra request per attachment-bearing mail.

## Related issues

closes #

## Checklist


connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +239,50 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+          getGraphClient()
+              .messages()
+              .byMessageId(messageId)
+              .attachments()
+              .get(
+                  requestConfiguration ->
+                      // Selecting only metadata fields ensures Graph does not return the
+                      // (potentially large) contentBytes payload for file attachments.
+                      requestConfiguration.queryParameters.select =
+                          new String[] {"id", "name", "contentType", "size", "isInline"});
```
vringar, line 266: What does isinline mean?

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/model/output/EmailMessage.java
```diff
@@ -24,13 +24,15 @@ public record EmailMessage(
     String body,
     String bodyContentType,
     OffsetDateTime receivedDateTime,
-    List<Document> attachments) {
+    List<Document> attachments,
+    List<EmailAttachmentMetadata> attachmentMetadata) {
```
vringar: I feel like we want to subclasses on with attachments and one with attachmentsMetadata. 
Because when the Documents are attached they already have everything but isInline set. And idk if anyone cares about isInline
So there is BaseMessage, MessageWithMetadata and MessageWithAttachments?

vringar (changes requested):

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
polenz: `isInline` means that attachments (e.g. images) are embedded directly into the email text, rather than being a separate attachment.

vringar (approved):

Szik:
**Runtime finding from QA verification of this PR** (SM 8.10-SNAPSHOT on GKE, connectors bundle image `8.10.0-pr8146-run31398807468-a1`)

An activation condition referencing the new `attachmentMetadata` cannot start a process instance, because the condition is evaluated twice against two different message shapes:

1. **Stage 1** — `MessageProcessor:37` calls `context.canActivate(message)` with `MessageWithMetadata`. `attachmentMetadata` is visible, the condition matches. ✅
2. **Stage 2** — `MessageProcessor:100` correlates with `.variables(GraphApiMapper.toMessageWithAttachments(...))`, and `InboundCorrelationHandler.correlate()` (connector-runtime-core, line 100) **re-evaluates the same activation condition** against those variables. `MessageWithAttachments` has no `attachmentMetadata`, so the FEEL expression yields null → `ActivationConditionNotMet`. ❌

### Reproduction (~10 min on any SM cluster running the bundle image)

- Deploy a process with an ms-email-inbound start event: `pollingConfig.filterCriteria.subjectContains=qa-pr8146`, folder `inbox`, `PT30S`, `onlyUnread=true`, correlation `notRequired`, postprocess `mark-read`.
- Activation condition: `=some a in attachmentMetadata satisfies a.contentType = "application/pdf"`
- Send a real mail matching the subject filter with a PDF attached.
- **Observed:** 0 process instances; the connector logs `Activation condition not met` once per poll cycle.

Controls run alongside it, which is what rules out an environment problem:

| Probe | Condition | Result |
|---|---|---|
| PDF attachment | `some a in attachmentMetadata satisfies a.contentType = "application/pdf"` | **0 PIs**, `Activation condition not met` each poll |
| control | `=true` | 1 PI created, `variables.attachments` populated with `{fileName, contentType, size}` |
| TXT attachment | same as row 1 | 0 PIs, `NoMatchingElement` at Stage 1 (correct negative) |
| no attachment | same as row 1 | 0 PIs, `NoMatchingElement`, no crash, pod healthy |
| mirror test | `some a in attachments satisfies a.name = "probe.pdf"` | 0 PIs, fails at **Stage 1** — `MessageWithMetadata` has no `attachments` |

The control proves metadata pre-fetch, attachment download, Document creation, doc-store upload and PI variables all work — only the condition path fails. The mirror test shows the asymmetry from the other side: neither field is visible at both gates, so no attachment-based activation condition can survive both.

### Side effect

With `consumeUnmatchedEvents = false` (the default), the Stage-2 failure maps to `ForwardErrorToUpstream` → `ShouldPostprocess.NO`, so mark-read is skipped and the same mail is re-polled indefinitely. With `consumeUnmatched = true` the mail is consumed, but there is still no process instance.

### Two possible directions — your call

1. Skip the second `canActivate` in the correlate path when Stage 1 already passed (e.g. a `correlate` overload that bypasses the internal re-check).
2. Carry `attachmentMetadata` through to `MessageWithAttachments` — which reverses the deliberate "so no separate metadata list is kept" decision documented on that record.

### Note on test coverage

`MessageProcessorTest` mocks `InboundConnectorContext` and asserts via `verify(context).canActivate(...)` / `verify(context).correlate(...)`. A mocked `correlate()` never executes the runtime's internal re-check, so this failure mode is structurally invisible to the current unit tests — it only appears against a real runtime.

Test-environment caveat, unrelated to this PR: client-credentials auth was unavailable on the test tenant (`AADSTS7000222`, expired app secret), so mails were sent and polled using a bearer token instead.

Full evidence (deployment, Graph API sends, connector pod logs, per-probe ledger) in the internal QA thread: https://camunda.slack.com/archives/C05K4TAEPDW/p1786393074658609?thread_ts=1786372416.806459&cid=C05K4TAEPDW

No re-run needed on our side until the PR is updated.

vringar:
Good catch, thanks for digging this deep (the mirror test is what convinced me the split was the actual bug, not an environment fluke).

Root cause is the second activation check inside `InboundCorrelationHandler.correlate()` (connector-runtime-core), which re-evaluates the condition against whatever variables it's handed, regardless of whether we already checked once. Skipping that re-check isn't really an option: RabbitMQ's and Kafka's consumers call `correlate()` directly without ever calling `canActivate()` themselves, so for them that internal check is the *only* place the activation condition gets enforced. So the bug was on our side, not the runtime's.

The actual bug: `MessageWithMetadata` and `MessageWithAttachments` were two different record shapes, and `attachmentMetadata` only lived on the first one, so it vanished by the time `correlate()` ran its re-check.

Fix: folded both back into a single `EmailMessage` record that carries `attachmentMetadata` from the start and just gets `attachments` filled in later once the content is downloaded (same idea as how `ReadEmailResponse.attachments` goes from empty to populated in the plain email connector). So `attachmentMetadata` is now the same field with the same value at both the poll-time check and the correlation re-check, no more asymmetry between the two stages.

Pushed to the branch. Whenever you get a chance, feel free to re-run.

Szik (approved):

vringar:
Sorry I just saw the big green button and went ahead. @camunda/connectors-experience do you also want to review?

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +239,50 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+        metadata.add(
+            new EmailAttachmentMetadata(
+                attachment.getId(),
+                attachment.getName(),
+                attachment.getContentType(),
+                attachment.getSize() == null ? null : attachment.getSize().longValue(),
+                attachment.getIsInline()));
+      }
+      return metadata;
+    } catch (Exception e) {
```
crobbins215: Degrading to an empty list turns a transient Graph failure into a silent false negative on the activation condition — and with the default config that consumes the mail.

Graph throttles this endpoint (429) and returns the occasional 5xx. When that happens for a mail that *does* carry a PDF, `attachmentMetadata` is empty, `some a in attachmentMetadata satisfies a.contentType = "application/pdf"` evaluates false, and we land in `NoMatchingElement`. From there `MessageProcessor` decides the mail's fate:

```java
if (e.discardUnmatchedEvents()) { yield ShouldPostprocess.NO; }  // stays unread → re-polled
yield ShouldPostprocess.YES;                                      // mark-read / move / delete
```

`discardUnmatchedEvents` comes straight off `consumeUnmatchedEvents` (`ActivationConditionEvaluator`), which defaults to `false`. So on the default path we take the `YES` branch and postprocess runs — the mail is marked read, moved, or (with `DeleteOperation`) deleted. A mail that should have started a process instance is consumed instead, and the only trace is a WARN in the pod logs. With `consumeUnmatchedEvents=true` it self-heals on the next poll, which is the safer of the two, but that isn't the default.

Suggestion: split transient from permanent. On 429/5xx/timeout, don't hand the message to `handler` at all — return early from `processMessageItem` so the mail is left untouched and re-polled next cycle. Keep the empty-list degrade for genuinely permanent errors (404, malformed response). Something like:

```java
} catch (ApiException e) when (isTransient(e)) {   // 429, 5xx, timeouts
  throw new TransientMetadataException(messageId, e);
} catch (Exception e) {
  LOGGER.warn("Failed to resolve attachment metadata for message {}", messageId, e);
  return List.of();
}
```

…with `processMessageItem` catching the transient case and returning `true` without accepting the message.

Worth surfacing through `context.log(...)` too, not just SLF4J — right now a user in Operate sees "activation condition not met" with no way to tell a real non-match from a throttled lookup.

Side note, not this PR's problem: the `discardUnmatchedEvents` branch above reads inverted against the flag name — `consumeUnmatchedEvents=true` is the branch that *doesn't* consume. Might be worth a separate look.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +239,50 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
   }
 
-  private static boolean processMessageItem(Message msg, Consumer<EmailMessage> handler) {
-    var myMsg = GraphApiMapper.toEmailMessage(msg, List.of());
+  private boolean processMessageItem(Message msg, Consumer<EmailMessage> handler) {
+    // Resolve lightweight attachment metadata (no content download) before the activation
+    // condition is evaluated, so conditions can filter on attachment properties such as file type.
+    List<EmailAttachmentMetadata> attachmentMetadata =
+        Boolean.TRUE.equals(msg.getHasAttachments())
+            ? fetchAttachmentMetadata(msg.getId())
```
crobbins215: This adds one serial Graph round-trip per attachment-bearing mail, inside the `PageIterator` page callback. A page is `PAGE_SIZE` messages, so a busy mailbox where most mail has attachments means that many sequential calls per poll cycle, on the same thread, before any of them is even known to be interesting. Two consequences worth thinking about: extra throttling pressure on a tenant that may already be near its Graph budget, and poll cycles that can outrun the configured interval.

Two things that would take the edge off:

**Reclaim the call you've made redundant.** `fetchAttachments` still opens with a dedicated count request:

```java
if (messageBuilder.attachments().count().get() == 0) {
  return List.of();
}
```

`msg.attachmentMetadata()` already answers that question by the time we get here, so this can become `if (msg.attachmentMetadata().isEmpty()) return List.of();` and the count call disappears. Net Graph traffic per attachment mail then stays roughly where it was before the PR. (Pre-existing line — your change is what makes it removable.)

**The gating question.** Metadata is fetched for every attachment-bearing mail regardless of whether any activation condition references `attachmentMetadata`, so users who never touch the field still pay a request per mail. Gating on the condition text is probably more trouble than it's worth, but if that's the accepted trade-off it'd be good to say so explicitly in the PR description — otherwise it reads like an oversight to the next person looking at poll latency.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +239,50 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+                  requestConfiguration ->
+                      // Selecting only metadata fields ensures Graph does not return the
+                      // (potentially large) contentBytes payload for file attachments.
+                      requestConfiguration.queryParameters.select =
+                          new String[] {"id", "name", "contentType", "size", "isInline"});
+      if (response == null || response.getValue() == null) {
+        return List.of();
+      }
+      var metadata = new ArrayList<EmailAttachmentMetadata>();
+      for (var attachment : response.getValue()) {
```
crobbins215, line 303: This maps the whole attachments collection, but `fetchAttachments` only ever converts `FileAttachment` — `ItemAttachment` and `ReferenceAttachment` are deliberately skipped there ("intentionally not supported"). So `attachmentMetadata` and `attachments` describe different sets.

The user-visible failure: a mail with a forwarded-message attachment (`itemAttachment`) or a OneDrive link (`referenceAttachment`) shows up in `attachmentMetadata`, a condition matches on it, the process instance starts — and `attachments` is empty. Condition said yes, payload has nothing. The skip itself is pre-existing and fine; what's new is that this PR makes it reachable, because before now nobody could see or filter on that list.

Cheapest fix is to filter here so the two lists agree:

```java
for (var attachment : response.getValue()) {
  if (!(attachment instanceof FileAttachment)) {
    continue;   // mirrors fetchAttachments, which only converts FileAttachment
  }
  ...
}
```

One wrinkle: with `$select` restricted to metadata fields, Graph may not populate enough of the payload for the Kiota deserializer to pick the `FileAttachment` subtype, in which case you'd want `@odata.type` in the `$select` and a discriminator check instead of `instanceof`. Worth verifying against a real mailbox with a forwarded-message attachment before settling on either.

Second, smaller thing: inline attachments (signature images and the like) land in both lists, so `attachments` on a plain HTML mail can be non-empty in a way users won't expect. `isInline` is exposed, so they *can* filter — but it needs to be documented, otherwise every condition written against `attachmentMetadata` quietly matches logo GIFs.

Related: there's no join between the two lists. `EmailAttachmentMetadata` carries `id`, `Document` doesn't, so a process that filters on metadata can't reliably pick out the matching document beyond comparing file names. Fine for now, but it'll come up.

crobbins215 (commented):
Light review — the approach is right and the fix for Szik's two-stage finding is the correct one (folding both shapes into one `EmailMessage` beats bypassing the runtime re-check, given Kafka and RabbitMQ depend on that check being the only gate). 

- **Metadata-fetch failure is a silent false negative, and the default config consumes the mail.** The `catch (Exception) → List.of()` degrade means a Graph 429 makes a PDF mail look attachment-free; `consumeUnmatchedEvents` defaults to `false`, which is the branch that runs postprocess, so the mail gets marked read / moved / deleted with no process instance. This is the one I'd call blocking.
- **One serial Graph round-trip per attachment mail, inside the page callback.** Throttling and poll-latency cost. Partly reclaimable: the metadata list makes the `attachments().count()` guard in `fetchAttachments` redundant, so that request can go.
- **`attachmentMetadata` and `attachments` describe different sets.** `fetchAttachments` is `FileAttachment`-only, but the metadata list isn't filtered, so a condition can match on an `itemAttachment` and then get an empty `attachments`. Pre-existing skip, newly reachable.

Two smaller notes, neither blocking. The `hasAttachments` gate and the degrade path in `MicrosoftMailClient` have no coverage — the mapping tests are good, but the risky branch is the untested one, and it's the same class of gap that let the original bug through to QA. And `attachmentMetadata` is a new user-facing FEEL variable with nothing pointing at it: the `activationCondition` tooltip in the three element templates doesn't hint it exists, and the docs checkbox is still unticked.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
vringar: Fixed. `fetchAttachmentMetadata` now catches `ApiException` and checks the status code — on 429/5xx it throws a `TransientMetadataLookupException` instead of degrading to an empty list, and `processMessageItem` catches that to leave the message untouched (skip calling the handler, but keep iterating) so it's re-evaluated on the next poll rather than risking the false-negative/consume scenario you described. Permanent failures still degrade to an empty list. Also added a `context.log(WARNING, ...)` alongside the SLF4J warn for the transient case so it's visible in Operate, not just pod logs.

Left the `discardUnmatchedEvents` naming issue as-is per your note that it's out of scope here.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
vringar: Fixed — `fetchAttachments` now checks `msg.attachmentMetadata().isEmpty()` instead of the separate `attachments().count().get()` call, so net Graph traffic per attachment mail is back to roughly where it was before this PR.

On the gating question: added a note to the PR description under "Known trade-offs" — determining whether the configured activation condition actually references `attachmentMetadata` would mean introspecting the FEEL expression, which felt like disproportionate complexity for the request savings, so we're accepting the extra per-mail request for now.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
vringar: Fixed the set mismatch — the metadata loop now skips anything that isn't a `FileAttachment`, mirroring `fetchAttachments`, so `attachmentMetadata` and `attachments` describe the same set.

I didn't add an `@odata.type`-based discriminator: went with the plain `instanceof` check since it's the same pattern `fetchAttachments` already relies on, and I don't have a way to verify against a real mailbox/forwarded-message attachment in this environment. Flagging that as something worth double-checking against a live tenant.

Also documented in `EmailMessage`'s javadoc that inline attachments (signature images etc.) show up in both lists and can be filtered via `isInline`.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
crobbins215: Thanks — and my suggestion was wrong here, sorry. Copilot's follow-up on `fetchAttachments` ([r3796049913](https://github.com/camunda/connectors/pull/8146#discussion_r3796049913)) is right, and the hole is the guard I asked you to add:

```java
if (msg.attachmentMetadata().isEmpty()) { return List.of(); }
```

An empty metadata list means two different things now — "this mail has no file attachments" and "the lookup failed and we don't know". The permanent-failure branch of `fetchAttachmentMetadata` still degrades to an empty list, and that mail *does* reach the handler. If it activates on something other than `attachmentMetadata` (a subject filter, or plain `=true`), we short-circuit here and its documents never get downloaded — where before this PR they would have been. I traded a Graph call for a correctness hole; should have caught that the guard needs a signal that can't mean "unknown".

Two ways out, either fine by me:

**Guard on `hasAttachments`.** It's already in `getSelect()`, just not carried on the record. Authoritative from Graph, can't mean "unknown":

```java
if (!Boolean.TRUE.equals(msg.hasAttachments())) { return List.of(); }
```

Bonus: it's a useful FEEL field in its own right, cheaper than `count(attachmentMetadata) > 0` for the common "any attachment at all" condition.

**Or just drop the guard.** Copilot's point that an empty collection already yields an empty document list is correct, and `fetchAttachments` only runs for mail that already activated — so the call we'd be saving was never on the hot path to begin with. Simplest, least state.

One correction to that Copilot comment though: the claim that inline-only messages also produce an empty metadata list is **not** true. Inline images are `fileAttachment` with `isInline=true`, and your filter is on type rather than on inline, so they stay in the list. Only the item/reference-attachment case yields an empty list legitimately — and that one matches pre-PR behaviour anyway, since those were skipped in the download loop regardless. Not worth chasing.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
crobbins215: Filter looks right, and fair enough on not adding the `@odata.type` discriminator without a mailbox to test against.

One thing worth raising though: that unverified bit is now load-bearing, because this fix and the `fetchAttachments` guard combine into a single point of failure. If the metadata-only `$select` does prevent Kiota from resolving the `FileAttachment` subtype, then `instanceof FileAttachment` is false for *every* attachment, so:

1. `attachmentMetadata` comes back empty for every mail — the feature silently does nothing, no condition can ever match; **and**
2. `fetchAttachments` short-circuits on `attachmentMetadata().isEmpty()`, so attachments stop being downloaded **at all**, including for processes that never referenced the new field.

Failure 2 is the one that bites, because it's a regression for users who aren't using this feature. Both are silent — no exception, no warning, just empty lists.

Cheap to settle: one mail with a PDF against a real mailbox, check `attachmentMetadata` is non-empty. If it is, we're fine and this whole comment is moot. If it isn't, `@odata.type` in the `$select` plus a discriminator check, as discussed.

Given the QA history on this PR I'd want that confirmed before merge rather than as a follow-up — it's the same shape as the original finding (condition matches in one place, silently doesn't in another), and it'd be a shame for it to come back a third time. @Szik may already have a tenant warm from the last round.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
crobbins215: This is exactly what I was after — the skip-and-retry keeps the mail intact, and routing it through `context.log` as well as SLF4J means someone debugging in Operate can actually tell a throttled lookup from a real non-match. Nice.

One question on the boundary, worth a check rather than a change: `isTransient` reads `ApiException.getResponseStatusCode()`, so it only classifies failures that arrived as an HTTP error *response*. What does the Kiota OkHttp adapter throw on a connection-level failure — socket timeout, connection reset, DNS blip? If those surface as something other than `ApiException` (an `IOException`, or a `RuntimeException` wrapping one), they'll fall through to the generic `catch (Exception)` and degrade to an empty list — which is the false-negative path this fix exists to close, reached by the most transient failure mode there is. Same for an `ApiException` constructed without a response, where `getResponseStatusCode()` is `0` and `isTransient` returns false.

I'm not certain what the adapter does here, so this may well be a non-issue. If it turns out IO failures don't arrive as `ApiException`, the fix is small — treat them as transient too:

```java
} catch (ApiException e) {
  if (isTransient(e)) { throw new TransientMetadataLookupException(messageId, e); }
  LOGGER.warn(...);
  return List.of();
} catch (IOException | UncheckedIOException e) {          // or whatever the adapter surfaces
  throw new TransientMetadataLookupException(messageId, e);
} catch (Exception e) {
  LOGGER.warn(...);
  return List.of();
}
```

Given the degrade branch is the one with teeth, I'd rather it be reachable only by failures we've positively identified as permanent, with everything unrecognised defaulting to retry. Cheaper to be wrong in the retry direction.

mathias-vandaele:
@polenz I believe comments can be checked/resolve and this can be merged

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -199,23 +220,48 @@ public void moveMessage(EmailMessage msg, Folder folder) {
...
-    var messageBuilder = constructCommonMessage(msg);
-    if (messageBuilder.attachments().count().get() == 0) {
+    // hasAttachments alone is not used as the guard: Graph excludes inline attachments (e.g.
+    // signature images) from that flag, so it would wrongly skip downloading them.
+    // attachmentMetadata alone is not enough either: fetchAttachmentMetadata degrades to an
+    // empty list on a permanent lookup failure too, which must not be mistaken for "no
+    // attachments" when Graph's own flag says otherwise - that would silently drop the email's
+    // real attachments before a later postprocess (mark-read/move/delete) step runs.
+    if (msg.attachmentMetadata().isEmpty() && !Boolean.TRUE.equals(msg.hasAttachments())) {
       return List.of();
```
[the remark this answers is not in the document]
vringar: Fixed. Replaced the `hasAttachments` backstop with an explicit `attachmentMetadataResolved` flag on `EmailMessage`: the download guard now only skips when the metadata lookup actually succeeded and confirmed an empty attachments collection (`attachmentMetadataResolved && attachmentMetadata.isEmpty()`), never based on `hasAttachments` alone. A permanent lookup failure now degrades to `resolved=false`, so this exact double-failure case (inline-only + failed lookup) no longer skips the download.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +280,132 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
-  private static boolean processMessageItem(Message msg, Consumer<EmailMessage> handler) {
-    var myMsg = GraphApiMapper.toEmailMessage(msg, List.of());
+  private boolean processMessageItem(Message msg, Consumer<EmailMessage> handler) {
+    // Resolve lightweight attachment metadata (no content download) before the activation
+    // condition is evaluated, so conditions can filter on attachment properties such as file
+    // type. Not gated on hasAttachments: Graph excludes inline attachments from that flag, so
+    // gating on it would hide inline-only messages from isInline-based conditions.
+    List<EmailAttachmentMetadata> attachmentMetadata;
+    try {
+      attachmentMetadata = fetchAttachmentMetadata(msg.getId());
```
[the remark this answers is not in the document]
vringar: Correct, and intentional: gating this on `hasAttachments` was the original design but caused the inline-attachment bug (see the other thread). Updated the PR description's "Known trade-offs" section to state explicitly that metadata is now fetched for every polled message, not just attachment-bearing ones, as the accepted cost of supporting `isInline` filtering correctly.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +280,132 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+      context.log(
+          activity ->
+              activity
+                  .withSeverity(Severity.WARNING)
+                  .withTag("attachment-metadata-lookup")
+                  .withMessage(
+                      "Could not resolve attachment metadata for email "
+                          + msg.getId()
+                          + " (transient Graph error), will retry on next poll"));
+      return true;
```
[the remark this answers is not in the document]
vringar: Fixed. `processMessageItem` now returns `false` on a `TransientMetadataLookupException`, stopping the current poll cycle instead of continuing to the next message. The current message and the rest of the page are left untouched and retried on the next poll cycle, rather than each one hammering the still-throttled/unavailable endpoint.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +280,132 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+    int status = e.getResponseStatusCode();
+    // 0 means the exception was constructed without an HTTP response at all (e.g. wrapping a
+    // lower-level failure) - treat unknown status the same as throttled/unavailable, since we'd
+    // rather retry a permanent failure than mistake a transient one for "no attachments".
+    return status == 0 || status == 429 || status >= 500;
+  }
+
+  private List<EmailAttachmentMetadata> fetchAttachmentMetadata(String messageId) {
+    try {
+      var client = getClient();
```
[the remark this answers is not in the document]
vringar: Fixed. Added a `GraphServiceClient` cached for the duration of one `poll()` call (`pollClient`, set at the start of `PageIteratorMessageFetcher.poll()` and cleared in a `finally`), reused by every message and attachment request within that poll instead of resolving a fresh client per message. One token exchange per poll cycle again for refresh-token authentication, matching the original intent stated in `createClientSupplier`'s comment.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +280,132 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
-    var myMsg = GraphApiMapper.toEmailMessage(msg, List.of());
+  private boolean processMessageItem(Message msg, Consumer<EmailMessage> handler) {
+    // Resolve lightweight attachment metadata (no content download) before the activation
+    // condition is evaluated, so conditions can filter on attachment properties such as file
+    // type. Not gated on hasAttachments: Graph excludes inline attachments from that flag, so
+    // gating on it would hide inline-only messages from isInline-based conditions.
+    List<EmailAttachmentMetadata> attachmentMetadata;
+    try {
+      attachmentMetadata = fetchAttachmentMetadata(msg.getId());
+    } catch (TransientMetadataLookupException e) {
```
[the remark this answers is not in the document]
vringar: Added poll-level tests driving `processMessageItem` through the real `constructMessageFetcher(...).poll(...)` entry point with a mocked `GraphServiceClient`: one asserts a successful metadata response is correctly mapped onto the handled `EmailMessage` (`attachmentMetadataResolved=true`, metadata populated), and one asserts a transient (429) failure means the handler is never invoked and the poll stops. Also added direct `isTransient` classification tests and a test proving the download guard's non-skip branch is actually reached (mocks the Graph attachments call rather than only inspecting the guard).

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +289,144 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+              .attachments()
+              .get(
+                  requestConfiguration ->
+                      // Selecting only metadata fields ensures Graph does not return the
+                      // (potentially large) contentBytes payload for file attachments.
+                      requestConfiguration.queryParameters.select =
+                          new String[] {"id", "name", "contentType", "size", "isInline"});
+      if (response == null || response.getValue() == null) {
+        return new AttachmentMetadataResolution(List.of(), true);
+      }
```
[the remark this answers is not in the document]
vringar: Fixed. Returns `resolved=false` for a null response/value now, rather than treating it as a confirmed-empty collection.

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -234,9 +289,144 @@ private MessageCollectionResponse fetchMessages(Folder folder, String filterStri
...
+                  .withSeverity(Severity.WARNING)
+                  .withTag("attachment-metadata-lookup")
+                  .withMessage(
+                      "Could not resolve attachment metadata for email "
+                          + msg.getId()
+                          + " (transient Graph error), will retry on next poll"));
+      // Stop this poll entirely rather than continuing to the next message: Graph is throttled or
+      // unavailable right now, so hitting it again for every remaining message in the page only
+      // amplifies the outage. This message and the untouched remainder are retried next poll.
+      return false;
```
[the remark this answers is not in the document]
vringar: Fixed. The transient-failure branch now rethrows (instead of returning false) so it propagates out of poll() to EmailPollingWorker.run(), which already calls context.reportHealth(Health.down(e)) on any uncaught exception. Confirmed empirically that PageIterator.iterate() does propagate an exception thrown from the callback rather than swallowing it.

vringar:
Docs update for `attachmentMetadata` (output mapping, activation condition example, FAQ disambiguation): camunda/camunda-docs#10010

connectors/microsoft/email-inbound/src/main/java/io/camunda/connector/microsoft/email/util/MicrosoftMailClient.java
```diff
@@ -199,23 +223,24 @@ public void moveMessage(EmailMessage msg, Folder folder) {
...
+        fetchAttachmentPage(
+            msg.id(),
+            null,
+            file ->
+                context.create(
+                    DocumentCreationRequest.from(file.getContentBytes())
+                        .fileName(file.getName())
+                        .contentType(file.getContentType())
+                        .build()));
+    return lookup.itemsOrEmpty();
```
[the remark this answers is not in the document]
vringar: Fixed. fetchAttachments() now rejects an Unresolved download result instead of falling through to an empty document list: it throws a ConnectorException rather than letting correlation/postprocessing proceed without the real attachments.

vringar:
@Szik idk if you are still responsible here, but could you give this another test. I've done quite a bit of refactoring and the E2E test is the only thing I trust, but I no longer have access to the appropriate credentials

Szik:
i will have a look later today @vringar

Szik (approved):

Szik:
LGTM

PR #7541, opened 2026-06-17
feat(etg): Configuration (credential) templates in element templates

chillleader:
## Description

Adds **Configuration** (credential) template support to the Element Template Generator (ETG) — the connectors side of the Credentials initiative ([PDP-3396](https://github.com/camunda/product-hub/issues/3396)).

"Configuration" is the generic core-modeling layer; a **credential** is a configuration template with `kind = CREDENTIAL`. With this change an element template can:

- declare a **configuration chooser** — a `{ "type": "Configuration", "configurationTemplate": … }` property the Modeler renders as a picker filtered to compatible configuration instances; and
- embed a **configuration template** — the field definitions a configuration editor (Hub / Modeler) renders — under a top-level `configurationTemplates` key.

New annotation-driven surface (rides `ADR-0001`): `@ConfigurationTemplate` (with `id` / `version` / required non-blank `name` / `kind`), `PropertyType.Configuration`, `@ElementTemplate.configurationTemplates()`, and the DSL/JSON (`ConfigurationProperty`, `ConfigurationTemplate`, the `property` binding, and a `secret` hint).

**Binding model:** the chosen configuration is bound **once, as a whole**, to a connector `configuration` input (`zeebe:input` outbound / `zeebe:property` inbound); the connector reads that object and interprets its structure itself. There is **no** deconstruction into per-field bindings on the element-template side.

### Why, and based on what

- **Design:** [camunda/connections-design](https://github.com/camunda/connections-design) — `CONFIGURATION_TEMPLATE.md`, `CONFIGURATION_ELEMENT_TEMPLATE.md`, `CREDENTIAL_TRANSITION.md`. In particular **GAP-001** replaced an earlier "deconstruct the credential into the connector's input paths" approach with whole-object binding, for decoupling: a configuration can later swap its authentication mechanism without changing the consuming model.
- **Schema:** the emitted templates are validated (draft-07, hermetic test) against the pre-release schema packages tracked in `bpmn-io/internal-docs#1331`:
  - [`@camunda/zeebe-configuration-templates-json-schema@0.2.0-alpha.0`](https://github.com/camunda/element-templates-json-schema/releases/tag/%40camunda%2Fzeebe-configuration-templates-json-schema%400.2.0-alpha.0)
  - [`@camunda/zeebe-element-templates-json-schema@0.43.0-alpha.0`](https://github.com/camunda/element-templates-json-schema/releases/tag/%40camunda%2Fzeebe-element-templates-json-schema%400.43.0-alpha.0)
[...]

PR #8149, opened 2026-07-29
feat(etg): credential-only authentication for OpenAPI and Postman generators

### LINKED ISSUE #8113: etg: OpenAPI and Postman collection generators don't emit the Configuration (credential) chooser for HTTP auth
## Context

The Credentials initiative ([PDP-3396](https://github.com/camunda/product-hub/issues/3396), design in [camunda/connections-design](https://github.com/camunda/connections-design)) added **Configuration** support to the annotation-driven Element Template Generator (`ADR-0004`, #7541): a `{"type":"Configuration", "configurationTemplate": …}` chooser property plus an embedded `configurationTemplates[]` document, so an element template field can bind a whole reusable credential instance instead of inlining auth fields.

#8072 (merged 2026-07-28) extended this to the hand-authored HTTP-family connectors: REST (`HttpJsonRequest`), GraphQL, and HTTP Polling all now expose an `authenticationConfiguration` chooser field typed `RestAuthenticationConfiguration` (`io.camunda.connectors:rest-authentication:1`, in `connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationConfiguration.java`), wrapping the existing sealed `Authentication` union (none/basic/bearer/apiKey/OAuth/OAuth-refresh). The runtime already implements configuration-over-inline precedence for these three connectors.

Related tracking: parent epic [PDP-3433](https://github.com/camunda/product-hub/issues/3433) and sibling task #8088 (credential validation via direct REST) are both part of the same Credentials rollout.

## Problem

The **OpenAPI** and **Postman collection** generators build element templates for the *same* `io.camunda:http-json:1` connector type, but neither emits the Configuration chooser or embeds `configurationTemplates`:

- `element-template-generator/openapi-parser/src/main/java/io/camunda/connector/generator/openapi/OpenApiOutboundTemplateGenerator.java`
- `element-template-generator/postman-collections-parser/src/main/java/io/camunda/connector/generator/postman/PostmanCollectionOutboundTemplateGenerator.java`

Both parse the spec's/collection's security scheme into an `HttpAuthentication` value and hand it to the shared builder:

```java
HttpOutboundElementTemplateBuilder.create(...)
    ...
    .authentication(authentication)   // List<HttpAuthentication>
```

`HttpOutboundElementTemplateBuilder.build()` (`element-template-generator/http-dsl/.../HttpOutboundElementTemplateBuilder.java:134`) only calls `PropertyUtil.authPropertyGroup(...)`, which emits **inline** `authentication.*` `zeebe:input` properties (`element-template-generator/http-dsl/.../PropertyUtil.java`, using the property prefabs in `HttpAuthentication.getPropertyPrefabs`). There is no code path in this builder that adds a `ConfigurationProperty` chooser or an embedded configuration template.

Net effect: a template hand-generated for REST already lets a user bind a reusable REST-authentication credential, but the *same connector type*, when its template is produced by importing an OpenAPI spec or a Postman collection, does not — the generated template is credential-incapable even though `HttpJsonRequest` (the connector actually executing it) already reads and prefers a bound `authenticationConfiguration`.

## Proposed change

Extend `OpenApiOutboundTemplateGenerator` and `PostmanCollectionOutboundTemplateGenerator` (most of the logic likely belongs in the shared `HttpOutboundElementTemplateBuilder`/`PropertyUtil`, since both generators route through it) so generated templates can also carry:

1. A `ConfigurationProperty` chooser, mirroring the hand-authored REST template exactly:
   ```json
   {
     "id": "authenticationConfiguration",
     "label": "Authentication credential",
     "optional": true,
     "group": "authentication",
     "binding": { "name": "authenticationConfiguration", "type": "zeebe:input" },
     "type": "Configuration",
     "configurationTemplate": "io.camunda.connectors:rest-authentication:1"
   }
   ```
2. The embedded `configurationTemplates[]` entry for `io.camunda.connectors:rest-authentication:1` (`kind: CREDENTIAL`) — reuse the existing extraction the annotation-driven generator already has for `@Configuration`-annotated classes (see `ClassBasedTemplateGenerator`) rather than re-deriving it from the OpenAPI/Postman security scheme, since the runtime-side shape (`RestAuthenticationConfiguration`) is now the canonical one for this connector type.
3. Keep the existing inline `authentication.*` properties as-is (fallback path) — non-breaking, per ADR-0004's "stay non-breaking" requirement and consistent with how REST/GraphQL/HTTP Polling retained inline auth alongside the chooser.

**Feasibility note:** this doesn't add a new module dependency. `http-dsl`'s own `HttpAuthentication.java` already imports `io.camunda.connector.http.base.model.auth.{BasicAuthentication, BearerAuthentication, ApiKeyAuthentication, OAuthAuthentication, NoAuthentication}` — i.e. `http-dsl` already depends on `http-base`, the module `RestAuthenticationConfiguration` lives in. Referencing `RestAuthenticationConfiguration` from the same builder is a same-dependency-edge addition, not a new one.

Gate the new output behind an opt-in switch so this doesn't change output for existing callers/snapshots by default — `GeneratorConfiguration` already has a `Map<GenerationFeature, Boolean> features` extension point (`element-template-generator/core/.../GeneratorConfiguration.java`) used for similarly-scoped toggles (`INBOUND_DEDUPLICATION`, `SYNCHRONOUS_RESPONSE`, …); adding e.g. `GenerationFeature.CREDENTIAL_CHOOSER` there is the natural fit and keeps both generators consistent.

## Non-goals

- Deriving a *bespoke* configuration/credential schema per OpenAPI security scheme or per Postman collection. Both generators target `io.camunda:http-json:1`, so v1 should just reference the single existing `io.camunda.connectors:rest-authentication:1` configuration template — the same one REST/GraphQL/HTTP Polling already use.
- Any runtime/connector-side change — `HttpJsonRequest` already consumes `authenticationConfiguration` with the right precedence; this is purely an ETG output gap.

## Acceptance criteria

- [ ] OpenAPI generator can emit the `authenticationConfiguration` chooser + embedded `configurationTemplates` entry for `io.camunda.connectors:rest-authentication:1`.
- [ ] Postman collection generator does the same.
- [ ] Inline `authentication.*` properties are still emitted unchanged (non-breaking default output, or documented default flip — team to decide).
- [ ] Emitted templates validate against the configuration-templates JSON schema, same conformance check added in #7541 (`@camunda/zeebe-configuration-templates-json-schema`).
- [ ] Existing dry-run/golden-file tests updated:
  - `element-template-generator/openapi-parser/src/test/java/io/camunda/connector/generator/openapi/OpenApiGenerationSourceTest.java`
  - `element-template-generator/postman-collections-parser/src/test/java/io/camunda/connector/generator/postman/PostmanCollectionsGeneratorDryRunExampleTest.java`
- [ ] ADR-0004 updated to note the OpenAPI/Postman generators are now in scope (it currently frames the work purely in terms of the annotation-driven path, ADR-0001).

## References

- ADR: `docs/adr/ADR-0004-configuration-templates-in-element-templates.md`
- #7541 — Configuration (credential) templates in element templates (annotation-driven ETG)
- #8072 — migrate GraphQL and HTTP Polling to Configuration credentials (2026-07-28, REST-family precedent)
- #8088 — sibling Credentials-rollout task (validation), parent epic PDP-3433
- `connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationConfiguration.java`
- `connectors/http/rest/src/main/java/io/camunda/connector/http/rest/model/HttpJsonRequest.java`

chillleader:
## Description

The OpenAPI and Postman collection generators build element templates for the same `io.camunda:http-json:1` connector type as the hand-authored REST template, but emitted only inline `authentication.*` `zeebe:input` properties — no Configuration (credential) chooser. So a template hand-authored for REST let you bind a reusable credential, while the *same connector type* generated from an imported spec did not, even though `HttpJsonRequest` already reads and prefers `authenticationConfiguration`.

This makes credential binding the **default** for both generators, and retains the inline path behind an opt-out flag.

### Design: credential-only, default on

The issue originally proposed emitting the chooser *alongside* retained inline auth, gated default-off. That was rejected during planning ([plan comment](https://github.com/camunda/connectors/issues/8113#issuecomment-5117854694)) — generated templates don't need both. Instead:

- **Default**: the `authentication` group contains exactly one `Configuration` chooser referencing the shared `io.camunda.connectors:rest-authentication:1` template, plus the embedded `configurationTemplates[]` entry. No inline `authentication.*` properties at all.
- **Opt-out**: `GenerationFeature.LEGACY_INLINE_AUTHENTICATION` restores today's inline output verbatim.

The flag is deliberately named for the *legacy* path, so absent ⇒ false ⇒ new behavior. That keeps the existing `== Boolean.TRUE` read idiom used by the other generation features rather than inventing `!= Boolean.FALSE` for one flag, and names the inline path as the one slated for removal.

The two paths are kept strictly separate — `PropertyUtil.authConfigurationPropertyGroup` is self-contained and shares no helper with the legacy fingerprint-grouping code; `authPropertyGroup` is untouched apart from a javadoc note. `build()` branches once.

**⚠️ This changes the authentication section of every OpenAPI/Postman-generated template by default.** That is the intended reversal of the issue's own AC#3.

### Credential-only is runtime-safe

With no inline properties, an unbound chooser leaves the `authentication` input absent, i.e. null. That is well-defined, not a crash: `HttpCommonRequest.authentication` carries only `@Valid` (no `@NotNull`), and `AuthenticationMapper.map(null)` returns `NoAuthentication`. No runtime/connector-side change is included or needed.

### Chooser is required when the source declares mandatory auth

Unlike the hand-authored REST chooser (always optional, because REST keeps an inline fallback), this chooser is required — `optional(false)` + `notEmpty` — when auth is mandatory. Rule: compute effective auth per operation exactly as `authPropertyGroup` does (`authenticationOverride()` when non-empty, else the global list), then required iff no effective list contains a `NoAuth`:

| Source | Chooser |
| --- | --- |
| declares only e.g. `bearer` | **required** |
| declares no security scheme (`SecurityUtil` returns `NoAuth + Basic + Bearer + OAuth2 + ApiKey`) | optional |
| `security: [{}, {bearer: []}]` — auth genuinely optional | optional |
| an operation override permits no auth | optional |
| empty auth list (nothing declared) | optional |

A template imported from an auth-requiring spec therefore shows a validation error until a credential is bound. That is the intended signal; the alternative is a template that silently issues unauthenticated requests and 401s at runtime.

### Accepted trade-offs

- **Spec-derived pre-fills are discarded.** A reusable credential can't be pre-filled from the spec, so OAuth token endpoint + scopes and API key name + location are folded into the chooser's `tooltip` instead of being lost silently.
- **Per-operation auth collapses to one credential.** Specs varying auth per operation previously emitted conditional auth groups; the distinct schemes are now surfaced in the tooltip.
- **No degraded path** on a Modeler/Hub without Configuration support — which is what the opt-out flag is for.

### Changes

1. **core** — extracted `ConfigurationTemplateUtil.fromAnnotatedClass(Class, TemplateGenerationContext)` out of the private `ClassBasedTemplateGenerator.buildConfigurationTemplates`, preserving validation and error messages. Needed because the public `extractConfigurationTemplatePropertiesFromType` returns only `List<PropertyBuilder>` — the `ConfigurationTemplate` wrapper was private, so http-dsl would otherwise have copy-pasted it and drifted. Now both paths share one construction site.
2. **core** — added `GenerationFeature.LEGACY_INLINE_AUTHENTICATION`.
3. **http-dsl** — `PropertyUtil.authConfigurationPropertyGroup` + `isAuthenticationRequired` + tooltip helpers; `HttpOutboundElementTemplateBuilder.legacyInlineAuthentication(boolean)` and the branch in `build()`. No new main-scope module dependency: http-dsl already depended on `connector-http-base`, where #8072 moved `RestAuthenticationConfiguration`.
4. **openapi-parser / postman-collections-parser** — thread the flag from `GeneratorConfiguration`.
5. **congen-cli** — repeatable `-f/--feature <NAME>` on `ConGen`, validated via `GenerationFeature.valueOf(...)`. Without this the flag would be unreachable: `generatorConfiguration()` hardcoded `Map.of()` with a `// todo: do we need to support feature overrides from the CLI?`, and no pom drives these two generators (congen-cli is their only consumer).
6. **ADR-0004** — records that the OpenAPI/Postman generators are now in scope.

### Verification

- `element-template-generator` core 211, http-dsl 42, openapi-parser 50, postman-collections-parser 34, congen-cli 2 — all green.
- **Refactor safety gate**: regenerated the hand-authored templates via `mvn -pl connectors/http/rest,connectors/http/graphql,connectors/http/polling process-classes` and confirmed `git status connectors/` is empty — the step-1 core refactor shifts no existing template. (Note the generator mojo binds to `process-classes`, not `compile`.)
- Test coverage includes a **equivalence assertion** that the embedded `ConfigurationTemplate` matches what `ClassBasedTemplateGenerator` emits for a `@Configuration`-annotated class, plus a legacy-path regression guard and all five optionality cases.

### Notes for reviewers

- `connector-core` was added to `http-dsl/pom.xml` in **test scope only** (the equivalence test needs `@OutboundConnector`/`OutboundConnectorFunction`; `analyze-only` requires the declaration). Main scope is unchanged.
- The chooser's `configurationTemplate` id is a literal in `PropertyUtil` rather than read from the annotation, since reading it would require `connector-core` on http-dsl's *main* classpath. A test pins the embedded template's id to the same string, so changing `@Configuration(id=…)` fails the build.
- Two `PropertyUtilTest` cases now opt into `legacyInlineAuthentication(true)`; `builder_conditionalAuth_*` genuinely changes behavior under credential-only, since a single chooser is always unconditional and group ordering flips.
- The generators' existing dry-run tests (`ExampleTest`, `PostmanCollectionsGeneratorDryRunExampleTest`) assert nothing beyond `isNotNull()` and there are no golden files in either parser module, so they were left alone rather than treated as coverage.

### Follow-ups (not in this PR)

- Remove the inline path and `LEGACY_INLINE_AUTHENTICATION` once Modeler/Hub credential support is GA.
- Web Modeler must pass the feature configuration and confirm it renders `type: "Configuration"` on OpenAPI/Postman import — there is no inline fallback anymore.

## Related issues

closes #8113

Part of the Credentials rollout: follows #7541 (annotation-driven Configuration templates) and #8072 (REST/GraphQL/HTTP Polling migration).

## Checklist


mathias-vandaele (approved):
thank you @chillleader 

Concerning copilot comment, this is better to put "_" instead of ignored, this is idiomatic, non blocking though

PR #8150, opened 2026-07-29
fix(inbound): retry publishing a lost process state change

### LINKED ISSUE #8148: Inbound connectors permanently fail to activate after a transient Orchestration Cluster outage (lost ProcessStateChanged event)
## Summary

A transient failure to reach the Orchestration Cluster REST API **permanently** prevents affected inbound connectors from being activated, until the connectors pod is restarted. The `ProcessStateChanged` signal is consumed exactly once and lost if publication fails, and nothing ever re-derives it.

Reported via [SUPPORT-33961](https://jira.camunda.com/browse/SUPPORT-33961) / INC-6835 (L2 — Critical). Customer: C8 SM 8.9.13, Helm on AWS, external RabbitMQ, Elasticsearch secondary storage.

**Symptom as observed by the customer:** during Zeebe pod rotation (worst during rolling updates), some — not all — RabbitMQ inbound connectors stop consuming from their queues and process instances get stuck. Restarting the connectors pod mitigates it.

## Root cause

`ProcessStateManagerImpl.update()` commits the state transition **before** the event that acts on it is published, and swallows publication failures.

https://github.com/camunda/connectors/blob/main/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateManagerImpl.java#L52-L66

```java
public void update(ImportResult processDefinitions) {
  StateUpdateResult result = processStateContainer.compareAndUpdate(processDefinitions); // (1) COMMITS
  if (result.isEmpty()) {
    LOG.debug("No process state changes detected");
    return;                                                                             // (4) forever after
  }
  for (var entry : result.affectedProcesses().entrySet()) {
    publishProcessStateChangedEvent(processRef, activeVersionKeys);                      // (2) may fail
  }
}
```

The sequence that bricks a connector:

1. `compareAndUpdate` sees a version that is `missingInState`, installs it, and returns `anyStateChanged = true` — the mutation is **already committed** to `processStates` ([`ProcessStateContainerImpl.java:156-161`](connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainerImpl.java#L156-L161)).
2. `publishProcessStateChangedEvent` → `getConnectors` → `ProcessDefinitionInspector.findInboundConnectors` does a cache-miss load that fetches the BPMN model over HTTP. During pod rotation this throws:
   ```
   org.springframework.cache.Cache$ValueRetrievalException: Value for key 'xx' could not be loaded ...
   Caused by: io.camunda.client.api.command.ClientException:
     org.apache.hc.client5.http.HttpHostConnectException:
       Connect to http://camunda-zeebe-gateway:8080 failed: Connection refused
   ```
3. That exception is caught and **only logged** ([`ProcessStateManagerImpl.java:106-113`](connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateManagerImpl.java#L106-L113)). No rollback, no retry, no dirty flag. The event never reaches `InboundExecutableRegistryImpl`.
4. On the next 5 s poll, the version is now in `presentInBoth`, so `anyStateChanged` stays `false` ([`ProcessStateContainerImpl.java:164-167`](connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainerImpl.java#L164-L167)) → `result.isEmpty()` → early return. **The state store believes the connector is active; the registry never heard about it.**

The registry is purely event-driven — there is no reconciliation loop that re-derives target executable state from `ProcessStateContainer`. So the divergence is permanent for the lifetime of the pod, exactly matching the customer's report.

### Why it affects only *some* queues

Only the processes whose BPMN fetch happens to land inside the connection-refused window lose their event. Everything else activates normally. This is why it "hints that the problem is internal to the Connectors" — the RabbitMQ side is entirely innocent here; the connector was simply never started.

### Why a retry would work

Caffeine does **not** negative-cache a failed load, so the `processDefinitions` cache is not poisoned. A second attempt would succeed. The only thing missing is something to trigger it.

### It is silent

Because the exception is caught *inside* `publishProcessStateChangedEvent`, `ImportSchedulers.pollOnePhysicalTenant` still returns `true` and `ready` stays `true` ([`ImportSchedulers.java:97-113`](connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/importer/ImportSchedulers.java#L97-L113)). The readiness probe does not go unhealthy and no metric reflects the broken state — the single ERROR log line is the only evidence. Operators have no way to know they need to restart.

## Regression scope

| Branch | Status |
|---|---|
| `main` (8.10.0-SNAPSHOT) | **affected** |
| `stable/8.9` | **affected** |
| `stable/8.8` | not affected |
| `stable/8.7` | not affected |

This is a **regression introduced in 8.9** by the state-management refactor in 4bc4463b40 ("feat: fetch active process definitions", #6056, 2026-01-28).

8.8's `ProcessStateStoreImpl` calls `getConnectors(...)` *inside* the `processStates.compute(...)` remapping function. When it throws, `ConcurrentHashMap.compute` propagates the exception **without installing the entry**, so the process stays absent from state, is re-detected as `newlyDeployed` on the next tick, and self-heals. The 8.9 refactor split "commit state" from "publish event" into two steps and lost that atomicity.

## Expected behaviour

A transient Orchestration Cluster outage must not permanently disable an inbound connector. The runtime should converge once the cluster is reachable again, without operator intervention.

## Suggested fix

**Immediate, backportable to `stable/8.9` — make the commit conditional on successful publication.** Either roll back the affected process's state when publication fails, or mark it dirty so the next `compareAndUpdate` re-reports it as changed. `publishProcessStateChangedEvent` must stop swallowing the failure and let `update()` act on it.

Note the ordering constraint: the elements must be fetched *before* the state is committed, since the fetch is the part that fails. Restoring 8.8's atomicity (resolve connectors → then commit → then publish) is the cleanest shape.

**Follow-up hardening (`main`) — add a reconciliation pass.** Periodically re-derive the target executable state from `ProcessStateContainer` and compare it against what the registry actually has running, healing any divergence. This closes the whole class of lost-event bugs rather than this one instance. Worth considering given the registry currently has no self-healing path at all.

**Observability, either way.** A failed state-change publication should be visible — a counter metric and/or a degraded-state indicator, so this is detectable without grepping logs. (Flipping readiness to `false` would surface it, but risks restart loops during a cluster outage; a metric is probably the safer call. Team's judgement.)

## Regression test

Deterministic and does not need a real broker: mock `ProcessDefinitionInspector.findInboundConnectors` to throw on the first invocation and succeed afterwards, then call `ProcessStateManagerImpl.update()` twice with the same `ImportResult`. Assert that `executableRegistry.publishEvent` is called with the `ProcessStateChanged` event. Today the second `update()` short-circuits at `result.isEmpty()` and the event is never published.

## Notes on triage

Marked `severity/high` rather than `mid`: restarting the connectors pod does mitigate it, but the failure is silent, so operators cannot know they need to. Happy to be re-triaged if the team reads the restart as a sufficient workaround.

Suspected duplicates checked — none found in this repo.

---

*Filed from incident channel #inc-support-33961-connectors-fails-to-reestablish-rabbitmq-queue-connection. Root cause analysis is from code inspection, not from a live reproduction — the customer-facing hypothesis in the support ticket (RabbitMQ reconnection logic) is not where the defect is.*

mathias-vandaele:
## Description

A transient Orchestration Cluster outage could permanently disable inbound connectors until the connectors pod was restarted. Reported via [SUPPORT-33961](https://jira.camunda.com/browse/SUPPORT-33961) / INC-6835 (L2 — Critical): during Zeebe pod rotation, some — not all — RabbitMQ inbound connectors stopped consuming and process instances got stuck.

RabbitMQ reconnection is not involved. The affected connectors were never started in the first place, so there was nothing to reconnect.

### The defect

A state transition is only ever reported once. `compareAndUpdate` commits it *before* the `ProcessStateChanged` event is published, and a later import of unchanged data produces no diff:

1. `compareAndUpdate` installs a new version and returns `stateChanged = true` — already committed.
2. Publishing fetches the BPMN model over HTTP, which fails during pod rotation (`Connection refused` to `camunda-zeebe-gateway:8080`).
3. That exception was caught and **only logged**.
4. Next poll: the version is now `presentInBoth`, so no diff → `result.isEmpty()` → early return, forever.

The state store believed the connector was active; the registry never heard about it. The registry is purely event-driven with no reconciliation pass, so the divergence was permanent. Only the processes whose fetch landed inside the outage window were affected — hence "some queues".

It was also invisible: the exception was caught below `pollOnePhysicalTenant`, so `ready` stayed `true` and the readiness probe never went unhealthy. A single ERROR log line was the only evidence, which is why "restart the pod" isn't really a workaround — operators can't tell they need to.

### Changes

The retry state lives in `ProcessStateManagerImpl` — the component that actually failed to publish — so the state container's `compareAndUpdate` diffing logic is untouched.

- **`ProcessStateManagerImpl`** — keeps a set of processes whose event could not be published and retries them on the next poll for their physical tenant, draining on read and re-adding on repeated failure so a persistent outage keeps retrying instead of being dropped once. `publishProcessStateChangedEvent` propagates instead of swallowing, and the handling moved up into the per-process loop so the failure has somewhere to be recorded. Per-process isolation is unchanged, not new — the old `catch` sat inside the publish method, so the loop already continued past a failure.
- **`ProcessStateContainer#getActiveVersions`** (new) — a read-only query, promoted from an existing private method (`private` → `@Override public synchronized`, a 2-line diff). The retry re-reads current versions through it rather than reusing the ones it failed with, so it cannot republish a set that a deployment has since superseded. A change reported by the current import also supersedes a pending retry for the same process.
- **New metric `camunda.connector.inbound.process-state-change.publish-failures`** — a rate that doesn't settle back to zero means connectors are failing to activate.
- **Inbound `README.md`** — documented the retry, which the event-flow description previously omitted.

Caffeine does not negative-cache a failed load, so the `processDefinitions` cache was never poisoned — a retry simply succeeds. The only thing missing was something to trigger it.

### Regression scope

Affects `main` and `stable/8.9` only. Introduced in 8.9 by #6056 (`4bc4463b40`): 8.8's `ProcessStateStoreImpl` resolved connectors *inside* the `processStates.compute(...)` remapping function, so a throw propagated without installing the entry and the process self-healed on the next tick. The refactor split "commit state" from "publish event" and lost that atomicity.

`stable/8.8` and `stable/8.7` do not have this bug and need no backport.

### Reviewer notes

- **Why not just wrap the publish in a Failsafe retry?** A bounded inline retry narrows the window but doesn't close it: a rolling update can outlast any backoff budget, and on exhaustion you're back to the same permanent latch. The queue is what makes it actually converge.
- **Why the container gains a getter at all.** An earlier version of this fix put the retry queue *inside* the container and folded it into `compareAndUpdate`. Same interface-method count, but it added mutable state to the diffing class and edited its hot path — the wrong risk profile for a fix that backports to a maintenance branch. Keeping the queue in the manager and re-reading versions through a getter leaves `compareAndUpdate` byte-for-byte unchanged, and avoids the staleness that remembering versions would introduce.
- **`getActiveVersions` is reentrant-safe.** It is `synchronized` and also called from within the already-`synchronized` `compareAndUpdate`; Java monitors are reentrant.
- **Concurrent polls and event ordering.** `compareAndUpdate` releases the container lock before publishing, so published versions are an earlier snapshot. Two polls overlapping on the same physical tenant could queue snapshots out of order and the stale one would win. That is a pre-existing property of the plain diff path, not something the retry introduces — it publishes outside the lock on `main` today too. It is unreachable under Spring Boot's default single-threaded task scheduler, which serializes the two `@Scheduled` importers, and this repo sets no custom `TaskScheduler` or `spring.task.scheduling.pool.size`.

  I initially added a per-physical-tenant lock here and then removed it: it would have closed the retry path while leaving the diff path three lines away open, which implies an enforced invariant that isn't. The assumption is now stated explicitly in the `ProcessStateManagerImpl` class javadoc, including that `@EnableScheduling` ships in the public starter so an embedding app raising the pool size would break it. Making the runtime genuinely safe under a multi-threaded scheduler is a broader change and belongs with the reconciliation pass below. Fan-out across physical tenants *within* one poll is safe — the retry queue is tenant-scoped, so those threads never contend for the same process.
- **Metric rather than readiness.** Flipping readiness to `false` would surface the degraded state more loudly, but risks restart loops during exactly the cluster outage that triggers this. Happy to change if the team prefers.
- **Reconciliation pass deliberately not included.** This PR fixes the reported bug. Periodically re-deriving target executable state from `ProcessStateContainer` would close the whole class of lost-event bugs and is worth its own PR — the registry currently has no self-healing path at all.

## Related issues

closes #8148

## Checklist

      *release has already been created*. The release branch will be merged back into stable/8.9 later, so the change
      will be included in future 8.9.x releases as well.

  `backport stable/8.9` added. Add `backport release-8.9.7` as well if this should go into that already-created release
  rather than waiting for the next 8.9.x — that is a release-management call. No 8.8/8.7 labels: those branches are not
  affected.

  9 new tests in `ProcessStateManagerImplTest`, using the real `ProcessStateContainerImpl` with a mocked inspector so
  they exercise the actual latch rather than a stand-in. `shouldRetryPublishingOnNextUpdateWhenFetchingConnectorsFailed`
  fails on `main` today. Also covered: repeated failures, one process failing without stranding its batch-mates, the
  registry itself rejecting the event, and cross-physical-tenant scoping, a retry publishing the versions active *now* rather than the superseded ones it failed
  with, and a process undeployed while its retry is still queued (publishes a deactivation, not the old version). Full
  `connector-runtime-spring` suite: 375 tests, 0 failures.

  The inbound runtime `README.md` is updated. The new metric is operator-facing, so it may warrant an entry in
  `camunda/camunda-docs` alongside the other inbound connector metrics — flagging rather than assuming.

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/InboundConnectorRuntimeConfiguration.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/InboundConnectorRuntimeConfiguration.java
@@ -224,7 +224,9 @@ public class InboundConnectorRuntimeConfiguration {
   public ProcessStateManager processStateManager(
       InboundExecutableRegistry registry,
       ProcessDefinitionInspector inspector,
-      ProcessStateContainer processStateContainer) {
-    return new ProcessStateManagerImpl(processStateContainer, inspector, registry);
+      ProcessStateContainer processStateContainer,
+      ConnectorsInboundMetrics connectorsInboundMetrics) {
+    return new ProcessStateManagerImpl(
+        processStateContainer, inspector, registry, connectorsInboundMetrics);
   }
 }
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/README.md
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/README.md
@@ -81,6 +81,15 @@ ImportSchedulers:
 These results are merged into the process state store. If the relevant versions for a process
 change, the runtime publishes a `ProcessStateChanged` event.
 
+A given state transition is only ever reported once: the state store commits it before the event is
+published, so a subsequent import of unchanged data yields no diff. Publishing can fail — it fetches
+the BPMN model from the Orchestration Cluster, which may be briefly unreachable during a rolling
+update — so `ProcessStateManagerImpl` queues any process whose event it could not publish and retries
+it on the next poll, re-reading the active versions from the state store rather than reusing the ones
+it failed with. Without that retry the process would be recorded as active while its connectors were
+never activated, and nothing would ever notice. Failures are counted by
+`camunda.connector.inbound.process-state-change.publish-failures`.
+
 Then the executable store is updated from those events:
 
 ```text
@@ -115,6 +124,7 @@ flowchart TD
     T --> importSchedulers
     S1 & S2 --> OC --> PS
     PS -->|versions changed| EV --> registry
+    EV -->|publish failed, retry next poll| PS
 ```
 
 ### When the listener receives an event, e.g. a new email in the inbox
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainer.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainer.java
@@ -17,7 +17,9 @@
 package io.camunda.connector.runtime.inbound.state;
 
 import io.camunda.connector.runtime.inbound.state.model.ImportResult;
+import io.camunda.connector.runtime.inbound.state.model.ProcessDefinitionRef;
 import io.camunda.connector.runtime.inbound.state.model.StateUpdateResult;
+import java.util.Set;
 
 /**
  * Container for the current process state. It is responsible for comparing the current state with
@@ -37,4 +39,18 @@ public interface ProcessStateContainer {
    *     deactivated
    */
   StateUpdateResult compareAndUpdate(ImportResult importResult);
+
+  /**
+   * Returns the process definition keys (versions) currently active for the given process, without
+   * modifying any state.
+   *
+   * <p>{@link #compareAndUpdate} only reports a given state transition once, so a caller that
+   * failed to act on one cannot obtain it again from a later import. This lets such a caller
+   * re-read the current state instead, so its retry acts on what is active now rather than on a
+   * remembered — and possibly superseded — set of versions.
+   *
+   * @param processRef the process definition to look up
+   * @return the active process definition keys, or an empty set if the process is not tracked
+   */
+  Set<Long> getActiveVersions(ProcessDefinitionRef processRef);
 }
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainerImpl.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateContainerImpl.java
@@ -105,7 +105,8 @@ public class ProcessStateContainerImpl implements ProcessStateContainer {
   }
 
   /** Returns the set of currently active process definition keys for a given process definition. */
-  private Set<Long> getActiveVersions(ProcessDefinitionRef processRef) {
+  @Override
+  public synchronized Set<Long> getActiveVersions(ProcessDefinitionRef processRef) {
     var versionsInState = processStates.get(processRef);
     if (versionsInState == null) {
       return Set.of();
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateManagerImpl.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/inbound/state/ProcessStateManagerImpl.java
@@ -24,13 +24,33 @@ import io.camunda.connector.runtime.inbound.executable.InboundExecutableRegistry
 import io.camunda.connector.runtime.inbound.state.model.ImportResult;
 import io.camunda.connector.runtime.inbound.state.model.ProcessDefinitionRef;
 import io.camunda.connector.runtime.inbound.state.model.StateUpdateResult;
+import io.camunda.connector.runtime.metrics.ConnectorsInboundMetrics;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Turns imported process data into {@code ProcessStateChanged} events.
+ *
+ * <p><b>Assumes the two {@code @Scheduled} importers in {@code ImportSchedulers} do not
+ * overlap.</b> {@code compareAndUpdate} releases the container's lock before we publish, so the
+ * versions we publish are a snapshot taken earlier. Two polls running concurrently for the same
+ * physical tenant could therefore queue snapshots out of order, and because the executable registry
+ * applies events in queue order, an older snapshot arriving last would win with no diff left to
+ * correct it. This holds for the retry path in this class and equally for the plain diff path,
+ * which has always published outside the lock.
+ *
+ * <p>Spring Boot's default single-threaded task scheduler serializes those importers, which is what
+ * makes the assumption hold today; nothing here enforces it. Note that {@code @EnableScheduling}
+ * ships in the public connectors starter, so an embedding application that raises {@code
+ * spring.task.scheduling.pool.size} would break it. Fan-out across physical tenants within a single
+ * poll is safe — {@code pendingRetries} is tenant-scoped, so those threads never contend for the
+ * same process.
+ */
 public class ProcessStateManagerImpl implements ProcessStateManager {
 
   private static final Logger LOG = LoggerFactory.getLogger(ProcessStateManagerImpl.class);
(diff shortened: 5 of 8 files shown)


mathias-vandaele:
Thanks — went through all four points. Three are fixed; on the race I landed somewhere different from the suggestion, so flagging the reasoning explicitly.

**Race between the two `@Scheduled` pollers — documented, not locked.**

The analysis is right: the atomic `remove` from `failedToPublish` doesn't cover the subsequent read-versions-and-publish, and it's non-triggering only because Spring Boot's default task scheduler is single-threaded. Confirmed there's no custom `TaskScheduler` bean and no `spring.task.scheduling.pool.size` anywhere in the repo.

I implemented the suggested per-physical-tenant lock, then backed it out. The reason: **this ordering hazard already exists on `main` in the non-retry path.** `compareAndUpdate` releases the container lock and returns a snapshot, then publishing happens outside the lock — so two overlapping polls can queue snapshots out of order and the stale one wins, with no diff left to correct it. Same silent divergence, no retry involved:

```java
StateUpdateResult result = processStateContainer.compareAndUpdate(processDefinitions); // lock released here
for (var entry : result.affectedProcesses().entrySet()) {
  publishProcessStateChangedEvent(processRef, activeVersionKeys);                      // snapshot from earlier
}
```

Locking only the retry path would have closed one instance while leaving the one three lines above it open — implying an enforced invariant that isn't there, which I think is worse than leaving both visible. So instead the assumption is now stated in the `ProcessStateManagerImpl` class javadoc: that the two importers must not overlap, that this holds today only by virtue of the default scheduler, and specifically that `@EnableScheduling` ships in the public connectors starter, so an embedding application raising the pool size would break it.

Making the runtime genuinely safe under a multi-threaded scheduler is a broader change than this PR, and largely subsumed by the reconciliation pass already noted as follow-up — a pass that re-derives target state would heal a stale-snapshot divergence too, which no amount of locking in this class does. Happy to add the lock anyway if you'd rather have belt-and-braces, but I didn't want to ship the appearance of a guarantee.

Fan-out across physical tenants *within* a single poll is safe regardless: the retry queue is tenant-scoped, so those virtual threads never contend for the same process.

**Overstated description — corrected.** You're right, and thanks for catching it. Per-process isolation already existed: the old `catch` sat inside `publishProcessStateChangedEvent`, so the loop already continued past a failure. Moving it into the loop only gives the failure somewhere to be recorded. Description and commit message both reworded.

**Null-check inconsistency — fixed.** Dropped the `Objects.requireNonNull` on `metrics`; the other three constructor params aren't checked and these are Spring-injected, so the check bought nothing.

**Undeploy-while-retry-pending — test added.** `shouldPublishDeactivationWhenProcessWasUndeployedWhileRetryPending`: fail on v1, then import without the process, assert a deactivation (empty version map) is published rather than v1 being resurrected. Correct by inspection as you say, now pinned.

Suite is at 375 tests, 0 failures.

chillleader (approved):

Szik (approved):

PR #8171, opened 2026-07-30
feat: help tooltip for connector response/output mapping

### LINKED ISSUE #154: Help tooltip for the response/output mapping in connectors

The response mapping in Connector templates/tasks is not self-explaining. A new user might drop off here, since it's completely different to how it works in other tasks, and there is no in-app guide. The UI is lacking following crucial information:
* Name of the variables (`status`, `header`, `body`) that they can map
* How to write the FEEL context to "destructure"/map the response

Besides that, the mapped variables are not available in variable autocompletion in the process after declaration with this FEEL context.

Provide autocompletion for the variables, as well as helpful tooltips. 

connector-properties-2

Related to https://github.com/bpmn-io/properties-panel/issues/202

Might even be a better (but major breaking) solution:

Use output variable mapping as people are already used to, and provide the response variables (`status`, `header`, `body`) in the autocompletion there.

connector-properties

__Additional context__

johnBgood:
## Description

Adds an in-app help tooltip to the connector "Result expression" (output mapping) property in generated element templates, addressing the lack of any in-modeler hint about available response variables or FEEL destructuring syntax.

Mechanism (generic, in `element-template-generator`):
- `@DataExample` gains a `DEFAULT_ID` constant, usable as an opt-in marker for whichever example should be treated as canonical when a class declares more than one. `id()`'s own default stays `""` — unchanged — since this annotation has existed since 2022 and external connector authors may already rely on that default in their own doc-template lookups.
- `ClassBasedDocsGenerator.resolvePrimaryExampleData` resolves the example explicitly marked `DEFAULT_ID`, falling back deterministically (sorted by method name) to the first declared example when none is marked.
- The resolved example is rendered into the generated `resultExpression` property's tooltip (merged with, never overwriting, any tooltip already present), using only HTML tags already proven safe elsewhere in this repo's committed templates (`<div>`, `<p>`, `<code>`, no `<pre>`).
- Every example's FEEL expression must itself be a valid Result expression — i.e. evaluate to a JSON object, since `ConnectorResultHandler` rejects scalar results at runtime. Wired accordingly for REST, Webhook, and AI Agent, plus the test fixtures.

Wired for three connectors whose `outputDataClass` already carries (or now carries) a `@DataExample`:
- **REST** (`HttpCommonResult`) — no version change; the tooltip is patched into the existing template.
- **Webhook** — new `WebhookOutputExample` holder class, shaped to match the connector's actual runtime FEEL-binding context (`request.body...`), not just what evaluates cleanly at doc-generation time — no version change.
- **AI Agent Task v1/v2** (and their derived Sub-process v1/v2 templates) — `AgentResponse`'s existing "text" example is now marked as the canonical one via `DEFAULT_ID` — no version change.

Per review feedback, none of the three connectors' template versions are bumped: the tooltip is a cosmetic, non-behavioral addition, so it's patched into the existing template revision rather than shipped as a new version with an archived `versioned/` snapshot of a version that never released.

`JdbcResponse` also has an `outputDataClass` but no `@DataExample`, confirmed unaffected — the mechanism is correctly opt-in.

## Related issues

closes #154

## Checklist

## Test plan

- New unit tests for `DataExample.DEFAULT_ID` resolution (default-id preference, deterministic fallback, empty case) and for the tooltip-merge output-group builders.
- New outbound/inbound generator fixtures + tests asserting the `resultExpression` tooltip is populated when `outputDataClass` carries a `@DataExample`, and stays `null` when it doesn't (regression check).
- Full reactor build (`./mvnw clean test -Dquickly -T 1C`) run locally: every module this PR touches passed, including the element-templates-vs-Java-source drift check CI enforces.


element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
```diff
@@ -307,12 +312,55 @@ private List<PropertyGroup> addServiceProperties(
...
+    var tooltip =
+        new StringBuilder("<div><p>Example response:</p><code>")
+            .append(escapeHtml(compactJson(example.json())))
+            .append("</code>");
+    if (StringUtils.isNotBlank(example.feel())) {
+      tooltip
+          .append("<p>Example FEEL expression: <code>")
+          .append(escapeHtml(example.feel()))
+          .append("</code> -&gt; <code>")
+          .append(escapeHtml(compactJson(example.feelResultJson())))
```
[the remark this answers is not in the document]
johnBgood: Confirmed against `ConnectorResultHandler.parseJsonVarsAsTypeOrThrow` — a scalar result expression is rejected at runtime (`"Result expression must return a JSON object"`). Fixed in 3dc0421b76: every wired example's `feel` now returns an object literal (e.g. `= { orderId: body.order.id }`), and templates were regenerated.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/ClassBasedTemplateGenerator.java
```diff
@@ -307,12 +312,55 @@ private List<PropertyGroup> addServiceProperties(
...
+          .append("</code></p>");
+    }
+    return tooltip.append("</div>").toString();
+  }
+
+  private static String compactJson(String json) {
+    return json == null ? "" : json.replaceAll("\\s+", " ").trim();
+  }
```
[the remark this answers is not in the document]
johnBgood: Good catch — switched `compactJson` to a Jackson tree round-trip (parse then `toString()`) instead of a whitespace regex, so string content is preserved verbatim. Fixed in 3dc0421b76.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/AgentTaskV1Function.java
```diff
@@ -38,7 +38,7 @@
     documentationRef =
         "https://docs.camunda.io/docs/8.10/components/connectors/out-of-the-box-connectors/agentic-ai-aiagent-task/",
     engineVersion = "^8.10",
-    version = 11,
+    version = 12,
```
maff: @johnBgood Let's please not bump the template version for this cosmetic change (at least for agentic AI). I think it's fine to patch it into the latest version as they are 8.10 only.

johnBgood: Done in 3c03468904 — reverted AgentTaskV1Function/V2Function back to version 11/1 and patched the tooltip into those templates directly (also removed the versioned/ archives that were only created because of the now-reverted bump). Left REST (17→18) and Webhook (15→16) version bumps as-is for now since you scoped this to agentic AI — happy to apply the same "patch in place" treatment there too if you'd like.

johnBgood: Done in a9a56a39ef — reverted REST (18->17) and Webhook (16->15) the same way, patched into the existing templates in place, and removed the now-unneeded versioned/ archives for those never-shipped versions. All three connectors now leave their template version untouched.

johnBgood:
Addressed Copilot's latest low-confidence suggestions (8cd474477f):

- Test fixtures (`MyConnectorOutput`, `MyConnectorResult`) had the same scalar-FEEL issue already fixed in the production examples — updated to evaluate to a JSON object too, consistent with what an actual Result expression requires at runtime.
- The version-bump suggestions for REST/Webhook/AI Agent are stale — those are intentionally *not* bumped anymore per @maff's feedback above; the PR description has been updated to reflect that.
- Kept the raw `ObjectMapper` for tooltip JSON compacting as-is: it's only used to parse-and-reprint already-valid JSON (`JsonNode.toString()`), which doesn't go through the writer configuration `ConnectorsObjectMapperSupplier` sets up anyway, so switching wouldn't change behavior — just adds an unnecessary dependency on the heavier shared mapper for a purely local operation.

johnBgood:
One more round addressed (c8cdb175e7): Copilot correctly caught that changing \`DataExample.id()\`'s own default from \`""\` to \`DEFAULT_ID\` is a real compatibility risk — \`@DataExample\` has existed since 2022 in this published module, so external connector authors relying on the old \`""\` default in their own doc-template lookups could break on upgrade. Fixed by keeping \`id()\`'s default at \`""\` and only exposing \`DEFAULT_ID\` as an opt-in marker (set explicitly on whichever example should be canonical, as already done for the three wired connectors). Verified zero diff in every regenerated template.

Left \`OUTPUT_GROUP_OUTBOUND\`/\`OUTPUT_GROUP_INBOUND\` as a clean rename rather than adding deprecated shims — unlike \`@DataExample\` (an author-facing annotation meant to be used directly by connector authors), these are internal DSL builder plumbing consumed only by this generator's own internals and one in-repo builder (\`HttpOutboundElementTemplateBuilder\`); there's no japicmp/revapi gate in this repo and no realistic external caller hand-assembling property groups this way.

johnBgood:
Two more low-confidence suggestions from Copilot:

- **PR description stale claim** — fixed, the description said \`id()\` "now defaults to" \`DEFAULT_ID\`, which was true of an earlier revision but not the final state (kept at \`""\`, see the c8cdb175e7 fix above). Corrected.
- **AgentResponse's "text" → "default" rename** — checked whether this has the same external-breakage shape as the annotation-default issue: it doesn't. \`exampleData["default"]\` is a Pebble build-time lookup key that never appears in the generated \`AI_AGENT.md\` output, and \`AgentResponse\` is Camunda's own connector implementation, not a class external authors extend or annotate — so the only consumer of that id string is \`AI_AGENT.peb\`, which is in this same repo and already updated in this same PR. No external contract to break. Left as-is.

Szik (approved):

chillleader (approved):

PR #8178, opened 2026-07-30
refactor(agentic-ai): migrate Google Vertex AI provider to google-genai SDK to enable Gemini 3 models

### LINKED ISSUE #8177: AI Agent task: Vertex AI "global" region builds wrong hostname (404 on global-aiplatform.googleapis.com)
## Description

A user reported while trying to use `gemini-3.5-flash-lite` on an AI Agent task:

- Regional (US/Europe) request for the new model: `NOT_FOUND` (expected - model not available in that region).
- `global` region request: `404`, with the client hitting `global-aiplatform.googleapis.com` instead of `aiplatform.googleapis.com`.

2.5 models work fine in regional mode; only `global`-only models (like `gemini-3.5-flash-lite`) are affected, since they can't be reached through a regional endpoint at all.

## Root cause

Three layers deep, and not Camunda-specific:

1. Google's own `google-cloud-vertexai` SDK (`VertexAI.java`) always builds the hostname as `"%s-aiplatform.googleapis.com".formatted(location)`, with no special case for `location = "global"`.
2. langchain4j's `VertexAiGeminiChatModel` inherits this unchanged.
3. The connector's `GoogleVertexAiChatModelProvider` passes `region` straight through to `.location(...)` and never overrides `.apiEndpoint(...)`.

Additionally, unlike Bedrock/Anthropic/Azure OpenAI/OpenAI-compatible, the Vertex AI provider's element template has no endpoint-override field, so there is no config-only workaround today.

## Resolution

Google has stopped maintaining the affected SDK (`google-cloud-vertexai`) in favor of the newer `google-genai` SDK, which handles the `global` region correctly. The fix lands via the migration to `langchain4j-google-genai` / `com.google.genai:google-genai` (see linked PR).

## Related issues

Related to #6064 (SDK deprecation/migration tracking).

nikonovd:
## Description

Migrates the Google Vertex AI provider from the sunset `langchain4j-vertex-ai-gemini` / `google-cloud-vertexai` stack to `langchain4j-google-genai` / `com.google.genai:google-genai`.

- Swaps the SDK in `ChatModelFactoryImpl`, adds `endpoint`/`timeouts` config, converts `Float` model parameters to `Double`.
- Bumps the AI Agent element template to v8.
- Fixes service-account credential scoping: `com.google.genai.ApiClient` only auto-scopes ADC it resolves itself, so user-supplied service-account credentials need `cloud-platform` applied explicitly, otherwise every service-account call fails with `invalid_scope`.
- Fixes Gemini 3 tool calling: the model requires its `thoughtSignature` echoed back on the next request. langchain4j carries this on `AiMessage.attributes()` keyed by tool call ID, but the shared `ChatMessageConverterImpl` didn't round-trip it through conversation history, so the second model call in any tool-calling turn 400'd.
- Hardens how `AiMessage.attributes()` is persisted: it's a generic, provider-populated map that was being dumped verbatim into the `AssistantMessage` process variable regardless of provider - a potential data leak for any provider/gateway that puts sensitive data in there. Replaced with an explicit per-provider decorator (`AssistantMessageMetadataDecorator`); only Google Vertex AI has one (scoped to `thought_signature_*` entries), every other provider drops attributes entirely. Attributes now live under their own `provider` metadata key, separate from the `framework`/`ChatResponseMetadata` key.
- Adds a manual, credential-gated acceptance test (`GoogleVertexAiRealModelManualTests`, excluded from CI) covering regional vs. global region, both credential modes, tool calling, and the model-unavailable-in-region failure mode.

**Behavior changes:**
- Retries removed (`maxRetries(0)` + `HttpRetryOptions.attempts(1)`) - transient 429/503 that previously self-healed now surface as incidents.
- ADC resolved eagerly at client construction, not on first call - surfaces as a `ConnectorInputException` at model-creation time instead of at call time.
- The 15s connect timeout is gone - `google-genai` relies solely on `callTimeout`; a black-holed TCP connect now hangs for the full call timeout.
- Thought signatures are persisted as base64 under `metadata.provider` on assistant messages, so conversation history grows slightly and is user-visible.

## Related issues

closes #8177

## Checklist


connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
```diff
@@ -222,15 +231,59 @@ protected CloseableChatModel createGoogleVertexAiChatModel(
...
+  private GoogleCredentials createGoogleServiceAccountCredentials(
       ServiceAccountCredentialsAuthentication sac) {
     try {
+      // Credentials read from a key file carry no scopes. google-genai only scopes the application
+      // default credentials it resolves itself and passes these through verbatim, so without this
+      // the token request fails with invalid_scope.
       return ServiceAccountCredentials.fromStream(
-          new ByteArrayInputStream(sac.jsonKey().getBytes(StandardCharsets.UTF_8)));
+              new ByteArrayInputStream(sac.jsonKey().getBytes(StandardCharsets.UTF_8)))
+          .createScoped(GOOGLE_CLOUD_PLATFORM_SCOPE);
```
nikonovd, line 303: 💡 this was done by `langchain4j-google-vertexai` beforehand and now needs to be done by ourselves.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
```diff
@@ -222,15 +231,59 @@ protected CloseableChatModel createGoogleVertexAiChatModel(
...
+            // genai's retry interceptor defaults to 5 attempts with backoff up to 60s, which
+            // would stack on top of langchain4j's maxRetries.
+            .retryOptions(HttpRetryOptions.builder().attempts(1).build());
+
+    // Applied only when explicitly configured. Deliberately NOT deriveTimeoutSetting(...), which
+    // would impose the default timeout on existing processes that run unbounded today.
+    Optional.ofNullable(connection.timeouts())
+        .map(TimeoutConfiguration::timeout)
+        .filter(Duration::isPositive)
+        .ifPresent(timeout -> httpOptions.timeout((int) timeout.toMillis()));
```
nikonovd: 🔧 apply the default timeout if not configured, as for the other providers

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
```diff
@@ -204,15 +213,15 @@ private void applyBedrockModelParametersIfPresent(
...
 
-    if (connection.authentication() instanceof ServiceAccountCredentialsAuthentication sac) {
-      builder.credentials(createGoogleServiceAccountCredentials(sac));
-    }
+    final var builder =
+        GoogleGenAiChatModel.builder()
+            .client(client)
+            .modelName(connection.model().model())
+            // google-genai brings its own retry interceptor, which is disabled below. Disable
+            // langchain4j's retries as well so the configured timeout is an exact upper bound.
```
nikonovd: 🔧 remove the comment

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
```diff
@@ -222,15 +231,59 @@ protected CloseableChatModel createGoogleVertexAiChatModel(
...
+    // client ourselves and close that instead - otherwise every job execution leaks the OkHttp
+    // dispatcher and connection pool.
+    return new CloseableChatModelDelegate(builder.build(), client);
+  }
+
+  private Client createGoogleGenAiClient(GoogleVertexAiConnection connection) {
+    final var httpOptions =
+        HttpOptions.builder()
+            // genai's retry interceptor defaults to 5 attempts with backoff up to 60s, which
+            // would stack on top of langchain4j's maxRetries.
```
nikonovd: 🔧 remove the comment

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/GoogleVertexAiRealModelManualTests.java
```diff
@@ -0,0 +1,288 @@
...
+      elementTemplate
+          // the shared fixtures default to OpenAI - drop those so only Vertex AI is configured
+          .withoutPropertyValueStartingWith("provider.openai.")
+          .property("provider.type", "google-vertex-ai")
+          .property("provider.googleVertexAi.projectId", requiredEnv(ENV_PROJECT_ID))
+          .property("provider.googleVertexAi.region", target.region())
+          .property("provider.googleVertexAi.model.model", target.model())
+          .property("provider.googleVertexAi.authentication.type", authMode.templateValue)
+          // needed so AgentResponse.responseText() is populated
+          .property("data.response.format.type", "text");
```
maff: 💭 This actually should be the default - is this really needed?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/GoogleVertexAiRealModelManualTests.java
```diff
@@ -0,0 +1,288 @@
...
+                              .sum();
+
+                      // exactly one tool call proves the round trip happened without retries
+                      assertThat(toolCallsMade).isEqualTo(1);
+                      assertThat(response.context().metrics().modelCalls()).isGreaterThan(1);
+                    }));
+  }
+
+  /** See {@link #targetsUnavailableInRegion()}. */
+  //  @ParameterizedTest(name = "[{index}] {0}")
```
maff: 🔧 Can we re-enable this, or make it disabled with a reason?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
```diff
@@ -222,15 +229,53 @@ protected CloseableChatModel createGoogleVertexAiChatModel(
...
+    Optional.ofNullable(connection.endpoint())
+        .filter(StringUtils::isNotBlank)
+        .ifPresent(httpOptions::baseUrl);
+
+    final var clientBuilder =
+        Client.builder()
+            .vertexAI(true)
+            .project(connection.projectId())
+            .location(connection.region())
+            .httpOptions(httpOptions.build());
```
maff, line 256: 🔧 This is now switching from gRPC to HTTP, correct? As we make the endpoint configurable, we also need to add support for the HTTP proxy here.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
```diff
@@ -47,7 +49,16 @@ public record GoogleVertexAiConnection(
...
               constraints = @TemplateProperty.PropertyConstraints(notEmpty = true))
           String region,
+      @HttpUrl
+          @TemplateProperty(
+              group = "provider",
+              description = "Optional custom API endpoint",
+              type = TemplateProperty.PropertyType.String,
+              feel = FeelMode.optional,
+              optional = true)
+          String endpoint,
```
maff, line 59: 🔧 Should we make this hidden by default or is there a standard case on Google to use a custom endpoint URL?

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryTest.java
```diff
@@ -664,81 +670,180 @@ void createsGoogleVertexAiChatModel() {
...
+            verify(builders.chatModelBuilder, never()).topP(anyDouble());
+            verify(builders.chatModelBuilder, never()).topK(anyInt());
           });
     }
 
+    /**
+     * Service account credentials must be scoped explicitly. google-genai only scopes the
+     * application default credentials it resolves itself and passes user-supplied credentials
+     * through verbatim, so an unscoped credential makes the token request fail with {@code
+     * invalid_scope}. The SDK we migrated away from applied this scope for us.
```
maff: 🔧 Let's remove sentences like the last one which only make sense in the context of this migration, not later when looking at the code without its history.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/Langchain4JAiFrameworkAdapterTest.java
```diff
@@ -114,15 +114,15 @@ class Langchain4JAiFrameworkAdapterTest {
   void setUp() {
     runtimeMemory = new DefaultRuntimeMemory();
     runtimeMemory.addMessages(INPUT_MESSAGES);
-    when(chatMessageConverter.map(runtimeMemory.filteredMessages())).thenReturn(L4J_MESSAGES);
+    when(chatMessageConverter.map(runtimeMemory.filteredMessages(), null)).thenReturn(L4J_MESSAGES);
```
maff: 🔧 Can we pass a mock or a dummy config instead of null here to reflect reality?

parent/pom.xml
```diff
@@ -638,6 +640,12 @@ limitations under the License.</license.inlineheader>
         <artifactId>google-api-services-sheets</artifactId>
         <version>${version.google-api-services-sheets}</version>
       </dependency>
+      <!-- keep in sync with langchain4j-google-genai -->
```
maff: 🔧 Redundant comment? 🤔

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/AssistantMessageMetadataDecorator.java
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Decorates the {@code AiMessage.attributes()} / {@code AssistantMessage.metadata()} round trip
+ * with provider-specific knowledge of what is safe to persist into a Camunda process variable.
+ *
+ * <p>{@code AiMessage.attributes()} is a generic map, populated only by langchain4j's own
+ * provider-specific integration code, and is documented as "typically provider-specific". Since the
+ * persisted {@code AssistantMessage} lives in an untyped process variable, it must never be dumped
+ * verbatim - each provider explicitly declares what it needs echoed back.
+ */
+public interface AssistantMessageMetadataDecorator {
```
maff: 💭 I think the approach and extraction/injection logic is good, but I'm just thinking if it's not overkill to add this infrastructure and the resolution logic per provider configuration vs. making it part of the `CloseableChatModel` (or a dedicated interface implemented by our chat models). For now it would only need to be implemented by the Google implementation (the others defaulting to DROP_ATTRIBUTES) + we'd have the flexibility to have configuration-aware implementations if there is the need to (the factory can decide on that when building the model).

The converter could just accept the chat model (or a dedicated interface, which could even be this one) instead of the provider configuration.

maff (commented):
Looks great overall 🚀 A couple of comments/thoughts in details.

We probably also need to port that forward to `main` to support the v1 templates. We might find a way to transparently switch to the v2 implementation, but it might be tricky with the persisted data.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/provider/GoogleVertexAiProviderConfiguration.java
nikonovd: https://docs.cloud.google.com/gemini-enterprise-agent-platform/machine-learning/general/vpc-service-controls

IMO it's like for other cloud providers, that you can restrict access to the API and therefor also will have a custom endpoint.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
nikonovd: My initial plan was to have it as a followup, but I'll contribute in that PR to have a roundup here.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/GoogleVertexAiRealModelManualTests.java
nikonovd: It's a failing test now and I kept it for visibility. But we can happily remove it.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/AssistantMessageMetadataDecorator.java
nikonovd: Fair point regarding the overkill. We could simplify it into a single class with a handler for google + default strategy otherwise.

I'd still point it to the `ProviderConfiguration` as `ChatModel` is a separate concern and a layer deeper than this converter. 

The typed `ProviderConfiguration` seems like a better fit.

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatModelFactoryImpl.java
nikonovd: Added proxy options, will test it on SaaS after the review being finished

maff (approved):

maff (approved):
Pre-approving as the PR looks good overall, Copilot comments might make sense to address.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/a2a/client/agentic/tool/A2aGatewayToolHandlerTest.java
```diff
@@ -346,8 +346,8 @@ void transformsA2aToolCalls_toA2aClientOperations() {
       var agentContext = AgentContext.empty();
       var toolCalls =
           List.of(
-              new ToolCall("call1", "A2A_a2a1", Map.of("message", "hello")),
-              new ToolCall("call2", "regular_tool", Map.of("arg", "value")));
+              new ToolCall("call1", "A2A_a2a1", Map.of("message", "hello"), Map.of()),
```
maff, line 349: 🔧 Nit: WDYT of adding a BC constructor to `ToolCall` to avoid having to pass empty metadata? We do this for the `Content` blocks as well.

connectors/agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ChatMessageConverterTest.java
```diff
@@ -222,81 +222,139 @@ void fromAssistantMessage_withMultipleContents_throwsException() {
...
+    final var toolCall = ToolCall.builder().id("toolCallId").name("toolName").build();
+    when(toolCallConverter.asToolCall(toolExecutionRequest)).thenReturn(toolCall);
+
     final var assistantMessage = chatMessageConverter.toAssistantMessage(chatResponse, VERTEX_AI);
 
-    assertThat(assistantMessage.metadata())
-        .containsEntry(
-            "provider",
-            Map.of(GoogleVertexAiProviderConfiguration.GOOGLE_VERTEX_AI_ID, expectedAttributes));
+    assertThat(assistantMessage.metadata()).doesNotContainKey("provider");
```
maff, line 262: 🔧 Is this relevant to test? I think this was only a temporary state on this branch, correct?

connectors/agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/framework/langchain4j/ToolCallMetadataDecorator.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * <p>Persisted entries are namespaced by the provider's {@code TemplateSubType} ID (currently only
+ * {@link GoogleVertexAiProviderConfiguration#GOOGLE_VERTEX_AI_ID}), so that switching a process
+ * instance's provider - e.g. via a config update or a process instance migration - cannot leak a
+ * previous provider's persisted metadata into an unrelated one.
+ */
+final class ToolCallMetadataDecorator {
+
+  private static final String THOUGHT_SIGNATURE_ATTRIBUTE_KEY_PREFIX = "thought_signature_";
+  private static final String THOUGHT_SIGNATURE_METADATA_KEY = "thoughtSignature";
```
maff, line 36: 🔧 Nit: given this class is generic, these could be prefixed with `GOOGLE_VERTEX_` or similar.

maff (approved):

PR #8089, opened 2026-07-27
feat: document store updates for multi-engine support

nicpuppa:
## Description

Propagates the physical tenant through document creation and resolution, so multi-client deployments no longer share a single global document store across every configured client.

- Adds an optional `physicalTenantId` field to `DocumentCreationRequest` and a shared `DocumentFactory.PHYSICAL_TENANT_ID_ATTRIBUTE` Jackson reader-attribute key, with a legacy constructor overload preserved for binary/source compatibility.
- Builds one `DocumentFactory`/`CamundaDocumentStoreImpl` per configured physical tenant for inbound connectors (mirroring the existing per-tenant `SearchQueryClient`/`ProcessInstanceClient` pattern); outbound already had this via the sibling multi-engine work on this branch.
- Adds a `Map<String, DocumentFactory>`-based constructor overload to the Jackson document deserializers, resolving the correct tenant's factory from the reader attribute at deserialize time — needed because the shared, singleton `ObjectMapper` backing inbound property binding can't be rebuilt per tenant the way outbound's per-worker mapper can.
- Wires the attribute through every inbound binding path: `InboundConnectorContextImpl.bindProperties`/`bindElementProperties` and `DefaultProcessInstanceContext.bind` (intermediate/polling connectors).
- `JobHandlerContext.create`/`InboundConnectorContextImpl.create` now stamp outgoing `DocumentCreationRequest`s with their own physical tenant ID (never overriding an explicit connector-supplied value), and `CamundaDocumentStoreImpl` validates it against its own configured tenant, catching a stale/cross-wired `DocumentFactory` instead of silently writing to the wrong cluster.

## Related issues

closes #6960 

## Checklist

- [ ] Backport labels are added if these code changes should be backported. No backport label is added to the latest
  release, as this branch will be rebased onto main before the next release. Example backport labels:
    - `backport stable/8.8`: for changes that should be included in the next 8.8.x release.
    - **or** `backport release-8.8.7`: for changes that should be included in the specific release 8.8.7, and this
      *release has already been created*. The release branch will be merged back into stable/8.8 later, so the change
      will be included in future 8.8.x releases as well.
- [ ] Tests/Integration tests for the changes have been added if applicable.
- [ ] If the change requires a documentation update, it has been added to the appropriate section in the documentation.

PR #8185, opened 2026-07-31
fix(runtime): reuse overridden document store for single-physical-tenant outbound wiring

maff:
## Description

Two outbound-side helpers in `OutboundConnectorRuntimeConfiguration` — `buildDocumentFactoriesByPhysicalTenantId` and `buildOutboundConnectorObjectMappersByPhysicalTenantId` — guarded their single-physical-tenant "reuse the overridden document store" path on `registry == null`. `CamundaClientRegistry` is registered unconditionally by the camunda-client Spring Boot starter (`@ConditionalOnMissingBean`), so it is **never actually null** in any real Spring context. The guard was dead code: outbound job processing always built a fresh, real, client-backed `CamundaDocumentStoreImpl`/`ObjectMapper` per physical tenant, silently bypassing any overridden `documentStore`/`documentFactory` bean — e.g. a test's `@MockitoSpyBean CamundaDocumentStore` / `@Primary DocumentFactory`, or a custom in-memory store in a non-multi-tenant deployment.

The correct version of this same check already exists inbound-side, in `PhysicalTenantIds.java`: it checks the actual resolved client count (`clientNames(...).size() <= 1`), not registry nullness. This PR mirrors that pattern on the outbound side.

### How this was found

Two e2e suites started failing on `main`:
- `connectors-e2e-test-intrinsic-functions`: `IntrinsicFunctionsTests#createLinkIntrinsic_returnsLink`
- `connectors-e2e-test-agentic-ai`: `AgentTaskMemoryStorageTests#camundaDocumentStorage`, `AgentSubProcessMemoryStorageTests#camundaDocumentStorage`

All three failed the same way — `awaitility` timeout waiting for process completion. The CI console log doesn't show the real exception (`redirectTestOutputToFile=true` sends test-JVM stdout to `target/surefire-reports/*-output.txt`, uploaded as a separate artifact, not printed to the job console). Pulling that artifact showed the real errors were engine-side REST failures:
- `createLink`: `ProblemException 403 Forbidden — "The in-memory document store does not support creating links"`
- `camundaDocumentStorage`: `ProblemException 404 Not Found — "Document with id '...' not found"`

Both are consistent with the outbound path silently talking to the *real* engine-backed document store instead of the test-overridden one (each test's own `@MockitoSpyBean CamundaDocumentStore` / test document-store config), because of the dead `registry == null` guard described above.

PR #8089 was the first suspect (it touched document-store/physical-tenant code and landed right before these started failing), but reverting it alone doesn't fix the tests — the actual guard bug was introduced earlier, in #8026 ("multi-engine support for outbound connectors"), and is unchanged (byte-for-byte identical) between #8026 and #8089. #8089 is blameless; it's just the first commit for which this e2e workflow happened to run in CI after #8026 landed.

## Fix

- `buildDocumentFactoriesByPhysicalTenantId`: `registry == null && injectedDocumentFactory != null` → `injectedDocumentFactory != null && clientNames(registry, legacyCamundaClient).size() <= 1`
- `buildOutboundConnectorObjectMappersByPhysicalTenantId`: `registry == null` → `documentFactoriesByPhysicalTenantId.size() <= 1` (its map is, by construction, already keyed per resolved physical tenant, so this is equivalent and needed no extra parameters)

## Verification

Rebuilt `connector-runtime-spring` + dependents and ran the three previously-failing tests locally — all pass:
- `IntrinsicFunctionsTests#createLinkIntrinsic_returnsLink` — PASS
- `AgentTaskMemoryStorageTests#camundaDocumentStorage` — PASS
- `AgentSubProcessMemoryStorageTests#camundaDocumentStorage` — PASS

## Related issues

Root-caused from CI failures reported in Slack (E2E test run on `main`, run [30603435913](https://github.com/camunda/connectors/actions/runs/30603435913)).

## Checklist

- [x] Tests/Integration tests for the changes have been added if applicable. (existing e2e tests now cover this — no new test needed, the regression is fully covered by `createLinkIntrinsic_returnsLink` and `camundaDocumentStorage`)

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
@@ -285,17 +285,19 @@ public class OutboundConnectorRuntimeConfiguration {
   }
 
   /**
-   * Builds the per-physical-tenant {@link DocumentFactory} map. In the single-physical-tenant case
-   * (no {@link CamundaClientRegistry}), the already-resolved {@code injectedDocumentFactory} bean
-   * is reused instead of always constructing a new client-backed one: this preserves pre-#6961
-   * behavior for single-client/custom runtimes that override the {@code documentFactory} bean (e.g.
-   * an in-memory document store in tests), which would otherwise be silently bypassed.
+   * Builds the per-physical-tenant {@link DocumentFactory} map. In the single-physical-tenant case,
+   * the already-resolved {@code injectedDocumentFactory} bean is reused instead of always
+   * constructing a new client-backed one: this preserves pre-#6961 behavior for single-client/
+   * custom runtimes that override the {@code documentFactory} bean (e.g. an in-memory document
+   * store in tests), which would otherwise be silently bypassed. Note this must check the actual
+   * client count rather than {@code registry == null}: {@link CamundaClientRegistry} is registered
+   * unconditionally by the camunda-client Spring Boot starter, so it is never actually null.
    */
   private static Map<String, DocumentFactory> buildDocumentFactoriesByPhysicalTenantId(
       CamundaClientRegistry registry,
       CamundaClient legacyCamundaClient,
       DocumentFactory injectedDocumentFactory) {
-    if (registry == null && injectedDocumentFactory != null) {
+    if (injectedDocumentFactory != null && clientNames(registry, legacyCamundaClient).size() <= 1) {
       return clientNames(registry, legacyCamundaClient).stream()
           .collect(
               toMapByPhysicalTenantId(
@@ -523,7 +525,7 @@ public class OutboundConnectorRuntimeConfiguration {
             registry, legacyCamundaClient, secretKeyCacheManager, secretFilterMode);
     var objectMappersByPhysicalTenantId =
         buildOutboundConnectorObjectMappersByPhysicalTenantId(
-            registry, documentFactoriesByPhysicalTenantId, outboundConnectorObjectMapper);
+            documentFactoriesByPhysicalTenantId, outboundConnectorObjectMapper);
     return new OutboundConnectorManager(
         jobWorkerManager,
         connectorFactory,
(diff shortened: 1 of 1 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -360,17 +360,19 @@ private static Map<String, CamundaDocumentStore> buildDocumentStoresByPhysicalTe
...
+   * store in tests), which would otherwise be silently bypassed. Note this must check the actual
+   * client count rather than {@code registry == null}: {@link CamundaClientRegistry} is registered
+   * unconditionally by the camunda-client Spring Boot starter, so it is never actually null.
    */
   private static Map<String, DocumentFactory> buildDocumentFactoriesByPhysicalTenantId(
       CamundaClientRegistry registry,
       CamundaClient legacyCamundaClient,
       DocumentFactory injectedDocumentFactory) {
-    if (registry == null && injectedDocumentFactory != null) {
+    if (injectedDocumentFactory != null && clientNames(registry, legacyCamundaClient).size() <= 1) {
```
johnBgood, line 300: @nicpuppa, worth checking you didn't use the same check in other places

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
```diff
@@ -631,10 +635,9 @@ public OutboundConnectorManager outboundConnectorManager(
    * "outboundConnectorObjectMapper"} instead of physical tenant IDs.
    */
   private static Map<String, ObjectMapper> buildOutboundConnectorObjectMappersByPhysicalTenantId(
-      CamundaClientRegistry registry,
       Map<String, DocumentFactory> documentFactoriesByPhysicalTenantId,
       ObjectMapper outboundConnectorObjectMapper) {
-    if (registry == null) {
+    if (documentFactoriesByPhysicalTenantId.size() <= 1) {
```
johnBgood, line 565: `isEmpty` ?

connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/OutboundConnectorRuntimeConfiguration.java
nicpuppa: what do you mean ?

johnBgood: the `registry == null ` check, apparently it can't be null:

> CamundaClientRegistry is registered unconditionally by the camunda-client Spring Boot starter

nicpuppa: there are 3 `registry == null` in this class

johnBgood: then they need to be replaced by the correct (clientNames <=1)

johnBgood:
@claude[agent]+claude-sonnet-5 check the comments, see if you can fix them

johnBgood (approved):

PR #8237, opened 2026-08-06
fix(connectors): retry instead of incidenting when error-expression evaluation is interrupted during shutdown

### LINKED ISSUE #8239: Runtime creates incident when restarted while an in-flight HTTP request is processing
__Describe the Bug__

When the Connectors runtime restarts while an HTTP-based connector request (which may be slow) is still in flight, this incorrectly results in an incident being raised. The restart interrupting an in-progress request is not itself a genuine job/validation failure, so raising an incident in this situation is a bug in the connectors runtime.

__Steps to Reproduce__

1. Start a process instance that executes an HTTP-based connector request (e.g. via the REST Connector) that takes some time to complete.
2. While that request is still in flight, restart the Connectors runtime.
3. An incident is raised, even though the only thing that happened is the runtime restarting mid-request.

__Expected Behavior__

A runtime restart while a slow/in-flight HTTP connector request is being processed should not, by itself, result in an incident being created.

__Status / Fix__

A fix is in progress and is targeted to be backported to `8.8` and released in the September 2026 release.

__Reference__

- Jira ticket SUPPORT-34047
- Reported via Slack incident thread by Stefanie Metzger.

ztefanie:
Closes #8239.

Incident: [#inc-support-34047-incident-created-while-evaluating-feel-expression](https://camunda.slack.com/archives/C0BMY3LPS10) ([SUPPORT-34047](https://jira.camunda.com/browse/SUPPORT-34047))

Backport of #8235.

## Original fix
#8006 (backport of #7960) already handles the incident's specific trigger on this branch: a job's thread gets interrupted during graceful shutdown, feel-scala throws a bare, message-less `InterruptedException` while evaluating the error/result expression, bypassing FEEL's normal error handling. #8006 detects this and **abandons the job** instead of raising a misleading incident, letting Zeebe's activation timeout reassign it.

## What this PR extends
Two remaining gaps in the same code path, for exceptions other than the interrupted-thread case:
1. `FeelEngineWrapperUtil.tryConvertToMap` only caught `IllegalArgumentException`, so other message-less exceptions could crash the whole evaluation instead of just the bad variable. Broadened to `RuntimeException`.
2. `OutboundConnectorExceptionHandler.handleFinalResultException` logged `ex.getMessage()` instead of `ex` (no stack trace ever captured) and hardcoded `retries=0`. Now logs the real throwable and retries with `job.getRetries() - 1`.

## Reproduced against
Rebuilt `camunda/connectors-bundle` from this fix, ran it in a real docker-compose stack with a slow mock endpoint, and did a real `docker stop`/`start` cycle mid-flight. Confirmed: clear reason instead of `null`, retries, and the retried attempt succeeds after restart.

## Test plan
- [x] `connector-runtime/connector-feel` tests pass (32/32)
- [x] `connector-runtime/connector-runtime-spring` tests pass (253/253)
- [x] End-to-end repro against a real docker-compose stack (crash → retry → success)

### DIFF
--- a/connector-runtime/connector-feel/src/main/java/io/camunda/connector/feel/FeelEngineWrapperUtil.java
+++ b/connector-runtime/connector-feel/src/main/java/io/camunda/connector/feel/FeelEngineWrapperUtil.java
@@ -84,9 +84,6 @@ public class FeelEngineWrapperUtil {
     try {
       return Optional.of(objectMapper.convertValue(o, MAP_TYPE_REFERENCE));
     } catch (RuntimeException ex) {
-      // A single variable failing to convert (e.g. its serialization is interrupted mid-flight
-      // during a runtime shutdown) must not take down the whole FEEL evaluation - drop it from
-      // the context instead.
       LOG.warn(ex.getMessage(), ex);
       return Optional.empty();
     }
--- a/connector-runtime/connector-feel/src/main/java/io/camunda/connector/feel/FeelEngineWrapperUtil.java
+++ b/connector-runtime/connector-feel/src/main/java/io/camunda/connector/feel/FeelEngineWrapperUtil.java
@@ -83,7 +83,10 @@ public class FeelEngineWrapperUtil {
       ObjectMapper objectMapper, Object o) {
     try {
       return Optional.of(objectMapper.convertValue(o, MAP_TYPE_REFERENCE));
-    } catch (IllegalArgumentException ex) {
+    } catch (RuntimeException ex) {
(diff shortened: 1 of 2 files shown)


ztefanie:
Closing: on `stable/8.9`, #8006 (backport of #7960) already handles this incident's specific trigger the same way `main` does — the interrupted-shutdown case is abandoned before reaching the code this PR touched. The remaining changes here were unreproduced speculative hardening, not tied to the actual incident. Superseded by #8236 for `stable/8.8`, which still needs the abandon-job fix directly.

PR #8242, opened 2026-08-06
ci(agentic-ai): add label-triggered PR run of the AI Agent real-LLM CPT suite

### LINKED ISSUE #8234: ci: allow `agentic-ai` label to trigger AI Agent CPT tests on a PR before merge queue
## Problem

The real-LLM AI Agent test suite only runs after a PR merges (via
`MERGE_QUEUE_HELM_TEST.yaml`'s `run-ai-agent-cpt` job) — never while
the PR is still open. If a PR breaks the AI Agent's actual LLM
behavior, nobody finds out until after it's already in `main`. The
only workaround before this ticket was manually triggering the entire
Helm-test workflow, which also builds/publishes Docker images and
kicks off two unrelated jobs.

This affects both real-LLM suites in this module: the OpenAI suite
(`AiAgentE2ETestIT`) and the Google Vertex AI suite
(`GoogleVertexAiE2ETestIT`).

## What was built

A new, standalone workflow, `.github/workflows/AI_AGENT_CPT_PR.yml`
(#8242). Apply the `ai-agent-cpt` label to a PR and it runs the suite;
it runs again on every later push while that label stays on. It's
separate from `MERGE_QUEUE_HELM_TEST.yaml` on purpose, so labeling a
PR doesn't also arm that file's other, unrelated jobs.

How it behaves:
- Only real cost when it's real work: a PR-check for touched paths
  runs first, so a labeled PR that doesn't touch
  `connectors/agentic-ai/**` or
  `connectors-e2e-test/connectors-e2e-test-agentic-ai/**` costs
  nothing — it skips straight through, including the image build.
- Never runs on forked-repo PRs — no secrets are available to a fork
  PR under this trigger, by design.
- The connector image is built and used locally on the runner, never
  pushed to a registry.
- Removing then re-adding the label cancels a stale in-progress run
  instead of piling up duplicates. Adding an unrelated label to the
  same PR does not touch an in-progress CPT run.
- A sanity check confirms the bundle jar was actually produced before
  the Docker build step runs, so a build problem shows up as a clear
  error instead of a confusing failure later.

Currently both real-LLM suites are wired up: OpenAI
(`OPENAI_API_KEY`) and Google Vertex AI (`IDP_GCP_SERVICE_ACCOUNT`,
`IDP_GCP_VERTEX_PROJECT_ID`, `IDP_GCP_VERTEX_REGION`), forwarded the
same way `MERGE_QUEUE_HELM_TEST.yaml` already does. Each provider row
skips itself if its secret isn't present, so nothing breaks if a
credential is ever missing.

## Current status

Open as a draft PR (#8242), rebased onto current `main`. Verified with
a real, disclosed run: the Maven build, local Docker image build, and
the real-LLM test step (OpenAI rows) all completed successfully — 15
tests, 0 failures. Testcontainers picked up the locally-built image
directly, no registry pull. Full run took ~27 minutes against the
30-minute timeout - tight, but sufficient for that run.

The Vertex AI wiring was added after that run, so it hasn't been
live-tested yet - same credential-forwarding pattern as the
already-working post-merge workflow, but worth a follow-up real run
before relying on the timeout holding with both providers' rows
running.

## Related problems found, not fixed here

These came up while building this workflow. None of them are caused
by this workflow or block it — they're pre-existing issues elsewhere
in this repo's CI, listed here so they don't get lost:

- `NIGHTLY_E2E.yml` has a bug in its fork-check condition: a labeled
  fork PR can get past the guard on a later push without the fork
  check actually applying. **Still present on `main`, not yet
  ticketed.**
- `MERGE_QUEUE_HELM_TEST.yaml` calls a GitHub Action that doesn't
  exist in this repo (`./.github/actions/observe-build-status`),
  silently swallowed because the step ignores errors. **Still present
  on `main`.** One attempt to fix this (PR #8266) was closed without
  merging, so this is still open and needs a new attempt or ticket.

## References

- Implementation: `.github/workflows/AI_AGENT_CPT_PR.yml` (#8242)
- Pattern this was modeled on: `.github/workflows/BUILD_PR_DOCKER_IMAGES.yml`
- Tracking issue: https://github.com/camunda/team-qa-engineering/issues/653

gbetances089:
## What this does

Adds a new GitHub Actions workflow, `AI_AGENT_CPT_PR.yml`, that runs the real-LLM AI Agent test
suite (OpenAI and Google Vertex AI) against a PR's own code when it's labeled
`ai-agent-model-e2e-test`. Today that suite only runs after merge, so a broken AI Agent change
isn't caught until it's already in `main`. This lets us catch it on the PR instead.

Closes #8234.

## How it's implemented

- Triggers only when the `ai-agent-model-e2e-test` label is applied — not on later pushes to the
  same PR, so re-running requires a maintainer to deliberately remove and re-apply the label after
  reviewing the new commits. This avoids a labeled PR silently re-running with live credentials on
  every subsequent push.
- Only runs if the PR touches `connectors/agentic-ai/**` or
  `connectors-e2e-test/connectors-e2e-test-agentic-ai/**`; otherwise it skips entirely, including
  the Docker image build.
- Builds the connector bundle image locally on the runner (`load: true`, never pushed to a
  registry) and runs the tests against it in the same job.
- Never runs on forked-repo PRs — forks get no secrets and no automatic run.
- A standalone workflow rather than an extension of the existing post-merge one
  (`MERGE_QUEUE_HELM_TEST.yaml`), since that one also feeds two unrelated jobs that shouldn't be
  armed by this label.
- 45-minute timeout, sized for a cold Maven cache plus both providers' test rows (a real run
  came in around ~27 minutes with OpenAI only).

## Checklist

- [ ] Tests/Integration tests for the changes have been added if applicable. (This PR *is* the CI workflow; verified
  with a real disclosed run against the OpenAI provider — see PR discussion for details.)


.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,182 @@
...
+on:
+  pull_request:
+    types: [ labeled, synchronize ]
+
+# Scoped to the PR number only (no label name in the key): removing-then-re-adding the label
+# cancels an in-flight run instead of racing a duplicate - the actual guard against label-flapping
+# cost abuse, since every run here spends real LLM API quota.
+concurrency:
+  group: ai-agent-cpt-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
gbetances089: Fixed in 38471e996: the group key now includes the label name (or event action as a fallback on synchronize), matching BUILD_PR_DOCKER_IMAGES.yml's pattern. An unrelated label now lands in its own group instead of colliding with a real CPT run; re-adding ai-agent-cpt still cancels the stale run as intended.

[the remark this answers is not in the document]
gbetances089: Re-flagged correctly on the latest review - my first fix (38471e996) only addressed the unrelated-label collision, not this one. Fixed properly in ef1d336cd: both cost-bearing trigger shapes (label applied, and a push while the label is already present) now collapse onto the same group suffix ("cpt"), reusing the exact same condition the job's own `if:` gate uses to decide what counts as a real CPT trigger, so the two can never drift apart again. An unrelated label still gets its own separate suffix. Verified against GitHub Actions' actual &&/|| return-the-operand semantics across 6 scenarios before pushing.

gbetances089:
Tested it in this PR: https://github.com/camunda/connectors/actions/runs/33626821613/job/100236420508?pr=8242

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,201 @@
...
+
+      - name: Run AI Agent E2E tests
+        if: steps.changes.outputs.ai == 'true'
+        env:
+          CONNECTORS_IMAGE_NAME: registry.camunda.cloud/team-connectors/connectors-bundle
+          CONNECTORS_IMAGE_VERSION: ${{ steps.tag.outputs.tag }}
+          OPENAI_API_KEY: ${{ steps.secrets.outputs.OPENAI_API_KEY }}
+          GOOGLE_VERTEX_AI_SERVICE_ACCOUNT: ${{ steps.secrets.outputs.IDP_GCP_SERVICE_ACCOUNT }}
+          GOOGLE_VERTEX_AI_PROJECT_ID: ${{ steps.secrets.outputs.IDP_GCP_VERTEX_PROJECT_ID }}
+          GOOGLE_VERTEX_AI_REGION: ${{ steps.secrets.outputs.IDP_GCP_VERTEX_REGION }}
```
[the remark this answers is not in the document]
gbetances089: Checked this carefully, including re-reading ProviderConfig.isEnabled() line by line (connectors-e2e-test-agentic-ai/.../AiAgentE2ETestIT.java:619-620) - you're right on the mechanism: setting these 3 env vars does flip isEnabled() true for the google-vertex-ai-v1 and google-gemini-vertex-ai-v2 rows, and the JSON does get forwarded as {{secrets.GOOGLE_VERTEX_AI_SERVICE_ACCOUNT}} into the connectors secret map. Not disputing that part.

But the premise is stale: the PR description doesn't say OpenAI-only/Vertex-deferred anymore - that was an earlier version of the body, corrected before this review ran. And #8595 (the credential-handling follow-up this comment points to) is closed - traced it back and the masking work it proposed turned out to be solving a problem that didn't exist: nothing in this repo decodes or logs a sub-field of this JSON anywhere, so the extra masking step was guarding against a leak vector that step itself would have introduced. Closed with that reasoning, not left pending.

On the actual safety question: these 3 vars use the exact same secret, same Vault path, same env-var forwarding, same vault-action whole-blob masking as MERGE_QUEUE_HELM_TEST.yaml's run-ai-agent-cpt job, which already runs this in production on every push to main. Wiring the same pattern into a second, PR-gated workflow doesn't change that credential's exposure at all - it's not a new or different risk shape.

Given that, not removing these - the concern this comment is based on (PR limits to OpenAI, #8595 pending) doesn't hold anymore. Happy to be shown otherwise if I'm missing something new, but as of right now this matches the current PR body and the resolved state of #8595.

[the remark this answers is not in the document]
gbetances089: Checked the math independently before acting on it - it holds up exactly. Confirmed 3 OpenAI rows enabled today (openAiV1, openAiResponsesV2, openAiCompletionsV2), and read providers() directly to confirm 4 more Vertex rows would activate (googleVertexAiV1 x2, googleGeminiVertexAiV2 x2) - 15 to 35 tests, matches. Pulled the real jobs step timestamps: 26m37s total, confirming the ~26.5min figure, and ~66s of post-step cleanup after the test step ends, confirming that too. Modeled the projected total for 35 tests two ways (linear scale-up and one discounting the roughly-fixed per-container startup cost) - both land at 29.5-30.8 minutes, i.e. right at or past the limit either way. Bumped to 45m in 5b6f14eba.

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,201 @@
...
+    timeout-minutes: 45
+    # Fork PRs get no secrets from `pull_request` by design (never widen this to
+    # pull_request_target - that would hand Vault + a real OPENAI_API_KEY/Vertex service
+    # account to unvetted fork code before review). Mirrors BUILD_PR_DOCKER_IMAGES.yml's gate,
+    # not NIGHTLY_E2E.yml's - that one has an operator-precedence bug where the fork check
+    # doesn't apply on `synchronize` (see #8234).
+    if: |
+      github.event.pull_request.state != 'closed' &&
+      github.event.pull_request.head.repo.full_name == github.repository && (
+        github.event.label.name == 'ai-agent-cpt' ||
```
nikonovd: 🤔 should we maybe name the label without having CPT in the name? I'm trying to wrap around a model where we are agnostic to the test framework, like `ai-agent-model-e2e-test` or something similar. WDYT?

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,201 @@
...
+      - uses: actions/checkout@v7
+        with:
+          # build the PR's actual code, not the synthetic merge ref
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      # First real step after checkout: everything below must be gated on this, including the
+      # image build - unlike run-ai-agent-cpt in MERGE_QUEUE_HELM_TEST.yaml (which gets its image
+      # for free from a separate build-image job), build and test are merged into this one job, so
+      # a labeled PR touching no agentic-ai files must not pay for a bundle build at all.
+      - name: Check for agentic-ai changes
```
nikonovd, line 67: 💭 this is good for now and I appreciate the solution.

A stretch we should look into later is if we can detect that a parent change, e.g. dependency update, that is not directly changing our module in source code, but implicitely, also should enable those tests.

By that we could catch regression introduced by e.g. dependency updates.

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,201 @@
...
+# Runs the real-LLM AI Agent CPT suite (OpenAI + Google Vertex AI, see #8234) on a PR before it
+# reaches the merge queue, gated by the `ai-agent-cpt` label so it isn't triggered accidentally.
+#
+# Deliberately a standalone workflow rather than adding `pull_request` to MERGE_QUEUE_HELM_TEST.yaml:
+# that file's build-image job also feeds trigger-sm-e2e/trigger-saas-e2e, which gate only on
+# build-image producing a version - reusing it here would arm both on every labeled PR.
+#
+# Build and test share a single job: the bundle image is built with `load: true` (never pushed),
+# so it's only ever available to this job's own Docker daemon - no push/pull round trip, no Harbor
+# credentials in this workflow's blast radius, and no orphan tags left behind in the registry.
```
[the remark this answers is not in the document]
gbetances089: Fixed in 4172436500: reworded to 'no Harbor login/push happens in this job' and clarified that the ARTIFACTORY_USR/ARTIFACTORY_PSW secrets it does import are only used for Maven/Nexus auth, not Docker registry access.

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,201 @@
...
+        if: steps.changes.outputs.ai == 'true'
+        run: |
+          ./mvnw --batch-mode install -DskipTests -DskipChecks \
+            --projects "apps/connector-runtime-application,apps/bundle/default-bundle,connectors-e2e-test/connectors-e2e-test-agentic-ai" \
+            --also-make
+
+      # Fails fast with a clear error if the bundle jar wasn't produced, rather than a confusing
+      # Docker COPY failure downstream. Checks the actual glob the Dockerfile depends on, not just
+      # that target/ exists. Same check MERGE_QUEUE_HELM_TEST.yaml runs after its own equivalent
+      # build step.
```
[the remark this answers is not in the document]
gbetances089: Fixed in 4172436500: reworded to note MERGE_QUEUE_HELM_TEST.yaml only runs a plain `ls -lh` on the directory and doesn't actually assert the jar exists, so it isn't an equivalent check.

.github/workflows/AI_AGENT_CPT_PR.yml
gbetances089: Agreed, and it's a good rationale - renamed the label from `ai-agent-cpt` to `ai-agent-model-e2e-test` in 4172436500 (5 occurrences: header comment, concurrency group, and the trigger `if:` condition). Left the `run-ai-agent-cpt` job id/name alone since that's a separate identifier also used as a job name in MERGE_QUEUE_HELM_TEST.yaml, unrelated to this label.

.github/workflows/AI_AGENT_CPT_PR.yml
gbetances089: Appreciate the idea - agreed it's worth tracking rather than losing once this thread closes. Filed https://github.com/camunda/connectors/issues/8970 with a starting sketch (scoping to agentic-ai-relevant dependency changes rather than any pom.xml edit, to keep the cost-control intent of the paths-filter gate). Not part of this PR's scope, just wanted a durable place for it.

.github/workflows/AI_AGENT_CPT_PR.yml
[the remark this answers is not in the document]
gbetances089: Following up on the blank/empty-credential angle Copilot's suppressed review comments flagged separately on this same line (isEnabled() only checking != null, not non-blank): confirmed that's a real gap in the current ProviderConfig.isEnabled() on main, but the real fix (RealLlmTestEnvironment.hasNonBlankValues()) already lands in #8713 immediately next in the stack, at the Java level - not something this workflow-only PR should carry, since that would pull a Java source change into a pure-CI PR. Leaving this PR as-is on that point; #8713 closes the gap right after.

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,205 @@
...
+    # Fork PRs get no secrets from `pull_request` by design (never widen this to
+    # pull_request_target - that would hand Vault + a real OPENAI_API_KEY/Vertex service
+    # account to unvetted fork code before review). Mirrors BUILD_PR_DOCKER_IMAGES.yml's gate,
+    # not NIGHTLY_E2E.yml's - that one has an operator-precedence bug where the fork check
+    # doesn't apply on `synchronize` (see #8234).
+    if: |
+      github.event.pull_request.state != 'closed' &&
+      github.event.pull_request.head.repo.full_name == github.repository && (
+        github.event.label.name == 'ai-agent-model-e2e-test' ||
+        (github.event.action == 'synchronize' && contains(github.event.pull_request.labels.*.name, 'ai-agent-model-e2e-test'))
```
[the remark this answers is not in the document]
gbetances089: Confirmed - validated this is real, not a false positive. The fork check only stops external forks from getting secrets; it does nothing once a same-repo contributor pushes new commits to an already-labeled PR, since GitHub's fork-approval gating doesn't apply to non-fork `pull_request` runs. Fixed in 9eba73d9b2 by dropping `synchronize` from the trigger entirely - re-running now requires a maintainer to deliberately remove and re-apply the label after reviewing the new commits, rather than it happening automatically on every push.

Also audited the rest of .github/workflows/ for the same shape (label + synchronize + real secrets imported): BUILD_PR_DOCKER_IMAGES.yml, PREVIEW-ENV-DEPLOY.yml, and TEST_FEATURE_BRANCH.yml (via NIGHTLY_E2E.yml) all share it and pre-date this PR. Left those three untouched here since changing three independently-owned, already-established workflows is a bigger repo-wide security decision than this PR's scope - tracked in https://github.com/camunda/connectors/issues/8971 instead.

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,210 @@
...
+    # listen for `synchronize` at all - see the top-of-file note on why.
+    if: |
+      github.event.pull_request.state != 'closed' &&
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      github.event.label.name == 'ai-agent-model-e2e-test'
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # build the PR's actual code, not the synthetic merge ref
+          ref: ${{ github.event.pull_request.head.sha }}
```
nikonovd, line 61: 🔧 should we apply `persist-credentials: false` here as well to be on the safe side, if some revision would introduce git operations here?

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,210 @@
...
+  run-ai-agent-cpt:
+    name: AI Agent E2E Tests (CPT)
+    runs-on: gcp-core-8-default
+    timeout-minutes: 45
+    # Fork PRs get no secrets from `pull_request` by design (never widen this to
+    # pull_request_target - that would hand Vault + a real OPENAI_API_KEY/Vertex service
+    # account to unvetted fork code before review). Mirrors BUILD_PR_DOCKER_IMAGES.yml's gate,
+    # not NIGHTLY_E2E.yml's - that one has an operator-precedence bug where the fork check
+    # doesn't apply on `synchronize` (see #8234). Moot here anyway since this workflow doesn't
+    # listen for `synchronize` at all - see the top-of-file note on why.
```
nikonovd, line 52: Let's cleanup the narration here so that we only mention what's relevant to the workflow, WDYT?

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,210 @@
...
+        run: ls -lh apps/bundle/default-bundle/target/*-with-dependencies.jar
+
+      - name: Compute image tag
+        if: steps.changes.outputs.ai == 'true'
+        id: tag
+        env:
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          set -euo pipefail
+          BASE_VERSION=$(./mvnw help:evaluate -Dexpression=project.version -q -DforceStdout)
```
nikonovd, line 164: I think we need -B argument in mvnw command here to avoid problems with the variable.

.github/workflows/AI_AGENT_CPT_PR.yml
```diff
@@ -0,0 +1,210 @@
...
+# that file's build-image job also feeds trigger-sm-e2e/trigger-saas-e2e, which gate only on
+# build-image producing a version - reusing it here would arm both on every labeled PR.
+#
+# Deliberately `labeled` only, not `labeled, synchronize`: a same-repo PR still gets full Vault +
+# OPENAI_API_KEY/Vertex-service-account credentials once the label is present, and unlike a fork,
+# nothing stops the PR author (or anyone else with push access to that branch) from pushing new
+# code afterwards. Auto-re-running on every later push would hand a live credential-bearing run to
+# whatever code lands on the next push, with no further human look - re-triggering only on the
+# `labeled` event means a maintainer has to deliberately remove and re-apply the label (reviewing
+# the new commits first) to arm another run.
```
nikonovd, line 18: This could be a bit overstated as technically people could just apply the label on the PR and then also execute their changeset here.

@johnBgood I'm not the biggest expert in github CI and especially not on the connectors CI setup.

Do you think we would need to take some measurements here to prevent people from labelling a maliscious PR with the new label and then hijack secrets via the maven lifecycle?

Would the same apply to `e2e-tests`? Curious to hear from your side.

johnBgood: You're right that the comment overstates it — applying the label just needs Triage/Write, which a same-repo PR author already has, so there's no actual review gate, just a deliberate action.

That said, I don't think it changes the trust boundary. The real control here (and everywhere else in this repo where a `pull_request`-triggered workflow imports secrets) is the fork check, not the label. Once you have push access to camunda/connectors, you can already get equivalent secrets into a run against your own code — e.g. by editing the workflow file itself in your branch, since GitHub runs the PR's own copy of it for non-fork `pull_request` events. The label is a cost/noise throttle (skip the ~30min real-LLM run on every push), not a privilege gate.

Same reasoning applies to `e2e-tests` (NIGHTLY_E2E.yml) and `docker-images-required`/`deploy-preview` (BUILD_PR_DOCKER_IMAGES.yml) — all three make the same "non-fork + label = trusted" assumption, so this isn't a new risk class this PR is introducing.

nikonovd: Thanks for your input. To rephrase it, we're fine here as the push permission to this repo is limited to camunda developers and external people are only able to fork, where we won't execute those workflows, even with the label.

If so, it's fine from my side as well 👍

gbetances089:
Follow-up on the latest review threads after f0c6c1ddb:

- The AWS/Bedrock credentials in the new PR workflow are intentionally pulled from Vault as the IDP/AWS values and mapped to the AWS_BEDROCK_* / ANTHROPIC_BEDROCK_* env vars used by the tests.
- The missing-credential risk was in the existing MERGE_QUEUE_HELM_TEST.yaml caller: it runs -Pit-real-llm without importing the additional AWS judge credentials. Fixed by keeping DocumentToolCallResultsIT out of the default it-real-llm profile, while the PR workflow still selects it explicitly with -Dit.test.
- The PR body now documents AWS Bedrock as part of the workflow scope and explains that document-tool-call rows are PR-workflow-explicit.
- Applied the requested workflow cleanup: persist-credentials: false, shorter workflow narration, --batch-mode for help:evaluate, and quoted the element-templates-cli install argument found during actionlint review.
- The fork/secrets thread was resolved as no code change needed based on the reviewer confirmation.

mathias-vandaele (approved):

nikonovd (approved):

PR #8243, opened 2026-08-06
feat(textract): add document source and response format dropdowns

ztefanie:
Textract document handling for camunda/product-hub#3665, following CSV (#8025). Base #8187; retarget to `main` once merged.

**Input** — outer dropdown relabeled **Document location**; `document` renders the standard **Document source** dropdown (Camunda Document / From URL).

**Output** — response format via `@DocumentReturnFormat({JSON, DOCUMENT})`, hidden on the Async path. JSON (and old templates) return today's result unchanged; Document returns a bare document reference.

**Generator** — new `@TemplateDocumentProperty.sources()` restricts which sources a connector offers, defaulting to all three. Declared order sets the dropdown default. Textract excludes Inline Content, whose UTF-8 text bytes it rejects.

## Checklist

- [ ] Live runtime QA
- [ ] Backport labels added if needed
- [x] Tests for the changes have been added


connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/model/TextractRequestData.java
```diff
@@ -52,17 +55,16 @@ public record TextractRequestData(
...
+    @TemplateDocumentProperty(
             group = "document",
-            label = "Camunda Document",
-            tooltip = "The Camunda document to be analyzed.",
-            feel = FeelMode.required,
-            type = TemplateProperty.PropertyType.String,
+            tooltip = "The document to be analyzed.",
+            sources = {DocumentSource.CAMUNDA, DocumentSource.EXTERNAL},
+            fileName = FieldVisibility.HIDDEN,
+            contentType = FieldVisibility.HIDDEN,
```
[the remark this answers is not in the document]
ztefanie: Good catch on the mechanism, and the suggested fix does work: `resolveComposerId` already prefers `annotation.id()` and only falls back to `targetPath + "__composer"` when it is blank, so `id = "input.document"` would keep the id stable with no generator change.

We're deliberately not taking it. This ships as a documented breaking migration instead — upgrading a task from a ≤v7 template does not carry the document expression over, and the model has to re-select the document source and re-enter the reference. That is now called out under "⚠️ Breaking change — template migration" in the PR description. The cost of a manual step on existing models was accepted knowingly, in favour of the new source dropdown being the single source of truth for that binding.

One technical point we'd like a second opinion on, though: the composer is a `Hidden` property carrying a fixed `value` (the `if … then … else` expression). On upgrade Modeler writes that template-supplied value regardless, so preserving the id looks like it avoids an *orphaned* property rather than actually carrying the user's old FEEL expression into the composer — the resulting `zeebe:input` targeting `input.document` ends up being the composer expression either way. If that reading is wrong and the id genuinely preserves the user's expression, we'd revisit the decision — please say so.

connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/TextractConnectorFunction.java
```diff
@@ -27,6 +32,7 @@
       "document",
       "input",
       "advanced",
+      "documentReturnFormat",
```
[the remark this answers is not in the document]
ztefanie: The versioning rule is right in general, but v7 isn't published anywhere yet, so there is no released version to preserve:

| branch | `@ElementTemplate` version | `versioned/` snapshots |
| --- | --- | --- |
| `main` | 6 | `-1` … `-5` |
| `stable/8.9` | 4 | `-1` … `-3` |
| `stable/8.8` | 4 | `-1` … `-3` |

v7 is introduced by #8187, the unmerged base this PR is stacked on. The 8.8/8.9 backport labels on that PR don't change the picture either: those branches are on v4, so their backports bump them to 5 and never reach 7 — version numbers are per-branch.

So editing v7 in place does not mutate a template any user has received. If a release is cut on `main` after #8187 merges but before this PR lands, this does become a real bump-to-8-plus-snapshot; that condition is recorded in the PR description as something to re-check before merge.

connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/model/DocumentLocationType.java
```diff
@@ -9,8 +9,8 @@
 import io.camunda.connector.generator.java.annotation.DropdownItem;
 
 public enum DocumentLocationType {
-  @DropdownItem(label = "Camunda Document")
+  @DropdownItem(label = "Uploaded document")
```
chillleader: nit: Not sure if this is the best label (not obvious where it comes from), I think I liked the old one better

connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/model/TextractRequest.java
```diff
@@ -6,12 +6,29 @@
...
 import io.camunda.connector.aws.model.impl.AwsBaseRequest;
+import io.camunda.connector.generator.java.annotation.DocumentReturnFormat;
+import io.camunda.connector.generator.java.annotation.FieldVisibility;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.NotNull;
 import org.apache.commons.lang3.builder.EqualsBuilder;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 
+@DocumentReturnFormat(
```
chillleader, line 19: Just an observation: our ETG DSL is becoming quite big, would be nice to have extensive documentation explaining all annotations. Should we create a ticket to document everything properly?

chillleader (approved):
Nice! Approving with one optional comment

Szik (approved):

connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/model/TextractRequest.java
ztefanie: @claude[agent] please create this ticket

ztefanie: @copilot create this ticket

connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/model/DocumentLocationType.java
ztefanie: "Camunda document" was inaccurate since that branch also covers "Camunda Document", "inline document" and "From URL," not just Camunda documents (from the store), but you are right, the wording wasn't so gut, I renamed it to "From Process or URL" and "From Amazon S3" so both options should be clear now :)

connectors/aws/aws-textract/src/main/java/io/camunda/connector/textract/model/TextractRequest.java
ztefanie: ok created  camunda/connectors#8300

PR #8251, opened 2026-08-06
feat(connector-sdk): add job lease support to Connector SDK (@OutboundConnector)

### LINKED ISSUE #8044: Add job lease support to Connector SDK (@OutboundConnector)
__What should we do?__

Add job lease support to the Connectors SDK.

Job lease tokens (see [ADR 0005-810](https://github.com/camunda/camunda/blob/main/zeebe/docs/adr/0005-810-job-lease.md)) are a new core concept for ensuring data correctness on job completion. The AI Agent connector needs this to safely push agent history & metrics during execution — a job must be requested with a lease token, and that token must be passed back on completion. The AI Agent connector also needs the token surfaced on `JobContext` to make its own agent history requests.

**Blocked on:** camunda/camunda#57812 — expose `withLease` in the Spring Boot SDK's `JobWorkerValue` (committed for alpha 5, since the connector runtime uses `jobWorkerManager` from the camunda-client, driven by the `@OutboundConnector` annotation, not the client directly).

Proposed approach (agreed in Slack thread):

- Add a lease option to `@OutboundConnector`, e.g. `withLease` / similar flag.
- Default value TBD — leaning toward `false` by default, with the AI Agent connector as first adopter opting in explicitly, unless there's reason to default it on for all connectors.
- Surface the lease token on `JobContext` so it can be passed back on job completion and used for agent history requests.

__Why should we do it?__

Without lease tokens, job completion (e.g. pushing agent history & metrics) can race and produce inconsistent data at production scale. This is a data-consistency fix, needed before the AI Agent connector can safely handle concurrent history/metric updates during execution.

johnBgood:
## Description

Adds job lease support to the Connector SDK. A connector can now opt in to activating its jobs with a lease, and read the resulting lease token to use in its own requests. This protects against a job being completed twice when a job is re-activated after a timeout while the original invocation is still finishing — without it, the stale invocation's result can silently overwrite the real one.

This is a prerequisite for the AI Agent connector to safely push agent history and metrics during execution at production scale. Adopting the flag on the AI Agent connector itself is a separate follow-up.

## Related issues

closes #8044

## Checklist


maff (approved):
LGTM, thanks for implementing!

PR #7767, opened 2026-07-02
feat(element-template-generator): support tooltip on discriminator properties

crobbins215:
## Description

`@TemplateDiscriminatorProperty` currently exposes `description()` but **no `tooltip()`**, so discriminator dropdowns cannot use tooltips. This is the gap @ztefanie flagged on #7651: the AWS **"Authentication"** type selector (`AwsAuthentication` in `aws-base`) is a discriminator, so it can only get a description today.

This PR:
- adds `tooltip()` to `@TemplateDiscriminatorProperty`
- threads it through `TemplatePropertiesUtil` (discriminator builder), mirroring the existing `description` handling
- adds test coverage in `OutboundClassBasedTemplateGeneratorTest`

`DropdownProperty` already serializes `tooltip`, so there are no DSL or JSON schema changes — the value simply flows through.

Split out from the description→tooltip migration (#7470) as its own generator PR, per review discussion. Unblocks migrating discriminator fields (AWS auth type) to tooltips.

## Related issues

related #7470

## Checklist

- [x] Test coverage added (`annotated_sealedType_followsAnnotations`)
- [x] `mvn spotless:apply` run

PR #8259, opened 2026-08-07
other(app-integrations): Improve App Integrations Connector

### LINKED ISSUE #8320: App Integrations connector missed the tooltip-first field-hint migration (#7470)
## Summary

`connectors/app-integrations` (`AppIntegrationsConnector`) shipped in #7588, merged 2026-07-10 — squarely between the two `tooltip-first element-template field hints` migration waves from #7470 (authored 2026-07-01, merged 2026-07-14, e.g. #7742–#7745). It landed after the migration's audit was done, so no slice ever touched it.

Before this fix: 36 `description=` properties, only 4 `tooltip=`.

## Fix

Applied the same `description=` → `tooltip=` decision rules used across #7470 to all 5 `@TemplateProperty`-annotated classes in the module:
- `AppIntegrationsConfiguration.java`
- `model/auth/OAuthAuthentication.java`
- `model/auth/ApiKeyAuthentication.java`
- `model/SendMessageRequest.java`
- `model/CreateChannelRequest.java`

Notable decisions:
- Merged redundant dual `description=`+`tooltip=` pairs (e.g. `baseUrl`, `apiKey`) into a single tooltip.
- Removed one field-hint entirely (`oauthTokenEndpoint`'s `"The OAuth token endpoint"` — redundant with its label, per the guide's "The [label]" pattern).
- Extracted two `has-example` hints into `placeholder=` (`scopes` → `read:contacts`, `email` → `=assigneeEmail`, `channelId` → `19:xxx@thread.tacv2`).
- Kept one `description=` — `displayName`'s "max 50 characters" — since it's a genuine `value-constraint` (must stay always-visible per the guide), split from its explanatory sentence which moved to `tooltip=`.
- Removed one redundant field-only hint (`description` field's "Optional description for the channel." — restates the label).

`TemplateLinkedResource`'s `resourceIdDescription` (on `SendMessageRequest`'s linked form) is intentionally left as-is — that annotation has no `tooltip` equivalent at the generator level (same category of hard exception as `@TemplateDiscriminatorProperty`), out of scope without a generator change.

No template version bump — this is a pure content/wording change, not a breaking one.

PR: (linked separately)

yhaskell:
## Description

Reworks the App Integrations connector's configuration and message model, and adds Slack as a
recipient.

### 1. Connection details move out of the element template

The `Configuration` and `Authentication` groups asked the modeler for a base URL and credentials —
20 of the template's 40 properties. In SaaS those values were silently discarded (the executor
already read env vars); in Self-Managed they were used, duplicating backend infrastructure into
every BPMN model. The backend is Camunda-operated in both modes, so its URL and credentials now come
from `APP_INTEGRATIONS_*` in the runtime environment only. OAuth wins when fully configured,
otherwise an API key; blanks count as absent.

Without those variables the connector is unusable on that runtime, so jobs fail via
`ConnectorRetryException` with `retries = 0` and error code `APP_INTEGRATIONS_NOT_CONFIGURED`. That
raises the incident immediately instead of burning the template's retries at its backoff — no retry
can supply a missing environment variable. The message names the missing variables but never their
values, since incident messages are visible in Operate.

### 2. Slack recipient, always-available message, per-platform additional content

Plain text and rich content were mutually exclusive, but the backend accepts both — a modeler could
not send "here's the summary" *plus* a card. `message` is now a top-level optional field, always
visible. Rich content moves to a separate **Additional content** switch whose options depend on the
recipient:

| Recipient | Additional content |
| --- | --- |
| Camunda | None · Form |
| Microsoft Teams | None · Adaptive card · Form |
| Slack | None · Block Kit · Form |

Picking at most one is what keeps a card/Block Kit payload mutually exclusive with a form —
structurally, rather than by cross-field validation. An empty message with "None" is rejected before
any HTTP call.

**Recipient source** gains Slack, addressed by a nested switch: a channel ID, or a user ID. **Create Channel** gains a Teams/Slack platform switch, with `teamId` and the Teams-URL
`groupId` extraction moving into the Teams subtype.

An element-template dropdown's choices are static — conditions govern whether a property renders,
never what a dropdown contains. So per-platform options mean one dropdown per recipient, which is
exactly what a sealed field nested inside a sealed subtype generates, with no generator change. It
does require each subtype to name its nested field distinctly (`camundaExtra` / `teamsExtra` /
`slackExtra`): subtype fields inherit the path of the sealed field they hang off with no subtype
segment added, so three fields named alike would all emit `recipient.<name>.type` and trip
`UniquePropertyIdRule`. All three carry the same Modeler label, so this is invisible to modelers.

`adaptiveCard` and `blocks` are `JsonNode` with `feel = required`, so the engine evaluates the
expression and the connector forwards real JSON — outbound connectors never evaluate FEEL
themselves. A value that still arrives as a string is parsed here; malformed or wrongly shaped input
fails with `VALIDATION_ERROR` before the backend is called.

The wire payload's `platform` field is read from the subtype constants the sealed hierarchies
already use (`Recipient.SlackRecipient.TYPE` and friends), so what the backend receives cannot drift
from the discriminator value Jackson and the element template agree on.

### 3. Template stays version 1, engine `^8.10`

v1 shipped: Modeler downloads its catalog from `marketplace.cloud.camunda.io`, fed by the npm package
this repo publishes, and it carries `io.camunda.connectors.AppIntegrations.v1` with the original
38-property shape.

The template deliberately **stays at `version = 1`** rather than bumping to 2. uniquet indexes each
connector's current template plus its `versioned/` snapshots, so a bump would keep both shapes
downloadable — the old v1 would remain in the catalog, and a diagram bound to it would go on working
against a runtime that no longer accepts its variables. Keeping version 1 overwrites the published
template in place, and nothing is written to `versioned/`, so the previously published v1 stops
existing at the next release rather than living on beside the rework.

**This is stronger than deprecation and is intentional:** a diagram already bound to the old v1 does
not keep working — the runtime rejects it, since the new request model binds a `recipient` the old
template never wrote. Acceptable only because the connector is pre-GA and nobody should have access
to it yet.

`engineVersion` moves from `^8.9` to `^8.10`, which is also what stops the replaced template from
being used on a runtime that cannot execute it. So **no 8.9 backport** — please do not add a
`backport stable/8.9` label.

### 4. Tooltip-first field hints

#8321 proposed migrating this connector's field hints to the tooltip-first convention from #7470,
which it missed by shipping between the two migration waves. Most of the properties that PR targeted
no longer exist after the rework above — the whole configuration and authentication block is gone —
so it was closed and the migration is carried here instead, applied to the current model and extended
to the fields this PR adds.

Always-visible descriptions are now reserved for genuine value constraints; everything explanatory
moves to a tooltip, and hints that were only a label restated plus an example become a placeholder.
The two channel-name fields keep their length and character rules visible; the Teams and Slack ID
fields drop prose that repeated their own label and keep just the example.

The seven discriminator dropdowns this PR introduces move to tooltips as well. #8321 recorded them as
having no tooltip support at the generator level — true when #7470's rules were written, but #7767
has since added it, so leaving them as descriptions would have pinned seven permanent blocks of text
in the properties panel.

Hints only: every changed line in both regenerated templates is a tooltip, description or placeholder,
with no reordering, no binding or condition change, and no template version bump.

### Fixes picked up along the way

- **Generator bug.** `OperationBasedConnectorUtil.mapCondition` merged the operation condition into
  an existing `AllMatch` by mutating it in place, throwing `UnsupportedOperationException` on the
  immutable list `mapPropertyCondition` returns. Only reachable with two levels of nested
  discriminators on an `@Operation` connector — a combination no connector had until now. Covered by
  a regression test.
- **`@TemplateLinkedResource` gains `conditions`**, a list of `@NestedPropertyCondition` ANDed with
  the operation scope, so a linked resource can be gated on discriminators instead of its own Yes/No
  toggle. It delegates to the same transform ordinary property conditions use, so `equals`,
  `equalsBoolean`, `oneOf` and `isActive` all work and malformed conditions fail with the same
  messages. Additive: every other connector's template regenerates byte-identical.
- Each of the three form linked resources is gated on **both** the recipient discriminator and its
  own additional-content discriminator. Gating on the inner one alone let a stale value from a
  previously selected branch keep matching, which could emit two form resources for one task.
- `CreateChannelRequest.membershipType` declared `choices` without `type = PropertyType.Dropdown`;
  since the field is not an enum the generator dropped the choices and rendered free text. Now a real
  dropdown — offering only `standard`, since private and shared Teams channels are not stable on
  Microsoft's side yet. They are switched off rather than removed: the field, the wire contract and
  the backend call are unchanged, the values just leave the dropdown and are refused by validation,
  so a hand-written variable cannot reach a path nobody has exercised. Re-enabling is moving a value
  out of `POSTPONED_MEMBERSHIP_TYPES` back into the dropdown's choices.
- `app-integrations` removed from `OperationMetadataIgnoreList`. Both `@Operation` methods declare
  keywords, so the connector now emits the `steps`/`presets` metadata that powers Modeler's operation
  search, and is held to the full rule set.
- Validation hardening from review: `@NotBlank` (not `@NotEmpty`) wherever the executor normalises
  blanks to `null`, so a whitespace-only channel ID, Slack target, channel name or team ID can no
  longer pass validation and then travel as an absent field. Candidate user/group lists reject blank
  entries the same way.

### Backend dependency

The wire contract carries an explicit `platform` discriminator, since Teams and Slack both address a
"channel" and the field shape alone is ambiguous. The backend has been changed in step and **now
accepts `platform`, `processDefinitionId`, `userId`, `blocks`, JSON cards, Slack recipients and Slack
channel creation**. Connector tests use stubs, so CI cannot catch a mismatch — the two sides still
have to move together.

Every message also carries `processDefinitionId`, the BPMN process ID of the process the job came
from (`JobContext.getBpmnProcessId()`), so the backend can evaluate notification rules scoped to a
process — `notification_rule.process_definition_id`. Without it, a rule targeting the current
process never delivers. It is sent for every recipient, not only the Camunda one, so a
process-scoped rule can also match a message addressed to Teams or Slack.

```json
{ "platform": "teams", "processDefinitionId": "order-process", "channelId": "19:…", "message": "Deploy done", "adaptiveCard": { … } }
{ "platform": "slack", "processDefinitionId": "order-process", "channelId": "C0123456789", "message": "Deploy done", "blocks": [ … ] }
{ "platform": "slack", "processDefinitionId": "order-process", "userId": "U0123456789", "message": "Ping" }
{ "platform": "camunda", "processDefinitionId": "order-process", "email": "a@b.c", "candidateUsers": ["alice"], "formResourceKey": "12345" }
```

Self-Managed deployments must now inject `APP_INTEGRATIONS_*` into the connector runtime, which needs
a Helm chart and docs update outside this repository.

### Verification

- `connectors/app-integrations`: 61 tests, green.
- `element-template-generator/core`: 254 tests, green.
- Both element templates regenerate with no diff, so the generator changes are behaviour-preserving
  for every other connector.

## Related issues

Follows up on #7588. Closes #8320, superseding #8321.

## Checklist

- [x] Backport labels are added if these code changes should be backported. **No backport** — the
  connector now requires engine `^8.10`.


connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/model/SendMessageRequest.java
```diff
@@ -13,60 +13,60 @@
...
+/**
+ * The plain-text {@code message} lives here rather than on each recipient because it is
+ * platform-independent and always available; only the rich "additional content" varies by platform,
+ * so that hangs off {@link Recipient}'s subtypes.
+ *
+ * <p>A form is a {@code zeebe:linkedResource}, not a variable, and its properties can only be gated
+ * on one property at a time. Since "Form" is an option on all three per-platform switches, three
+ * linked resources are declared — one per switch, each gated on its own discriminator. The
+ * conditions are mutually exclusive, so at most one {@code zeebe:linkedResource} block is ever
+ * written to the BPMN, and {@code AppIntegrationsConnector.formResourceKey} picks it up by resource
```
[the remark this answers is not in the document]
yhaskell: Good catch — fixed in 316727e3f, this was a real defect.

Each form linked resource is now gated on **both** discriminators, e.g.

```
operation==sendMessage && recipient.type==teams && recipient.teamsExtra.type==form
```

You are right that the generator already does this for ordinary nested properties (`recipient.teamsExtra.adaptiveCard` carries both), which made the linked resources the inconsistent ones. I had been relying on Modeler pruning the conditioned-out discriminator, which is not something correctness should depend on.

Supporting two gates meant the condition API had to hold more than one condition, so `conditionProperty`/`conditionEquals` became `conditions = {@NestedPropertyCondition...}`, ANDed. Those attributes were introduced earlier in this same PR and never released, so nothing external breaks. Verified in the regenerated template: all twelve form properties across the three branches now carry both conditions.

connectors/app-integrations/README.md
```diff
@@ -1,83 +1,148 @@
...
+| Variable | Required | Notes |
+| --- | --- | --- |
+| `APP_INTEGRATIONS_BASE_URL` | yes | Backend base URL. |
+| `APP_INTEGRATIONS_API_KEY` | for API key auth | Sent in the `X-API-KEY` header. |
+| `APP_INTEGRATIONS_OAUTH_TOKEN_ENDPOINT` | for OAuth | OAuth 2.0 token endpoint. |
+| `APP_INTEGRATIONS_OAUTH_CLIENT_ID` | for OAuth | |
+| `APP_INTEGRATIONS_OAUTH_CLIENT_SECRET` | for OAuth | |
+| `APP_INTEGRATIONS_OAUTH_AUDIENCE` | no | Target API identifier. |
+| `APP_INTEGRATIONS_OAUTH_SCOPES` | no | |
+| `APP_INTEGRATIONS_OAUTH_CLIENT_AUTHENTICATION` | no | `credentials-body` (default) or `basic-auth-header`. |
```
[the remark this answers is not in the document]
yhaskell: Correct, fixed in 316727e3f. `OAuthConstants` defines `credentialsBody` and `basicAuthHeader`; the table documented hyphenated values that would have thrown `IllegalArgumentException` before token acquisition. The README now lists the exact literals.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/model/SlackTarget.java
```diff
@@ -0,0 +1,53 @@
...
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.appintegrations.model;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import io.camunda.connector.generator.java.annotation.TemplateDiscriminatorProperty;
+import io.camunda.connector.generator.java.annotation.TemplateProperty;
+import io.camunda.connector.generator.java.annotation.TemplateSubType;
+import jakarta.validation.constraints.NotEmpty;
```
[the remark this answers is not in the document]
yhaskell: Agreed, fixed in 316727e3f — `@NotBlank` on both Slack targets. The failure mode was exactly as described: `@NotEmpty` accepts `"   "`, `blankToNull` then turns it into `null`, and the request goes out with no recipient identifier while validation reports success. `whitespaceOnlyRecipients_failValidation` covers channel, user and the Teams channel ID.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/model/Recipient.java
```diff
@@ -0,0 +1,137 @@
...
+      return (email != null && !email.isBlank())
+          || (candidateUsers != null && !candidateUsers.isEmpty())
+          || (candidateGroups != null && !candidateGroups.isEmpty());
+    }
+  }
+
+  /** A Microsoft Teams channel, addressed by its channel ID. */
+  @TemplateSubType(id = TeamsRecipient.TYPE, label = "Microsoft Teams")
+  record TeamsRecipient(
+      @NotEmpty
```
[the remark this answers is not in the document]
yhaskell: Same fix as the Slack targets, in 316727e3f — `@NotBlank` on the Teams `channelId`, with a whitespace case in the tests. I also applied it to `CreateChannelRequest.displayName` and `TeamsChannelPlatform.teamId`, which had the same `@NotEmpty` + blank-normalisation combination.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsConnector.java
```diff
@@ -34,7 +35,7 @@
 @ElementTemplate(
     id = "io.camunda.connectors.AppIntegrations.v1",
     name = "App Integrations Connector",
-    version = 1,
+    version = 2,
```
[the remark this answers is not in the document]
yhaskell: Fixed in 316727e3f. The connector-level description and keywords now read "Microsoft Teams and Slack" and include `slack`, `block kit` and `form`, so marketplace-level discovery matches the operation-level metadata. Both templates regenerated.

element-template-generator/core/src/main/java/io/camunda/connector/generator/java/util/LinkedResourcePropertiesUtil.java
```diff
@@ -82,13 +82,34 @@ static List<PropertyBuilder> buildLinkedResourcePropertiesCore(
                 + linkedResource.linkName()
                 + "'. Each linked resource on the same class must have a unique linkName.");
       }
+      for (var condition : linkedResource.conditions()) {
+        if (condition.property().isBlank() || condition.equals().isBlank()) {
+          throw new IllegalArgumentException(
+              "@TemplateLinkedResource(linkName='"
+                  + linkedResource.linkName()
+                  + "') declares a condition with a blank property or equals value. Both are required.");
```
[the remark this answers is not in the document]
yhaskell: Fair point, fixed in d133d4dbd — the annotation advertised four condition forms and the implementation honoured one.

Rather than narrowing the type, `conditions` now delegates to the existing public `TemplatePropertyAnnotationProcessor.transformToNestedCondition`, so `equals`, `equalsBoolean`, `oneOf` and `isActive` all work and the hand-rolled validation is gone — that method already validates, with the same messages ordinary property conditions produce. Only the operation-scope ID prefix is reapplied afterwards (`withIdPrefix`).

Two checks that this was behaviour-preserving: the app-integrations templates regenerate byte-identical, and a new test asserts a `oneOf`-gated linked resource really emits a `OneOf` condition. The invalid-condition fixture moved to a case the shared validator rejects (`equals` without `property`), which throws `IllegalStateException` like every other malformed property condition.

yhaskell:
Review round addressed — all six Copilot findings were legitimate, none were false positives. Details are in the individual threads; summary here.

**Substantive fixes** (`316727e3f`, `d133d4dbd`)

| Finding | Fix |
| --- | --- |
| Form gates not mutually exclusive | Each form linked resource now gated on **both** `recipient.type` and its own `*Extra.type`, matching what the generator already does for ordinary nested properties |
| `@NotEmpty` accepts whitespace | `@NotBlank` on Teams `channelId`, both Slack targets, `displayName`, `teamId` — plus whitespace test cases |
| Connector metadata Teams-only | Description and keywords now cover Slack, Block Kit and forms |
| README OAuth literals | `credentialsBody` / `basicAuthHeader`, not the hyphenated forms |
| `conditions` advertised more than it supported | Delegates to the existing `transformToNestedCondition`, so `equals` / `equalsBoolean` / `oneOf` / `isActive` all work |

The first one was a genuine defect, not just a consistency nit: gating only on the inner discriminator let a stale value from a previously selected recipient branch keep matching, which could emit a second `zeebe:linkedResource` — after which `formResourceKey` silently takes whichever came first.

**CodeQL (6 alerts)** — unread pattern variables in the additional-content switch became zero-component record patterns (`Form()` / `None()`); the new test fixtures now reference their `@Operation` parameters.

**CI notes**

- `check-pull-request-name` was failing on the `!` breaking-change marker — the regex expects `: ` straight after the optional scope. Title adjusted; the breaking change is described in the PR body.
- `run-tests` failed once on `McpRemoteClientRegistryTest.doesNotCacheIfCacheDisabledOrConfiguredToZeroCacheSize`, which is unrelated to this branch: it touches no `agentic-ai` files, the test passes locally, and it sleeps 10ms between two cache lookups. Re-run and green. Flagging in case it turns up elsewhere — see also #5784.
- `QA-approval` is a human gate and stays red until a QA reviewer signs off.

**Verification after the rebase onto main**: 51 connector tests, 253 element-template-generator tests, element-template validator clean at 674 templates, spotless clean, and both templates regenerate byte-identical to what is committed.

Still outstanding and unchanged: the backend does not yet accept `platform` / `userId` / `blocks` / JSON cards / Slack, so this cannot work end-to-end until that lands.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
```diff
@@ -79,24 +100,156 @@ class AppIntegrationsExecutor {
...
+  private MessagePayload messagePayload(SendMessageRequest request, String formResourceKey) {
+    String platform;
+    String email = null;
+    List<String> candidateUsers = null;
+    List<String> candidateGroups = null;
+    String channelId = null;
+    String userId = null;
+    switch (request.recipient()) {
+      case Recipient.CamundaRecipient camunda -> {
+        platform = PLATFORM_CAMUNDA;
```
mathias-vandaele: `platform = Recipient.CamundaRecipient.TYPE;`

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
```diff
@@ -79,24 +100,156 @@ class AppIntegrationsExecutor {
...
+        email = blankToNull(camunda.email());
+        candidateUsers = emptyToNull(camunda.candidateUsers());
+        candidateGroups = emptyToNull(camunda.candidateGroups());
+      }
+      case Recipient.TeamsRecipient teams -> {
+        platform = PLATFORM_TEAMS;
```
mathias-vandaele: `platform = Recipient.TeamsRecipient.TYPE;`

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
```diff
@@ -79,24 +100,156 @@ class AppIntegrationsExecutor {
...
+        channelId = blankToNull(teams.channelId());
+      }
+      case Recipient.SlackRecipient slack -> {
+        platform = PLATFORM_SLACK;
```
mathias-vandaele: `platform = Recipient.SlackRecipient.TYPE; `

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
```diff
@@ -61,11 +70,23 @@ class AppIntegrationsExecutor {
...
+  private static final String OAUTH_CLIENT_AUTHENTICATION_ENV_VAR =
+      "APP_INTEGRATIONS_OAUTH_CLIENT_AUTHENTICATION";
+
+  static final String NOT_CONFIGURED_ERROR_CODE = "APP_INTEGRATIONS_NOT_CONFIGURED";
+
+  // Sent as an explicit discriminator: Teams and Slack both address a "channel", so the field shape
+  // alone does not tell the backend which platform to target.
+  private static final String PLATFORM_CAMUNDA = "camunda";
+  private static final String PLATFORM_TEAMS = "teams";
+  private static final String PLATFORM_SLACK = "slack";
```
mathias-vandaele: We can remove this block

mathias-vandaele (approved):
minor nit peak but approved :+1:

mathias-vandaele:
```
This is deliberate and stronger than deprecation: a diagram already bound to v1 loses its
template binding rather than merely warning. Acceptable only because the connector is days old and
pre-GA. If we would rather keep compatibility, the alternative is to commit
versioned/app-integrations-connector-1.json byte-identical to main's template and add a
deprecated block in a follow-up.
```

So we deliberately accept that no one will have access to the old ET in 8.9 ? 🤔 
no versioning at all ? why would we do this ?

yhaskell:
> ```
> This is deliberate and stronger than deprecation: a diagram already bound to v1 loses its
> template binding rather than merely warning. Acceptable only because the connector is days old and
> pre-GA. If we would rather keep compatibility, the alternative is to commit
> versioned/app-integrations-connector-1.json byte-identical to main's template and add a
> deprecated block in a follow-up.
> ```
> 
> So we deliberately accept that no one will have access to the old ET in 8.9 ? 🤔 no versioning at all ? why would we do this ?

@mathias-vandaele We don't actually deploy connector in 8.9 at all, it'll only exist in 8.10+, and the first version of connector didn't work at all, so it's totally safe to remove it altogether :)

*Update:* after a short slack discussion we downgraded the version to v1 so it'll silently replace the old version, which is what we want.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
yhaskell: Done — the block is gone. It was a second copy of values the model already owns, so the executor could have drifted from the discriminator that Jackson and the element template agree on. Both payload builders now read the subtype constants directly, and `ChannelPlatform` gained the `TYPE` constants that `Recipient` already had (also feeding `@TemplateSubType(id = ...)`, so the literal exists once per subtype). The "why an explicit platform discriminator" note moved onto `MessagePayload`, where the field lives. Templates regenerate byte-identical.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
yhaskell: Applied — `platform = Recipient.CamundaRecipient.TYPE;`, and the local `PLATFORM_*` constants are removed.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
yhaskell: Applied — `platform = Recipient.TeamsRecipient.TYPE;`. `createChannelPayload` does the same via `ChannelPlatform.TeamsChannelPlatform.TYPE`, which the subtype gained for this.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/AppIntegrationsExecutor.java
yhaskell: Applied — `platform = Recipient.SlackRecipient.TYPE;`, with `ChannelPlatform.SlackChannelPlatform.TYPE` on the channel side.

connectors/app-integrations/src/main/java/io/camunda/connector/appintegrations/model/SendMessageRequest.java
```diff
@@ -8,65 +8,74 @@
...
-            optional = true)
-        String email,
-    @TemplateProperty(
-            group = "message",
-            label = "Channel ID",
-            description =
-                "Microsoft Teams channel ID to send to directly, e.g. 19:xxx@thread.tacv2. Use when sending to a channel rather than a user.",
-            optional = true)
-        String channelId,
+    @NotNull @Valid Recipient recipient,
```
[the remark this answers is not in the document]
yhaskell: Resolving this differently than the PR originally described, after checking with @mathias-vandaele: the template stays at `version = 1` instead of shipping a v2. A bump would have kept the already-published v1 in the Modeler catalog next to the rework — uniquet indexes the current template plus any `versioned/` snapshots — so both shapes would stay downloadable and a v1-bound diagram would keep deploying against a runtime that no longer accepts its variables. Keeping version 1 overwrites the published template in place and writes nothing to `versioned/`, so the old shape stops existing at the next release, and `engineVersion ^8.10` keeps the replacement off runtimes that cannot execute it. So yes, this is a coordinated breaking migration rather than a compatibility story, and deliberately so: the connector is pre-GA and access to it is not expected to exist yet.

yhaskell:
@mathias-vandaele After implementation of the corresponding app-integrations part I had to redo parts of this PR, so I'm asking for a new review :)

mathias-vandaele (approved):
Thank for the modification @yhaskell

Szik (approved):

Szik (approved):

Szik (approved):

Szik (approved):

PR #8269, opened 2026-08-10
[Backport stable/8.8] refactor(agentic-ai): migrate Google Vertex AI provider to google-genai SDK

nikonovd:
## Description

Manual backport of #8178 to `stable/8.8`. The automated backport bot failed to cherry-pick because 8.8 diverged too far from the PR's base (no A2A gateway, no per-tool-call metadata, no `TimeoutConfiguration`/`ChatModelHttpProxySupport`/`CloseableChatModel` infra), so this reimplements the relevant slice of the migration on top of 8.8's current code instead of cherry-picking directly.

**Core SDK migration:**
- `dev.langchain4j:langchain4j-vertex-ai-gemini` wraps `com.google.cloud:google-cloud-vertexai`, which Google sunset on 2026-06-24. It always builds the hostname as `<location>-aiplatform.googleapis.com`, so `region=global` 404s and global-only models (e.g. Gemini 3) can't be used at all - no config-only workaround exists.
- Switches to `langchain4j-google-genai` / `com.google.genai`, which maps `global` to `https://aiplatform.googleapis.com` at the source.
- Unlike #8178, the timeout is **not** configurable here - it stays the fixed `DEFAULT_MODEL_CALL_TIMEOUT` (3 min) already used by every provider on this branch, since 8.8 has no per-connection timeout config infra. Custom endpoint and HTTP proxy support are likewise out of scope (that infra doesn't exist on 8.8).
- No `CloseableChatModel` wrapper - `GoogleGenAiChatModel`'s underlying client (OkHttp dispatcher/pool) isn't explicitly closed, matching 8.8's existing behavior for every other provider (none of them close today).
- Kept from #8178: SDK-internal retries disabled (`maxRetries(0)` + `HttpRetryOptions.attempts(1)`, avoids stacking with langchain4j's own retry), eager service-account credential scoping to `cloud-platform` (required or the token request fails with `invalid_scope`).

**Gemini 3 thought signature preservation:**
- Gemini 3 rejects a follow-up request whose `functionCall` parts are missing the `thoughtSignature` the model returned with them, so tool calling failed on the second model call. langchain4j carries the signature on `AiMessage.attributes()` keyed by tool call ID, but `ChatMessageConverterImpl` only copied text and tool requests, dropping it once persisted to the conversation history.
- `ToolCall` gains a `metadata` map (BC 3-arg constructor kept for existing callers) to carry provider round-trip data attached to the specific tool call it belongs to.
- `ToolCallMetadataDecorator` un-flattens/re-flattens the attribute and namespaces it by provider ID, so every non-Google provider keeps dropping attributes entirely and a provider switch can't leak a previous provider's persisted metadata.
- `ChatMessageConverter`/`ChatMessageConverterImpl`/`Langchain4JAiFrameworkAdapter` thread `ProviderConfiguration` through the conversation-history round trip so the decorator can dispatch on it.

Not backported (infra doesn't exist on 8.8, or out of scope): configurable endpoint/timeout, HTTP proxy support, A2A/MCP gateway `ToolCall` changes (neither the A2A gateway nor the newer MCP metadata plumbing exist on this branch), the element-template version bump.

## Related issues

closes #

Backport of #8178.

## Checklist

### DIFF
--- a/connectors/agentic-ai/pom.xml
+++ b/connectors/agentic-ai/pom.xml
@@ -113,9 +113,21 @@
       <groupId>dev.langchain4j</groupId>
       <artifactId>langchain4j-open-ai</artifactId>
     </dependency>
+    <!--
+      Brings in com.fasterxml.jackson.module:jackson-module-kotlin at runtime scope, which is
+      REQUIRED: google-genai shades KotlinModule into com.google.genai.shaded.* but ships a
+      META-INF/services entry pointing at the unshaded class, so
+      ObjectMapper.findAndRegisterModules() (called by langchain4j-core) throws
+      ServiceConfigurationError without it. Do not exclude it.
+    -->
     <dependency>
       <groupId>dev.langchain4j</groupId>
-      <artifactId>langchain4j-vertex-ai-gemini</artifactId>
+      <artifactId>langchain4j-google-genai</artifactId>
+    </dependency>
+    <!-- main code imports com.google.genai.* directly (we build the Client ourselves) -->
+    <dependency>
+      <groupId>com.google.genai</groupId>
+      <artifactId>google-genai</artifactId>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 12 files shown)


maff (commented):
Do we also want to port the thought signature support to 8.8 (which IIRC was the main reason for this change)?

nikonovd:
> Do we also want to port the thought signature support to 8.8 (which IIRC was the main reason for this change)?

yep definitely! I added the passthrough of those.

maff (approved):
LGTM, thanks for backporting!

Issue #8053, opened 2026-07-24
Agentic AI: evaluate moving v1 provider implementations to native SDK implementations


maff:
Once the v2 native providers land, evaluate migrating the corresponding v1 implementations to the native SDK implementations where one exists — on a per-provider basis (mix & match).

For example, Anthropic and OpenAI v1 could adopt the native implementation while Google Gemini / Vertex v1 stays on LangChain4j until a native path is ready.

LangChain4j remains the supported v1 implementation and the custom-provider registration path regardless. Optional / opportunistic.

PR #8279, opened 2026-08-10
feat(agentic-ai): add native OpenAI chat model provider

### LINKED ISSUE #8061: Agentic AI: OpenAI provider — OpenAI API (direct) & custom/compatible backends
OpenAI provider — `direct` (OpenAI API; Responses + Chat Completions) and `compatible` (arbitrary base URL + API key / none auth + custom headers/query/body) backends.

Wire mappers: OpenAI Responses and OpenAI Chat Completions (`api` family selector). Ships with reasoning effort + prompt caching. **HTTP proxy support mandatory.**

Backend of the OpenAI provider (#7219).

maff:
## Description

Adds a native OpenAI provider to the agentic AI connector's v2 chat model layer, sibling to the native Anthropic provider.

Two orthogonal, independently configurable axes model OpenAI's actual API surface:

- **API family** — `completions` (Chat Completions) or `responses` (Responses API, default). Each has its own request/response converter and streaming assembler, since the wire shapes and feature sets genuinely differ (e.g. only Responses supports server-side reasoning with encrypted-content replay).
- **Backend** — `openai-api` (OpenAI's own endpoint) or `custom` (any OpenAI-compatible endpoint, e.g. self-hosted/proxied). Kept separate from API family because a custom backend can speak either family.

Both axes follow the sealed-type "backend-subtype wrapping convention" already established by the Anthropic provider, so template generation and config parsing stay consistent across providers.

Key implementation choices:

- Both families always stream internally (`createStreaming`), even for a single non-streaming logical call, and delegate accumulation to the vendor SDK's own `ResponseAccumulator`/`ChatCompletionAccumulator` rather than hand-rolling event folds — this is what the vendor SDK is designed for and keeps the deterministic wire-format tests meaningful.
- Reasoning (Responses only) uses encrypted, stateless content (`store(false)` + `include(reasoning.encrypted_content)`) so turn continuation carries no server-side state dependency.
- Server-tool/provider-specific response fragments round-trip through the existing opaque `ProviderContent` type rather than being modeled as first-class domain concepts, matching the provider-namespaced-metadata convention used elsewhere.
- Terminal failure conditions (context-length exceeded, content-filter/refusal) are raised as typed unchecked exceptions (`ContextWindowExceededException`, `ContentFilteredException`) carrying a `PartialResult` with whatever assistant message/metrics were already built for the turn, rather than folded into the normal stop-reason enum — a refusal in particular surfaces no dedicated stop/finish reason on either API family, so it's detected from the message content itself and treated the same way. Stop-reason mapping otherwise covers the remaining normal outcomes (completion, tool calls, truncation) without throwing.
- Token usage subtracts cached tokens back out of the prompt total, since `AgentMetrics.TokenUsage.inputTokenCount` is a cross-provider invariant meaning "new, non-cached tokens only."
- The assistant message id is taken from the Responses API's own per-message id (`msg_*`), not the enclosing response's envelope id (`resp_*`) — the two are different objects, and only the message id is the semantically correct answer when a turn produces no assistant message at all (tool-only/reasoning-only turns).

E2e coverage mirrors the Anthropic provider's pattern: WireMock wire-format smoke tests for both families/backends, a dedicated advanced-features test for Responses (effort, encrypted-reasoning round-trip, server-tool content round-trip), per-feature sub-process tests for Completions (reasoning effort, prompt caching), and gated real-API acceptance rows for both families.

## Related issues

closes #8061

## Checklist


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
```diff
@@ -0,0 +1,320 @@
...
+    if (!system.isBlank()) {
+      builder.instructions(system);
+    }
+  }
+
+  private void applyMessages(ResponseCreateParams.Builder builder, List<Message> messages) {
+    final List<ResponseInputItem> items = new ArrayList<>();
+    for (final Message message : messages) {
+      switch (message) {
+        case SystemMessage ignored -> {} // hoisted to top-level instructions
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/anthropic/AnthropicContentConverter.java
```diff
@@ -68,8 +69,20 @@
...
         case ObjectContent obj -> blocks.add(ContentBlockParam.ofText(toTextBlockParam(obj)));
-        case ReasoningContent rc -> blocks.add(toReasoningContentBlockParam(rc));
-        case ProviderContent pc -> blocks.add(toProviderContentBlockParam(pc));
+        // A ReasoningContent/ProviderContent block from a different provider (e.g. a prior turn
+        // on OpenAI, left behind by a provider switch) carries a payload shaped for that other
+        // vendor's SDK; convertValue-ing it against Anthropic's ContentBlockParam would either
+        // throw or silently produce garbage, so it's dropped rather than replayed.
+        case ReasoningContent rc when ANTHROPIC_ID.equals(rc.provider()) ->
+            blocks.add(toReasoningContentBlockParam(rc));
+        case ReasoningContent rc -> {
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/anthropic/AnthropicContentConverter.java
```diff
@@ -68,8 +69,20 @@
...
+          // dropped: foreign provider, see comment above
+        }
+        case ProviderContent pc when ANTHROPIC_ID.equals(pc.provider()) ->
+            blocks.add(toProviderContentBlockParam(pc));
+        case ProviderContent pc -> {
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiChatModelFactory.java
```diff
@@ -0,0 +1,151 @@
...
+    final var timeout = connection.timeouts() != null ? connection.timeouts().timeout() : null;
+
+    final var client = buildClient(connection.backend(), timeout, httpProxySupport);
+    final var strategy = strategyFor(connection.api());
+    return new OpenAiChatModel(client, model, strategy);
+  }
+
+  private OpenAiApiFamilyStrategy strategyFor(OpenAiChatModelConfiguration.OpenAiApi api) {
+    return switch (api) {
+      case OpenAiCompletionsApi ignored -> completionsStrategy;
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiChatModelFactory.java
```diff
@@ -0,0 +1,151 @@
...
+      case OpenAiResponsesApi ignored -> responsesStrategy;
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
```diff
@@ -0,0 +1,348 @@
...
+    final List<ResponseInputItem> items = new ArrayList<>();
+    final List<Content> plainContent = new ArrayList<>();
+    for (final Content content : assistant.content()) {
+      switch (content) {
+        case ReasoningContent reasoning
+            when OpenAiChatModelConfiguration.OPENAI_ID.equals(reasoning.provider()) ->
+            items.add(
+                ObjectMappers.jsonMapper()
+                    .convertValue(mergeReasoningText(reasoning), ResponseInputItem.class));
+        case ReasoningContent ignored -> {} // foreign provider, see class-method Javadoc
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
```diff
@@ -0,0 +1,348 @@
...
+        case ProviderContent providerContent
+            when OpenAiChatModelConfiguration.OPENAI_ID.equals(providerContent.provider()) ->
+            items.add(
+                ObjectMappers.jsonMapper()
+                    .convertValue(providerContent.payload(), ResponseInputItem.class));
+        case ProviderContent ignored -> {} // foreign provider, see class-method Javadoc
```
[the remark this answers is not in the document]
maff: Not fixing: this is a known CodeQL false-positive on this codebase. Renaming the switch pattern-binding to `ignored` does not silence "unread local variable" on Java 21 — that only works with Java 22's `_` binding, which is not available here. No code change made.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesResponseConverter.java
```diff
@@ -0,0 +1,294 @@
...
+              raw);
+        }
+        content.add(ProviderContent.providerContent(OPENAI_PROVIDER, raw));
+      }
+    }
+
+    return AssistantMessage.builder()
+        .content(content)
+        .toolCalls(toolCalls)
+        .messageId(response.id())
```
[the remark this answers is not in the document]
maff: Confirmed and fixed in f6f74c2898: `OpenAiResponsesResponseConverter` was stamping `Response#id()` (the `resp_*` envelope id) onto `AssistantMessage#messageId()`. `OpenAiResponsesRequestConverter` replays that as `ResponseOutputMessage.id`, which the API validates against the `msg_*` output-item namespace, so it now captures the id off the message output item itself instead. Updated the request/response converter tests accordingly (existing fixtures already had distinct `resp_123`/`msg_1` ids, so the assertion was silently checking the wrong one).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesResponseConverter.java
```diff
@@ -0,0 +1,294 @@
...
+    if (incompleteReason.isPresent()) {
+      final var reason = incompleteReason.get();
+      return switch (reason.value()) {
+        case MAX_OUTPUT_TOKENS -> StopReason.LENGTH;
+        case CONTENT_FILTER -> StopReason.CONTENT_FILTERED;
+        default -> new StopReason.UnknownStopReason(reason.asString());
+      };
+    }
+
+    return hasToolCalls ? StopReason.TOOL_USE : StopReason.STOP;
```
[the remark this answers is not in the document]
maff: Confirmed and fixed in 60f81a24c7: a terminal `response.failed` event carries the same `Response` shape as a normal completion, with `status: failed` and `error` populated instead — no `incomplete_details`, so `mapStopReason` fell through to a plain `STOP`/`TOOL_USE` and the converter reported a successful `ChatResult.Completed`. Added an upfront check that throws a `ConnectorException` under the existing `FAILED_MODEL_CALL` error code, mirroring how `OpenAiChatModel` already reports transport/SDK-level failures. Added a unit test covering the failed-response case.

connectors/agentic-ai/docs/reference/ai-agent.md
```diff
@@ -1831,11 +1832,105 @@ base class). Reference implementation: `AnthropicChatModelFactory` with
...
+whenever effort is set, additionally sets `store(false)` and requests
+`include: ["reasoning.encrypted_content"]`; `store(false)` is correct because the connector persists
+conversation state itself, so OpenAI-side state would only compete with the agent context for
+authority. `OpenAiResponsesResponseConverter` captures the returned `reasoning` output item as
+`ReasoningContent` whose `payload` is the **full raw item**, built with the SDK's own
+`com.openai.core.ObjectMappers.jsonMapper()` rather than the injected app `ObjectMapper` (the wrong
+mapper leaks a spurious field onto the wire), and replays it verbatim on the next turn. Unlike
+Anthropic — which lifts the human-readable thinking text out into `ReasoningContent.text` and merges it
+back before replay — OpenAI's reasoning summary stays inside the raw payload, so `text` is left `null`
+and replay is a straight `convertValue` with no reassembly step. There is no runtime validation that a
```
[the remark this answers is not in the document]
maff: Fixed in 86384eacff: the doc was wrong, `OpenAiResponsesResponseConverter` does lift a non-empty `summary` onto `ReasoningContent.text` the same way Anthropic's `thinking` blocks are handled, and conditionally strips it back out of `payload` when reconstructable. Rewrote the paragraph to describe the actual behavior.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiChatModelFactory.java
```diff
@@ -0,0 +1,151 @@
...
+  private static void applyCustomBackend(
+      OpenAIOkHttpClient.Builder builder, OpenAiCustomBackend custom) {
+    final var connection = custom.custom();
+    builder.baseUrl(connection.endpoint());
+
+    final String apiKey =
+        connection.authentication() instanceof ApiKeyAuthentication apiKeyAuth
+            ? apiKeyAuth.apiKey()
+            : NO_AUTH_PLACEHOLDER_API_KEY;
+    builder.apiKey(apiKey);
```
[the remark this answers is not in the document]
maff: Investigated. There is no supported way to build a client that sends zero auth headers with this SDK version. `credential` (or `workloadIdentity`/`adminApiKey`) is a required field on the builder — the client throws `IllegalStateException: credential is required, but was not set` otherwise, verified directly against the pinned 4.50.0 jar.

4.50.0 does add `httpRequestAuthenticator(HttpRequestAuthenticator)`, which looks like exactly the right hook (a per-request `HttpRequest -> HttpRequest` no-op would send nothing), but it is compiled with the JVM `ACC_SYNTHETIC` flag (Kotlin `@JvmSynthetic`) — confirmed with `javap -v` and reproduced in isolation. javac refuses to resolve synthetic members from Java source at all ("cannot find symbol"), so it is not actually part of this SDK's supported Java API surface in this version, only reachable via reflection or from Kotlin callers.

Given that, the placeholder-`apiKey` approach already in `OpenAiChatModelFactory` is the best available option under the current pinned SDK version — no code change. Left a note in team memory to revisit if/when we bump the openai-java dependency past whatever version stops marking that method synthetic.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/anthropic/AnthropicContentConverter.java
```diff
@@ -143,20 +156,24 @@ private ContentBlockParam toContentBlockParam(Object payload) {
...
   }
 
+  /**
+   * Converts a tool result's structured content into Anthropic tool-result blocks. Unlike {@link
+   * #toContentBlockParams}, a document here is flattened to a JSON reference rather than embedded
+   * natively as an image/document block: the composer's synthetic {@code <doc/>} fallback message
+   * already delivers the actual document bytes for tool results (see {@code
+   * AgentConversationTurnInputComposerImpl}), so embedding it here as well would send it to the
+   * model twice.
+   */
```
maff, line 146: ℹ️ Reverted a fix added while implementing the Anthropic provider - this makes sure we don't send documents twice as part of a tool call result (synthetic message + tool call result) by not supporting it in the converter for the moment. Will be adapted when we can conditionally create the synthetic message based on model capabilities.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/v2/AnthropicCustomEndpointAuthentication.java
```diff
@@ -15,14 +15,18 @@
 import jakarta.validation.constraints.NotBlank;
 
 /**
- * Authentication strategies shared by "custom"-backend endpoints. Extensible: more schemes can be
```
maff, line 18: ℹ️ Moved from a shared object to an anthropic-specific one. OpenAI does not allow custom endpoints without setting any credentials (the SDK enforces at least one of the known authentication methods), so we are not able to support "none" on OpenAI.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/tool/ToolCallResultContent.java
```diff
@@ -89,6 +91,7 @@ static List<Content> contentFromObject(@Nullable Object content) {
     return switch (content) {
       case null -> List.of();
       case String s -> s.isBlank() ? List.of() : List.of(TextContent.textContent(s));
+      case Document document -> List.of(DocumentContent.documentContent(document));
```
maff, line 94: ℹ️ Part of the fix revert mentioned in the anthropic content converter.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/anthropic/AnthropicContentConverter.java
```diff
@@ -198,51 +191,19 @@ private enum DocumentBlockKind {
...
+    if (DocumentMimeTypes.isPdf(contentType)) {
       return DocumentBlockKind.PDF;
     }
-
-    final var mime = parsed.getMimeType();
-    if (mime.startsWith("text/")
-        || isCompatibleWithAnyOf(parsed, ADDITIONAL_TEXT_FILE_CONTENT_TYPES)
-        || mime.equals("application/x-yaml")
-        || mime.endsWith("+json")
-        || mime.endsWith("+xml")) {
```
maff, line 228: ℹ️ Extracted to `DocumentMimeTypes` helper.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiCompletionsPromptCachingTests.java
nikonovd, line 1: 💭 IDK if this test adds much benefit besides unit testing, because we only assert that once a response (which is mocked here as well) provides a value for cached tokens, they show up in the agent context.

This is also verified with unit tests. What's your opinion here?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiResponsesAdvancedFeaturesTests.java
```diff
@@ -0,0 +1,216 @@
...
+                .isReady()
+                .hasResponseText(responseText));
+  }
+
+  // ---------------------------------------------------------------------------
+  // Byte-identical encrypted-reasoning round-trip
+  // ---------------------------------------------------------------------------
+
+  @Test
+  void reasoningItemWithEncryptedContentRoundTripsByteIdenticalOnFollowUpRequest()
```
nikonovd: 💭 Could we merge this test with `configuredEffortAppearsAsReasoningEffortOnTheWire` to avoid having a second test execution just to verify that reasoning effort appears on wire?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
```diff
@@ -0,0 +1,89 @@
...
+        .property("provider.type", "openai")
+        .property("provider.openai.api.type", "completions")
+        .property("provider.openai.backend.type", "custom")
+        .property("provider.openai.backend.custom.endpoint", wireMock.getHttpBaseUrl() + "/v1")
+        .property("provider.openai.backend.custom.authentication.type", "apiKey")
+        .property("provider.openai.backend.custom.authentication.apiKey", "dummy")
+        .property("provider.openai.model.model", defaultModel());
+  }
+
+  static LoggedRequest soleRecordedRequest() {
```
nikonovd: 🔧 i think the method name is accidentally not clear enough
 WDYT about splitting the assertion and a retrieval method up into two blocks?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
```diff
@@ -0,0 +1,89 @@
...
+   */
+  protected String defaultModel() {
+    return DEFAULT_MODEL;
+  }
+
+  private ElementTemplate configureOpenAiCompletionsBackend(ElementTemplate template) {
+    return template
+        .property("provider.type", "openai")
+        .property("provider.openai.api.type", "completions")
+        .property("provider.openai.backend.type", "custom")
```
nikonovd: ❓ Do we need to use `custom` here if the endpoint is a hidden property on the template?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiResponsesSubProcessTest.java
```diff
@@ -0,0 +1,53 @@
...
+   */
+  protected String defaultModel() {
+    return DEFAULT_MODEL;
+  }
+
+  private ElementTemplate configureOpenAiResponsesBackend(ElementTemplate template) {
+    return template
+        .property("provider.type", "openai")
+        .property("provider.openai.api.type", "responses")
+        .property("provider.openai.backend.type", "custom")
```
nikonovd: ❓ Do we need to use `custom` here if the endpoint is a hidden property on the template?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/openai/OpenAiCompletionsV2SseChatModelStubs.java
```diff
@@ -0,0 +1,258 @@
...
+ * delta), and a trailing usage-only chunk ({@code choices: []} with the final {@code usage}) -
+ * mirroring real {@code stream_options.include_usage=true} behavior, which the native provider's
+ * request converter always sets.
+ *
+ * <p>{@link #stubConversation(TurnStub...)} always renders {@code prompt_tokens_details} / {@code
+ * completion_tokens_details} with zero-valued sub-fields, since the generic {@link TurnStub} SPI
+ * has no dial for them; use {@link #stubConversation(UsageDetailsTurnStub)} for a single turn whose
+ * cache-read/reasoning token counts need to be non-zero.
+ */
+public final class OpenAiCompletionsV2SseChatModelStubs {
```
nikonovd, line 69: 💭 Not blocking, we can also think of a followup.

But I have two points I'd like to explore:

1. Do we really need V1 and V2 stubs, as those should mostly be identical from the API perspective - it's rather Completions and Responses that differentiate
2. Do we go the easy road with `SseStubs` and `HttpStubs` or do we integrate both into a single builder with different output serialization?

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/openai/OpenAiResponsesV2WireFormatFixture.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @Override
+  public Function<ElementTemplate, ElementTemplate> configureProvider(
+      WireMockRuntimeInfo wireMock) {
+    return template ->
+        template
+            .property("provider.type", "openai")
+            .property("provider.openai.api.type", "responses")
+            .property("provider.openai.backend.type", "custom")
```
nikonovd: ❓ Do we need to use `custom` here if the endpoint is a hidden property on the template?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/anthropic/AnthropicMessageResponseConverter.java
```diff
@@ -65,6 +73,19 @@ public ChatResult toResult(Message message, Duration executionTime) {
     final AgentMetrics metrics =
         toMetrics(message, assistantMessage.toolCalls().size(), executionTime);
 
+    if (assistantMessage.stopReason() instanceof UnknownStopReason unknown
+        && StopReason.MODEL_CONTEXT_WINDOW_EXCEEDED.asString().equals(unknown.value())) {
+      throw new ContextWindowExceededException(
+          "Model's context window was exceeded before it could finish generating a response.",
+          new PartialResult(assistantMessage, metrics));
```
nikonovd, line 73: 🚀 awesome to see it being explicitly raised as an error!

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/LangChain4JChatModel.java
```diff
@@ -89,6 +93,14 @@ public ChatResult execute(io.camunda.connector.agenticai.aiagent.chatmodel.ChatR
     final AssistantMessage assistantMessage =
         chatMessageConverter.toAssistantMessage(chatResponse, providerConfiguration);
     final var metrics = buildMetrics(chatResponse, assistantMessage, executionTime);
+
+    if (assistantMessage.stopReason() instanceof UnknownStopReason unknown
+        && FinishReason.CONTENT_FILTER.name().equals(unknown.value())) {
+      throw new ContentFilteredException(
+          "Model response was blocked by provider content filtering.",
+          new PartialResult(assistantMessage, metrics));
+    }
```
nikonovd, line 102: 💭 I know it got introduced in this PR, but could we maybe delegate fixes around non-related parts to other PRs?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
```diff
@@ -0,0 +1,329 @@
...
+      builder.temperature(params.temperature());
+    }
+    if (params.topP() != null) {
+      builder.topP(params.topP());
+    }
+  }
+
+  /**
+   * Maps the {@code effort} dial onto the SDK's {@code reasoning_effort} param. Unlike the
+   * Responses sibling's {@code applyReasoning}, there is no encrypted-content include: Completions
```
nikonovd: 🔧 as already discussed the javadoc should be isolated to its area of responsibility, not narrating through the implementation journey

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
```diff
@@ -0,0 +1,329 @@
...
+  private void applyReasoning(
+      ChatCompletionCreateParams.Builder builder, @Nullable CompletionsParameters params) {
+    final OpenAiEffort effort = params == null ? null : params.effort();
+    if (effort == null) {
+      return;
+    }
+    builder.reasoningEffort(ReasoningEffort.of(effort.name().toLowerCase(Locale.ROOT)));
+  }
+
+  private void applyMessages(ChatCompletionCreateParams.Builder builder, List<Message> messages) {
```
nikonovd, line 152: ⛏️ we should tweak our agentic coding guidelines, sometimes the traditional for-loops are convoluted in contrast to streams. WDYT

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsResponseConverter.java
```diff
@@ -0,0 +1,206 @@
...
+
+    final List<Content> content = new ArrayList<>();
+    message
+        .content()
+        .filter(text -> !text.isBlank())
+        .ifPresent(text -> content.add(TextContent.textContent(text)));
+    // A refusal has no dedicated domain content type; kept as TextContent so the declination stays
+    // visible in the partial result once toResult() throws ContentFilteredException for it (see
+    // hasRefusal).
+    message.refusal().ifPresent(refusal -> content.add(TextContent.textContent(refusal)));
```
nikonovd, line 118: 💭 With this mapping we don't have any explicit mapping on the request conversion, so the refusal itself gets lost in the conversation history.

Is that intended?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
```diff
@@ -0,0 +1,358 @@
...
+            items.add(
+                ObjectMappers.jsonMapper()
+                    .convertValue(mergeReasoningText(reasoning), ResponseInputItem.class));
+        case ReasoningContent ignored -> {} // foreign provider, see class-method Javadoc
+        case ProviderContent providerContent when OPENAI_ID.equals(providerContent.provider()) ->
+            items.add(
+                ObjectMappers.jsonMapper()
+                    .convertValue(providerContent.payload(), ResponseInputItem.class));
+        case ProviderContent ignored -> {} // foreign provider, see class-method Javadoc
+        default -> plainContent.add(content); // Text/Object/Document: replayed as a message below
```
nikonovd, line 217: ❓ Do we then add documents twice here accidentally?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/OpenAiRequestCustomizations.java
```diff
@@ -0,0 +1,44 @@
...
+          new OpenAiRequestCustomizations(
+              apiBackend.openai().headers(),
+              apiBackend.openai().queryParameters(),
+              apiBackend.openai().bodyProperties());
+      case OpenAiCustomBackend custom ->
+          new OpenAiRequestCustomizations(
+              custom.custom().headers(),
+              custom.custom().queryParameters(),
+              custom.custom().bodyProperties());
+    };
```
nikonovd: 🔧 we could inverse the Customizations construction to the Backends with an interface helper, as there is no implementation that doesn't yield any of those.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesResponseConverter.java
```diff
@@ -0,0 +1,332 @@
...
+    return ObjectMappers.jsonMapper()
+        .convertValue(item, new TypeReference<Map<String, Object>>() {});
+  }
+
+  private AgentMetrics toMetrics(Response response, int toolCalls, Duration executionTime) {
+    final AgentMetrics.TokenUsage tokenUsage =
+        response.usage().map(this::toTokenUsage).orElseGet(AgentMetrics.TokenUsage::empty);
+
+    return AgentMetrics.builder()
+        .modelCalls(1)
```
nikonovd, line 300: 💭 Followup idea: having a `AgentMetrics.afterModelCall()` to avoid `builder().modelCalls(1)`

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiChatModel.java
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      return strategy.call(client, configuration, request);
+    } catch (ConnectorException e) {
+      // already coded (e.g. OpenAiContentConverter's unsupported content type); avoid
+      // double-wrapping as a generic "Model call failed" below.
+      throw e;
+    } catch (BadRequestException e) {
+      // an over-length request is rejected outright with an HTTP 400, on both API families - this
+      // error code is the only signal for it. No partial result exists: the request was rejected
+      // before any response body could be converted.
```
nikonovd, line 71: If that's the case, why do we / does the provider show a stop reason that exceeds the context window?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/message/StopReason.java
```diff
@@ -19,20 +19,17 @@
...
     permits StopReason.KnownStopReason, StopReason.UnknownStopReason {
 
   StopReason STOP = KnownStopReason.STOP;
   StopReason LENGTH = KnownStopReason.LENGTH;
   StopReason TOOL_USE = KnownStopReason.TOOL_USE;
-  StopReason CONTENT_FILTERED = KnownStopReason.CONTENT_FILTERED;
-  StopReason CONTEXT_WINDOW_EXCEEDED = KnownStopReason.CONTEXT_WINDOW_EXCEEDED;
-  StopReason GUARDRAIL = KnownStopReason.GUARDRAIL;
-  StopReason ERROR = KnownStopReason.ERROR;
-  StopReason ABORTED = KnownStopReason.ABORTED;
```
nikonovd, line 35: ❓ So if I got it correct, we now only carry valid stop reasons as domain models, that are not leading to exceptional errors, correct?

connectors/agentic-ai/connector-agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentConversationTurnInputComposerImplTest.java
```diff
@@ -406,11 +406,12 @@ void toolResultTurn_resultsContainDocuments_emitsToolCallDocumentMessage() {
   }
 
   @Test
-  void toolResultTurn_bareDocumentResult_deliversDocumentOnlyViaFallbackMessage() {
+  void toolResultTurn_bareDocumentResult_liftsToDocumentContentAndAlsoDeliversFallbackMessage() {
     // a tool call result whose whole content is a Document (not nested in an object): the
-    // ToolCallResultContent lift must not turn this into a first-class DocumentContent, since a
-    // native provider tool-result converter would then embed the same bytes a second time — the
-    // fallback <doc/> message stays the document's only delivery channel
+    // ToolCallResultContent lift turns this into a first-class DocumentContent, same as every
```
nikonovd, line 411: 💭 Shouldn't we now inverse this relationship (in a follow up after code freeze)?

The results are _always_ plain tool call results with document contents and the provider specific strategy decides if they want to do a fallback message or not?

I guess it's something for the multimodality epic. Should we document this as a separate issue?

nikonovd (commented):
Looks good, I left some comments. Let's clarify them.

Some general remarks from my side:
* The javadoc and comments are very verbose, let's take a look on our AGENTS.md setup 🍪 
* As discussed in our weekly, some wording is counterintuitive (lift, canningResponse, etc) and we should take a look here as well.

Pre-approving as nothing is blocking from my side 🚀

nikonovd (approved):
Looks good, I left some comments. Let's clarify them.

Some general remarks from my side:
* The javadoc and comments are very verbose, let's take a look on our AGENTS.md setup 🍪 
* As discussed in our weekly, some wording is counterintuitive (lift, canningResponse, etc) and we should take a look here as well.

Pre-approving as nothing is blocking from my side 🚀 

I didn't check every minor detail here.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/openai/OpenAiCompletionsV2SseChatModelStubs.java
maff: Unfortunately the difference is also streaming vs non-streaming, that's why we have a completely different path here. What I could imagine is to land these changes first, but then route existing v1 connectors to the new implementations as part of https://github.com/camunda/connectors/issues/8053 - this would allow us to get rid of 2 sets of stubs completely.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/message/StopReason.java
maff: Exactly - as model providers differ in the way they respond on terminal situations (stop reason on a 200 OK response vs a 400 response without an assistant message), these stop reasons now only cover what we can actually return as part of an assistant message, combined with defined domain exceptions under `ChatModelRejectedException`.

connectors/agentic-ai/connector-agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/agent/AgentConversationTurnInputComposerImplTest.java
maff: Yes, exactly - we should tackle this with the multimodality and model awareness. Right now, every document in the tool call results message is the document reference + the exctracted document in the synthetic user message. With the document awareness epic we can decide where to put that based on capabilities.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
maff: Fixed — javadoc now states this method's own contract only, no more sibling-comparison narration. Also swept the same pattern across the rest of the new Anthropic/OpenAI provider files (`4ab8bce16f`).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/OpenAiRequestCustomizations.java
maff: Done — `OpenAiBackend` now exposes `headers()`/`queryParameters()`/`bodyProperties()` directly (implemented by both `OpenAiApiBackend` and `OpenAiCustomBackend`), so `OpenAiRequestCustomizations.from()` just delegates instead of switching (`2dd6790aec`).

maff: Update: reworked this further per feedback -- the backend now returns the OpenAiRequestCustomizations object itself via a single requestCustomizations() method (moved into the model package alongside OpenAiChatModelConfiguration so it can build it directly), rather than exposing 3 leaf accessors for the family package to assemble. from(OpenAiConnection) is gone entirely. Superseded 2dd6790aec with 78ca6cc354.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
maff: Agreed on the general point, but leaving this loop as-is here -- not the scope of this PR to revisit the coding guidelines. Happy to pick that up as its own discussion separately.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesResponseConverter.java
maff: Good idea, filing it as a followup rather than doing it here.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Good catch -- checked, and it turns out neither OpenAI nor Anthropic e2e coverage ever exercises the branded/native backend type (`openai`/`anthropic`), only `custom`, even though the branded backend's `endpoint` is a hidden-but-settable property. Not fixing the existing tests (would only swap which single path gets covered), but worth a followup to add dedicated branded-backend e2e coverage.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiResponsesSubProcessTest.java
maff: Same as the Completions-family thread -- confirmed there's no e2e coverage of the branded backend type at all on either provider, only `custom`. Filing a followup for dedicated branded-backend e2e coverage rather than changing this test.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/openai/OpenAiResponsesV2WireFormatFixture.java
maff: Same answer as the other two `custom` threads -- real gap, but a followup rather than a change here.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/openai/OpenAiCompletionsV2SseChatModelStubs.java
maff: Agreed not blocking -- leaving both questions (V1/V2 stub duplication, SseStubs/HttpStubs unification) as a followup rather than reshaping this now.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiCompletionsPromptCachingTests.java
```diff
@@ -0,0 +1,77 @@
+/*
```
maff: Keeping it -- it's the regression guard for the prompt-caching investigation (no repro found, fix deferred), and it exercises the full SSE-accumulator-to-metrics path rather than just a converter's mapping shape, which is more than the unit tests cover on their own.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiResponsesAdvancedFeaturesTests.java
maff: Done -- merged into the byte-identical round-trip test (now configuredEffortRoundTripsByteIdenticalOnFollowUpRequest), which already used effort("high"). The wire-level reasoning.effort/include[] assertions now run against the first request there; the standalone test is gone (ccfbf5f1ab).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Renamed to assertSoleRecordedRequest(), with a javadoc pointing at recordedLoggedRequests() for the bare-retrieval case -- name now makes clear it does both the assert and the retrieval (ccfbf5f1ab).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/OpenAiRequestCustomizations.java
maff: Done - now exposed by the backend directly.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiResponsesAdvancedFeaturesTests.java
maff: Done -- merged into the byte-identical round-trip test (now configuredEffortRoundTripsByteIdenticalOnFollowUpRequest), which already used effort("high"). The wire-level reasoning.effort/include[] assertions now run against the first request there; the standalone test is gone (ccfbf5f1ab).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/OpenAiRequestCustomizations.java
maff: Done - now exposed by the backend directly.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Split into assertSoleRecordedRequest() (void, just the size==1 assertion) and soleRecordedRequest() (returns recordedLoggedRequests().get(0), no assertion) -- (e7d8c99cc8).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Switched to the openai-api backend with its hidden endpoint override instead of custom, here and in the other 3 v2 OpenAI e2e config spots (e7d8c99cc8).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiResponsesAdvancedFeaturesTests.java
maff: Done -- merged into the byte-identical round-trip test (now configuredEffortRoundTripsByteIdenticalOnFollowUpRequest), which already used effort("high"). The wire-level reasoning.effort/include[] assertions now run against the first request there; the standalone test is gone (ccfbf5f1ab).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/OpenAiRequestCustomizations.java
maff: Done - now exposed by the backend directly.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Switched to the openai-api backend with its hidden endpoint override instead of custom, here and in the other 3 v2 OpenAI e2e config spots (a2dfc3c4ee).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiResponsesAdvancedFeaturesTests.java
maff: Done -- merged into the byte-identical round-trip test (now configuredEffortRoundTripsByteIdenticalOnFollowUpRequest), which already used effort("high"). The wire-level reasoning.effort/include[] assertions now run against the first request there; the standalone test is gone (ccfbf5f1ab).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/OpenAiRequestCustomizations.java
maff: Done - now exposed by the backend directly.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Switched to the openai-api backend with its hidden endpoint override instead of custom, here and in the other 3 v2 OpenAI e2e config spots (a2dfc3c4ee).

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/BaseOpenAiCompletionsSubProcessTest.java
maff: Combined it with parseBody instead -- parseSoleRecordedRequest() now asserts exactly-one and returns the parsed body directly (45f030ea75).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsResponseConverter.java
maff: Good catch - updated to return the rejection as error variables & we can address ingesting the metrics and potentially storing the message in a follow-up.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiChatModel.java
maff: In this case we don't show a stop reason anymore. As we have no assistant message, we also can't propagate the partial result.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
maff: We definitely can add guidelines to prefer streams - in this case I'm not sure if it would improve as we add multiple items for the tool call results (probably resulting in some `flatMap` construct).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
maff: Checked - no double-add. Each `Content` block in `assistant.content()` is visited exactly once by the for-loop; the switch is exhaustive and mutually exclusive, so a block lands in exactly one arm (reasoning-this-provider / reasoning-other / provider-this-provider / provider-other / default-plain). `plainContent` is built once and converted to wire items exactly once at the end. The double-send bug elsewhere (tool-call-result documents vs. the composer's synthetic `<doc/>` fallback) is a different method (`toolResultInputItems`), unrelated to this path.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
maff: We definitely can add guidelines to prefer streams - in this case I'm not sure if it would improve as we add multiple items for the tool call results (probably resulting in some flatMap construct).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsRequestConverter.java
```diff
@@ -0,0 +1,315 @@
...
+  private ChatCompletionAssistantMessageParam assistantMessage(AssistantMessage assistant) {
+    final var builder = ChatCompletionAssistantMessageParam.builder();
+
+    final List<Content> plainContent =
+        assistant.content().stream()
+            .filter(c -> !(c instanceof ReasoningContent) && !(c instanceof ProviderContent))
+            .toList();
+    if (!plainContent.isEmpty()) {
+      builder.content(toTextOutput(plainContent));
+    }
```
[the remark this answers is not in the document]
maff: Fixed — `assistantMessage()` now returns `null` (message omitted) when both the filtered content and tool calls are empty, instead of building an empty assistant message.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
```diff
@@ -0,0 +1,336 @@
...
+   */
+  private void applyReasoning(
+      ResponseCreateParams.Builder builder, @Nullable ResponsesParameters params) {
+    final OpenAiEffort effort = params == null ? null : params.effort();
+    if (effort == null) {
+      return;
+    }
+    builder.reasoning(Reasoning.builder().effort(mapEffort(effort)).build());
+    builder.addInclude(ResponseIncludable.REASONING_ENCRYPTED_CONTENT);
+  }
```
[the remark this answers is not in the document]
maff: Fixed — `applyReasoning` now requests `REASONING_ENCRYPTED_CONTENT` unconditionally, keeping only `reasoning.effort` itself gated on explicit `effort` config.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/completions/OpenAiCompletionsResponseConverter.java
```diff
@@ -0,0 +1,195 @@
...
+      case LENGTH -> StopReason.LENGTH;
+      default -> new StopReason.UnknownStopReason(finishReason.asString());
+    };
+  }
+
+  private void toToolCall(ChatCompletionMessageToolCall call, List<ToolCall> toolCalls) {
+    if (call.function().isEmpty()) {
+      // Only function tool calls have a provider-neutral representation; custom tool calls are
+      // not supported by the domain model and are silently skipped.
+      return;
```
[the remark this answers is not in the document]
maff: Fixed — `toToolCall` now preserves a custom tool call losslessly as `ProviderContent` (matching the Responses converter's fallback for unrecognized output items) instead of dropping it.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/v2/AgentSubProcessOpenAiResponsesAdvancedFeaturesTests.java
```diff
@@ -0,0 +1,192 @@
...
+
+    final var zeebeTest =
+        awaitProcessCompletion(createProcessInstance(Map.of("userPrompt", userPrompt)));
+
+    final var recorded = OpenAiResponsesV2RecordedConversation.recorded();
+    assertThat(recorded.modelCallCount()).isEqualTo(1);
+
+    final var request = recorded.lastRequest();
+    assertThat(request.reasoningEffort()).as("reasoning.effort").isEmpty();
+    assertThat(request.include()).as("include[]").isEmpty();
```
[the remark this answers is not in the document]
maff: Fixed — renamed to `unsetEffortOmitsReasoningEffortButStillRequestsEncryptedContent` and updated the assertion: `include[]` now expects `reasoning.encrypted_content`, only `reasoning.effort` stays empty.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/family/responses/OpenAiResponsesRequestConverter.java
```diff
@@ -0,0 +1,341 @@
...
+                  .content(
+                      EasyInputMessage.Content.ofResponseInputMessageContentList(
+                          contentConverter.toResponsesContentParts(plainContent)))
+                  .build()));
+    }
+    for (final ToolCall toolCall : assistant.toolCalls()) {
+      items.add(
+          ResponseInputItem.ofFunctionCall(
+              ResponseFunctionToolCall.builder()
+                  .callId(toolCall.id())
```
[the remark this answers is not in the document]
maff: Confirmed real, but the fix can't live here: the ordering is already lost in `OpenAiResponsesResponseConverter#toAssistantMessage`, which splits each turn's ordered output items into two separate lists (`content`/`toolCalls`) with no cross-list position preserved -- and `AssistantMessage` itself has no field to carry that interleaving. This is a shared domain-model limitation (Anthropic's converter does the same split), not specific to this file, so out of scope for this PR. Tracking as a follow-up.

PR #8280, opened 2026-08-11
refactor(agentic-ai): migrate v1 Google Vertex AI provider to google-genai SDK

nikonovd:
## Description

Upstream port of #8178 (already shipped on `stable/8.9`, backported to `stable/8.8` in #8269) to `main`. `main` has since restructured the AI Agent connector into v1 (langchain4j-based, this PR) and v2 (native per-provider implementations, currently Anthropic only) - **this only touches v1**. Google Vertex AI has no v2 implementation yet, so nothing under `model/request/v2/` or `chatmodel/provider/anthropic/` is touched.

**Core SDK migration** (`GoogleVertexAiChatModelFactory`):
- `dev.langchain4j:langchain4j-vertex-ai-gemini` wraps `com.google.cloud:google-cloud-vertexai`, which Google sunset on 2026-06-24. It always builds the hostname as `<location>-aiplatform.googleapis.com`, so `region=global` 404s and global-only models (e.g. Gemini 3) can't be used at all - no config-only workaround exists.
- Switches to `langchain4j-google-genai` / `com.google.genai`, which maps `global` to `https://aiplatform.googleapis.com` at the source.
- Adds `endpoint` and `timeouts` to `GoogleVertexAiProviderConfiguration` (v1), matching every other v1 provider (`deriveTimeoutSetting`, `ChatModelHttpProxySupport.createGoogleGenAiProxyOptions`), and bumps the v1 AI Agent element template from version 11 to 12.
- SDK-internal retries disabled (`maxRetries(0)` + `HttpRetryOptions.attempts(1)`, avoids stacking with langchain4j's own retry), eager service-account credential scoping to `cloud-platform` (required or the token request fails with `invalid_scope`).

**Gemini 3 thought signature preservation:**
- Gemini 3 rejects a follow-up request whose `functionCall` parts are missing the `thoughtSignature` the model returned with them, so tool calling failed on the second model call. langchain4j carries the signature on `AiMessage.attributes()` keyed by tool call ID, but `ChatMessageConverterImpl` only copied text and tool requests, dropping it once persisted to history.
- Unlike the 8.9/8.8 ports of this fix, `main`'s langchain4j layer already gives each provider its own `CloseableChatModel` instance (`LangChain4JChatModelFactory` / `CloseableChatModelDelegate`), so instead of threading a `ProviderConfiguration` through the shared converter (which would couple it to v1's config type), the round trip is wired through that per-provider instance:
  - `CloseableChatModel` gains `decorateOnWrite`/`decorateOnRead` default (no-op) methods - every provider drops attributes unless it overrides them.
  - `GoogleVertexAiCloseableChatModel` overrides them with the thought-signature logic, namespaced by provider ID so a provider switch or process instance migration can't leak a previous provider's persisted metadata into an unrelated one.
  - `ChatMessageConverter`/`Impl` and `LangChain4JChatModel` thread the `CloseableChatModel` through the round trip instead of a `ProviderConfiguration`.

## Related issues

Upstream port of #8178 / #8269.

## Checklist


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/CloseableChatModel.java
```diff
@@ -16,4 +17,33 @@
...
+
+  /**
+   * Called on the read path with the already-unwrapped {@code toolCall.metadata()} (never null, may
+   * be empty, untyped since it round-tripped through a process variable) of a single tool call.
+   * Returns the entry to restore onto the outgoing {@code AiMessage.attributes()}, keyed the way
+   * langchain4j expects.
+   */
+  default Map<String, Object> decorateOnRead(String toolCallId, Map<?, ?> toolCallMetadata) {
+    return Map.of();
+  }
```
maff: 💭 Why do we deviate from the 8.8/8.9 pattern and keep this logic now as part of the actual chat model? Primarily asking as we had discussed it on the original PR. Not generally opposed to keeping it here, but we should probably make the method naming more explicit.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/ToolCallMetadataDecorator.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * <p>Persisted entries are namespaced by the provider's {@code TemplateSubType} ID (currently only
+ * {@link GoogleVertexAiProviderConfiguration#GOOGLE_VERTEX_AI_ID}), so that switching a process
+ * instance's provider - e.g. via a config update or a process instance migration - cannot leak a
+ * previous provider's persisted metadata into an unrelated one.
+ */
+final class ToolCallMetadataDecorator {
+
+  private static final String THOUGHT_SIGNATURE_ATTRIBUTE_KEY_PREFIX = "thought_signature_";
+  private static final String THOUGHT_SIGNATURE_METADATA_KEY = "thoughtSignature";
```
maff: 🔧 We added a `GOOGLE_` prefix on the 8.9 variant here IIRC?

maff (approved):

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/langchain4j/CloseableChatModel.java
nikonovd: Right take and fixed as discussed.

Issue #6961, opened 2026-04-14
Outbound Connector Adaptation for Multi-Engine Support


johnBgood:
**Context:**
Outbound connectors require minimal changes to support multi-Engine. The main update is leveraging the new registry to start job workers for every engine.

**Tasks:**
- Update the outbound connector job worker logic to use the new registry.
    - See reference: [OutboundConnectorManager.java#L119](https://github.com/camunda/connectors/blob/9a2abb544cd7b09555e32424cf6245204b9cf1f4/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/lifecycle/OutboundConnectorManager.java#L119)
- Additional connector isolation per engine is needed.
    - Connectors must be re-created per physical tenant, for isolation

## Prerequistes from OC
`ActivatedJob` needs to provide the `physicalTenantId` in addition to the [existing properties](https://github.com/camunda/camunda/blob/a2764245de80f7cf05cfbfb5205c4627d8c1787b/clients/java/src/main/java/io/camunda/client/api/response/ActivatedJob.java#L25).

deepthidevaki:
@johnBgood Could you also write down the requirements from Camunda Spring Boot client? Mainly the API contracts. For example, `JobWorkerManager#createJobWorker` creates a worker for the given physical tenant or all physical tenants.

johnBgood:
> [@johnBgood](https://github.com/johnBgood) Could you also write down the requirements from Camunda Spring Boot client? Mainly the API contracts. For example, `JobWorkerManager#createJobWorker` creates a worker for the given physical tenant or all physical tenants.

@deepthidevaki there should be no other requirement than the one I just added in the description 👍

PR #8281, opened 2026-08-11
feat: outbound connector management endpoints for multi-engine support

### LINKED ISSUE #7965: Outbound Connector Management Endpoints for Multi-Engine Support
**Context:**
Following on from #6961 (per-engine `CamundaClient` registry), the outbound connector *management* endpoint — used to list active outbound connectors and expose metrics/broker-stream connectivity — currently reads from a single `OutboundConnectorsService`, backed by a static `OutboundConnectorFactory` list and one `BrokerJobStreamClient` wrapping a single injected `CamundaClient`. Once there is one client per physical tenant (engine), this endpoint needs to aggregate across all of them.

**Tasks:**
- `OutboundConnectorsRestController` ([list active outbound](https://github.com/camunda/connectors/blob/2157eccf97d1fa8effa568308762bb366ca28da8/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/controller/OutboundConnectorsRestController.java#L68), [metrics](https://github.com/camunda/connectors/blob/2157eccf97d1fa8effa568308762bb366ca28da8/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/controller/OutboundConnectorsRestController.java#L94)) — update [`OutboundConnectorsService`](https://github.com/camunda/connectors/blob/2157eccf97d1fa8effa568308762bb366ca28da8/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/instances/service/OutboundConnectorsService.java#L37) to use the new per-PT `CamundaClient` registry (from #6961) instead of the single [`BrokerJobStreamClient`/`CamundaClient`](https://github.com/camunda/connectors/blob/2157eccf97d1fa8effa568308762bb366ca28da8/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/jobstream/BrokerJobStreamClient.java#L54); aggregate broker topology/stream connectivity across all engines.
- Add a `physicalTenantIds` (list) filter query param to `/outbound` and `/outbound/metrics`.
- Add a `physicalTenantId` field to [`OutboundConnectorResponse`](https://github.com/camunda/connectors/blob/2157eccf97d1fa8effa568308762bb366ca28da8/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/controller/OutboundConnectorResponse.java#L40) — today it only carries `runtimeId` (pod identity), nothing for engine identity.

**Notes:**
- Mirrors the aggregation approach used for inbound management endpoints — keep the two consistent (same filter param name/semantics, same DTO field name for engine identity).

**Depends on:** #6961 (per-engine `CamundaClient` registry must exist before this can aggregate across them).

nicpuppa:
## Description

Follow-up to #6961 on the outbound side, mirroring what #8184 did for inbound.

The `/outbound` management endpoints read from a single `OutboundConnectorFactory` list and one `BrokerJobStreamClient` wrapping one injected `CamundaClient`. With several engines configured, they only ever reported whichever client happened to be `@Primary`, and nothing in the response said which engine an entry belonged to.

**Aggregate broker connectivity across engines.** `OutboundConnectorsService` now takes the configured physical tenants plus one `BrokerJobStreamClient` per tenant, and lists each registered connector once per (connector, physical tenant) pair — every engine runs a job worker for every registered connector — each entry carrying that engine's own broker/stream connectivity. A tenant whose brokers are unreachable, or that has no broker monitoring at all, still appears with `UNKNOWN` rather than dropping out, so one engine's failure never hides another's connectors.

`OutboundConnectorRuntimeConfiguration` builds the per-tenant client map from the `CamundaClientRegistry`, via a plain static helper rather than a `Map`-typed `@Bean` parameter, for the reasons already documented in that class. Topology-discovery mode resolves each engine's brokers through that engine's own client; with a single tenant, or in explicit-addresses mode (where the addresses are global config rather than per-engine), the existing scalar `brokerJobStreamClient` bean is reused as-is — so single-engine deployments, and overrides of that bean, behave exactly as before.

**`physicalTenantId` on `OutboundConnectorResponse`.** It carried only `runtimeId` (pod identity), nothing for engine identity. Omitted (`NON_NULL`) for legacy wiring that supplies no registry, preserving the pre-multi-engine response shape.

**`physicalTenantIds` filter** on `/outbound`, `/outbound/{type}`, `/outbound/metrics` and `/outbound/metrics/{connectorType}` — same parameter name and null/empty-means-unfiltered semantics as the inbound endpoints. The filter narrows what each pod reports, not the pod fan-out: `InstanceForwardingRouter` forwarding is untouched and every pod still answers.

**Metrics tagged by physical tenant.** The filter would otherwise have been a no-op on `/outbound/metrics`: outbound meters had no tenant dimension at all. Invocation counters, the execution-time timer, and the `last-completed`/`last-failed`/`max-execution-time` gauges are now tagged with the tenant the job worker was opened for — the same ID the listing and the per-physical-tenant document/secret beans are keyed by, so a meter and the connector entry it belongs to always agree. This also fixes a latent bug: two engines running the same connector type registered the same gauge id, so the second registration silently collided with the first and that engine's timestamps were never visible.

One thing worth a reviewer's opinion: a **filtered** metrics response omits the `worker` section. Those counters (`camunda.client.worker.*`) are recorded by the Camunda client, whose `JobWorkerMetricsFactoryContext` exposes only the job type, so they cannot be attributed to an engine. Reporting the pod-wide totals inside an otherwise filtered response would misattribute them; reporting zeros would read as "no jobs activated". Unfiltered responses are unchanged.

## Related issues

closes #7965

### DIFF
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/instances/service/OutboundConnectorsService.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/instances/service/OutboundConnectorsService.java
@@ -24,41 +24,115 @@ import io.camunda.connector.runtime.outbound.controller.OutboundConnectorRespons
 import io.camunda.connector.runtime.outbound.jobstream.BrokerJobStreamClient;
 import io.camunda.connector.runtime.outbound.jobstream.BrokerStreamsResult;
 import io.camunda.connector.runtime.outbound.jobstream.StreamConnectivity;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/**
+ * Backs the {@code /outbound} management endpoints. Every configured physical tenant (engine) runs
+ * a job worker for every registered outbound connector, so the registered connectors are listed
+ * once per (connector, physical tenant) pair, each entry enriched with the broker/stream
+ * connectivity observed on that tenant's own brokers.
+ */
 public class OutboundConnectorsService {
 
   private static final Logger LOG = LoggerFactory.getLogger(OutboundConnectorsService.class);
 
   private final OutboundConnectorFactory connectorFactory;
-  private final BrokerJobStreamClient brokerJobStreamClient;
 
+  /**
+   * All configured physical tenants, in iteration order, or a single {@code null} entry for the
+   * legacy single-engine wiring (whose responses carry no {@code physicalTenantId}, i.e. exactly
+   * the pre-multi-engine response shape). Deliberately kept separate from {@link
+   * #brokerJobStreamClientsByPhysicalTenantId}: broker monitoring can be disabled entirely (leaving
+   * no clients at all), and the endpoint must still report every engine's connectors.
+   */
+  private final List<String> physicalTenantIds;
+
+  private final Map<String, BrokerJobStreamClient> brokerJobStreamClientsByPhysicalTenantId;
+
+  /** Used for the {@code null} physical tenant of the legacy single-engine wiring. */
+  private final BrokerJobStreamClient legacyBrokerJobStreamClient;
+
+  /** Legacy single-engine constructor without broker monitoring. */
   public OutboundConnectorsService(OutboundConnectorFactory connectorFactory) {
-    this(connectorFactory, null);
+    this(connectorFactory, (BrokerJobStreamClient) null);
   }
 
+  /**
+   * Legacy single-engine constructor: one unnamed engine, whose responses carry no {@code
+   * physicalTenantId}.
+   */
   public OutboundConnectorsService(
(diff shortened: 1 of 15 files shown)


connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/instances/service/OutboundConnectorsService.java
```diff
@@ -24,63 +24,174 @@
...
-    return connectorFactory.getRuntimeConfigurations().stream()
-        .map(config -> toResponse(config, runtimeId, brokerStreams))
-        .toList();
+    return findAll(runtimeId, null);
+  }
+
+  /**
+   * @param physicalTenantIdFilter when non-null and non-empty, restricts the result to these
+   *     physical tenants (engines); {@code null}/empty returns every configured engine, matching
+   *     the pre-existing (unfiltered) behavior of this endpoint.
```
chillleader, line 110: :wrench: let's add a test case for this behavior (null filter)

chillleader (approved):
Looks good, and nice extensive test coverage! Just one minor comment and good to go.

Szik (approved):

PR #8288, opened 2026-08-11
feat(jdbc): enhance credential validation with database selection and connection checks

### LINKED ISSUE #8286: Roll out credential validation to the remaining configuration types (REST authentication, JDBC connection)
## Summary

[#8027](https://github.com/camunda/connectors/pull/8027) shipped the credential-validation
subsystem (`ConfigurationValidator` SDK contract, startup registry, resolution service, REST
endpoint) and adopted it for one configuration type: the AWS credential. Every other configuration
type in the repo still answers `UNSUPPORTED`, so the Hub cannot check those credentials at
creation time.

This issue covers rolling validation out to the remaining configuration types, closing the
"Connector adoption" checkbox of #8088.

## Current state

Three configuration types exist today (`@Configuration`-annotated classes):

| Configuration | Id | Consumed by | Validator |
|---|---|---|---|
| `AwsCredentialConfiguration` | `io.camunda:aws-credential:1` | all AWS connectors + IDP extraction | ✅ `AwsCredentialValidator` (STS `GetCallerIdentity`) |
| `RestAuthenticationConfiguration` | `io.camunda.connectors:rest-authentication:1` | REST, GraphQL, HTTP Polling | ❌ missing |
| `JdbcConnectionConfiguration` | `io.camunda.connectors:jdbc-connection:1` | SQL (JDBC) | ❌ missing |

Because a validator is keyed on the configuration type rather than on the connector, the two
missing validators cover four more connectors.

### REST authentication

Validate the credential by exercising whatever the credential itself can reach:

- OAuth 2.0 client credentials and OAuth 2.0 refresh token carry a token endpoint, so the check is
  a real one — request a token and require an `access_token` back. Reuses the same token-request
  construction the connector uses at execution time, so validation cannot drift from execution.
- Basic, Bearer and API key carry no endpoint. There is nothing to call, so the credential is
  reported as not validatable rather than given a green tick nobody verified.
- No authentication is usable by definition.

### JDBC connection

A connection credential currently records host, port, database name and login, but not which
database product it points at — the product is picked in the process, not in the credential. That
makes the credential impossible to validate on its own, and arguably incomplete: the same
host/port/login is only meaningful for one product.

Add the database product to the credential, and validate by opening a real connection and closing
it again — authentication happens during connect, so no query is needed and none is run. The
connection is opened through the same code path execution uses, so a credential that validates
cannot fail at execution over a driver or URL scheme the two assembled differently.

A credential naming a database takes precedence over the connector's own database field at
execution time, with the inline field as the fallback — the same consumption rule the AWS credential
already applies to its authentication and region, and the rule this connector already applies to the
connection itself. Without it the credential would supply the host, port and login while the inline
field kept choosing the driver, so a credential validated against PostgreSQL could be executed
against the same host with the MySQL driver and nothing would report the mismatch. Requiredness of
the database moves from the field to the effective value, so a diagram that gets its database from a
credential is no longer asked to fill a duplicate dropdown.

A credential that names no database — every credential stored before the field existed — falls back
to the inline field, so those diagrams keep executing unchanged. This is the one deliberate
difference from the AWS rule, which lets a bound credential win unconditionally.

Two limits fall out of this and are recorded rather than fixed here:

- A configuration template cannot carry a map (the generator disables FEEL on its properties), so a
  credential carries no driver properties. SQL Server needs `encrypt=false` to accept an untrusted
  server certificate, so a SQL Server connection credential cannot be reached at all — for
  execution as much as for validation. Tracked separately.
- The connection has no login timeout, matching execution. A validation request against a
  firewalled host blocks for the OS TCP timeout.

### Out of scope

- Anything already listed as out of scope in #8088 (hybrid runtimes, Modeler-triggered validation,
  re-validation freshness).
- New configuration types.

## Acceptance criteria

- [ ] A REST authentication credential using OAuth 2.0 (either grant) is validated against its
      token endpoint; a rejected client secret or an expired refresh token is reported as
      unauthorized, not as a generic error.
- [ ] A REST authentication credential whose type cannot be checked out of band is reported as not
      validatable, and never as valid.
- [ ] A JDBC connection credential is validated by connecting to the database; a wrong password is
      reported as unauthorized, and an unreachable host is not. Verified against real MariaDB,
      MySQL, PostgreSQL and Oracle servers (SQL Server excluded by the driver-properties limit
      above).
- [ ] The database a credential names is the one execution connects with, verified against those
      same real servers, including that a mismatched inline database really would fail without the
      credential overriding it.
- [ ] A credential that names no database still executes against the connector's own field.
- [ ] Validation never surfaces resolved secret material in a client-facing message, matching the
      message-safety policy already documented on `ConfigurationValidationService`.
- [ ] Every new validator is registered for `ServiceLoader` discovery and covered by unit tests.
- [ ] Element templates affected by the JDBC credential change are regenerated.

## Links

- Framework: #8027
- Phase 1 tracking issue: #8088
- Parent epic: [PDP-3433](https://github.com/camunda/product-hub/issues/3433)

mathias-vandaele:
This pull request introduces a new out-of-band validator for REST authentication credentials, improves the documentation and handling of unsupported configuration validation, and updates the JDBC connector's configuration templates to make the database selection optional and more user-friendly.

**Authentication validation improvements:**

* Added `RestAuthenticationValidator` to provide out-of-band validation for REST authentication credentials, supporting OAuth 2.0 (with real token requests) and returning "unsupported" for Basic, Bearer, and API Key authentication since these cannot be checked out-of-band. The validator is registered as a service and includes comprehensive test coverage. [[1]](diffhunk://#diff-372b4a78c04588abbcdc7ff6cf76faac01b3ed0a62bbd63678850389014b3824R1-R176) [[2]](diffhunk://#diff-b095f81e3e92d48560344c916a62d4fb466aa27dd6c718ccb3019589e28ce8d3R1) [[3]](diffhunk://#diff-3b0bf8f52045710f9fc46976968188b7be8ee6db0202797766c8ae291a90c982R1-R358)
* Improved the documentation and semantics of the `UNSUPPORTED` status in `ConfigurationValidationResult`, clarifying when and why it is returned, and added a static `unsupported()` method for validators to use when a configuration cannot be checked. [[1]](diffhunk://#diff-c9160f856edf95d275a9ecff6edc46e2d84425540bc199a6383fb5355d3c4b12L27-R31) [[2]](diffhunk://#diff-c9160f856edf95d275a9ecff6edc46e2d84425540bc199a6383fb5355d3c4b12R81-R89)

**JDBC connector template enhancements:**

* Made the `database` property optional in `jdbc-outbound-connector-hybrid.json` to allow greater flexibility in configuration.
* Added a new dropdown property for selecting the database type (MariaDB, MSSQL, MySQL, PostgreSQL, Oracle) to the JDBC connection template, improving clarity and user experience.


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationValidator.java
```diff
@@ -0,0 +1,178 @@
...
+ * name the API it is meant for:
+ *
+ * <ul>
+ *   <li>Both OAuth 2.0 variants carry their own token endpoint, so the check is a real one: request
+ *       a token exactly as connector execution would, and require an access token back.
+ *   <li>Basic, bearer and API key carry a secret and nothing to present it to, so there is no
+ *       out-of-band check to run and the result is {@link
+ *       ConfigurationValidationResult#unsupported() unsupported} — reporting success would confirm
+ *       a secret nobody verified.
+ *   <li>No authentication is usable by definition.
```
chillleader: Should we also include the URL into the REST credential? I think this would make sense as you actually need a combination of URL and auth to verify anything.

connectors/jdbc/src/main/java/io/camunda/connector/jdbc/model/request/JdbcRequest.java
```diff
@@ -18,16 +18,19 @@
 import jakarta.validation.constraints.NotNull;
 
 public record JdbcRequest(
-    @NotNull
-        @TemplateProperty(
+    // Not @NotNull, and no notEmpty constraint: the database may come from a bound connection
```
chillleader: Should we always include it into the credential? It would need an isEmpty condition then that I am adding in my other PR: https://github.com/camunda/connectors/pull/8307

chillleader (commented):
Looks good but I think we need to merge my other PR first (https://github.com/camunda/connectors/pull/8307) and then make some changes here

Szik (approved):

johnBgood (approved):

PR #8296, opened 2026-08-12
fix(ci): skip versioned element template check on release branches

ztefanie:
## Summary
- `check-versioned-element-templates` compares newly-added `versioned/*.json` snapshots against `origin/main`, hardcoded regardless of the PR's actual base branch.
- Release branches (`stable/*`, `alpha/*`) accumulate their own versioned template history independently of `main` and routinely diverge from it, so this comparison produces false-positive mismatches for files completely unrelated to the PR being checked (observed on #8283: 16 of 17 flagged files aren't touched by that PR at all).
- Skip the check when the PR/push target branch matches `stable/*` or `alpha/*` — it's only meaningful for PRs targeting `main`.

## Test plan
- [x] Ran the updated script locally with `TARGET_BRANCH=alpha/8.10` — exits 0 immediately with a skip message.
- [x] Verified `TARGET_BRANCH=main` (or unset) still falls through to the existing check logic.
- [ ] CI passes on this PR.

### DIFF
--- a/.github/workflows/TEST_FEATURE_BRANCH.yml
+++ b/.github/workflows/TEST_FEATURE_BRANCH.yml
@@ -370,6 +370,8 @@ jobs:
         shell: bash
 
       - name: Check versioned element templates match main
+        env:
+          TARGET_BRANCH: ${{ github.event.pull_request.base.ref || github.ref_name }}
         run: |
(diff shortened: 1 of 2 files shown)


mathias-vandaele (approved):
Thank you @ztefanie

PR #8301, opened 2026-08-13
ci: temporary canary to verify qa/ci/common Vault access

hilalbursalii:
Temporary, throwaway PR to confirm this repo's Vault role can now read
`secret/data/products/qa/ci/common` after the grant in
camunda/infra-core#13882.

Adds a single-step workflow that reads one key
(`SLACK_BOT_USER_OAUTH_TOKEN`) via `hashicorp/vault-action` and never
prints the value — pass/fail on the Vault step is the only signal we need.

Will close without merging once the check result is confirmed.

### DIFF
--- a/.github/workflows/vault-access-check-qa-common.yaml
+++ b/.github/workflows/vault-access-check-qa-common.yaml
@@ -1,0 +1,27 @@
+name: Vault Access Check - qa/ci/common
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    paths:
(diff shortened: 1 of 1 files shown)


hilalbursalii:
Confirmed: Vault Access Check - qa/ci/common run succeeded (https://github.com/camunda/connectors/actions/runs/31684693404) — this repo's Vault role can read secret/data/products/qa/ci/common. Closing this temporary canary without merging.

Issue #7595, opened 2026-06-22
AI Agent Connector: Agent Instance Implementation


nikonovd:
This is an aggregate issue to finalize the AI agent interactions with the OC agent instance APIs: https://github.com/camunda/product-hub/issues/3462

Initial agent instance status and metrics updates as well as conversation history population were implemented in:
* #7198 
* #7199 
* #7450

korthout:
@nikonovd Can we close this issue or do you want to keep it to track follow-up work?

nikonovd:
@korthout we can close it and I'll extract the remaining issue 👍 thanks.

PR #8318, opened 2026-08-14
feat(agentic-ai): add stable, self-generated ids to conversation messages

### LINKED ISSUE #8304: Add stable, self-generated ids to conversation history messages
Prepares the AI Agent conversation model for dedup on the redesigned agent instance history API
(camunda/camunda#58789).

Conversation messages currently have no independent identity: `AssistantMessage.messageId` exists
but is provider-sourced and used only for AWS AgentCore round-tripping; `UserMessage`,
`SystemMessage`, and `ToolCallResultMessage` have no id at all. Tool call ids remain separately
provider/gateway-sourced and stay the correlation key for AHSP/Zeebe — this change does not touch
them.

This adds a random, self-generated `id` to every `Message`, assigned once at construction and
persisted going forward (stable via storage, not re-derivation), with an on-load backfill for
messages persisted before this change ships. The `AgentInstanceClient` interface is prepared to
carry the id through now, though the current wire API has no field to receive it yet — only the
client's internals will need to change once the redesigned API lands.

Part of #7595.

nikonovd:
## Description

Adds a random, self-generated `id` to every conversation `Message`, generated once and stable through persistence, independent of provider ids. Fixes `AwsAgentCoreConversationStore` not round-tripping it (hand-built messages, not Jackson). Prepares for id-based history dedup once the redesigned agent instance history API lands (camunda/camunda#58789).

ADR: connectors/agentic-ai/docs/adr/012-stable-self-generated-message-ids.md

## Related issues

Part of #7595. closes #8304

## Checklist


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/memory/conversation/awsagentcore/mapping/AwsAgentCoreConversationMapper.java
```diff
@@ -255,15 +256,23 @@ private Optional<Message> extractMessageFromPayloads(
     return switch (messageRole) {
       case USER -> {
         var builder = UserMessage.builder().content(content).metadata(metadata);
-        if (properties != null && properties.get(PROPERTY_USER_NAME) instanceof String name) {
-          builder.name(name);
+        if (properties != null) {
+          if (properties.get(PROPERTY_ID) instanceof String id) {
+            builder.id(id);
+          }
```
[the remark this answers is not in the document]
maff: I think that's ok as agentcore never landed in a stable release.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/message/MessageUtil.java
```diff
@@ -10,10 +10,18 @@
...
 
 public class MessageUtil {
   private MessageUtil() {}
 
+  /**
+   * Builder default for {@link Message#id()}; also backfills ids on pre-existing persisted data.
+   */
+  public static String generateId() {
+    return UUID.randomUUID().toString();
+  }
```
maff, line 26: 💭 Not blocking: just thinking of potential storage backends (index performance) and sortability, WDYT of defaulting to something like UUIDv7 as message ID format?

connectors/agentic-ai/connector-agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/model/message/MessageIdTest.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package io.camunda.connector.agenticai.aiagent.model.message;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.camunda.connector.agenticai.aiagent.model.tool.ToolCallResultContent;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class MessageIdTest {
```
maff, line 17: 🔧 Can we also add tests here for JSON deserialization adding an ID. This would cover the case of an older conversation (no message IDs) being migrated to this newer state.

connectors/agentic-ai/connector-agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/model/message/MessageSerializationTest.java
```diff
@@ -30,16 +31,17 @@ void messagesCanBeSerializedAndDeserialized() throws Exception {
...
     assertThat(deserialized.messages())
         .usingRecursiveFieldByFieldElementComparator()
-        .containsExactlyElementsOf(TestMessagesFixture.testMessages());
+        .containsExactlyElementsOf(wrapper.messages());
   }
 
   @Test
   void messagesCanBeDeserializedFromFixture() throws IOException {
-    // test that representation stored in fixture file is stable
+    // fixture predates Message#id(); ids are backfilled on load, so ignore them here.
```
maff: 🔧 Same remark as on the message ID tests - as they are backfilled, we could at least assert on the presence.

connectors/agentic-ai/docs/adr/012-stable-self-generated-message-ids.md
```diff
@@ -0,0 +1,137 @@
...
+
+**Implemented**
+
+## Context and Problem Statement
+
+Conversation `Message`s (`SystemMessage`, `UserMessage`, `AssistantMessage`,
+`ToolCallResultMessage`) have no independent identity. `AssistantMessage.messageId` exists but is
+provider-sourced (present only when the LLM provider returns one) and is used exclusively for AWS
+AgentCore round-tripping — it is not a general-purpose identity mechanism, is absent for other
+providers, and does not exist at all on the other three message types.
```
maff: The AWS part is not entirely true - it is added as part of the native provider integrations now. Worth investigating if we could use the `messageId` instead of adding an additional `id` field or remove the `messageId` in favor of this ID field. Otherwise we can keep both and state ID = synthetic, always generated by us, messageId = provider relation/third party key. We just need to be aware of the additional storage implications.

connectors/agentic-ai/docs/adr/012-stable-self-generated-message-ids.md
```diff
@@ -0,0 +1,137 @@
...
+* **Stability**: once assigned, an id must not change across reloads of the same persisted
+  message — dedup depends on it being the same value every time.
+* **Minimal surface area**: avoid new persisted state beyond the id itself; avoid a parallel
+  migration mechanism if the existing serialization contract can absorb the change.
+* **Non-null domain model**: `@NullMarked` is the convention in this module (see `AGENTS.md`
+  "Null safety"); a field that's sometimes absent should either be `@Nullable` or handled so the
+  domain type never has to represent "no id" as a real state.
+* **No new identity for tool calls**: `ToolCall.id` already serves the correlation needs of
+  Zeebe/AHSP element activation and the provider round-trip (a tool result must echo back the
+  exact id the provider issued for a `tool_use` block); it must not be touched by this change.
```
maff: We generate an ID for tool call result messages now - does that oppose this statement?

maff (approved):
Awesome to have stable message IDs now! 🎉

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/message/MessageUtil.java
nikonovd: Good point. As the JDK lacks of native UUIDv7 support, we should decide together how we want to integrate support for that.

I'm not aware of connectors-wide solutions for that.

connectors/agentic-ai/docs/adr/012-stable-self-generated-message-ids.md
nikonovd: It's an additional context - idempotency on the agent instance API will depent on the toolCall.ids as our tool call result messages are aggregates of those, they can't serve as a unique key, as we generate the result message only once before the conversation, while the tool results are mapped to one-message-per-tool-call on the Agent Instance.

While we do that, anthropic seems to be the only one where this aggregation is really present on the API-evel, so maybe we rethink this state of the art structure on our side.

connectors/agentic-ai/connector-agentic-ai/src/test/java/io/camunda/connector/agenticai/aiagent/model/message/MessageIdTest.java
nikonovd: Good point, added those!

PR #8332, opened 2026-08-17
ci(e2e): parallelize Maven builds and cache full .m2 repo for e2e test matrix

nikonovd:
## Description

Two runtime tweaks to the e2e branch-run workflow (`.github/workflows/E2E_BRANCH_RUN.yml`):

1. **Parallel Maven builds**: add `-T1C` (one thread per core) to both mvnw invocations that build/run the e2e reactor, so the build and each matrix test job compile faster.
2. **Full `.m2` repo cache reuse**: the build job previously uploaded only the first-party artifacts (`~/.m2/repository/io/camunda/connector`) for the test matrix to download. Now the entire `~/.m2/repository` (first-party + all resolved third-party dependencies) is uploaded/downloaded, so test jobs stop re-resolving their dependency tree from Nexus on every matrix run. Maven stays online, so anything missing still falls back to Nexus.

## Checklist

- [x] Tests/Integration tests for the changes have been added if applicable. (CI workflow change only; no application code affected.)

### DIFF
--- a/.github/workflows/E2E_BRANCH_RUN.yml
+++ b/.github/workflows/E2E_BRANCH_RUN.yml
@@ -146,14 +146,6 @@ jobs:
             secret/data/products/connectors/ci/nexus USER | CI_NEXUS_USR;
             secret/data/products/connectors/ci/nexus PASSWORD | CI_NEXUS_PSW;
 
-      - name: Restore cache
-        uses: actions/cache@v6
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
-          restore-keys: |
-            ${{ runner.os }}-maven-
-
       - uses: actions/setup-java@v5
         with:
           distribution: 'temurin'
@@ -200,7 +192,7 @@ jobs:
       # Run only this module's tests. No -am: base/connector SNAPSHOTs come from the downloaded
       # artifact, so nothing upstream is rebuilt.
       - name: Test e2e module
-        run: ./mvnw --batch-mode -T1C test -pl "${{ matrix.entry.module }}" -f connectors-e2e-test/pom.xml
+        run: ./mvnw --batch-mode test -pl "${{ matrix.entry.module }}" -f connectors-e2e-test/pom.xml
 
       - name: Publish Test Report
         if: always()
--- a/.github/workflows/E2E_BRANCH_RUN.yml
+++ b/.github/workflows/E2E_BRANCH_RUN.yml
@@ -67,7 +67,7 @@ jobs:
       # POM and all first-party e2e artifacts land in ~/.m2 for the test matrix). Tests are skipped
       # here; connector tests run in the next step, e2e tests run isolated in the matrix.
       - name: Build connectors
-        run: ./mvnw --batch-mode -U clean install -PcheckFormat -DskipTests
+        run: ./mvnw --batch-mode -T1C -U clean install -PcheckFormat -DskipTests
 
       # Non-PR (nightly/manual): run unit/integration tests for the connector modules. The whole
       # connectors-e2e-test subtree is excluded by enumerating every e2e pom path: excluding only the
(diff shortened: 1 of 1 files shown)


.github/workflows/E2E_BRANCH_RUN.yml
```diff
@@ -183,18 +187,20 @@ jobs:
...
+      # third-party dependencies). This lets the test run resolve its dependency tree locally
+      # instead of re-downloading it from Nexus on every matrix job. Maven stays online, so any
+      # artifact not present still falls back to Nexus.
+      - name: Download Maven repository
         uses: actions/download-artifact@v8
         with:
-          name: m2-firstparty-${{ steps.sanitize.outputs.branch }}-${{ github.run_id }}
-          path: ~/.m2/repository/io/camunda/connector
+          name: m2-repo-${{ steps.sanitize.outputs.branch }}-${{ github.run_id }}
+          path: ~/.m2/repository
```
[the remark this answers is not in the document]
nikonovd: Fixed — dropped the test job's `actions/cache` restore; the downloaded artifact is now the sole, authoritative source for the full `~/.m2/repository` in that job (build job's cache restore is untouched).

.github/workflows/E2E_BRANCH_RUN.yml
```diff
@@ -183,18 +187,20 @@ jobs:
...
 
       # Run only this module's tests. No -am: base/connector SNAPSHOTs come from the downloaded
       # artifact, so nothing upstream is rebuilt.
       - name: Test e2e module
-        run: ./mvnw --batch-mode test -pl "${{ matrix.entry.module }}" -f connectors-e2e-test/pom.xml
+        run: ./mvnw --batch-mode -T1C test -pl "${{ matrix.entry.module }}" -f connectors-e2e-test/pom.xml
```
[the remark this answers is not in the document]
nikonovd: Fixed — removed `-T1C` from the test job's mvnw call. Agreed: with `-pl` and no `-am` the reactor is a single module per matrix job, so there's nothing to parallelize there. Kept `-T1C` on the build job, which does build the full multi-module reactor.

.github/workflows/E2E_BRANCH_RUN.yml
```diff
@@ -104,12 +104,16 @@ jobs:
...
+      # everything from the artifact instead of re-downloading their dependency tree from
+      # Nexus. Maven stays online in the test jobs, so anything missing still falls back to
+      # Nexus - the artifact just eliminates the bulk of the per-job resolution.
+      - name: Upload Maven repository
         uses: actions/upload-artifact@v7
         with:
-          name: m2-firstparty-${{ steps.sanitize.outputs.branch }}-${{ github.run_id }}
-          path: ~/.m2/repository/io/camunda/connector
+          name: m2-repo-${{ steps.sanitize.outputs.branch }}-${{ github.run_id }}
+          path: ~/.m2/repository
```
maff, line 116: Not objecting to change, I'm just wondering how this is different to the previous restore cache approach? 🤔

maff (approved):

.github/workflows/E2E_BRANCH_RUN.yml
nikonovd: We're now caching all maven artifacts in the repository to ensure that all dependencies from the `Build` step are available in the e2e test execution of all downstream jobs.

This reduces the time as we're not downloading dependencies again during verification.

nikonovd: To be more precise - because I now saw that you're referring to the deleted `Restore cache` action.

That one was shared across branches and runs - where the POM hash is the important key. Those caches can also be evicted, or fall back to a stale partial match via `restore-keys` (an older snapshot from a previous pom version, not just an empty cache).

The cache we're creating here is really the SSOT for the further executions of the e2e tests - guaranteed to be present and to match this exact run.

Note the build job's own `Restore cache` step is untouched - we only dropped the redundant restore in the test job; the build job still benefits from it.

PR #8342, opened 2026-08-18
feat(agentic-ai): opt AI Agent connectors into job leasing

### LINKED ISSUE #8341: AI Agent: opt v2 connectors into job leasing and plumb lease token to agent instance client
Sub-issue of #7595. Consumes the Spring SDK job-lease work in camunda/camunda#57812.

## Context

The AI Agent runs a distributed, supersession-prone loop: each tool-call round is a **separate job activation**, and a completing tool creates a new job while a stale in-flight job may still be running (see `ai-agent.md` §10). camunda/camunda#57812 exposes `withLease` in the Spring Boot SDK so a worker's activations carry a per-activation `leaseToken`, fencing complete/fail/throw-error against a stale, superseded activation of the same job.

Job **completion** is already safe in the connector runtime — it completes via the `ActivatedJob`-based overloads, so a lease token carried on the job fences completion automatically once leasing is opted in. What is **not** fenced without extra work is the agent-instance side-effect writes (`CamundaAgentInstanceClient` update + create-history) that a superseded activation can still issue, corrupting status / metrics / conversation history.

## Goal

Opt the **v2** AI Agent connectors into job leasing and fence their agent-instance writes with the lease token.

## Scope / approach

- **Opt in via annotation:** `withLease = true` on `AgentTaskV2Function` and `AgentSubProcessV2Function`. Always-on; accepts the version-skew tradeoff (over REST a pre-leasing gateway rejects activation with HTTP 400; over gRPC the token drops to `null`). Wiring already exists (`OutboundConnectorManager` maps `withLease()` → `JobWorkerValue.setWithLease`).
- **Forward the lease** from `CamundaAgentInstanceClient` via `jobLease(...)` on the update and create-history commands whenever a token is present.

> **Scope revised during implementation.** The original plan was to plumb the token to a single seam that logs a WARN and leaves the command unchanged (forwarding deferred pending an endpoint redesign). This is **not viable together with `withLease = true`**: the create-history command already sends `jobKey`, and once the job is leased the engine correlates the write to the active job by `jobKey` and rejects it (`NOT_FOUND`, *"supplied lease does not match"*) unless the matching lease is supplied. Warn+noop → every history write 404s → the job fails → the process hangs (reproduced across all v2 e2e tests). The lease is therefore forwarded now, not deferred.

Completion fencing is active as a side effect of opting in — a superseded completion is fenced rather than surfacing as `NOT_FOUND`.

## Tasks

- [x] `withLease = true` on `AgentTaskV2Function` + `AgentSubProcessV2Function`.
- [x] Forward the lease token via `jobLease(...)` on `CamundaAgentInstanceClient` update + create-history writes (only when a token is present; v1 / gRPC-skew send none).
- [x] Unit tests: forwarding when leased; no forwarding when unleased (both call sites).
- [x] E2E regression: v2 suites pass with leasing on.
- [x] Docs: `ai-agent.md` §10 (leased v2 activations, fenced completion + writes) + §23 (lease forwarding, engine lease-match rejection); supersession gotcha in `AGENTS.md`.

## Out of scope

- **v1** agent connectors and the **gateway / MCP / A2A / ad-hoc** connectors — left unleased (they don't write agent-instance state; their completion is already safe; keeps the version-skew surface minimal).
- ADR (rationale captured in PR + reference docs).

## Acceptance criteria

- v2 agent connectors activate jobs with a lease; completion is fenced against superseded activations.
- Agent-instance status/metrics/history writes carry the lease and are rejected by the engine when it doesn't match (a superseded activation cannot corrupt them).
- Existing v2 e2e suite passes with leasing enabled; unit tests cover the present/absent-token paths for both write call sites.

nikonovd:
## Description

Opts the AI Agent connectors into job leasing and fences their agent-instance writes against superseded activations.

- **`withLease = true`** on all four AI Agent functions — `AgentTaskV1Function`, `AgentTaskV2Function`, `AgentSubProcessV1Function`, `AgentSubProcessV2Function`. Each activation now carries a per-activation `leaseToken`; the runtime completes/fails via the `ActivatedJob`-based command overloads, so job completion is fenced against a superseded activation automatically. v1 is leased too because agent-instance handling doesn't differentiate by version: a custom v1 template can still carry an agent definition and emit visibility data, so it has the same supersession-corruption exposure.
- **`CamundaAgentInstanceClient` forwards the lease** via `jobLease(...)` on the update and create-history commands whenever a token is present. The engine correlates each write to the active job by `jobKey` and rejects it (`NOT_FOUND`, "supplied lease does not match") when the lease doesn't match — so a superseded activation's status/metrics/history writes are discarded server-side rather than committed.
- No lease is forwarded in the **gRPC version-skew** case (a pre-leasing gateway drops the field, token `null`); those writes proceed unfenced. Note the version-skew contract on `OutboundConnector#withLease`: over REST an older gateway rejects the activation (HTTP 400).

## Related issues

closes #8341

## Checklist

  release, as this branch will be rebased onto main before the next release.


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -383,9 +387,27 @@
...
   }
 
+  /**
+   * Seam for job-lease fencing of agent-instance writes. Called only for an activation that carries
+   * a lease token; the token is deliberately not forwarded to the write command, because the
+   * agent-instance write endpoint is being redesigned and we do not bind to its current {@code
+   * jobLease(...)} command step. When the new endpoint lands, forward the token from this single
+   * point. Returns the command unchanged.
+   */
+  private <T> T applyJobLease(T command, String leaseToken) {
```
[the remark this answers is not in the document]
nikonovd: Dismissed as we implemented an actual application of the jobLease token

nikonovd:
Re: Copilot's comments about `jobLease` forwarding vs the WARN/no-op seam — the scope in #8341 was revised during implementation, and the issue is now updated to match.

The warn+noop/deferred approach is not viable together with `withLease = true`: `newCreateAgentHistoryItemCommand` already sends `.jobKey(jobKey)`, and once the job is leased the engine correlates the write to the active job by `jobKey` and rejects it (`NOT_FOUND`, "supplied lease does not match") unless the matching lease is supplied. Leaving it unforwarded makes every history write 404, failing the job and hanging the process — reproduced across all v2 e2e tests (`AgentTaskCustomProviderToolCallingTests`, `AgentSubProcessCustomProviderToolCallingTests`, `ProviderWireFormatSmokeTests` AnthropicV2 fixture). Forwarding the lease is what makes leasing work against the current engine, so the deferral was dropped and the token is forwarded on both writes. Docs and tests reflect the forwarding contract. CodeQL 'useless parameter' is resolved (the seam helper was removed).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/AgentSubProcessV2Function.java
```diff
@@ -22,7 +22,8 @@
       AgentSubProcessV2Function.TOOL_CALL_RESULTS_VARIABLE,
       AgentSubProcessV2Function.PROVIDER_VARIABLE,
       AgentSubProcessV2Function.DATA_VARIABLE
-    })
+    },
+    withLease = true)
```
maff, line 26: 💭 Are we sure about only doing this for v2 job workers? Does the agent instance API require setting the lease in the future? There will still be a gate with the check for an agent definition, but would anything speak against also using the leases on the v1 connectors?

maff: Background: as we don't differentiate between v1 and v2 in the agent instance handling, a custom template targeting a v1 job worker could still contain an agent definition and opt into sending visibility data. If the API requires the lease, the request would fail in that case.

nikonovd: Nothing against it, but I was basing it on the decision yesterday to have agent instance API only for the latest version of connector templates.

It's no hassle to integrate it into the v1 templates as well

nikonovd: Good call — done in 2502172. Set `withLease = true` on the v1 functions too (`AgentTaskV1Function`, `AgentSubProcessV1Function`), same rationale you gave: agent-instance handling doesn't differentiate v1/v2, so a custom v1 template with an agent definition gets the same fencing. v1 e2e (agent-instance + tool-calling, both flavors) pass with leasing on. Gateway/MCP/A2A stay unleased. Docs updated.

connectors/agentic-ai/docs/reference/ai-agent.md
```diff
@@ -1784,6 +1786,8 @@ path yet). The item carries the current turn's
 `iterationKey` and the active `jobKey`; the engine discards superseded/non-completed items by
 observing job completion (`jobLease` enforcement is a planned follow-up, camunda/camunda#55033).
 
+**Job-lease fencing.** With the AI Agent connectors leased ([§10](#10-concurrency-challenges--race-conditions)), the activation's `leaseToken` reaches `CamundaAgentInstanceClient`, which forwards it via `jobLease(...)` on both write commands (`newUpdateAgentInstanceCommand`, `newCreateAgentHistoryItemCommand`) whenever a token is present. The engine correlates a write to the active job by `jobKey` and rejects it (`NOT_FOUND`, "supplied lease does not match") if the supplied lease doesn't match — so a superseded activation's status/metrics/history writes are discarded rather than committed. A token is absent (and no lease forwarded) only in the gRPC version-skew case (a pre-leasing gateway drops the field); those writes proceed unfenced.
```
maff: 💭 Do you think it would make sense to include a reference to the general job leasing concept in the monorepo, e.g. https://github.com/camunda/camunda/blob/main/zeebe/docs/adr/0005-810-job-lease.md?

maff (approved):

connectors/agentic-ai/docs/reference/ai-agent.md
nikonovd: Good idea — added a link to the monorepo ADR [0005-810-job-lease](https://github.com/camunda/camunda/blob/main/zeebe/docs/adr/0005-810-job-lease.md) from the job-leasing note in §10 (f3ad871).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -225,6 +225,10 @@ private void executeUpdate(
       cmd = cmd.tools(toolMapper.mapTools(tools));
     }
 
+    final String leaseToken = executionContext.jobContext().getLeaseToken();
+    if (leaseToken != null && !leaseToken.isBlank()) {
+      cmd = cmd.jobLease(leaseToken);
```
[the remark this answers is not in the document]
nikonovd: Investigated — the fix is to remove jobLease from the update path, not add jobKey. Per the client javadoc, on `UpdateAgentInstanceCommand` both `jobKey` ("required whenever history() is non-empty; otherwise irrelevant") and `jobLease` ("history items submitted under a superseded lease are discarded") only govern a batched `history(...)` list. This update sends status/metrics/tools and no history batch (history goes through the dedicated `newCreateAgentHistoryItemCommand`), so forwarding the lease here fenced nothing. Dropped the inert `jobLease` from the update; kept it on the create-history command (which sends jobKey + jobLease and is actually fenced). Docs now state status/metrics updates are not lease-fenced (best-effort). 50d2d0c.

PR #8356, opened 2026-08-19
fix(agentic-ai): replay OpenAI Responses assistant text without 400ing

maff:
## Description

The native OpenAI **Responses** provider crashed with a 400 (`Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.`) whenever a conversation replayed a completed assistant *text* turn — e.g. a second agent turn after the user gives feedback and asks a follow-up. Tool-call turns replay fine (`function_call` items), which is why this went unnoticed: the bug only surfaces once a turn ends in plain text and the conversation continues.

Assistant text is now replayed as a structured output message (`output_text` parts, one per text block, under a valid `msg_*` id), matching what the Responses API actually accepts for the assistant role.

Also hardens the coverage that let this ship:
- The wire-format e2e suite normalized away the very detail (`input_text` vs `output_text`) that would have caught this, and only compared joined text, never content-part *type*. It now keeps the raw wire encoding and asserts no assistant-role item ever carries an input-only part.
- The real-provider acceptance suite never exercised a multi-turn conversation with a completed text answer in the middle — every scenario was single-shot. It now includes a user-feedback loop that forces a text turn to be replayed, for every provider row.

## Related issues

closes #

## Checklist


nikonovd (approved):
🚀 Looks good, checked the API docs again and seems to fit.

PR #8358, opened 2026-08-19
fix(agentic-ai): auto-inject additionalProperties:false for OpenAI native strict schemas

maff:
## Description

Both native OpenAI converters (`OpenAiResponsesRequestConverter` and `OpenAiCompletionsRequestConverter`) set
`.strict(true)` on the user-supplied structured-output schema unconditionally, with no normalization. OpenAI's strict
mode requires `additionalProperties: false` on every object level, including the root, and 400s instead of
defaulting it when missing. Hit on live testing gpt-5.5 Responses:

```
Model call failed: 400: Invalid schema for response_format 'Response': In context=(), 'additionalProperties' is required to be supplied and to be false.
```

Fix: `OpenAiStrictJsonSchemas` recursively injects `additionalProperties: false` into the schema (root, `properties`,
array `items`, `anyOf`, `$defs`) before it's converted to the SDK's `Schema` type, overriding any explicit value.
This mirrors what `langchain4j`'s OpenAI/Anthropic clients already do for the v1 provider path — only the v2 native
converters were missing it, since they bypass `langchain4j` entirely.

Both converters' existing `configuresStructuredOutputFromJsonSchema` tests are extended with a nested object and an
array of objects to cover the recursion and override. `RealProviderApiSmokeIT#RESPONSE_SCHEMA` no longer hardcodes
`additionalProperties: false`, so it now exercises the fix against the real API.

## Checklist

### DIFF
--- a/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiStrictJsonSchemas.java
+++ b/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiStrictJsonSchemas.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. Licensed under a proprietary license.
+ * See the License.txt file for more information. You may not use this file
+ * except in compliance with the proprietary license.
+ */
+package io.camunda.connector.agenticai.aiagent.chatmodel.provider.openai;
+
+import static io.camunda.connector.agenticai.common.JsonSchemaConstants.PROPERTY_ADDITIONAL_PROPERTIES;
+import static io.camunda.connector.agenticai.common.JsonSchemaConstants.PROPERTY_ANYOF;
+import static io.camunda.connector.agenticai.common.JsonSchemaConstants.PROPERTY_DEFINITIONS;
+import static io.camunda.connector.agenticai.common.JsonSchemaConstants.PROPERTY_ITEMS;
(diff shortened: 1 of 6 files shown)


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiStrictJsonSchemas.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private OpenAiStrictJsonSchemas() {}
+
+  public static JsonNode forStrictMode(Map<String, Object> schema, ObjectMapper objectMapper) {
+    final JsonNode root = objectMapper.valueToTree(schema);
+    enforceAdditionalPropertiesFalse(root);
+    return root;
+  }
+
+  private static void enforceAdditionalPropertiesFalse(@Nullable JsonNode node) {
```
nikonovd: 💭 This is a recursion and with potentially deeply nested schemas we could run into stack size limitations.

I'd prefer to rework it into a recursion-free loop, if possible, to avoid that option.

nikonovd (approved):
Looks good! Please take a look at my comment, otherwise feel free to merge.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/openai/OpenAiStrictJsonSchemas.java
maff: Good point, updated to an iterative solution.

PR #8375, opened 2026-08-20
feat(agentic-ai): unified agent instance updates during conversation turns

### LINKED ISSUE #7596: AI Agent: Apply unified agent instance updates during conversation turns
## Problem

Before each LLM call, `BaseAgentRequestHandler` sends multiple separate API calls:

1. One per tool call result item
2. One for each other message that is newly added to the LLM conversation.

After each LLM call, `BaseAgentRequestHandler` sends two separate API calls derived from the same LLM response:

1. `agentInstanceClient.update(…)` — status + token/tool-call delta
2. `agentInstanceClient.createHistoryForAssistantMessage(…)` — assistant history item

From the PR review (@maff on `BaseAgentRequestHandler.java:333`):

> Why does the metrics update be separate from the assistant message history update? Isn't that always the same situation after the LLM call and should be combined in a singular API call? Right now we massively increase the error surface by having to send multiple granular API calls for different part of the data which are derived from the same source.

A failure on either call leaves the agent instance in a partially updated state with no recovery path.

## Goal

The API for agent instance history and config updates is being redesigned (see issues below). By that we will be enabled to supply updates to the both, agent instance config and history items + metrics in a single call.

## Requirements

1. When the agent instance is created the initial configuration (system prompt, limits, model, etc) are being passed on the create request
2. When `TOOL_DISCOVERY` is finished, and/or the agent immediately proceeds to the first conversation request, the agent instance is being updated with:
   a. the potentially updated system prompt
   b. all tool definitions 
   c. the first batch of messages (constructed via the turn composer)
3. After an LLM request was finished, the agent instance is immediately (not on job completion listener) being updated with
   a. the assistant message
   b. the request metrics
   c. the new status

All requests are to be bound by `jobLease` and `jobKey` where applicable. This will ensure that stale data (e.g. due to job supersession, network failures, incidents, ...) is going to be discarded on the engine.

## Implementation Notes

The new API design and its hooks must be based on https://github.com/camunda/camunda/issues/58789 and https://github.com/camunda/camunda/issues/59784

nikonovd:
## Description

Redesigns agent-instance history reporting to use the engine's batched `history[]` list on the `UPDATE` command (camunda/camunda#58789), collapsing per-turn engine round-trips and lease-fencing them so a superseded job activation is rejected instead of corrupting the instance. See [ADR 013](connectors/agentic-ai/docs/adr/013-unified-agent-instance-turn-updates.md) for the full design.

- `AgentConfiguration.fingerprint()` + per-turn fingerprint tracking, so a `CONFIGURATION` history item is only emitted when the system prompt or tool list actually changed.
- `AgentInstanceClient.applyTurnStart` / `applyTurnCompletion` / `applyToolCallResults`: each builds one batched, lease-fenced `update()` call. `applyTurnStart` (before the LLM call) carries `THINKING` + input items + an optional `CONFIGURATION` item; `applyTurnCompletion` fires once per LLM round (including continuations), carrying that round's status + assistant item + metrics.
- `BaseAgentRequestHandler` rewired onto the two-call-per-turn design: deletes the deferred, completion-listener-driven status/metrics patch entirely (`notifyThinking`, both `notifyMetrics` overloads, `createMetricsCompletionListener`, `createToolDiscoveryCompletionListener`, `shouldUpdateAgentInstanceBeforeJobCompletion`) — every agent-instance write is now synchronous.
- A `404` on a batched update now maps to a non-retryable `ConnectorRetryException` (`AGENT_INSTANCE_SUPERSEDED`) instead of the ordinary retry path.
- `AgentInstanceHistoryItemIds` for `historyItemId` derivation (message id / tool-call id / deterministic fallback for id-less event results), enabling dedup once the engine's per-item dedup lands.
- Assistant history content is ordered `text -> reasoning -> object -> document -> everything else`.
- Deletes the now-unused per-item client methods (`createHistoryForInputMessages`, `createHistoryForAssistantMessage`, `createHistoryForToolCallResults`) and their tests.
- ADR 013 added; ADR 005/011 marked superseded; `ai-agent.md` and `AGENTS.md` updated to match.

## Related issues

closes #7596

## Checklist

- [x] Tests/Integration tests for the changes have been added.


connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
```diff
@@ -275,7 +275,7 @@ void shouldReportFastToolResultToHistoryWhileSlowToolStillInFlight() throws Exce
             Map.of("userPrompt", "Calculate the superflux product and download a file")));
 
     verify(agentInstanceClient)
```
maff, line 277: 💭 Would we potentially also be able now to query the actual API for the data to ensure it was stored as we expect?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentMetadata.java
```diff
@@ -19,17 +20,51 @@
...
  */
 @AgenticAiRecord
 public record AgentMetadata(
     Long processDefinitionKey,
     Long processInstanceKey,
     @Nullable Long agentInstanceKey,
-    @Nullable Integer lastIterationKey)
+    @Nullable Integer lastIterationKey,
+    Map<Integer, String> configurationFingerprintHistory)
     implements AgentMetadataBuilder.With {
```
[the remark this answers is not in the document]
nikonovd: Confirmed, fixed in 0709821. Jackson binds the missing field to null for pre-existing persisted AgentContext, and configurationFingerprintAt() then NPEs on entrySet(). Normalized null to an empty map in the record's compact constructor.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConversation.java
```diff
@@ -205,25 +213,37 @@ public List<Message> allMessages() {
...
       return withMetrics;
     }
-    return withMetrics.withMetadata(metadata.withLastIterationKey(currentTurn.iterationKey()));
+    var updatedMetadata = metadata.withLastIterationKey(currentTurn.iterationKey());
+    var precedingFingerprint =
+        previousTurns.isEmpty() ? null : previousTurns.getLast().configurationFingerprint();
+    if (!Objects.equals(currentTurn.configurationFingerprint(), precedingFingerprint)) {
+      var history = new LinkedHashMap<>(updatedMetadata.configurationFingerprintHistory());
+      history.put(currentTurn.iterationKey(), currentTurn.configurationFingerprint());
+      updatedMetadata = updatedMetadata.withConfigurationFingerprintHistory(history);
```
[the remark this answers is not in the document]
nikonovd: Confirmed, fixed in f6b2bd2. The wire-level CONFIGURATION decision in applyTurnStart is unaffected (it compares against the durably-reconstructed previous turn, before any continuation round exists), but toAgentContext()'s persisted-history bookkeeping was comparing against previousTurns.getLast(), which after nextContinuationRound() is a same-invocation round carrying the same fingerprint as currentTurn — always equal, so a real change was never recorded. Now captures the fingerprint/iterationKey that preceded the invocation once, at rehydrate(), and compares against that instead of the mutated list.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
```diff
@@ -43,4 +60,27 @@ public int maxModelCalls() {
...
+   * send with unchanged content dedups for free.
+   */
+  public String fingerprint() {
+    return Integer.toHexString(
+        Objects.hash(
+            chatModel.model(),
+            chatModel.provider(),
+            systemPrompt.prompt(),
+            limits != null ? limits.maxModelCalls() : null,
+            tools));
```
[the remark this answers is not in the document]
nikonovd: Confirmed, fixed in d19b196. Switched to a SHA-256 digest over a canonical string of the same fields.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceHistoryItemIds.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public static String forMessage(Message message) {
+    return message.id().toString();
+  }
+
+  public static String forToolCallResult(ToolCallResultContent result) {
+    if (StringUtils.isNotBlank(result.id())) {
+      return result.id();
+    }
+    return Integer.toHexString(Objects.hash(result.elementId(), result.completedAt()));
```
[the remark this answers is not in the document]
nikonovd: Confirmed, fixed in ec40d28. Same fix: SHA-256 digest over the element id and completion timestamp instead of Integer.hashCode().

connectors/agentic-ai/docs/adr/013-unified-agent-instance-turn-updates.md
```diff
@@ -0,0 +1,110 @@
...
+* The engine's create-agent-instance command still carries a single history-item write, not a batch
+  (Req 1, deferred — see below), so the first turn's create is not yet lease-fenced the same way.
+
+## Deferred
+
+**History batch on create (Req 1).** camunda#59784 (engine) has no PR yet. Until it lands, `create()`
+keeps sending its mandatory `model`/`provider`/`systemPrompt` (+ `limits`) fields directly on the create
+command, and the change-detection fingerprint is seeded from those same values so the first turn's
+`CONFIGURATION` item is only emitted when they already differ from what create sent. When camunda#59784
+lands, create should carry a `CONFIGURATION` item instead and this seeding drops out.
```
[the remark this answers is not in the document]
nikonovd: Confirmed, corrected in 4f19476. No seeding exists in code today — AgentMetadata starts with an empty configurationFingerprintHistory, so the first turn always emits CONFIGURATION, redundant with what create() already sent. Rewrote the paragraph to describe that instead of the seeding mechanism that was never implemented.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
```diff
@@ -27,11 +35,23 @@ public record AgentConfiguration(
     @Nullable MemoryConfiguration memory,
     @Nullable LimitsConfiguration limits,
     @Nullable EventHandlingConfiguration events,
-    @Nullable ResponseConfiguration response) {
+    @Nullable ResponseConfiguration response,
+    List<ToolDefinition> tools) {
```
maff: 🔧 Let's align the naming to `toolDefinitions` as we use it in the agent context.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceClient.java
```diff
@@ -39,57 +42,57 @@ void update(
...
    */
-  void createHistoryForInputMessages(
+  void applyTurnStart(
       AgentExecutionContext executionContext,
       @Nullable AgentInstanceKey agentInstanceKey,
       AgentConversationTurn turn,
       Optional<AgentConversationTurn> previousTurn,
-      OffsetDateTime turnIngestionTimestamp);
+      OffsetDateTime turnIngestionTimestamp,
+      AgentConfiguration configuration);
```
maff: 🔧 Nit: order-wise would it make sense to move the configuration further up the parameters list (e.g. after executionContext) to align this with the request handler?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
```diff
@@ -421,18 +450,39 @@ private ConnectorException buildUpdateException(
...
-    return new ConnectorException(ERROR_CODE_AGENT_INSTANCE_HISTORY_ITEM_FAILED, message, cause);
+    if (reason == FailureReason.PERMANENT_ERROR && isNotFound(cause)) {
+      return ConnectorRetryException.builder()
+          .errorCode(ERROR_CODE_AGENT_INSTANCE_SUPERSEDED)
+          .message(
+              "Agent instance update rejected: the job activation has been superseded: %s"
+                  .formatted(cause.getMessage()))
+          .cause(cause)
+          .retries(0)
+          .build();
```
maff, line 469: 💭 Will this raise an incident or silently complete the job?

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceHistoryMapper.java
```diff
@@ -145,8 +149,24 @@ private Map<String, Object> argumentsForResult(
                         .formatted(result.id())));
   }
 
+  /**
+   * A history-item content ordering: {@code text}, {@code reasoning}, {@code object}, {@code
+   * document}, then everything else, stable within each group.
+   */
+  private static final Comparator<Content> ASSISTANT_CONTENT_ORDER =
```
maff: 💭 I'm not sure if reordering the content here is really a good idea. One part is how we display "technical" content (encrypted reasoning) in the UI, but otherwise I think the natural order mostly makes sense. Hypothetical example of content blocks in an assistant message (`t` = text, `r` = reasoning):

```
t0: Yay, finally some work to do.
r0: The user asked me to work on task <task>, let's think about how to solve this.
r1: A-ha, I think I found a solution. Let's think harder to verify.
t1: Still working on it...
r2: <encrypted blob>, no plain text
r3: <encrypted blob> + "Confirmed, my solution is good"
t2: Hey user, here is your solution: <solution>
```

By re-ordering we are re-shuffling this, which might lead to harder understanding of the actual content:

```
t0: Yay, finally some work to do.
t1: Still working on it...
t2: Hey user, here is your solution: <solution>
r0: The user asked me to work on task <task>, let's think about how to solve this.
r1: A-ha, I think I found a solution. Let's think harder to verify.
r2: <encrypted blob>, no plain text
r3: <encrypted blob> + "Confirmed, my solution is good"
```

⏩ I'd prefer to keep the stored order and look into how to best present this in the UI. What we could do as long as the API does not have the primitives for reasoning/provider content it is to only send human-readable reasoning output from `ReasoningContent.text` and omit the rest of `ReasoningContent` (+ potentially also `ProviderContent`) for the moment, or just show a synthetic message (e.g. `Thinking...` or `Encrypted thinking content` or similar).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
```diff
@@ -208,19 +217,8 @@ private R proceed(
...
-      notifyMetrics(executionContext, storedConversation, agentResponse, true);
-      return buildConnectorResponse(
-          executionContext, storedConversation, agentResponse, messageStorageCompletionListener);
-    }
-
     return buildConnectorResponse(
-        executionContext,
-        storedConversation,
-        agentResponse,
-        AgentJobCompletionListener.compose(
```
maff, line 222: 🔧 Is the `compose` implementation still used? Otherwise we could remove it.

connectors/agentic-ai/docs/adr/013-unified-agent-instance-turn-updates.md
```diff
@@ -0,0 +1,114 @@
...
+create, and a final status/metrics update. Each was retried independently via `CamundaApiRetry` with no
+all-or-nothing guarantee, so a mid-turn failure could leave the agent instance with some items
+committed and others not, with no recovery path. Metrics and the assistant history item derive from the
+same LLM response and were reported through two unrelated code paths, sometimes in the same call and
+sometimes deferred to a job-completion listener depending on which connector flavor and turn shape was
+in play — a distinction that existed only to decide whether a PATCH was safe to defer, not because the
+data itself differed.
+
+The engine gained a batched `history[]` list on the `UPDATE` command plus `jobKey`/`jobLease` and
+per-item `historyItemId` (camunda#58789, camunda#59714, camunda#59756), making it possible to collapse
```
maff: 🔧 Nit: these issue links don't work - replace with camunda/camunda#58789, etc.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceHistoryMapper.java
nikonovd: Let's remove the ordering as discussed.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: It won't as the failJob command is ignored engine wise, as the job is not existing anymore.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
nikonovd: Should be possible as the Agent instance query APIs now support RDBMS

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/AgentConfiguration.java
nikonovd: Renamed to `toolDefinitions` (0a3e4235) to match `AgentContext`/`ConversationSnapshot`/the converter chain. Fingerprint hashes the field value, not its name, so this doesn't change existing fingerprints.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceClient.java
nikonovd: Moved `configuration` next to `executionContext` (d3dc8814).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agent/BaseAgentRequestHandler.java
nikonovd: Confirmed dead — this diff's rewire removed the only call site, and it wasn't test-covered on its own. Removed (a43bbd88).

connectors/agentic-ai/docs/adr/013-unified-agent-instance-turn-updates.md
nikonovd: Fixed (8ad29839) — also fixed the same pattern at lines 102/109/113 in this ADR and the one in ai-agent.md, which had the identical issue.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/AgentInstanceHistoryMapper.java
nikonovd: Reverted the sort (5b9e3bdc) — keeping stored order per your ⏩. Left the display-filtering half (reasoning-text-only vs synthetic placeholder) as a separate, un-actioned question — that's a UI-presentation call, not something I want to fold into this revert.

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: Raises an incident, doesn't silently complete. Traced it: `ConnectorRetryException.retries(0)` → `OutboundConnectorExceptionHandler.handleConnectorRetryException` → `ErrorResult(..., retries=0)` → `SpringConnectorJobHandler` fails the job via `newFailCommand(job).retries(0)`, explicitly bypassing BPMN error-expression routing (per the comment at that call site — this path is Java-exception-driven, not a modeled `ConnectorError`). Zeebe raises an incident on a job failed with 0 remaining retries.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
nikonovd: Not right now — `agentInstanceClient` in this e2e harness is a Mockito mock, there's no real engine-backed agent-instance API to query against in this test setup yet. Worth revisiting once/if that becomes available; for now the mock-verification approach is what the harness supports.

maff (approved):
Thanks for addressing the comments. Pre-approving with the Claude comment on the incident/supersession still open. Otherwise looks good from what I can judge! 🚀

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/agentinstance/CamundaAgentInstanceClient.java
nikonovd: Correction to my last reply — I was wrong, checked deeper. The `jobLease` sent on the update is `executionContext.jobContext().getLeaseToken()`, i.e. Zeebe's own per-activation job lease (`CamundaAgentInstanceClient.java:418-420`). A 404 here means that exact token is stale. Zeebe's job-lease ADR (camunda/camunda `zeebe/docs/adr/0005-810-job-lease.md`, D4) states complete/fail/throw-error on a leased job require the matching token, `INVALID_STATE` otherwise, non-retryable. The runtime's `newFailCommand(job)` auto-attaches that same (now-stale) token (`JobClientImpl`: `withLeaseToken(job.getLeaseToken())`), so the ensuing FailJobCommand hits the identical fencing check and is rejected by the engine — it never reaches the retries-to-zero path that raises an incident. Per that ADR's Consequences section: "Workers must treat a lease-mismatch rejection as the expected outcome of losing an activation race — drop the work, log at debug — not as an error." So: no incident, the fail command is silently rejected. Sorry for the noise on the earlier answer.

PR #8472, opened 2026-08-27
ci: report SaaS and cancelled AlwaysGreen failures to Slack

hilalbursalii:
## Description

SaaS E2E failures on this branch currently produce **no AlwaysGreen feedback at all** — no Slack
message, no PR comment. Two independent causes, both fixed here.

**1. There is no SaaS feedback job on this branch.** `saas-e2e-debug-summary` exists only on `main`
and was never backported, so `sm-e2e-debug-summary` is the only feedback job here and it is gated on
the SM result alone. A SaaS failure has nothing listening to it. This PR adds the mirror job, using
this branch's own pinned action versions (`checkout@v6`, `setup-node@v4`, node 20) rather than
main's, and reusing the `post-failure-feedback` action already present here.

**2. `trigger-saas-e2e` times out before its own wait loop does.** The job carries
`timeout-minutes: 25`, but the "Wait for downstream workflow to finish" loop is dimensioned for
40 minutes (`for _ in {1..80}` × `sleep 30`). Any downstream run over ~25 minutes therefore ends the
job as **`cancelled`**, not `failure` — and a job timeout also drops the remaining steps, so an
`if: always()` status-recording fallback never runs either. Raising the job timeout to 50 lets the
loop hit its own limit and `exit 1`, which surfaces correctly as `failure`.

As a backstop, both feedback gates now treat `cancelled` as red **on `push` only**. On a protected
branch the concurrency group is keyed by `github.sha`, so nothing supersedes a run and a cancellation
means a timeout or an abort. `merge_group` runs are superseded routinely, so folding `cancelled` in
there would alert on every queue reshuffle.

### Worked example

Run [33001673998](https://github.com/camunda/connectors/actions/runs/33001673998) (push, `stable/8.8`):

| Time | Event |
|---|---|
| 19:00:38 | `Trigger SaaS E2E tests` starts |
| 19:00:47 | Downstream run [33002856062](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/33002856062) created |
| 19:25:52 | Upstream job killed by `timeout-minutes: 25`, downstream still `in_progress` |
| 19:32:15 | `Run tests (chromium-v1)` **fails** |
| 19:37:08 | `Run tests (chromium-v2)` **fails** |
| 19:38:11 | Downstream concludes `failure` |

The watcher was dead 6–11 minutes before the failures existed, and nothing on this branch would have
reported them even if it had survived.

### Known follow-up, not fixed here

Delivery is still blocked on a Slack-side change: the bot behind `CONNECTORS_SLACK_BOT_TOKEN`
(**Connector Slack Automation**, `U05Q4U3PX7Z`) is **not a member of `#alwaysgreen-reliability`**
(`C0AK0AVKRL0`). Every post returns `not_in_channel` and is swallowed — `post-failure-feedback` logs
`[feedback] Slack post failed: not_in_channel` under `continue-on-error: true`, and
`connectors-streak-detector.yml` passes `errors: false`, so both report success while delivering
nothing. The app must be invited to the channel before any of this reaches Slack; the PR-comment path
works regardless.

Automatic triage/fix-agent dispatch is deliberately **not** backported — `alwaysgreen-triage.yml`
and `alwaysgreen-fix.yml` are absent on the stable branches. This PR is notifications only.

## Related issues

Surfaced while triaging https://github.com/camunda/connectors/actions/runs/33001673998

## Checklist

### Companion PRs

- main: camunda/connectors#8475
- stable/8.9: camunda/connectors#8473
- stable/8.8: camunda/connectors#8472
- stable/8.7: camunda/connectors#8474

### DIFF
--- a/.github/workflows/MERGE_QUEUE_HELM_TEST.yaml
+++ b/.github/workflows/MERGE_QUEUE_HELM_TEST.yaml
@@ -171,7 +171,13 @@ jobs:
   sm-e2e-debug-summary:
     name: AlwaysGreen failure feedback (SM E2E)
     needs: trigger-sm-e2e
-    if: always() && needs.trigger-sm-e2e.result == 'failure'
+    # `cancelled` is included for push only: on a protected branch the concurrency
+    # group is keyed by sha so nothing supersedes a run, meaning a cancellation is a
+    # timeout or an abort rather than a newer commit. merge_group runs are superseded
+    # routinely, so folding `cancelled` in there would alert on every queue reshuffle.
+    if: >-
+      always() && (needs.trigger-sm-e2e.result == 'failure' ||
+      (github.event_name == 'push' && needs.trigger-sm-e2e.result == 'cancelled'))
     runs-on: ubuntu-latest
     timeout-minutes: 5
     continue-on-error: ${{ github.event_name == 'merge_group' }}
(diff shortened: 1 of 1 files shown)


hilalbursalii:
Superseded by #8469, which wires the full AlwaysGreen stack (including `saas-e2e-debug-summary`) into `stable/8.8` and does it better — referencing `alwaysgreen-triage.yml` and `post-failure-feedback` at `@main` instead of duplicating them onto the branch.

Closing this in favour of #8469. The one thing #8469 does not carry is the `timeout-minutes: 25` → `50` fix and the push-scoped `cancelled` gates — that stays on #8475 (`main`) and will be backported by label once #8469 has landed. Verified by dry-run: with #8469 applied, all four hunks of #8475 apply cleanly to `stable/8.8`, so no manual backport is needed.

PR #8486, opened 2026-08-28
fix: override reactor-core and reactor-netty to fixed versions

mathias-vandaele:
## Summary

Overrides `io.projectreactor:reactor-core` and `io.projectreactor.netty:reactor-netty-core`/`reactor-netty-http`
transitive versions (pulled in via the Azure SDK) to fixed versions ahead of what the current
Spring Boot BOM manages, with an enforcer guard to revisit once Spring Boot ships the fix natively.

Security fix. Details in the internal tracking issues: camunda/team-connectors#1268, camunda/team-connectors#1269

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -95,6 +95,11 @@ limitations under the License.</license.inlineheader>
          Ref: https://github.com/rabbitmq/rabbitmq-java-client/security/advisories/GHSA-jh4v-gfqj-7rhx -->
     <version.amqp-client>5.35.0</version.amqp-client>
     <tomcat.version>10.1.59</tomcat.version>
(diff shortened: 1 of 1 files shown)


mathias-vandaele:
Closing — CVE-2026-47863 (the reason for this override) is not reachable in our deployment. Verified by searching the compiled bytecode of azure-core 1.59.0, azure-core-http-netty 1.16.6, azure-identity 1.18.5, and azure-ai-documentintelligence 1.0.10 for `bufferTimeout`/`fairBackpressure` — zero matches in any of them, and our own code never calls it either. CVE-2026-47874 was already assessed as not-affected (reactor-netty used only as an outbound client here, never a server). No security fix needed on this branch; tracking issue updated accordingly.

PR #8487, opened 2026-08-28
fix: override reactor-core and reactor-netty to fixed versions

mathias-vandaele:
## Summary

Overrides `io.projectreactor:reactor-core` and `io.projectreactor.netty:reactor-netty-core`/`reactor-netty-http`
transitive versions (pulled in via the Azure SDK) to fixed versions ahead of what the current
Spring Boot BOM manages, with an enforcer guard to revisit once Spring Boot ships the fix natively.

Security fix. Details in the internal tracking issues: camunda/team-connectors#1268, camunda/team-connectors#1269


mathias-vandaele:
Closing — same reasoning as camunda/connectors#8486: CVE-2026-47863 is not reachable in our deployment (verified via bytecode search of the actual Azure SDK jars we depend on — no bufferTimeout/fairBackpressure usage anywhere in the chain), and CVE-2026-47874 was already not-affected. No security fix needed on this branch; tracking issue updated accordingly.

PR #8493, opened 2026-08-28
fix: allow-list secrets on a SubProcess/AdHocSubProcess element itself

### LINKED ISSUE #8490: Outbound secret allow-list never includes the SubProcess/AdHocSubProcess element itself (breaks AI Agent Sub-process secrets under STRICT filter)
## Description

The connector runtime's outbound secret allow-list never includes a `SubProcess`/`AdHocSubProcess` element itself — only elements *inside* one. Any outbound connector whose `zeebe:ioMapping` is attached directly to the sub-process boundary (which is exactly how the AI Agent Sub-process connector works) gets an empty allow-list for every secret it references. Under the `STRICT`/`LAX` secret filter (live on SaaS since PR #8368, 2026-08-26), an empty allow-list denies every secret name, and a denied secret resolves to its own literal, unresolved placeholder text rather than failing loudly.

## How this was found

Running the on-demand SaaS 8.10 workflow against a branch that migrates AI Agent BPMN fixtures from the v1 to the v2 element template ([camunda/c8-cross-component-e2e-tests#3206](https://github.com/camunda/c8-cross-component-e2e-tests/pull/3206), run [33193916713](https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/33193916713)). The AI Agent **sub-process** flavor test failed with a real Operate incident (screenshot confirms "1 incident", not a UI timing issue):

```
Ad hoc sub process error (no retries left).
errorType: AD_HOC_SUB_PROCESS_NO_RETRIES
errorMessage: Configured region (%7B%7Bsecrets.IDP_AWS_REGION%7D%7D) resulted in an invalid URI:
              https://bedrock-runtime.%7B%7Bsecrets.IDP_AWS_REGION%7D%7D.amazonaws.com. This is
              usually caused by an invalid region configuration.
              | Error variables: {error={message=..., type=software.amazon.awssdk.core.exception.SdkClientException}}
```

`%7B%7B...%7D%7D` is URL-encoded `{{...}}` — the literal, unresolved `{{secrets.IDP_AWS_REGION}}` placeholder was passed straight into the AWS SDK as the region, which then fails building a URI from it. The secret never resolved; the connector received the placeholder text itself as its input value.

Not a regression from that BPMN migration: the fixture's provider input mapping (`{{secrets.IDP_AWS_REGION}}` on `provider.bedrock.region`) is unchanged from the working v1 template — same secret reference syntax the AI Agent **task**-flavor fixtures use successfully in the same run. Only the sub-process flavor fails, because of where its `zeebe:ioMapping` is attached (see below).

## Root cause

`ProcessDefinitionSecretKeyCache` (`connector-runtime/connector-runtime-spring/.../outbound/secret/ProcessDefinitionSecretKeyCache.java`) builds the outbound secret allow-list per element. Its element-collection walk (`collectFlowElements`):

```java
private Collection<FlowElement> collectFlowElements(
    final Collection<FlowElement> processFlowElements, final Collection<FlowElement> buffer) {
  for (FlowElement element : processFlowElements) {
    if (element instanceof SubProcess subprocess) {
      buffer.addAll(retrieveEligibleElementsFromSubprocess(subprocess));
      continue;   // <-- the SubProcess element itself is never added to buffer
    }
    buffer.add(element);
  }
  return buffer;
}
```

When it meets a `SubProcess` (which `AdHocSubProcess` is a subtype of), it recurses into the sub-process to collect eligible *children*, then `continue`s — the sub-process element itself never enters `buffer`, so it's never checked against `OUTBOUND_ELIGIBLE_TYPES` and never gets an allow-list entry, regardless of whether it carries its own `zeebe:ioMapping`.

This is exactly the failure mode already described in #8359 ("Outbound secret filter denies all secrets for tasks without a Modeler element template"): an element missing from the allow-list map gets `Collections.emptyList()` from `getSecretKeys`. The live implementation backing the filter is `LazyLoadingSecretFilter`, not `SecretFilter.allowOnly` directly, but the semantics are identical: a non-null, empty `Set` denies every name (only a *null* lookup result, meaning the lookup itself failed, allows everything under `LAX`). `SecretHandler` then logs "placeholder left unreplaced" and returns `null` for the denied name, and `SecretUtil.resolveSecretValue` falls back to `matcher.group()` — the literal, unresolved placeholder text — instead of raising anything. #8359's fix (in the same PR #8368) removed the `&& isElementTemplate(element)` guard that had been excluding tasks authored without a Modeler template. It did not touch `collectFlowElements` or the `OUTBOUND_ELIGIBLE_TYPES` list at all — the type restriction there predates PR #8368 and was never in scope for that fix, which is why the sub-process-element gap is a distinct, still-open issue in the same function.

The AI Agent Sub-process connector (both v1 `io.camunda.connectors.agenticai.aiagent.jobworker.v1` and v2 `io.camunda.connectors.agenticai.ai-agent-subprocess.v2`) attaches its `zeebe:ioMapping` — including all provider/secret configuration — directly to the `bpmn:adHocSubProcess` element, not to a child task inside it. It is exactly the shape this gap misses.

## Why this surfaced now

- PR #8368 (merged 2026-08-26) is what sets `camunda.connector.secret-resolver.secret-filter.mode=STRICT` for the SaaS bundle (`apps/bundle/camunda-saas-bundle/src/main/resources/application.properties`). Filter ships `DISABLED` everywhere else — this bug has no effect until the filter is active.
- #8441 (whole-cluster secret loading broken, closed 2026-08-27) was independently causing every secret-dependent job to incident out for an unrelated reason, which likely masked this gap until that was fixed.
- No existing unit test exercises this shape: `ProcessDefinitionSecretKeyCacheTest` has 6 cases, all against Service/Send/Script/BusinessRule tasks (`outbound-with-secrets.bpmn`, `outbound-multiple-tasks-with-secrets.bpmn`, `outbound-no-secrets.bpmn`, `outbound-no-template.bpmn`, `outbound-task-types.bpmn`) — none attach a `zeebe:ioMapping` to a `SubProcess`/`AdHocSubProcess` element itself. Confirmed via `gh api` listing of the test resource directory: no such fixture exists.

## Impact

Any outbound connector configured directly on a sub-process/ad-hoc-sub-process boundary element and referencing `{{secrets.*}}` will silently receive the literal placeholder text instead of the resolved secret, once the SaaS `STRICT`/`LAX` filter is active. Confirmed affected: AI Agent Sub-process connector (both v1 and v2 templates) on any provider/auth mode using secret references. Likely affects any other connector or hand-modeled sub-process using the same "ioMapping on the sub-process element" pattern — not yet checked against other connector types.

## Suggested fix

Two changes are required together — adding the element to `buffer` alone is a no-op, since the sole caller (`retrieveOutboundEligibleElementsFromProcess`) filters that buffer against `OUTBOUND_ELIGIBLE_TYPES` via `iet.isInstance(element)`, and `SubProcess`/`AdHocSubProcess` isn't in that list:

1. Add `SubProcess.class` (covers `AdHocSubProcess` via inheritance) to the `OUTBOUND_ELIGIBLE_TYPES` static block.
2. In `collectFlowElements`, add the `SubProcess` element itself to `buffer` before/alongside recursing into `retrieveEligibleElementsFromSubprocess(subprocess)` for its children — currently it only adds the children and `continue`s past the sub-process element.

Worth calling out explicitly in the PR: this widens eligibility to *every* embedded/event/transaction sub-process, not just ad-hoc ones — low risk since only elements that actually carry a `zeebe:ioMapping` yield any keys, but a maintainer should be aware of the wider blast radius. An alternative is adding `AdHocSubProcess.class` specifically instead of the broader `SubProcess.class`, trading a smaller blast radius for not covering plain/embedded sub-processes that might have the same issue.

Add a test fixture with an outbound connector's `zeebe:ioMapping` attached directly to a `SubProcess`/`AdHocSubProcess` element (not a child task), mirroring the existing `outbound-with-secrets.bpmn` pattern.

## Reproduce

1. On a SaaS 8.10 cluster with `secret-filter.mode=STRICT` (default for SaaS since 2026-08-26), deploy a BPMN with an `AdHocSubProcess` whose `zeebe:ioMapping` references `{{secrets.SOME_SECRET}}` directly on the sub-process element (e.g. the AI Agent Sub-process connector template).
2. Start an instance.
3. The job receives the literal `{{secrets.SOME_SECRET}}` string as its input value instead of the resolved secret, and fails downstream with whatever error the unresolved placeholder text causes at the target system (in this case, an invalid AWS region URI).

## References

- Failing run: https://github.com/camunda/c8-cross-component-e2e-tests/actions/runs/33193916713
- Related: #8359 (same allow-list gap, task-without-template case — fixed), #8441 (unrelated whole-cluster secret loading failure, fixed, may have masked this), PR #8368 (enabled STRICT for SaaS + fixed #8359, but not this)

Fingerprint: nightly-product-bug fp=38dc6cee

Fingerprint: nightly-product-bug fp=1a524d4b

gbetances089:
## What this fixes

Fixes #8490.

If a connector's configuration (including secret references like `{{secrets.MY_SECRET}}`) is set directly on a sub-process element in a BPMN diagram — rather than on a task inside it — the runtime silently ignores every secret reference on that element. It doesn't throw an error. Instead, the literal, unresolved text `{{secrets.MY_SECRET}}` gets used as the actual value wherever the secret was supposed to go.

This is exactly what happened to the AI Agent Sub-process connector on a live SaaS 8.10 cluster: its AWS region field ended up literally containing the text `{{secrets.IDP_AWS_REGION}}` instead of the real region name, which then broke when the AWS SDK tried to build a URL out of it. See #8490 for the full incident trace.

The AI Agent Sub-process connector is the one we caught this on, but the bug applies to *any* connector configured this way, not just that one.

## Why this happens

The runtime keeps a list of which secrets each element in a process is "allowed" to use, built from scanning the deployed diagram. That list is what lets the secret filter (enabled on SaaS today) tell a legitimate secret reference apart from something that shouldn't have access. The code that builds this list checks every task in the diagram — but it was skipping sub-process elements themselves. It would look *inside* a sub-process for tasks, but never checked the sub-process boundary itself, even when a connector's configuration lived right there.

So any sub-process configured this way ended up with an empty "allowed secrets" list, and an empty list is treated as "this element isn't allowed to use any secrets at all" — which silently produces the wrong value instead of a clear error.

## The fix

The scanning logic now also considers the sub-process element itself, not just what's inside it. Two related changes were needed together for this to actually work; either one alone wouldn't have been enough.

## Testing

Added a small test BPMN diagram with a connector configuration directly on a sub-process element (mirroring the existing test-fixture pattern in this file), plus two new test cases:
- the sub-process element correctly picks up its own secrets
- a task inside that sub-process that has no secrets of its own still correctly gets nothing (proving the fix doesn't accidentally leak the parent's secrets to unrelated children)

All 8 tests in `ProcessDefinitionSecretKeyCacheTest` pass (6 existing + 2 new). Also ran the wider test suite for this module locally with no unrelated regressions (one pre-existing, unrelated failure that requires Docker, not present in this environment).

We also verified this end-to-end on a real, running Camunda 8 cluster (self-managed, GKE) — reproduced the original bug on the unfixed code, confirmed it's actually fixed, and checked several related edge cases. See the verification comment below for the full write-up and the run link.

## Backport

This bug is present on `main` and on `stable/8.9`/`stable/8.10` identically. It's currently only *live* in practice on SaaS 8.10, because that's the only place the secret filter that exposes it is turned on today — but flagging for a decision on whether to backport, since the underlying code is broken everywhere the filter could be enabled.

## Reviewer checklist

- [ ] Confirm whether this should backport to `stable/8.9` and/or `stable/8.10`
- [x] Confirm the scope of the fix (it now applies to every kind of sub-process, not just the ad-hoc kind used by the AI Agent connector) is the right call, or should be narrowed — **verified on a real cluster**: embedded, event, multi-instance, and nested sub-processes all resolve secrets correctly with this change; transaction sub-processes aren't a valid Camunda 8 element at all, so they're unaffected either way. See the verification comment below for details and the full run.

### DIFF
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CorrelateMessageCommandDummy.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CorrelateMessageCommandDummy.java
@@ -74,6 +74,16 @@ public class CorrelateMessageCommandDummy
   }
 
   @Override
+  public CorrelateMessageCommandStep3 addVariable(String key, Object value) {
+    return this;
+  }
+
+  @Override
+  public CorrelateMessageCommandStep3 addVariables(Map<String, Object> variables) {
+    return this;
+  }
+
+  @Override
   public CorrelateMessageCommandStep3 tenantId(String tenantId) {
     return this;
   }
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CorrelateMessageCommandDummy.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CorrelateMessageCommandDummy.java
@@ -74,16 +74,6 @@ public class CorrelateMessageCommandDummy
   }
 
   @Override
-  public CorrelateMessageCommandStep3 addVariable(String key, Object value) {
-    return this;
-  }
-
-  @Override
-  public CorrelateMessageCommandStep3 addVariables(Map<String, Object> variables) {
-    return this;
-  }
-
-  @Override
   public CorrelateMessageCommandStep3 tenantId(String tenantId) {
     return this;
   }
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CreateCommandDummy.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CreateCommandDummy.java
@@ -167,16 +167,6 @@ public class CreateCommandDummy
   }
 
   @Override
-  public CreateProcessInstanceCommandStep3 addVariable(String key, Object value) {
-    return this;
-  }
-
-  @Override
-  public CreateProcessInstanceCommandStep3 addVariables(Map<String, Object> variables) {
-    return this;
-  }
-
-  @Override
   public CreateProcessInstanceCommandStep3 tenantId(String tenantId) {
     return this;
   }
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CreateCommandDummy.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/CreateCommandDummy.java
@@ -167,6 +167,16 @@ public class CreateCommandDummy
   }
 
   @Override
+  public CreateProcessInstanceCommandStep3 addVariable(String key, Object value) {
+    return this;
+  }
+
+  @Override
+  public CreateProcessInstanceCommandStep3 addVariables(Map<String, Object> variables) {
+    return this;
+  }
+
+  @Override
   public CreateProcessInstanceCommandStep3 tenantId(String tenantId) {
     return this;
   }
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/PublishMessageCommandDummy.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/PublishMessageCommandDummy.java
@@ -82,6 +82,16 @@ public class PublishMessageCommandDummy
   }
 
   @Override
+  public PublishMessageCommandStep3 addVariable(String key, Object value) {
+    return this;
+  }
+
+  @Override
+  public PublishMessageCommandStep3 addVariables(Map<String, Object> variables) {
+    return this;
+  }
+
+  @Override
   public PublishMessageCommandStep3 tenantId(String tenantId) {
     return this;
   }
--- a/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/PublishMessageCommandDummy.java
+++ b/connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/testutil/command/PublishMessageCommandDummy.java
@@ -82,16 +82,6 @@ public class PublishMessageCommandDummy
   }
 
   @Override
-  public PublishMessageCommandStep3 addVariable(String key, Object value) {
-    return this;
-  }
-
-  @Override
-  public PublishMessageCommandStep3 addVariables(Map<String, Object> variables) {
-    return this;
-  }
-
-  @Override
   public PublishMessageCommandStep3 tenantId(String tenantId) {
     return this;
   }
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
@@ -53,6 +53,10 @@ public class ProcessDefinitionSecretKeyCache implements SecretKeyCache {
     OUTBOUND_ELIGIBLE_TYPES.add(SendTask.class);
     OUTBOUND_ELIGIBLE_TYPES.add(ScriptTask.class);
     OUTBOUND_ELIGIBLE_TYPES.add(BusinessRuleTask.class);
+    // Covers AdHocSubProcess too (AdHocSubProcess extends SubProcess): a connector's
+    // zeebe:ioMapping can be attached directly to the sub-process boundary element itself
+    // (e.g. the AI Agent Sub-process connector), not just to a child task inside it.
+    OUTBOUND_ELIGIBLE_TYPES.add(SubProcess.class);
   }
 
   private final String physicalTenantId;
@@ -161,9 +165,12 @@ public class ProcessDefinitionSecretKeyCache implements SecretKeyCache {
   private Collection<FlowElement> collectFlowElements(
       final Collection<FlowElement> processFlowElements, final Collection<FlowElement> buffer) {
     for (FlowElement element : processFlowElements) {
-      // if we detect a subprocess, we have to expand it
-      // its building blocks to identify where are connectors
+      // if we detect a subprocess, we have to expand it to identify where are connectors in its
+      // building blocks, but the subprocess element itself may also carry a connector directly
+      // (e.g. an ad-hoc sub-process connector's zeebe:ioMapping lives on the boundary element,
+      // not on a child), so it must be added to the buffer too, not skipped via continue.
       if (element instanceof SubProcess subprocess) {
+        buffer.add(element);
         buffer.addAll(retrieveEligibleElementsFromSubprocess(subprocess));
         continue;
       }
--- a/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCacheTest.java
+++ b/connector-runtime/connector-runtime-spring/src/test/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCacheTest.java
@@ -112,6 +112,34 @@ class ProcessDefinitionSecretKeyCacheTest {
   }
 
   @Test
+  void getSecretKeys_adHocSubProcessWithSecretsOnBoundaryElement_returnsItsOwnSecrets()
+      throws IOException {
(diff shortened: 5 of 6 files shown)


gbetances089:
## Extra verification on a real cluster

On top of the unit tests, we ran this fix against a real, running Camunda 8 cluster (self-managed, on GKE) to confirm it actually behaves correctly in practice, not just that the code compiles and a mocked test passes. Full run: https://github.com/camunda/team-qa-engineering/actions/runs/33369946033

### The core bug, reproduced and fixed

We built a small process with a connector configured directly on an ad-hoc sub-process element (the same shape the AI Agent Sub-process connector uses), gave it two secrets to resolve, and turned on the same strict secret filter that's active on SaaS today.

- On the old code (built from `main`, no fix): the connector silently sent the literal, unresolved text `{{secrets.SPBOUNDARY_ALPHA}}` as part of its outbound request. No error, no incident. This is exactly the failure described in the original bug report.
- On this PR's code: both secrets resolved to their real values, and nothing unresolved showed up anywhere in the outbound request.

We captured the actual outbound HTTP request in both cases (via a small echo endpoint inside the cluster) to see the real payload each connector sent, rather than just checking whether the process instance succeeded or failed.

### Checked the fix isn't too broad or too narrow

The fix works by telling the runtime "sub-process elements can carry their own secrets too," which technically applies to every kind of sub-process in BPMN, not just the ad-hoc one this bug was found on. We wanted to make sure that didn't quietly change behavior for other sub-process types, so we tried the same setup against:

- A regular embedded sub-process — worked correctly.
- An event sub-process — worked correctly.
- A multi-instance sub-process — worked correctly.
- A sub-process nested inside another sub-process — worked correctly.
- A transaction sub-process — this one isn't actually a valid Camunda 8 element at all (rejected by the same BPMN linter Modeler uses), so it was never reachable either before or after this change.

We also checked the fix doesn't accidentally let a child task inside a sub-process see or reuse its parent's secrets: it doesn't. Each element only resolves what it declares for itself.

And we checked the filter itself hasn't been loosened: a secret name that isn't declared anywhere in the process still gets rejected and stays as literal text, exactly as it should.

### One risk we specifically ruled out

Widening which elements the runtime scans for secrets could, in theory, make an existing edge case more likely to hit: if two different parts of the same diagram (two separate pools in a collaboration) reused the same element ID, the runtime's secret lookup could throw an unhandled error and break secret resolution for the whole process definition.

We tested this directly by deploying a diagram built exactly that way. Camunda 8 itself rejects it at deploy time with a validation error, before the process can even be created. So this was never actually reachable, either before or after this fix, and needs no follow-up.

### One unrelated thing we noticed, not a regression

If a connector job tries to run in the very first moment after its process is deployed, before the deployment is fully visible to the runtime yet, the strict secret filter can fail that specific job attempt with an error. We confirmed this happens the same way on a plain, ordinary service task with no sub-process involved at all, so it has nothing to do with this fix. In real usage this is invisible because Camunda automatically retries failed jobs. Noting it here for visibility, not as something this PR needs to address.

### Bottom line

15 checks run, all passed, including a genuine negative control that proves the old code really does have the bug this PR fixes. No follow-up ticket needed for the risks we went looking for.

gbetances089:
Closing this in favor of #8509, which fixes the same root cause (#8490) and is already sequenced into a broader multi-branch rollout (#8509→main, #8510→stable/8.9, #8511→release-8.10.0-alpha3), tied to the STRICT-by-default work in #8506/#8507/#8508.

One difference worth carrying over: this PR's fix was broader — `OUTBOUND_ELIGIBLE_TYPES.add(SubProcess.class)` rather than `AdHocSubProcess.class` specifically — so it also covers embedded, event, multi-instance, and nested sub-processes, not just the ad-hoc flavor. I verified this on a real Camunda 8.10 cluster (self-managed, GKE) before closing:

- Embedded, event, multi-instance, and nested sub-processes with a connector configured directly on the sub-process boundary all correctly resolve secrets with the `SubProcess.class` change.
- Transaction sub-processes aren't a valid Camunda 8 element at all (rejected by the platform's own BPMN linter), so they're unaffected either way regardless of which class is used.
- A genuine negative control confirmed the original bug on unfixed `main`: literal `{{secrets.X}}` text sent in the outbound payload, no incident.
- Checked and ruled out a theoretical duplicate-element-ID collision risk in the same allow-list code (`Collectors.toMap` on element IDs across pools) — Zeebe itself rejects duplicate IDs at deploy time (`cvc-id.2`), so this was never reachable before or after either fix.

Full run: https://github.com/camunda/team-qa-engineering/actions/runs/33369946033

If #8509's narrower `AdHocSubProcess`-only scope is intentional (e.g. to keep the change minimal), that's a reasonable call — just flagging that the broader fix is verified safe in case it's useful, or in case another sub-process flavor turns out to need the same coverage later.

PR #8518, opened 2026-08-31
fix: include AdHocSubProcess, IntermediateThrowEvent, and EndEvent in outbound secret key extraction

johnBgood:
## Summary

Replaces #8511, which was opened against `release-8.10.0-alpha3` — the correct target for this work is `release-8.10.0-alpha5`. Force-pushing the existing branch to retarget it isn't possible (branch protection blocks force-push), so this is a fresh branch with the same fix cherry-picked from main (`a4b95f32da`, `bf7f5302b9`) directly onto `release-8.10.0-alpha5`, which is already in sync with `main` for every file this touches.

`ProcessDefinitionSecretKeyCache`'s element traversal (`collectFlowElements`) replaces every `SubProcess` with only its children, and `OUTBOUND_ELIGIBLE_TYPES` doesn't include `AdHocSubProcess` either way. The current "AI Agent Sub-process" v2 element template targets `bpmn:AdHocSubProcess` directly and declares its own secrets (e.g. an LLM provider API key) in that element's own `zeebe:ioMapping` — so its allow-list always comes back empty, and every secret it references gets silently blocked once secret filtering is active.

The shipped Send Message connector templates target `bpmn:IntermediateThrowEvent` and `bpmn:EndEvent`, which had the same gap — neither type was in `OUTBOUND_ELIGIBLE_TYPES`, so their declared secrets were silently unresolvable too.

## Changes
- Add `AdHocSubProcess.class` to `OUTBOUND_ELIGIBLE_TYPES`.
- Special-case `AdHocSubProcess` in `collectFlowElements`: add the element itself to the eligible-elements buffer (so its own `zeebe:ioMapping` is inspected), while still recursing into its children (who may declare their own secrets too, same as any other subprocess).
- Add `IntermediateThrowEvent.class` and `EndEvent.class` to `OUTBOUND_ELIGIBLE_TYPES` (targeted by the shipped Send Message connector templates).

## Test plan
- [x] Added `outbound-adhoc-subprocess.bpmn` fixture: an `AdHocSubProcess` with its own secret plus a nested child task with a different secret, and a test confirming both resolve their own secret keys.
- [x] Added `outbound-message-events.bpmn` fixture and a test covering the intermediate-throw-event and end-event case.
- [x] Full suite for `connector-runtime-spring`: all green.
- [x] `spotless:check` clean.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@
 
 # Worktrees
 .worktree
- 
+
 # BlueJ files
 *.ctxt
 
--- a/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
+++ b/connector-runtime/connector-runtime-spring/src/main/java/io/camunda/connector/runtime/outbound/secret/ProcessDefinitionSecretKeyCache.java
@@ -20,6 +20,7 @@ import io.camunda.client.CamundaClient;
 import io.camunda.connector.runtime.core.secret.SecretUtil;
 import io.camunda.zeebe.model.bpmn.Bpmn;
 import io.camunda.zeebe.model.bpmn.BpmnModelInstance;
+import io.camunda.zeebe.model.bpmn.instance.AdHocSubProcess;
 import io.camunda.zeebe.model.bpmn.instance.BaseElement;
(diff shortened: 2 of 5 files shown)


johnBgood:
Closing per plan: main (#8509) now carries the `backport release-8.10.0-alpha5` label so the backport bot will create this automatically once #8509 merges, rather than maintaining it by hand.

PR #8519, opened 2026-08-31
fix: default outbound secret filter mode to STRICT

johnBgood:
## Summary

Replaces #8508, which was opened against `release-8.10.0-alpha3` — the correct target for this work is `release-8.10.0-alpha5`. Force-pushing the existing branch to retarget it isn't possible (branch protection blocks force-push), so this is a fresh branch with the same fix cherry-picked from main (`6ae298f893`, `b9ff127486`) directly onto `release-8.10.0-alpha5`, which is already in sync with `main` for every file this touches — including `OutboundConnectorExceptionHandler`, which alpha5 already carries in its fully-reworked (masking-safe) form, so none of the legacy-overload compatibility/security fixes from #8508 are needed here.

Follow-up to #7749 (the release-8.10.0-alpha3 backport of #7568), which shipped the outbound secret filter with a `DISABLED` default. A fresh install therefore stays unfiltered until an operator explicitly sets `camunda.connector.secret-resolver.secret-filter.mode=STRICT` (or `LAX`), which makes the security fix opt-in rather than on by default.

## Changes
- Flips the four independent `@Value` default declarations to `STRICT`: the three beans in `OutboundConnectorRuntimeConfiguration` (`secretFilterFactory`, `secretFilterFactoriesByPhysicalTenantId`, `outboundConnectorManager`) and the `legacyFallbackSecretFilterGuard` bean in `ConnectorsAutoConfiguration`, which independently reads its own default and must agree with the others or a `legacy.mode=FALLBACK` deployment with no explicit filter property would refuse to start.
- Updates ADR-0007, which still said the filter "ships disabled".

## Test plan
- [x] `OutboundConnectorsAutoConfigurationTest#shouldDefaultSecretFilterModeToStrict`, strengthened to assert on the legacy scalar bean, `secretFilterFactoriesByPhysicalTenantId`, and the factories actually held by `OutboundConnectorManager`.
- [x] `LegacySecretFallbackWiringTest` covers a `FALLBACK` deployment with no explicit `secret-filter.mode` booting successfully, proving the guard and the real filter bean agree on the default.
- [x] Full suite for `connector-runtime-spring` and `spring-boot-starter-camunda-connectors`: all green (189 tests).
- [x] `spotless:check` clean.


johnBgood:
Closing per plan: main (#8506) now carries the `backport release-8.10.0-alpha5` label so the backport bot will create this automatically once #8506 merges, rather than maintaining it by hand.

PR #8538, opened 2026-09-01
feat(runtime): extend secret filtering to inbound connectors

### LINKED ISSUE #7730: feat(runtime): extend secret filtering to inbound connectors
## Context

PR #7568 introduced per-job secret allow-listing for outbound connectors, restricting secret resolution to secrets declared in BPMN Zeebe input mappings.

Inbound connectors also resolve secrets (e.g. HTTP polling connector can fetch process variables — see #7308), so the same restriction should apply there too.

## Work

- Extend `SecretKeyCache` / `ProcessDefinitionSecretKeyCache` to also extract inbound connector secret keys from BPMN input mappings
- Wire a `SecretFilterFactory` into `InboundConnectorContextImpl` (analogous to how it is wired into `SpringConnectorJobHandler` for outbound)
- Honour the same `SecretFilterMode` config param (or a separate inbound-scoped one) so the feature can be enabled/disabled independently
- Add tests

## References

- PR #7568 (outbound implementation)

mathias-vandaele:
## Summary

- Restricts inbound secret resolution to the secret names declared in that connector's own deployed `zeebe:property` text, instead of always allowing every secret. This mirrors the outbound job path's model-derived allow-list (added in #7568), and honours the existing `camunda.connector.secret-resolver.secret-filter.mode` property (`DISABLED`/`LAX`/`STRICT`) — `DISABLED` keeps the prior allow-all behavior, `LAX`/`STRICT` both enable the filter (they only differ outbound, where the allow-list needs a remote BPMN lookup that can fail; inbound's allow-list is built from data already in memory).
- No new BPMN fetch, cache, or config bean is needed: `InboundConnectorContextImpl` already has each element's raw `zeebe:property` map in memory, so the allow-list is computed locally from `SecretUtil.retrieveSecretKeysInInput`, the same helper that builds the outbound allow-list.
- Adds a `secretFilterEnabled` overload to `InboundConnectorContextImpl` and `DefaultInboundConnectorContextFactory` (existing overloads delegate with `false`, matching the codebase's existing binary-compatibility-overload convention), so none of the ~50 existing test call sites needed updating.

Closes #7730

## Test plan

- [x] `InboundConnectorContextImplTest`: two new tests — a secret declared in the element's own property text still resolves; a secret not declared anywhere is left as a placeholder and the `SecretProvider` is never called.
- [x] Full existing suites pass unchanged: `connector-runtime-core`, `connector-runtime-spring` (506 tests), `spring-boot-starter-camunda-connectors` (189 tests), `connector-http-polling`.

---

## Risk ledger

Every material correctness/security concern raised on this PR, its status, and how it was falsified. "Broken to confirm" means the fix was reverted in place and the named test observed to fail — after two tests in this branch turned out to pass vacuously, a green test alone is not treated as evidence.

| # | Concern | Source | Status | Falsification evidence |
|---|---|---|---|---|
| 1 | Second-order chaining: `replaceSecrets` runs the brace pass then the bare pass **over that pass's output**, so a resolved secret value containing reference-shaped text yields a lookup for a name no model declares — reaching the cluster store under `FALLBACK` | QA re-test | **Fixed** — this is the filter's actual purpose | Broken to confirm: filter disabled ⇒ `leavesASecretNamedOnlyByAnotherSecretsValue` fails, `expected "secrets.CHAINED" but was "leaked-value"`. Live: same diagram flips BASIC auth 200↔401 on mode alone |
| 2 | Allow-list unioned names across sibling elements, so a name only a sibling declared was resolvable while binding another element | QA re-test / Copilot | **Fixed** — allow-list scoped per call to the text it filters | Broken to confirm: union restored ⇒ `leavesASecretOnlyASiblingElementDeclares` fails, `expected "secrets.ONLY_B" but was "leaked-b-value"` |
| 3 | Retained correlation result: a `BindableProcessElement` held past a hot swap was judged against the new details and denied a secret its own text declares | QA ledger C | **Fixed** — details read once per correlation, snapshot travels with the element; now unrepresentable, since the allow-list comes from the text in hand | Broken to confirm: binder pointed back at the live field ⇒ `resolvesASecretOfARetainedElementDroppedByAHotSwap` fails, `"secrets.KEPT"` vs `"kept-value"` — reproducing the reported failure exactly |
| 4 | Torn read within one request: `correlate()` and `bindProperties()` read the details separately, split by the broker round trip | QA ledger B / Copilot High | **Fixed** by the same single read as #3; unrepresentable after #2 | Mechanism removed structurally (one read). No unit test — reproducing it needs the 20-thread harness, and a test that only fails at ~47k swaps/s is not worth carrying |
| 5 | Reported 4.6% false-block rate | QA ledger B | **Rejected as stated** | 934,599 swaps in 20s ≈ 47,000 redeploys/second; production is a `@Scheduled(fixedDelay = 1000)` poll firing on actual deployment. Mechanism accepted (#4), magnitude is a harness artifact |
| 6 | Hot-swap visibility: `connectorDetails` written by the scheduler thread, read by Tomcat/connector threads with no shared lock | Copilot Medium | **Fixed** — `volatile` | Writer confirmed at `processEventQueue` (`@Scheduled`), readers at `InboundWebhookRestController:248`, `WebhookConnectorRegistry:109/131`, `InboundExecutableQueryService:135/147`. Initial publication was already safe via the state store's `ConcurrentHashMap` |
| 7 | `SecretFilterMode` → boolean wiring untested; no live probe can catch a regression | QA ledger D / Copilot ×4 | **Fixed** — two tests: a direct bean-method call (arg dropped / inverted), and a `@SpringBootTest` with the property omitted, which is the only one that evaluates the `@Value` default string itself | Broken to confirm **three times**: arg dropped ⇒ fails; boolean inverted ⇒ fails; default typo'd from `STRICT` to `DISABLED` ⇒ the direct-call test stays green (proving QA's point) while `InboundSecretFilterDefaultModeWiringTest` fails with `SecretNotAvailable` |
| 8 | ADR-0007 described inbound as unconditionally allow-all and #7730 as unstarted | Copilot Low | **Fixed** — Amendment 2 + supersede pointers at lines 108/156 | — |
| 9 | Amendment 2's own claim "the filter is a no-op on every current call site" | QA re-test | **Fixed** — Amendment 2 rewritten with an explicit correction | Falsified by #1. The earlier probe it rested on had not exercised a chained value; both it and this document were wrong |
| 10 | `ConnectorsAutoConfiguration` guard javadoc still argued inbound had "no injected name for a filter to reject", contradicting Amendment 2 in the same branch | QA re-test (R11) | **Fixed** | Also corrected a second error: the guard gates one property both directions read, so it was never outbound-only. Repo-wide grep for the stale phrasings now returns nothing outside the correction itself |
| 11 | `LAX` and `STRICT` behave identically inbound, so `LAX` deployments gain a restriction | Author-raised | **Not yet signed off** — this row records that it was raised, not that anyone has approved it | Safe because the allow-list is a superset of the names its own filtered text declares. Only `DISABLED` opts out. **Needs an explicit sign-off from a human reviewer before merge** — it is a behaviour change for existing `LAX` deployments |
| 12 | Brace spelling `{{secrets.X}}` uncovered by the filter's tests | Author-raised | **Covered**, low value | Honest limit: the bare pattern also matches inside braces, so its branch is redundantly covered. Trim symmetry is pinned at the right layer by `SecretUtilTests.shouldReportTheSameNameReplacementLooksUp` |
| 13 | Cross-form widening: allow-list extraction matched the new `camunda.secrets.<name>` form too, so a name declared only there authorized legacy-path resolution `replaceSecrets` never performs for that form | QA re-test | **Fixed** — extraction switched to `retrieveLegacySecretKeysInInput`, which already existed for exactly this reason | Broken to confirm: reverted to `retrieveSecretKeysInInput` ⇒ `leavesASecretDeclaredOnlyInTheNewForm` fails, `expected "secrets.API_KEY" but was "leaked-via-legacy-path"` — reproducing the reported leak |

### Not closed — needs a cluster

- **`io.camunda:http-polling` response-body path is unverified.** QA's injection matrix passed every row except this one; it had no evidence because the run budget expired before a pollable in-cluster endpoint existed. Nothing in this branch closes it, and it should not be marked done until it is actually exercised against a cluster.

### Residual risk, accepted

- **A future call site that filters runtime-derived text would allow-list its own names.** Scoping the allow-list to the text in hand is what closes #1–#4; it also means the filter cannot catch a *new* call site that starts replacing over a payload. That trade was taken deliberately: it closes a present-day escalation at the cost of a hypothetical future one.
- **Outbound was re-verified only under `LAX`** in the QA run. Unchanged by this branch, but worth a `STRICT` outbound pass before merge.


mathias-vandaele:
### Self-review follow-up

Reviewed the first commit against the two code paths that actually run secret replacement on inbound. Found and fixed one real defect, and replaced the tests, which were not exercising either path.

**Defect: the allow-list was stale after an element hot-swap.**

`connectorDetails` is not immutable — `updateConnectorDetails` swaps it in place when a new process version is hot-swapped onto a live executable (`InboundExecutableRegistryImpl#processHotSwaps`), while `AbstractConnectorContext.secretFilter` is `final` and `getSecretHandler()` memoizes the handler built from it. `ValidInboundConnectorDetails#checkCompatibility` only pins the **first** element's properties, and a declared deduplication scope (the webhook connector has one) lets grouped elements differ on out-of-scope properties. So a newly added element could declare a secret none of the original elements did, and the snapshotted allow-list denied it — silently, leaving the placeholder in the bound value rather than failing.

The filter now reads a holder refreshed on every details change. Regression test `secretFilterEnabled_resolvesASecretIntroducedByAHotSwappedElement` fails on the first commit (`expected "added-value" but was "secrets.ADDED"`) and passes now.

**Allow-list source made exact.** It is now derived from precisely the two texts replacement runs over — `rawPropertiesWithoutKeywords` (connector-level binding) and each element's own property map (element-scoped binding after correlation) — instead of `InboundConnectorElement#rawProperties`, which is a third, separate accessor that merely happens to hold the same map today.

**Tests.** The two original tests called `getSecretHandler().replaceSecrets(...)` directly, so they exercised neither real path, and their fixture built the element with an empty property map, leaving the element-scoped path uncovered. Replaced with four: resolution through `getProperties()`, a blocked injected name, resolution of a secret declared **only** by the activated element of a multi-element executable (this one fails if the allow-list is taken from the connector-level map alone), and the hot swap above.

**Checked and found not to be issues:**
- Property **keys** are not a gap. Replacement runs over the serialized JSON of the whole map, keys included, but `readWrappedProperties` splits keys on `.` into nested objects, so `secrets.X` in a key never appears contiguously in that JSON and cannot match either legacy pattern.
- Name **trimming** is consistent: `retrieveSecretKeysInInput` and `SecretUtil`'s replacer both trim, so `{{ secrets.FOO }}` allow-lists and looks up the same `FOO`.
- The activated element is always one of `connectorElements()` (`InboundCorrelationHandler#findMatchingElement` selects from the list it was given), so the element-scoped path cannot present a text outside the union.
- No secret-resolution path bypasses this filter: the only two `getSecretHandler()` callers on inbound are both in this class, and `InboundIntermediateConnectorContextImpl` delegates.
- Concurrency: the holder is an `AtomicReference` holding an immutable set, so a hot swap concurrent with a webhook thread resolving secrets swaps atomically.

**One deliberate decision worth a reviewer's confirmation:** `LAX` and `STRICT` behave identically inbound (only `DISABLED` turns the filter off). Outbound they differ only because the allow-list needs a remote BPMN lookup that can fail; inbound's is computed from data already in memory. The consequence is that a deployment on `LAX` gains the inbound restriction. That is safe only because the allow-list is a superset of the names its own filtered text declares — which is what the reworked tests now pin.

**Known pre-existing behaviour, unchanged:** `propertiesWithSecrets` is memoized and not invalidated on `updateConnectorDetails`. Harmless today, since compatibility requires `rawPropertiesWithoutKeywords` to be equal, so the resolved connector-level map cannot change. Not touched here.

Full suites re-run green: `connector-runtime-core`, `connector-runtime-spring` (506), `spring-boot-starter-camunda-connectors` (189), `connector-http-polling`.

connector-runtime/connector-runtime-core/src/test/java/io/camunda/connector/runtime/core/inbound/InboundConnectorContextImplTest.java
```diff
@@ -299,6 +299,93 @@ void bindProperties_secretContextCarriesTheElementsPhysicalTenantId() {
...
+            getInboundConnectorDefinition(Map.of("token", "secrets.DECLARED")), secretProvider);
+
+    assertThat(replace(context, "secrets.INJECTED")).isEqualTo("secrets.INJECTED");
+    verify(secretProvider, never()).getSecret(eq("INJECTED"), any());
+  }
+
+  @Test
+  void secretFilterEnabled_resolvesASecretIntroducedByAHotSwappedElement() {
+    var scope = List.of("shared");
+    var first = elementWithProperties("a", Map.of("shared", "x", "token", "secrets.ORIGINAL"));
```
johnBgood: should we test {{secrets... syntax as well?

johnBgood (approved):
Left 1 comment, and some copilot comments are valid as well, other than that all good, pre approving. 
We need to backport down to 8.6, good luck

mathias-vandaele:
> Left 1 comment, and some copilot comments are valid as well, other than that all good, pre approving. We need to backport down to 8.6, good luck

@johnBgood Will add tests, copilot comment is already resolved, `connectorDetails` is now volatile

Szik (approved):

PR #8544, opened 2026-09-01
fix: bump reactor-netty to 1.3.7

ztefanie:
## Summary

Bumps `io.projectreactor.netty:reactor-netty-http` / `reactor-netty-core` from `1.2.18` to `1.3.7`
via a new `version.reactor-netty` property and `dependencyManagement` overrides in
`parent/pom.xml`, mirroring the existing `reactor-core` override.

1.3.7 is the reactor-netty that reactor-bom pairs with `reactor-core 3.8.7`, which this branch is
already pinned to.

Verified with `mvn dependency:tree`: `reactor-netty-http` now resolves to `1.3.7`.

Security fix. Details in the internal tracking issue: camunda/team-connectors#1273

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -136,6 +136,12 @@ limitations under the License.</license.inlineheader>
          ships a reactor-bom with reactor-core >= 3.8.7
          Ref: https://spring.io/security/cve-2026-47857/ -->
     <version.reactor-core>3.8.7</version.reactor-core>
+    <!-- CVE-2026-47874: reactor-netty transitive override — remove once spring-boot-dependencies
+         ships a reactor-bom with reactor-netty >= 1.3.7. The 1.2.x line fix (1.2.19) is
+         Enterprise-only and not published to Maven Central, so 1.3.7 is the only reachable fix.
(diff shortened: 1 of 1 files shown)


ztefanie:
Good catch — added in the follow-up commit, following the existing convention in this file (one `requireProperty` guard per override, as with the amqp-client CVE-2026-63337 and reactor-core CVE-2026-47857 guards) rather than folding it into the reactor-core message.

Verified both directions with `mvn -N validate`:

- current `version.spring-boot` (`3.5.16-spring-boot-3.5.19`) → BUILD SUCCESS
- bumped to `3.5.16-spring-boot-3.5.18` → BUILD FAILURE, `Rule 5 ... failed: version.spring-boot was changed! Check if the reactor-netty CVE-2026-47874 override can be removed.`

The message names the exact removal steps (`version.reactor-netty` property, both `reactor-netty-http` / `reactor-netty-core` dependencyManagement entries, and the rule itself) and tells the reader to confirm with `dependency:tree` rather than trusting the property, matching the wording of the neighbouring guards.

johnBgood (approved):

PR #8552, opened 2026-09-01
ci: dispatch a fix agent for setup-only SaaS failures

hilalbursalii:
## Description

`saas-provisioning` — every failing spec in `test-setup.spec.ts` — was report-only, on the
assumption that a setup failure is always the environment. [Run 33520908181](https://github.com/camunda/connectors/actions/runs/33520908181)
on `stable/8.8` is the counter-example: three users failed `Create Project Folder`, triage
classified it correctly, and then had nothing to offer.

It's the wrong assumption. The surface name describes a *symptom*. Three users failing one
step is equally consistent with a dead org **or** with one stale locator in the shared
helper all three call — and only the screenshot separates them, which is the agent's job
rather than the classifier's.

### Flipping the flag alone would not have worked

Provisioning candidates had their specs cleared one line after classification, so they'd
have been suppressed again as `no-failing-specs-extracted`. Spec retention is now per
surface:

| Surface | Specs it carries |
|---|---|
| `saas-smoke-e2e` | non-setup only — setup failures in a mixed report are withheld and re-triaged as their own candidate |
| `saas-provisioning` | setup specs only |
| `saas-infra` | none — no report, or nothing failed |

### The manual carries the part that decides whether this helps

A new section on how to tell a moved control from an unreachable app (screenshot first;
the SaaS org is deleted after the run, so there's nothing to check live), and that a dead
environment is **escalated, not accommodated**. It names the three traps this surface
invites: weakening a setup assertion, waiting longer for an org that doesn't exist, and
reading "several users failed the same step" as proof of an environment fault.

### A second defect, found by replaying that run

A SaaS run publishes one report per Tasklist generation (`-v1`, `-v2`) and an SM run one
per shard, and the specs were concatenated — **six entries for three failures**, with
duplicate `fp=` lines in the coverage block. Now merged by file and test name, with
attempt histories concatenated so a spec that passed in either generation still reads as
flaky rather than depending on which report was parsed first.

### Verification

Replaying the real run, before and after:

```
before:  suppressed  connectors:stable/8.8:saas-provisioning  surface-not-in-increment
after:   dispatch    connectors:stable/8.8:saas-provisioning
         3 specs, 3 unique fingerprints, attempts=4 each, deterministic=true
```

97 unit tests pass (was 94 before this change's own tests), including the real run's shape
as a fixture case. `actionlint` clean.

The label-length matrix is narrowed to this copy's real key space — `SOURCE` derives from
`REPO` and the fix workflow rejects any other source, so the speculative entries could
never produce a label. Widest real key is 46 of 50 characters.

### Residual risk, stated plainly

This increases dispatch volume, and it points the agent at a surface where the honest
answer is often "not determined". The guardrail is the manual plus the existing never-mask
rules; the failure mode to watch for is a PR that relaxes a setup assertion, which should
be rejected on sight.

## Related issues

Follow-up to #8267. Prompted by the triage verdict on run 33520908181.

closes #

## Checklist

- [ ] Backport labels are added if these code changes should be backported. **Not needed** —
  the stable branches call `alwaysgreen-triage.yml@main`, so updating `main` is enough.

### DIFF
--- a/.github/scripts/alwaysgreen/classify.py
+++ b/.github/scripts/alwaysgreen/classify.py
@@ -110,9 +110,18 @@ SURFACE_CI_INFRA = "ci-infra"
 #: is reported and routed, never dispatched — see DISPATCHABLE_SURFACES.
 SURFACE_CONNECTORS_AI = "connectors-ai-e2e"
 
-#: Surfaces the first increment dispatches. Everything else is recorded and
-#: reported but not handed to the agent yet.
-DISPATCHABLE_SURFACES = frozenset({SURFACE_SM_E2E, SURFACE_SAAS_E2E})
+#: Surfaces handed to the agent. Everything else is recorded and reported only.
+#:
+#: `saas-provisioning` is included because the name describes a symptom, not a cause:
+#: every failing spec being in test-setup.spec.ts can mean the SaaS org or cluster is
+#: genuinely dead, or it can mean the setup spec itself has gone stale against a changed
+#: Modeler UI — and the second is as fixable as any other spec. Only the screenshot and
+#: the trace separate them, which is the agent's job, not the classifier's. The manual
+#: requires it to escalate rather than weaken a setup assertion when the environment is
+#: the one at fault.
+DISPATCHABLE_SURFACES = frozenset(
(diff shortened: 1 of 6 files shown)


johnBgood (approved):

johnBgood:
@hilalbursalii There are conflicts, feel free to merge when ready :)

PR #8617, opened 2026-09-03
feat(agentic-ai): report reasoning and cache token metrics to agent instance

### LINKED ISSUE #8605: AI Agent Connector: report reasoning and cache token metrics to Agent Instance API
## Context

Camunda AI Agent connectors support prompt caching and reasoning. Today, only input and output token counts are reported to the Agent Instance API. Reasoning-token, cache-creation-token, and cache-read-token counts are already computed internally by the connector on every model call, but are not forwarded.

This means a real part of token consumption is currently invisible in Agent Instance metrics, which are used for cost calculation and observability. The gap grows the more caching or reasoning is used, since those tokens are billed differently from plain input/output tokens.

## Goal

Extend the connector's Agent Instance API reporting so reasoning, cache-creation, and cache-read token counts are included alongside the existing input/output token counts.

## Scope

- In scope: reporting the three additional token counts (already tracked internally by the connector) to the Agent Instance API.

## Acceptance criteria

- [ ] Agent Instance metrics reported by the AI Agent connector include reasoning, cache-creation, and cache-read token counts alongside input/output tokens.
- [ ] Existing input/output token reporting is unaffected.

nikonovd:
## Description

Agent instance metrics reported by the AI Agent connector now include reasoning, cache-creation, and cache-read token counts, alongside the existing input and output token counts. Callers of the Agent Instance API can see the full per-call and aggregated token breakdown, including tokens spent on prompt caching and model reasoning, which the connector previously dropped when reporting metrics.

- The connector reports the three new token counts for every model call.
- The Agent Instance query API returns the same three counts in the aggregated metrics once the engine persists them.

## Related issues

closes #8605

## Checklist

### DIFF
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
@@ -319,4 +319,41 @@ class AgentSubProcessAgentInstanceTests extends BaseAgentSubProcessTest {
                     results.stream().anyMatch(r -> "fast-002".equals(r.id()))),
             any(AgentConversationTurn.class));
   }
+
+  /**
+   * Single model call whose usage carries non-zero cache-read and reasoning token counts, to verify
+   * the connector reports them to the Agent Instance API and the engine's aggregated metrics
+   * reflect them on read-back.
+   */
+  @Test
+  void shouldUpdateAgentInstanceMetricsWithCacheAndReasoningTokenCounts() throws Exception {
+    OpenAiCompletionsChatModelStubs.stubConversation(
+        new OpenAiCompletionsChatModelStubs.UsageDetailsTurnStub(
+            "The superflux calculation of 5 and 3 is complete.", 10, 20, 50, 7));
+
+    enqueueUserFeedback(userSatisfiedFeedback());
+
+    final var zeebeTest =
+        awaitProcessCompletion(
+            createProcessInstance(
+                Map.of("userPrompt", "Calculate the superflux product of 5 and 3")));
+
+    final var expectedMetrics =
+        new AgentMetrics(1, new AgentMetrics.TokenUsage(10, 20, 50, 0, 7), 0);
+    final var agentInstanceKey = new AtomicLong();
+    assertAgentResponse(
+        zeebeTest,
+        agentResponse -> {
+          AgentSubProcessResponseAssert.assertThat(agentResponse)
+              .isReady()
+              .hasAgentInstanceKey()
+              .hasMetrics(expectedMetrics);
+          agentInstanceKey.set(agentResponse.context().metadata().agentInstanceKey());
+        });
+
+    AgentInstanceEngineVerifier.verify(camundaClient, agentInstanceKey.get())
+        .hasStatus(AgentInstanceStatus.COMPLETED)
+        .hasMetrics(expectedMetrics)
+        .verify();
+  }
 }
--- a/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
+++ b/connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
@@ -338,8 +338,62 @@ class AgentSubProcessAgentInstanceTests extends BaseAgentSubProcessTest {
             createProcessInstance(
                 Map.of("userPrompt", "Calculate the superflux product of 5 and 3")));
 
-    final var expectedMetrics =
-        new AgentMetrics(1, new AgentMetrics.TokenUsage(10, 20, 50, 0, 7), 0);
+    final var expectedTokenUsage =
+        AgentMetrics.TokenUsage.builder()
+            .inputTokenCount(10)
+            .outputTokenCount(20)
+            .cacheReadTokenCount(50)
+            .cacheCreationTokenCount(0)
+            .reasoningTokenCount(7)
+            .build();
+    final var expectedMetrics = new AgentMetrics(1, expectedTokenUsage, 0);
+    final var agentInstanceKey = new AtomicLong();
+    assertAgentResponse(
+        zeebeTest,
+        agentResponse -> {
+          AgentSubProcessResponseAssert.assertThat(agentResponse)
+              .isReady()
+              .hasAgentInstanceKey()
+              .hasMetrics(expectedMetrics);
+          agentInstanceKey.set(agentResponse.context().metadata().agentInstanceKey());
+        });
+
+    AgentInstanceEngineVerifier.verify(camundaClient, agentInstanceKey.get())
+        .hasStatus(AgentInstanceStatus.COMPLETED)
+        .hasMetrics(expectedMetrics)
+        .verify();
+  }
+
+  /**
+   * Two model calls, each with distinct non-zero cache-read and reasoning token counts, chained via
+   * the feedback loop (follow-up, then satisfied) so no tool call is needed to drive a second model
+   * call. A single-turn test cannot distinguish the engine summing these counts across history
+   * items from it merely keeping the latest turn's values; this asserts the summed totals on
+   * read-back, so only the former passes.
+   */
+  @Test
+  void shouldSumCacheAndReasoningTokenCountsAcrossMultipleTurns() throws Exception {
+    OpenAiCompletionsChatModelStubs.stubConversation(
+        new OpenAiCompletionsChatModelStubs.UsageDetailsTurnStub(
+            "A haiku about the sea.", 10, 20, 30, 5),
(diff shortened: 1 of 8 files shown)


connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
```diff
@@ -319,4 +319,41 @@ void shouldReportFastToolResultToHistoryWhileSlowToolStillInFlight() throws Exce
...
+
+    enqueueUserFeedback(userSatisfiedFeedback());
+
+    final var zeebeTest =
+        awaitProcessCompletion(
+            createProcessInstance(
+                Map.of("userPrompt", "Calculate the superflux product of 5 and 3")));
+
+    final var expectedMetrics =
+        new AgentMetrics(1, new AgentMetrics.TokenUsage(10, 20, 50, 0, 7), 0);
```
HanselIdes: 🔧 Label the metrics so you can track which is input/output/cache read/etc
It's slightly more verbose but adds clarity

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/task/AgentTaskAgentInstanceTests.java
```diff
@@ -187,4 +187,42 @@ void shouldUpdateAgentInstanceStatusAndMetricsAcrossMultipleToolCallRounds() thr
...
+    enqueueUserFeedback(userSatisfiedFeedback());
+
+    final var zeebeTest =
+        awaitProcessCompletion(
+            createProcessInstance(
+                e -> e.property("provider.openai.model.model", "gpt-4o"),
+                Map.of("userPrompt", "Calculate the superflux product of 5 and 3")));
+
+    final var expectedMetrics =
+        new AgentMetrics(1, new AgentMetrics.TokenUsage(10, 20, 50, 0, 7), 0);
```
HanselIdes: 🤖 💭 `cacheCreationTokenCount` is asserted as `0` here, and both e2e variants have no way to do otherwise: OpenAI Completions carries no cache-creation field on the wire, so `UsageDetailsTurnStub` can only dial `cached_tokens` and `reasoning_tokens`. That field is produced only by the Anthropic, Bedrock and Gemini converters.

Net effect: of the three counts this PR forwards, cache-creation is the one with no coverage proving it survives converter -> connector -> engine -> query API. The converters have their own unit tests for it, and `AgentInstanceHistoryMapperTest` covers the mapping in isolation, but nothing joins the two ends.

Not asking you to change this test. Worth knowing the gap exists, and if you want to close it, the cheapest route looks like a `UsageDetailsTurnStub` equivalent on `StreamingAnthropicMessagesSseChatModelStubs` with one e2e case pointed at it.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/task/AgentTaskAgentInstanceTests.java
```diff
@@ -187,4 +187,42 @@ void shouldUpdateAgentInstanceStatusAndMetricsAcrossMultipleToolCallRounds() thr
...
+
+  /**
+   * Single model call whose usage carries non-zero cache-read and reasoning token counts, to verify
+   * the connector reports them to the Agent Instance API and the engine's aggregated metrics
+   * reflect them on read-back.
+   */
+  @Test
+  void shouldUpdateAgentInstanceMetricsWithCacheAndReasoningTokenCounts() throws Exception {
+    OpenAiCompletionsChatModelStubs.stubConversation(
+        new OpenAiCompletionsChatModelStubs.UsageDetailsTurnStub(
```
HanselIdes, line 199: 🤖 🔧 Single model call, so this test cannot distinguish "the engine sums the new counts across history items" from "the engine copies the last item's values". The read-back assertion passes either way.

`TokenUsage.add()` covers the summing at unit level in `AgentMetricsTest`, but the engine-side aggregation of `reasoningTokenCount` / `cacheReadTokenCount` / `cacheCreationTokenCount` has no coverage. The existing `shouldUpdateAgentInstanceStatusAndMetricsAcrossMultipleToolCallRounds` would be the place for it, except it uses the generic `Turn` API which has no dial for these fields, so it asserts zeros.

Suggested fix: add a varargs or list overload `stubConversation(UsageDetailsTurnStub... turns)` alongside the existing single-turn one (the state-machine wiring in the generic `stubConversation` already does this), then extend the multi-round test with two turns carrying different non-zero cache-read and reasoning values and assert the summed totals on read-back. That makes the aggregation the thing under test rather than a single pass-through.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
nikonovd: Fixed - both e2e test files now build the TokenUsage via the named builder instead of positional args.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/task/AgentTaskAgentInstanceTests.java
nikonovd: Closed. Added `AgentSubProcessAnthropicPromptCachingTests.enablePromptCachingSurfacesCacheTokenCountsInAgentInstanceMetrics`, wired through a new `StreamingAnthropicMessagesSseChatModelStubs.UsageDetailsTurnStub` that dials real `cache_creation_input_tokens`/`cache_read_input_tokens` on the wire (previously hardcoded null in two places). That closes the converter -> connector -> engine -> query API chain for cache-creation specifically.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/task/AgentTaskAgentInstanceTests.java
nikonovd: Closed, via a new test rather than extending this one. Added `OpenAiCompletionsChatModelStubs.stubConversation(UsageDetailsTurnStub...)` (varargs overload mirroring the existing `Turn...`/`TurnStub...` chaining), then `AgentSubProcessAgentInstanceTests.shouldSumCacheAndReasoningTokenCountsAcrossMultipleTurns`: two model calls with distinct non-zero cache-read/reasoning values, chained via the sub-process feedback loop (follow-up, then satisfied) so no tool call is needed. Asserts the summed totals on read-back, so a same-value-both-turns bug or a keep-latest bug would both be caught.

Put it in the sub-process flavor rather than extending `shouldUpdateAgentInstanceStatusAndMetricsAcrossMultipleToolCallRounds` here: that test's multi-round shape is tool-calls-only, and `UsageDetailsTurnStub` has no tool-call variant, so retrofitting it would mean extending the stub's shape too. The feedback loop gives a cleaner two-model-call path with plain text turns. The aggregation being proved is engine-side and flavor-independent, so one flavor is enough.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/subprocess/AgentSubProcessAgentInstanceTests.java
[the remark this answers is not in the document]
nikonovd: Closed. Added a new Anthropic-based test (`AgentSubProcessAnthropicPromptCachingTests.enablePromptCachingSurfacesCacheTokenCountsInAgentInstanceMetrics`) that dials real `cache_creation_input_tokens` on the wire and asserts the non-zero value survives to the Agent Instance query API read-back. OpenAI has no cache-creation signal, so this test's own cacheCreationTokenCount stays 0 by design; Anthropic is what closes the gap.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/task/AgentTaskAgentInstanceTests.java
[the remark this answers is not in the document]
nikonovd: Closed. Added a new Anthropic-based test (`AgentSubProcessAnthropicPromptCachingTests.enablePromptCachingSurfacesCacheTokenCountsInAgentInstanceMetrics`) that dials real `cache_creation_input_tokens` on the wire and asserts the non-zero value survives to the Agent Instance query API read-back. OpenAI has no cache-creation signal, so this test's own cacheCreationTokenCount stays 0 by design; Anthropic is what closes the gap.

HanselIdes (approved):
LGTM

gbetances089 (approved):

PR #8619, opened 2026-09-03
ci: share AlwaysGreen failure detection with the stable branches

hilalbursalii:
## Description

The merge-queue helm pipeline can conclude `success` while carrying a red job. Two
independent reasons:

* the SM Playwright matrix in camunda-platform-helm runs its `full` leg with
  `continue-on-error: ${{ !matrix.blocking }}`, so a `full`-only failure leaves
  `needs.trigger-sm-e2e.result == 'success'`;
* this workflow is itself `continue-on-error` on `merge_group`, which does the same one
  level up.

Either one alone hides a real failure from every `needs.*.result` the AlwaysGreen gates
read. main was fixed by scanning the jobs API instead — but that scan is a job inside
main's own workflow, and the stable branches still gate on `needs.*.result`. Run
[33727876677](https://github.com/camunda/connectors/actions/runs/33727876677) on
stable/8.9 is the consequence: a failing Playwright job, a run that concluded `success`,
and no Slack alert, no PR comment and no fix agent.

This PR extracts the scan into a reusable workflow so the stable branches can call it
the way they already call `alwaysgreen-triage.yml@main`, rather than each carrying a
copy that has to stay in step with the ref normalisation and supported-ref list those
workflows enforce.

**No behaviour change on main** — same evidence, same four outputs, same gates. The job
keeps its id, name, `needs` list and `permissions`; only its body moves.

Merge this before the three branch-local PRs that call it:

* stable/8.9 — #8620
* stable/8.8 — #8621
* stable/8.7 — #8622

Considered and rejected: a `workflow_run` listener on main, which would need no stable
edits at all (verified that one does fire for stable runs — the streak detector's
listener run 33730238377 reports `BASE_REF: stable/8.9`). It cannot gate the SM and
SaaS feedback jobs, which are plain jobs in each caller, and it would mean restructuring
main's working in-pipeline triage. Worth revisiting for stable/8.10, whose workflow
still carries `push: branches: [main]` and so never runs on its own pushes.

## Related issues

Detected on:

* https://github.com/camunda/connectors/actions/runs/33727876677 (stable/8.9, concluded `success` with a red job)
* https://github.com/camunda/connectors/actions/runs/33366607997 (main, the failure that prompted the original jobs-API scan)

## Checklist

### DIFF
--- a/.github/workflows/MERGE_QUEUE_HELM_TEST.yaml
+++ b/.github/workflows/MERGE_QUEUE_HELM_TEST.yaml
@@ -168,25 +168,9 @@ jobs:
       scenario: alwaysgreen
       shortname: agrn
 
-  # The failure signal for everything AlwaysGreen, read from the jobs API rather than
-  # from `needs.*.result`.
-  #
-  # `needs.*.result` cannot see two of the ways a job in this pipeline fails:
-  #
-  #   * A nested job of a called workflow that carries continue-on-error. The SM
-  #     Playwright matrix in camunda-platform-helm is
-  #     `continue-on-error: ${{ !matrix.blocking }}`, and its `full` leg is
-  #     `blocking: false` — so when only that leg fails, INTEGRATION_TEST.yml still
-  #     reports success and `needs.trigger-sm-e2e.result` is `success`.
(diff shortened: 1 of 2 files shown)


hilalbursalii:

PR #8620, opened 2026-09-03
ci: gate AlwaysGreen on this run's job conclusions (stable/8.9)

hilalbursalii:
## Description

The AlwaysGreen gates on this branch read `needs.*.result`, which cannot see two of the
ways a job here fails:

* the SM Playwright matrix in camunda-platform-helm runs its `full` leg with
  `continue-on-error: ${{ !matrix.blocking }}`, so a `full`-only failure leaves
  `needs.trigger-sm-e2e.result == 'success'`;
* this workflow is itself `continue-on-error` on `merge_group`, which does the same one
  level up.

Run [33727876677](https://github.com/camunda/connectors/actions/runs/33727876677) on this branch is the case: the Playwright `full` leg failed, the run concluded
`success`, and `AlwaysGreen failure feedback (SM E2E)` and `AlwaysGreen triage and fix
dispatch` were both `skipped` — no Slack alert, no PR comment, no fix agent for a real
e2e failure.

Gate on this run's job conclusions instead, read from the jobs API — the same evidence
`.github/scripts/alwaysgreen/discover.py` reads, so the gate and the classifier can no
longer disagree. The scan itself is the reusable workflow added on main in #8619, called
`@main` exactly like `alwaysgreen-triage.yml`, so future changes to detection stay
main-only.

Triage also stops being push-only, matching main: a merge-queue entry runs the same SM
and SaaS suites against the same clusters, so a red one is the same regression seen ~10
minutes earlier. Safe to let through — there is no `required_status_checks` rule on this
branch, so a triage failure cannot block a queue entry — and dedupe is unchanged,
because the queue ref normalises to its base branch and the follow-up push run derives
the same dispatch key.

Authored per branch rather than backported: main's copy of this file has diverged, so
the patch does not apply. Verified with `actionlint`.

**Merge #8619 first** — this calls a workflow that does not exist on main yet.

## Related issues

Detected on: https://github.com/camunda/connectors/actions/runs/33727876677

Companion PRs: #8619 (main, the shared workflow) and the sibling stable branches.

## Checklist


hilalbursalii:

PR #8623, opened 2026-09-03
ci: release an idle PR's hold on the spec files it touches

hilalbursalii:
## Description

AlwaysGreen will not dispatch a fix for a failing spec whose file an open PR in the e2e
repository already touches, so two agents cannot rewrite one spec at the same time. That
claim is deliberately author-agnostic — another agent's branch, a human fix and a bot PR
all count — but nothing ever released it. It held for as long as the PR stayed open,
covering every file the PR touched and every failure in those files, whatever the PR was
actually for.

Run [33727363856](https://github.com/camunda/connectors/actions/runs/33727363856) is the
consequence. Triage found a real SaaS failure, and its plan reports:

```json
{"surface": "saas-smoke-e2e", "reason": "spec-path-claimed-by-open-pr",
 "detail": "tests/8.10/smoke-tests.spec.ts (#3153)"}
```

`c8-cross-component-e2e-tests#3153` is a draft opened 2026-08-26 with one commit and no
activity since. It touches ten spec files, so it had been silently holding all ten shut
for eight days.

Expire the claim after a day of inactivity, measured from the PR's last activity rather
than from when it was opened: a PR being written, reviewed or rebased touches its own
timestamp well inside a day, so an active one keeps its files, while one nobody has
touched since yesterday is not work a second agent could collide with. The coarse
dispatch-key lock already expires this way (#8604); this is the same release applied to
the finer per-file claim, and the key lock plus the in-flight check still bound duplicate
work.

The TTL is checked before the file-list truncation guard, so a stale PR large enough to
truncate `gh`'s per-PR file list no longer fails the lookup for every surface.

Overridable via `ALWAYSGREEN_PATH_CLAIM_TTL_HOURS`; `0` restores the old never-expiring
claim.

Not in scope: the SM leg of that same run was suppressed as
`tracked-by-open-product-bug` (camunda/camunda#61565), which is correct and untouched.

## Related issues

Detected on: https://github.com/camunda/connectors/actions/runs/33727363856

Follows #8604, which gave the dispatch-key lock its TTL.

## Checklist

### DIFF
--- a/.github/scripts/alwaysgreen/discover.py
+++ b/.github/scripts/alwaysgreen/discover.py
@@ -48,23 +48,36 @@ KEY_LABEL_PREFIX = planning.KEY_LABEL_PREFIX
 #: two lookups below fail closed, a repository the token cannot reach would wedge every
 #: dispatch shut rather than merely returning nothing.
 FIX_PR_REPOS = [REPO, E2E_REPO]
+
+
+def _ttl_hours(env_var: str, default: int) -> int:
+    """An overridable TTL dial, read defensively.
+
+    Matches the degrade-don't-crash bias of everything else here: an unreadable
+    timestamp keeps the lock and a failed lookup suppresses, so a malformed dial must
+    not raise at import and take down every entry point in this module. Set to 0 to
+    restore the old never-expiring lock.
+    """
+    try:
+        return int(os.environ.get(env_var, "").strip() or default)
+    except ValueError:
+        print(
+            f"::warning::unparseable {env_var}; using the default ({default}h).",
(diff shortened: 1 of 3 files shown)


hilalbursalii:

hilalbursalii:
Superseded: this commit is now part of #8624, which needs it — without the TTL the watcher hands the failure to triage and the dispatch is suppressed one layer later by a 14-day-idle draft. Verified in https://github.com/camunda/connectors/actions/runs/33740413464: `dispatches=1 suppressed=0`, 37 idle claims released.

PR #8641, opened 2026-09-03
fix: read every secret reference in a single scan

### LINKED ISSUE #8602: SecretUtil chain-resolution edge cases: bypass, cross-pass double-lookup, cubic DoS (reverted from backport PRs, needs own scoping)
## Context

While backporting outbound secret filtering (#7568) across main/stable-8.9/stable-8.8/stable-8.7/end-of-life-8.6, Copilot review rounds surfaced a family of real bugs in `SecretUtil.java`'s two-pass (parentheses + bare) chain-resolution design, going well beyond the backport's own scope. Fixes were implemented, verified, and briefly landed on all 5 branches, then **reverted** to keep those PRs scoped to the backport itself. This issue tracks the reverted work for separate follow-up.

The revert point on every branch is the commit `fix: don't treat a chain-introduced bracketed reference as denied by the parentheses pass` — everything after that, up to and including the commits below, was reverted from `SecretUtil.java` + `SecretUtilTests.java` only (no other files touched).

## What was found and fixed (then reverted)

1. **Generated-bracket bypass via full-name denial.** A chain-generated bracket (e.g. `A` resolves to `{{secrets.FOO:BAR}}`) whose full name (`FOO:BAR`) is denied, but whose bare prefix (`FOO`) is separately allowed, let the bare pass resolve the truncated prefix and leak `{{REAL_VALUE:BAR}}` — recreating the exact bypass class the original two-pass design was meant to prevent, just for a name that wasn't in the original input for the parentheses pass to deny. This also had a multi-hop variant (`A -> B:C -> D:E -> FULL_VALUE`) needing a bounded convergence loop rather than a single pass.

2. **Cross-pass double provider lookup.** The bare pass asking the replacer directly whether a bracket's full name resolves (the fix for #1) could re-fetch a name the parentheses pass had already resolved or denied. Not just a performance nit: `secretReplacer` is not guaranteed side-effect-free in production (a metered provider aggregator counts each resolution).

3. **Cubic-complexity DoS via denied special-character references.** The bare pass's per-match nesting-exclusion check rescanned the full denied-bracket list on every match, and reran once per original bare match — cubic in the number of denied `{{secrets.X:Y}}`-shaped references. Traced the actual call path: `SecretUtil.replaceSecrets` runs over `job.getVariables()` (the full raw job-variable JSON via `JobHandlerContext.getJsonReplacedWithSecrets`), **not just deploy-time BPMN text** — so this is reachable by anyone who can start a process instance or complete a task and influence a variable's runtime value, no deploy rights required.

4. **Two smaller edge cases** on top of the above: (a) a self-reference or multi-name cycle can leave a truncating bracket literal in the text while its *name* has a cached non-null resolution (because it resolved to a *different* bracket that itself never got applied) — asking "is this name denied" wrongly said no; (b) a chain-generated bracket whose name starts entirely outside the bare pattern's character class (e.g. `:B`) produces no bare match at all, so nothing gated a loop iteration for it and it was left unresolved forever.

## Where the reverted code lives

The fixes (and their regression tests) are fully implemented and were verified green on all 5 branches before being reverted. Rather than re-deriving them, whoever picks this up can look at the revert commit's parent on each branch:

- main: `761a6005eb` (parent of `be2831e1c8`)
- stable/8.9: `351a899509` (parent of `966259c8e7`)
- stable/8.8: `6857b104af` (parent of `d6feedbdef`)
- stable/8.7: `95b710038b` (parent of `8942d1b32a`)
- end-of-life/8.6: `59e8e0a839` (parent of `07495c004d`)

`git show <sha> -- '*/SecretUtil.java' '*/SecretUtilTests.java'` on any of those gives the full incremental history of fixes + tests from the boundary commit forward.

## Suggested approach for a future pass

Given how many rounds of incremental patching this took (each fix's own edge case became the next finding), a **from-scratch redesign** of the chain-resolution logic is probably a better use of time than re-cherry-picking the incremental fixes above -- the two-pass parentheses/bare split with ad-hoc truncating-bracket handling has proven hard to reason about completely. Whoever picks this up should treat the list above as the acceptance criteria / regression-test source material, not as a mandate to reapply the same patches.

mathias-vandaele:
Closes #8602.

## What was wrong

`SecretUtil` read the same text several times with patterns that disagree about what a secret name is — two replacement passes (bracketed, then bare) plus one scan per pattern in the key extractor. The bare form's character class stops at `:`, so every extra read could recover a **truncated prefix of a name the text never declared**, and each consumer drew a different conclusion from it.

Before writing any code I reproduced each claim in #8602 against `main` (`67d3995749`) with a standalone harness built from that file verbatim. Results, and where they stand now:

| | before | after |
|---|---|---|
| `retrieveSecretKeysInInput("{{secrets.PROD:API}}")` | `[PROD:API, PROD]` — admits the undeclared `PROD` onto the STRICT allow-list | `[PROD:API]` |
| `retrieveSecretKeysInInput("{{secrets.camunda.secrets.FOO}}")` | `[camunda.secrets.FOO, FOO]` | `[camunda.secrets.FOO]` |
| `{{secrets.PROD:API}}`, `PROD:API` denied / `PROD` allowed | **`{{p4ssw0rd:API}}`** — the bare pass resolves a prefix of the name the bracketed pass just denied | `{{secrets.PROD:API}}` |
| the same through a chained value (`A` → `{{secrets.PROD:API}}`) | **`{{p4ssw0rd:API}}`** | `{{secrets.PROD:API}}` |
| 3200 denied references (79 KB) | 20.5M replacer calls, **4016 ms** of one thread | 3200 calls, **2 ms** |
| 3 occurrences of one name | 3 replacer calls | 1 |
| a name that trims to blank | writes the literal text `null` into the payload | asked normally; the filter decides |

The quadratic case is the one that matters operationally: each pass re-ran a whole-string replacement once per match in *its* original input, and `SecretUtil.replaceSecrets` runs over `job.getVariables()` (`JobHandlerContext:95`) — the raw runtime variable JSON, not just deploy-time BPMN text. No deploy rights needed; influencing one variable value is enough. It is quadratic, not the cubic reported in the issue — the cubic factor came from the reverted fix's own per-match rescan.

Items 2, 4a and 4b of #8602 are **not** reachable on `main`: they were defects of the reverted implementation, not of the code it was reverted to. Item 2's underlying concern is real in a weaker form (no memoization at all), and is addressed here.

## The change

One pattern holding all three reference forms, and one left-to-right scan:

```
{{\s*secrets\.(?<braced>\S+?)\s*}}  |  camunda\.secrets\.(?<reference>[\p{Alnum}_-]+)  |  secrets\.(?<bare>([a-zA-Z0-9]+[\/._-])*[a-zA-Z0-9]+)
```

A match consumes its whole span and the scan resumes past it, so a narrower alternative can never re-read the inside of a wider match — the entire bug family dies structurally, with no nesting-exclusion check, no cursor, and no depth cap. `replaceSecrets` and `keysIn` consume the same scan, so an allow-list can no longer admit or deny a name the resolver reads differently; that disagreement *was* the bug class.

Gone: `replaceSecretsWithParentheses`, `replaceSecretsWithoutParentheses`, `resolveSecretValue`, both rescan loops, and the `(?<!camunda\.)` lookbehind (redundant now that the `camunda.secrets.` alternative consumes its own span). `replaceSecrets` scans through the existing public `replaceTokens` rather than reimplementing its loop, so this file's public surface is unchanged from main. Production code goes from 159 to 131 lines; cyclomatic complexity is 5 for `replaceSecrets` and 2 for everything else. A per-call memo (`containsKey`, so a denied name's `null` is not re-asked) means the replacer sees each name at most once — the answer to the metered-provider concern in item 2.

Per the issue's own advice this is a redesign, not a re-cherry-pick of the reverted patches; that list served as the acceptance criteria.

## Behaviour changes, deliberate

- **A resolved value is no longer rescanned**, so a secret whose value spells another reference is no longer treated as a template. Undocumented, untested, and it only ever worked in one direction (`{{secrets.A}}` → `"secrets.B"` chained; `"secrets.A"` → `"{{secrets.B}}"` did not), at a depth set by the match count of the original text. It is also what made a chain-generated bracket — and the quadratic rescan — possible at all.
- **The bracketed capture now ends before the trailing whitespace**, so a name is clean at the source and needs no `trim()` on either side. That closes the trim/duplicate-match coupling recorded in #8602's comment structurally rather than by convention: `{{ secrets.FOO }}` resolves `FOO` *and* reports `FOO`, without depending on a redundant bare match to supply the clean spelling.

## Testing

- `LegacySecretCharacterizationTest`'s **30 pins pass unchanged** — the evidence the legacy path did not shift.
- 9 tests added to `SecretUtilTests` for the issue's items plus the edge cases. The DoS regression asserts a lookup *count* (5000 references → 5000 lookups), not a duration, so it cannot flake. Two of them were themselves corrected under review: the memo test now uses a *denied* name repeated three times (an allowed name memoizes under `computeIfAbsent` too, so it pinned nothing), and the blank-name test now uses a NUL name — `{{secrets.}}` matches neither old pattern, so it passed on `main` and pinned no regression, where a NUL name is what made the old code write the literal text `null` into the payload.
- Full suites green: **592** `connector-runtime-core`, **513** `connector-runtime-spring`, **190** `spring-boot-starter-camunda-connectors`.

## Note for review

Two comments survived, both because losing them reintroduces a defect: three lines above the pattern (splitting it back into per-form or per-caller scans restores the bypass) and one above `resolve` (rewriting `containsKey` as `computeIfAbsent` silently restores per-occurrence lookups for denied names). Everything else is code.

An earlier revision of this PR deleted the public `SecretUtil.replaceTokens` as unused. Copilot flagged that three times (once inline, twice as suppressed comments) and QA carried it as an unverifiable compatibility risk — correctly: no cluster test can detect a `NoSuchMethodError` in an embedder, and the build has no japicmp/revapi check to catch it. It is restored, and `replaceSecrets` now scans through it, so it is used rather than dead-but-public. Its definition is byte-identical to main's.

Affected release branches, to be handled one at a time after this lands: `stable/8.10` carries a byte-identical file (it took the feature through the alpha5 merge from main, so #8602 does not list it) and this commit cherry-picks onto it cleanly. `stable/8.9` and `stable/8.8` share a third state that already has the `isNotNestedInAny` fix — their DoS is cubic rather than quadratic — and `stable/8.7` and `stable/8.6` each have their own shape (no JSON escaping; on 8.6 no `SecretContext` either). #8602 names that last branch `end-of-life/8.6`, which no longer exists on the remote — the 8.6 line is `stable/8.6`. So each backport is an adaptation of this design, not a cherry-pick.


johnBgood (approved):

Szik (approved):

mathias-vandaele:
Manual backports opened, since the automatic ones could not cherry-pick: #8657 (stable/8.9), #8658 (stable/8.8), #8659 (stable/8.7), #8660 (stable/8.6).

Issue #8638, opened 2026-09-03
Outbound secret redaction re-reads the store instead of using the substituted values; inbound has no redaction


chillleader:
## Summary

Outbound secret redaction masks against a **fresh read** of the secret store performed *after* the connector has run, not against the values that were actually substituted into the job's input. If a secret rotates during the connector call, the re-read returns the new value, the message contains the old one, and the old one is published.

Separately, and more broadly: **inbound connectors have no redaction at all.**

Both point at the same fix — capture the values at substitution time — which is why they are filed together.

## The rotation race

1. `JobHandlerContext.getJsonReplacedWithSecrets()` resolves `{{secrets.FOO}}` to `old-value` via `SecretHandler`'s replacer (`SecretHandler.java:35-43`).
2. The connector calls an API with `old-value`. The API rejects it and echoes it back.
3. `FOO` rotates in the store to `new-value`.
4. `OutboundConnectorExceptionHandler.fetchSecretsForMasking` does its own `fetchAll(["FOO"])` and gets `new-value`.
5. `hideSecretsFromMessage` replaces `new-value`, which is not in the message. `old-value` is published in the incident.

Found by Copilot during review of #8634 ([thread](https://github.com/camunda/connectors/pull/8634#discussion_r3924680431)).

### Exposure by provider

| provider | exposed? |
| --- | --- |
| `EnvironmentSecretProvider`, `ConsoleSecretProvider` | No — process env / static config cannot rotate mid-process |
| `AwsSecretProvider`, `GcpSecretProvider` | Only when the connector call outlives the cache; `AbstractSecretProvider` uses `refreshAfterWrite`, default 5s (`CAMUNDA_CONNECTOR_SECRETS_CACHE_MILLIS`) |
| `CentralStoreSecretProvider` (legacy `FALLBACK`) | **Yes, no mitigation** — `SecretReferenceResolver` reads live |

The cache is an accidental mitigation, not a designed one.

## Inbound has no redaction

`InboundConnectorContextImpl`'s use of `retrieveLegacySecretKeysInInput` is allow-listing (`SecretFilter.allowOnly`, line 189-190), not masking. The only `replaceSecrets` call on the inbound path (`InboundPropertyHandler.java:118`) is resolution. Meanwhile inbound surfaces errors through `ActivityLogRegistry`/`ActivityLogEntry` and `reportHealth`, both operator-visible. So an inbound connector whose error text carries a resolved secret publishes it unredacted.

[...]

PR #8642, opened 2026-09-03
refactor: share the secret-masking string primitive via SecretUtil

chillleader:
## Description

Pure relocation: `hideSecretsFromMessage(String, List<String>)` moved from `OutboundConnectorExceptionHandler` (connector-runtime-spring) to `SecretUtil` (connector-runtime-core, which spring already depends on). No behavior change — outbound now delegates to the shared method instead of keeping its own copy.

This is prep for #8638: inbound redaction (next PR in the stack) needs the same longest-first, empty-skip masking logic. Sharing it means one implementation to keep correct instead of two copies that can drift apart over time.

### Verification

All existing outbound tests pass unchanged, confirming this is behavior-preserving:
- `OutboundConnectorExceptionHandlerTest`: 31/31 passing
- `SpringConnectorJobHandlerTest`: 140/140 passing

## Related issues

Stacked on #8634. Part of #8638.

## Checklist


chillleader:
Consolidated into #8643 — this refactor is now the first commit of that PR, which combines it with the inbound redaction fix and the outbound rotation-race fix.

PR #8718, opened 2026-09-08
[Backport stable/8.8] fix(http): align SSL error code and root-cause handling with main (#8515)

ztefanie:
## Summary

Follow-up backport of #8515


ztefanie:
Fair catch — added a direct unit test (`shouldFallBackToClassName_whenRootCauseMessageIsNullOrBlank`) exercising `rootMessage()` via reflection with a null-message and a blank-message cause, asserting the fallback to the exception's simple class name. Pushed in 98c9405. (Note: the same fallback ships on main via #8515 without this coverage either — worth a follow-up there too.)

johnBgood (approved):

Issue #8752, opened 2026-09-09
Docs link version bumper strands links that miss a release cycle, and no CI check rejects unversioned links


crobbins215:
Follow-up from #8736, where seven connectors were found linking to documentation pinned to 8.6 and 8.7. Investigating why the existing safeguards did not prevent that turned up one bug and one missing check.

## 1. The release bumper strands any link that misses a cycle

`.github/workflows/scripts/version_bump_all_element_templates_docs_links_versions.sh`, called from the `version-bump-docs-links` job in `RELEASE.yaml`, rewrites only the exact previous minor:

```
s|(https://docs.camunda.io/docs/)$FROM_VERSION/|\1$TO_VERSION/|g
```

`RELEASE.yaml` computes `previous_minor = current_minor - 1`, so the 8.9.0 release rewrote only `8.8` to `8.9`. A link already at 8.6 or 8.7 matches nothing and is never rewritten again — it is stranded permanently, and each subsequent release widens the gap.

This accounts for every stale link found in #8736: SQL, SageMaker and Textract at 8.6; Comprehend, S3, Box and Gemini at 8.7. By then the 8.6 pages had moved to the `unsupported.docs.camunda.io` archive domain, so users opening the documentation link from the properties panel landed on content marked unsupported.

Suggested fix: match any older minor rather than only N-1, and run a one-off catch-up pass for links already stranded.

## 2. No CI check rejects unversioned links

The only link check in CI is the forbidden-`/next/` grep at `TEST_FEATURE_BRANCH.yml:181-189`. Nothing fails a build for adding an unversioned link.

`check_versioned_element_templates.sh` does not cover this — it verifies that newly added `versioned/` snapshots match their non-versioned counterpart, which is a different concern.

Because unversioned links are also invisible to the bumper in item 1, they never get corrected once introduced.

## 3. Existing unversioned links to convert

Counting only the real sources — hand-written templates and the Java `@ElementTemplate` / `@TemplateProperty` annotations that generate the rest, so generated JSON is not double-counted:

| Source | Files | Versionless links |
| --- | --- | --- |
| Hand-written element templates | 32 | 164 |
| Java sources | 69 | 95 |
| **Total** | **101** | **259** |

Largest single files: `github-connector.json` (28), `twilio-connector.json` (11), `google-maps-platform-connector.json` (9), `easy-post-connector.json` (9).

Worth landing the check in item 2 first, otherwise a bulk conversion drifts again straight away.

## Suggested order

[...]

PR #8757, opened 2026-09-09
fix(databricks): point the connector at its Camunda documentation

crobbins215:
Follow-up to #8736, which repaired stale and broken documentation links across the connectors but left Databricks out of scope.

## Connector documentation link

The template's `documentationRef` pointed at the Databricks REST API reference:

```
https://docs.databricks.com/api/workspace/introduction
```

So the documentation link on the element in the Modeler opened Databricks' API docs rather than guidance on using the Camunda connector. That URL has also rotted — it now drops to the generic `/api/workspace/` index, with no introduction page.

A Camunda page for this connector exists, so it is linked instead:

```
https://docs.camunda.io/docs/8.10/components/connectors/out-of-the-box-connectors/databricks/
```

Pinned to 8.10 because the page has so far only been published to the unreleased documentation — the same situation as the AgentCore Runtime connector in #8736. The numbered form resolves today, keeps the literal `/next/` out of the template as the CI check requires, and becomes stable when 8.10 ships. The unversioned and 8.9 forms both 404.

## Relocated Databricks link

The model serving query tooltip pointed at `/api/workspace/servingendpoints/query`, which Databricks has moved to `/api/model-serving-query/v1/query`. Updated to the new location.

The OAuth M2M link is left as-is — it still resolves, and a vendor API reference is the right target for that tooltip.

## Verification

- All four remaining links checked with curl: 200, no redirect away from the intended page.
- `element-template-validator`: `Scanned 716 template(s). No findings.`
- Forbidden-`/next/` check passes.
- Template parses as valid JSON.
- This template is hand-maintained — `connectors/databricks/` declares no generator plugin — so there is no Java source to update and nothing to regenerate.
- No template `version` bump: the change does not affect how the template applies, and `CurrentVersionBumpRule` passes.

## Not included

The two generic `use-connectors` links in this template are unversioned, as they are in most templates. That is the repo-wide cleanup tracked in #8752, deliberately not started here.

### DIFF
--- a/connectors/databricks/element-templates/databricks-connector.json
+++ b/connectors/databricks/element-templates/databricks-connector.json
@@ -18,7 +18,7 @@
     "oauth m2m",
     "service principal"
   ],
-  "documentationRef" : "https://docs.databricks.com/api/workspace/introduction",
+  "documentationRef" : "https://docs.camunda.io/docs/8.10/components/connectors/out-of-the-box-connectors/databricks/",
   "category" : {
     "id" : "connectors",
     "name" : "Connectors"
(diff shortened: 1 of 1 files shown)


crobbins215:
@ztefanie one I missed, the docs werent live when I created it first time around

ztefanie (approved):

Issue #8750, opened 2026-09-09
test(agentic-ai): add a portable host and container real-provider test launcher


gbetances089:
__What should we do?__

Add one repository-owned launcher for the AI Agent real-provider CPT suites that can run the Maven
test process either:

- directly on the developer or CI host
- inside a reproducible test-runner container

For native and document suites, the launcher should accept an execution-registry row or a direct
JUnit capability-tag/provider-group selection. Bundle-backed runs should replay their static registry
row without pretending that `AiAgentE2ETestIT` is capability-tag partitioned. Both execution modes
must invoke the same `it-real-llm` Maven contract. GitHub Actions should call the launcher in host mode
after importing the selected credential profiles instead of reconstructing Maven selection arguments
in workflow YAML.

Keep GitHub and Vault concerns outside the launcher. Define one canonical environment-variable
contract, normalize aliases such as
`GOOGLE_VERTEX_AI_SERVICE_ACCOUNT`/`GOOGLE_VERTEX_AI_SERVICE_ACCOUNT_JSON`, and keep genuinely
different auth types separate: a Bedrock provider API key is not interchangeable with the AWS
access/secret pair used by the document judge.

Provide a non-billable validation mode that resolves the requested registry row, validates required
nonblank variable names, performs narrow structural validation for known structured values such as
Vertex service-account JSON, and reports the selected suites/provider rows without printing values or
calling a provider.

The test-runner container should:

- use Java 21 and Node 24
- execute the repository Maven wrapper
- install the `element-templates-cli` version pinned in `.github/workflows/package.json` for every
  suite, because BPMN preparation shells out to it
- accept Maven settings and repository-cache mounts at runtime when internal repository access is
  required
- receive provider, judge, Maven, and registry credentials only at runtime, never in image layers or
  command-line arguments
- document Docker socket and Testcontainers host configuration for Linux and Docker Desktop
- reject host-only `localhost` defaults for Ollama or Docker Model Runner in container mode unless an
  explicit reachable host endpoint is configured

For bundle-backed tests, support an explicitly configured
`CONNECTORS_IMAGE_NAME`/`CONNECTORS_IMAGE_VERSION`. Building the current default bundle locally
requires access to its private `reg.mini.dev` base image. Document that prerequisite and the option to
use an already-built image accessible to the caller, for example through authenticated registry
access. Do not claim that a supported public-base build variant exists: adding such a variant is
outside this ticket unless it is designed and reviewed separately.

__Why should we do it?__

[...]

Issue #8749, opened 2026-09-09
docs(agentic-ai): document the E2E test structure and local/container execution


gbetances089:
__What should we do?__

Create `connectors-e2e-test/connectors-e2e-test-agentic-ai/README.md` as the developer-facing
reference for the AI Agent test module. No README currently exists in that module.

Document the implemented test structure and execution contract after #8722, #8750, and #8745,
including:

- the distinction between deterministic mocked/process tests and opt-in real-provider tests
- the native in-process capability suites, the dedicated tool-result-document suite, and the
  bundle-backed `AiAgentE2ETestIT`
- which document scenarios are active, disabled, manual-only, or provider-excluded, with links to
  their tracking issues rather than implying all declared scenarios execute
- which tests run under Surefire versus the `it-real-llm` Failsafe profile
- how JUnit capability tags, provider groups, the Java provider catalogs, the CI execution registry,
  and credential profiles divide responsibility
- how to run a direct capability/provider selection or replay an execution-registry row
- how to run Maven on the host and how to run Maven inside the test-runner container
- prerequisites: Java 21, Docker/Testcontainers, Node 24, the repository-pinned
  `element-templates-cli`, Maven settings where internal repository access is required, and a
  configured bundle image for bundle-backed tests
- the canonical environment variables for provider credentials, judge credentials, provider
  selection, bundle image selection, and local-provider endpoints
- how missing credentials, intentional exclusions, and zero-row selections behave
- the distinction between deterministic validation, paid real-provider execution, and manual-only
  scenarios such as opportunistic prompt-cache hits
- common failures: missing template CLI, missing/nonblank credentials, Docker socket access,
  Docker Desktop versus Linux Testcontainers host configuration, container `localhost` behavior,
  private base-image access, WireMock/document setup, and unexpected zero-test runs

Use placeholder environment values only. Do not place secrets in command-line arguments, Docker build
arguments, image layers, committed env files, or example output.

Do not copy a static provider/model/capability matrix into the README. The Java catalogs remain the
source of truth and the CI registry remains the execution-plan source of truth; document how a
developer inspects those sources so the README does not drift each time a model row changes.

Update the E2E section in `connectors/agentic-ai/AGENTS.md` to link to the new README and remove
duplicated or stale operational commands, including the currently unpinned
`npm i -g element-templates-cli` example. Keep the test-selection guidance for coding agents in
`AGENTS.md`; keep human setup, execution, credential, and troubleshooting instructions in the README.

[...]

Issue #8745, opened 2026-09-09
test(agentic-ai): migrate document tool-result CPT into the capability suite structure


gbetances089:
## What should we do?

Make `DocumentToolCallResultsIT` part of the capability-oriented real-provider CPT structure.

The class is already focused on one capability: documents returned by tools. Its single, multiple,
and nested-document methods are scenarios of that capability, so they should remain together rather
than becoming three suites.

The migration should:

1. Rename the class to the repository's `*E2ETestIT` convention so the `it-real-llm` profile can
   discover it without a class-specific Failsafe include.
2. Reuse the provider metadata, builders, environment selection, model construction, and lifecycle
   support extracted by #8722 where the execution models overlap.
3. Keep suite-specific setup such as the Bedrock judge, PDF fixtures, and document-tool BPMN local
   to the document suite.
4. Preserve the suite's v1, v2, and local-provider definitions as intentional extensions of the
   shared native-provider catalog.
5. Consolidate the overlapping `documentInToolResultIsReadByModel` scenario currently in
   `RealProviderApiSmokeIT` into the dedicated document-tool-result suite.

## Why?

Leaving this class outside the migration would preserve a second provider catalog and a second set
of environment/model helpers beside the new capability suites.

The current overlap also creates two tests for essentially the same nested PDF flow:

- `RealProviderApiSmokeIT.documentInToolResultIsReadByModel`
- `DocumentToolCallResultsIT.nestedStructureDocumentsFromToolCallResult`

They do not currently have identical provider coverage. The smoke class includes native v2 rows such
as OpenAI Foundry, Gemini, and Vertex, while the dedicated document suite includes v1 and local rows.
The duplicate must therefore not be deleted until the dedicated suite has inherited or intentionally
excluded every relevant provider/backend row.

## Acceptance criteria

- [ ] The document-tool-result suite follows the `*E2ETestIT` naming convention and no longer needs
      a class-specific Failsafe include.
- [ ] Shared provider/build/lifecycle support is reused where applicable; the suite does not retain
      a second copy of builders already owned by the capability-test support.
- [ ] Bedrock judge configuration, PDF fixtures, and document-specific BPMN/setup remain owned by
      the document suite.
- [ ] Document test setup clears `InMemoryDocumentStore.INSTANCE` before and after every test; this
      cleanup is not lost when shared support or class structure changes.
- [ ] Single-, multiple-, and nested-document scenarios remain together in the dedicated suite.
[...]

Issue #8722, opened 2026-09-08
test(agentic-ai): split native real-provider smoke tests into capability suites


gbetances089:
## Current status

PR #8758 implements the capability-split and registry portion of this work.

Completed in PR #8758:

- Native real-provider tests are split into independently selectable capability suites.
- Shared provider metadata, setup, and assertions are reused by those suites.
- JUnit capability tags and a checked-in CI execution registry are in place.
- Registry validation checks provider groups, tags, credential profiles, and execution constraints.
- Blocking CI builds its matrix from the validated registry.
- Strict CI mode fails clearly when no enabled provider is selected.
- Prompt-caching remains available for manual/unsharded runs but is not a blocking provider-sharded requirement.
- The bundle remains a separate Docker-backed test path with all five scenarios. No bundle scenario was removed.
- The live bundle multiple-tool assertion checks tool execution and the final response without assuming a provider-specific model-call count.

Still tracked separately:

- Portable host/container launcher: #8750
- Developer-facing E2E README: #8749
- Final document tool-result migration: #8745
- Final CI runtime/cost measurements and registry balancing: #8721

## Original scope

The native real-provider AI Agent tests were difficult to maintain and schedule independently because one smoke class combined provider definitions, setup, and several capabilities.

The goal is to make native real-provider coverage independently selectable by capability, then use those boundaries to build CI jobs with clear provider, credential, and failure boundaries.

The capability groups are:

- core tool-calling and feedback behavior
- structured output
- reasoning
- prompt caching
- multimodal documents attached to user messages
- documents returned by tools

`AiAgentE2ETestIT` remains a separate bundle-backed suite. It is not a duplicate of the native suites: it checks the packaged Docker runtime, V1/V2 templates, provider wiring, and end-to-end behavior.

## Local provider and capability selection

The same split can be used locally with Maven. First provide the required credentials through the
shell or a secret manager; do not put real credentials in committed files or command examples.

Run one provider group and one capability:

```bash
RUN_NATIVE_LLM_E2E=true \
REAL_LLM_PROVIDER_GROUP=openai \
./mvnw verify \
  -pl connectors-e2e-test/connectors-e2e-test-agentic-ai \
  -Pit-real-llm \
  -Dgroups=reasoning
```

Run two document capabilities for another provider group:

```bash
RUN_NATIVE_LLM_E2E=true \
REAL_LLM_PROVIDER_GROUP=bedrock \
./mvnw verify \
[...]

PR #8758, opened 2026-09-09
test(ai-agent): split real-provider tests by capability

gbetances089:
## What changed

This PR splits the native real-provider Agentic AI tests into separate capability suites so CI can run only the capabilities and providers needed for each job.

It:

- shares the provider catalog, configuration, setup, and assertions between suites;
- adds stable JUnit tags for core smoke tests, structured output, reasoning, prompt caching, multimodal documents, and document tool results;
- adds a checked-in registry that defines CI rows, provider groups, tags, and credential profiles;
- validates the registry and generates the blocking CI matrix from it;
- adds deterministic checks for provider selection and zero-provider failures;
- keeps prompt-caching rows available for manual/unsharded runs instead of making opportunistic cache hits blocking;
- keeps the bundle E2E suite on its separate Docker path with all five scenarios;
- makes the bundle multiple-tool check verify both tools and the final response without requiring an exact number of model calls;
- runs the capability matrix with a maximum of four parallel jobs.

## Important boundaries

- No bundle scenarios were removed or migrated. The bundle tests cover packaged Docker wiring, V1/V2 templates, provider integrations, and end-to-end behavior.
- The live-provider tests do not require a fixed number of model rounds because providers can make the same valid request in different numbers of rounds.
- Deterministic same-round tool-call behavior remains covered by the existing mocked/process test.
- The portable launcher, developer README, and the remaining document-tool-result migration are separate follow-up work.

## Local provider/capability selection

The split can also be exercised locally without running every provider or capability. Set the
needed credentials in the shell or another secret manager first; do not commit them or place them
in command history.

For example, run only the OpenAI reasoning capability:

```bash
RUN_NATIVE_LLM_E2E=true \
REAL_LLM_PROVIDER_GROUP=openai \
./mvnw verify \
  -pl connectors-e2e-test/connectors-e2e-test-agentic-ai \
  -Pit-real-llm \
  -Dgroups=reasoning
```

To run the document capabilities for Bedrock instead, change the provider group and capability
expression:

```bash
RUN_NATIVE_LLM_E2E=true \
REAL_LLM_PROVIDER_GROUP=bedrock \
./mvnw verify \
  -pl connectors-e2e-test/connectors-e2e-test-agentic-ai \
  -Pit-real-llm \
  -Dgroups='multimodal-documents | document-tool-results'
```

These commands call real provider APIs and may require provider and document-judge credentials.
The provider group selects the provider family; the JUnit group expression selects the capability.
Blocking CI additionally sets `REQUIRE_NATIVE_LLM_PROVIDER=true` so an empty selection fails
instead of appearing successful.

## Verification

- The AI Agent capability workflow passed at commit `21cb14165`:
  - registry validation;
  - Vertex, OpenAI, and Bedrock native capability jobs;
  - OpenAI and Bedrock document capability jobs;
  - the bundle E2E job.
- The affected Maven module and `RealProviderSelectionTest` passed locally.
- The bundle job completed with all five scenarios; no bundle scenario was removed.
- Follow-up validation hardening rejects malformed registry shapes with clear errors, keeps the
  reasoning prompt in one place, and uses the canonical `multimodal-documents` capability name.

## Related issues

- Capability split and registry work: #8722
- CI partition and measurement: #8721
- Portable launcher: #8750
- Developer documentation: #8749
- Document tool-result migration: #8745

## Checklist

- [ ] If the change requires a documentation update, it has been added to the appropriate section.


gbetances089:
Addressed all four Copilot findings in commit dede3d6bc5:

- `.github/workflows/AI_AGENT_CPT_PR.yml`: guarded matrix expansion with an empty-array fallback when `setup` is skipped.
- `.github/ai-agent-cpt/registry.json`: removed the unused native Vertex service-account variable; the bundle workflow export remains unchanged for the separate bundle suite.
- `.github/ai-agent-cpt/validate_registry.py`: added required row-key and type validation before matrix rendering.
- `RealProviderPromptCachingE2ETestIT.java`: removed the stale Re-entry Javadoc.

Validation passed: Python compilation, registry validation/rendering, malformed-row rejection, YAML parsing, and `git diff --check`.

gbetances089:
Reviewed the second Copilot round and challenged both findings.

- **Prompt-caching registry rows — fixed in `c930cfec2b`:** valid. The selector intentionally removes prompt-caching providers whenever `REAL_LLM_PROVIDER_GROUP` is set, so the three `ci: false` manual rows would otherwise select zero providers if used with their provider groups. Their `providerGroup` values are now empty so they remain unsharded/manual.
- **Parentheses/negation in registry groups — not applied:** the validator intentionally supports the limited grammar currently emitted by this registry (`tag | tag` and `tag & tag`). Stripping punctuation would accept malformed expressions rather than parse Maven/JUnit syntax correctly. If the registry needs negation or grouping, the validator should gain a real expression parser and tests; broadening validation heuristically would be worse.

Additional internal finding: all parameterized native suites use `allowZeroInvocations = true`, so missing credentials or a selector mismatch can produce a green job with zero provider tests. I did not change that contract in this PR; it should be tracked as a follow-up if blocking coverage must guarantee at least one executed provider.

gbetances089:
Addressed the zero-test false-green risk in commit `63a49bf6e2`.

The blocking workflow now sets `REQUIRE_NATIVE_LLM_PROVIDER=true`. Capability provider selectors fail with a clear error when they resolve to zero enabled providers. The flag is opt-in, so local/manual runs with partial credentials retain their existing behavior. Added a deterministic selection test covering the strict prompt-caching case.

Validation: `RealProviderSelectionTest` 4/4 passed; registry validation, YAML parsing, and `git diff --check` passed.

gbetances089:
Addressed the latest review findings in `237ac4992`:

- Removed the duplicate tool-result document scenario from `RealProviderMultimodalE2ETestIT`; that coverage remains in `DocumentToolCallResultsIT`.
- Made the bundle row declare all provider/judge credential profiles it needs.
- Changed provider-secret imports to use `matrix.credential-profiles` exclusively.
- Removed `structured-output` from the Vertex and Bedrock native rows. The Java provider catalog has no enabled structured-output provider for either group; strict CI correctly exposed that registry mismatch.

The failures in run [34355769540](https://github.com/camunda/connectors/actions/runs/34355769540) had two different causes:

- Vertex native: the structured-output row had zero eligible providers and failed the new strict guard.
- Bedrock native: the same structured-output mismatch occurred, and the core tool-call scenario separately hit `converseResponse() called before the stream completed (no messageStop event received)` for `openai.gpt-oss-120b-1:0`. The latter is a provider/runtime streaming failure, not caused by the registry split, and remains separate.

Validation after the fix: registry matrix rendering, `git diff --check`, and `RealProviderSelectionTest` (4/4) passed.

gbetances089:
Addressed in `21cb14165bfbc940e746d6eccbfe6e91e04d53e2`. Strict mode now reports a dedicated error when no enabled providers are selected, instead of describing the base selection as lacking an `any` capability. Added deterministic coverage in `RealProviderSelectionTest`; the capability-specific error remains unchanged.

gbetances089:
Addressed in `8fb69583f`. The registry validator now rejects malformed root values, non-object rows, non-string credential-profile entries, and non-string credential-list entries with clear validation errors instead of uncaught type errors. The reasoning system prompt is now defined once, and multimodal strict-mode errors use the canonical `multimodal-documents` tag name. Targeted provider-selection tests and malformed-input validation checks pass.

mathias-vandaele (approved):

PR #8765, opened 2026-09-09
fix: validate static auth credentials against endpoint

mathias-vandaele:
This pull request updates the authentication validation logic for REST connectors, making the validation more robust by actively testing credentials against real endpoints. The main changes include switching from static validation for basic, bearer, and API key authentication to actually making HTTP requests to the configured endpoint, refining error handling, and updating tests to reflect the new behavior.

**Authentication validation logic changes:**

* Basic, bearer, and API key authentication are now validated by making a real HTTP request to the configured endpoint, rather than being marked as "unsupported" by default. Only "No Authentication" is rejected outright.
* The validator now distinguishes between different failure types: a 401 Unauthorized response is treated as a credential failure, while other error codes result in an "unsupported" verdict.
* Error messages and logging have been improved for clarity and sensitivity, ensuring that sensitive details are not leaked in validation messages. [[1]](diffhunk://#diff-372b4a78c04588abbcdc7ff6cf76faac01b3ed0a62bbd63678850389014b3824R16-R37) [[2]](diffhunk://#diff-372b4a78c04588abbcdc7ff6cf76faac01b3ed0a62bbd63678850389014b3824L47-L78)

**Test updates and improvements:**

* Unit tests have been updated to expect the new validation outcomes, including the rejection of "No Authentication" and the real validation of basic, bearer, and API key credentials.
* New integration tests using WireMock validate the behavior of the validator against real endpoints, covering successful authentication, unauthorized responses, forbidden responses, unsupported methods, redirects, and unreachable endpoints.

**Code and documentation cleanup:**

* The documentation and comments in `RestAuthenticationValidator` have been updated to reflect the new validation strategy, clarifying the behavior for each authentication type.
* Test helper methods were simplified to always require an explicit URL, matching the new validation approach.

### DIFF
--- a/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationValidator.java
+++ b/connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationValidator.java
@@ -13,26 +13,29 @@ import io.camunda.connector.api.validation.ConfigurationValidator;
 import io.camunda.connector.http.client.authentication.OAuthConstants;
 import io.camunda.connector.http.client.authentication.OAuthService;
 import io.camunda.connector.http.client.client.apache.CustomApacheHttpClient;
+import io.camunda.connector.http.client.mapper.ResponseMapper;
+import io.camunda.connector.http.client.model.HttpClientRequest;
+import io.camunda.connector.http.client.model.HttpMethod;
 import java.util.Map;
 import java.util.Set;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Only the OAuth client-credentials variant can be checked out-of-band: it carries its own token
- * endpoint, so a token is actually requested. Basic, bearer and API key hold a secret with nothing
- * to present it to, and checking the refresh-token grant would consume a token the provider may
- * rotate — both return {@link ConfigurationValidationResult#unsupported() unsupported} rather than
- * an unverified success.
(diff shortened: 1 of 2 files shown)


connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationValidator.java
```diff
@@ -44,32 +47,47 @@ public class RestAuthenticationValidator
...
+      case ApiKeyAuthentication ignored -> callEndpoint(configuration);
       case OAuthAuthentication oauth -> requestToken(oauth);
       case OAuthRefreshTokenAuthentication ignored -> ConfigurationValidationResult.unsupported();
     };
   }
 
+  private static ConfigurationValidationResult callEndpoint(
+      RestAuthenticationConfiguration configuration) {
+    var request = new HttpClientRequest();
+    request.setMethod(HttpMethod.GET);
```
johnBgood, line 66: ❓ Shouldn't this be taken from the configuration? If the server doesn't support GET we might have false-positive error

johnBgood (commented):
Left 1 comment

connectors/http/http-base/src/main/java/io/camunda/connector/http/base/model/auth/RestAuthenticationValidator.java
mathias-vandaele: method is not part of configuration unfortunately..

johnBgood (approved):

PR #8854, opened 2026-09-11
fix(e2e): harden reusable workflow caller

johnBgood:
## Description

Prevents fork PR synchronizations from invoking the secrets-dependent E2E workflow by applying the fork guard to both supported pull-request actions.

Also explicitly grants the reusable workflow the read access needed for repository and Actions data, plus write access for test report checks. This avoids relying on changing repository defaults.

## Related issues

Follow-up to #8790.

### DIFF
--- a/.github/workflows/NIGHTLY_E2E.yml
+++ b/.github/workflows/NIGHTLY_E2E.yml
@@ -7,6 +7,11 @@ on:
   pull_request:
     types: [ labeled, synchronize ]
 
+permissions:
+  actions: read
+  contents: read
+  checks: write
+
 jobs:
 
   setup:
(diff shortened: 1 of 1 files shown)


johnBgood:
The cache warning does not require a change. `actions/cache` authenticates to the cache service with the runner-provided `ACTIONS_RUNTIME_TOKEN`, not the workflow `GITHUB_TOKEN`; therefore the `actions: read` permission declared here does not make cache saves read-only. Cache write eligibility is determined separately by GitHub from the triggering context (for example, untrusted fork events receive a read-only cache token).

Keeping `actions: read` is intentional: the reusable workflow needs to query Actions job data for failure diagnostics, while granting `actions: write` would add unnecessary privilege. References: [actions/toolkit cache client](https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/cacheHttpClient.ts) and [GitHub dependency caching reference](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching).

chillleader (approved):

PR #8867, opened 2026-09-11
fix(agentic-ai): apply configured default-timeout to native chat-model factories

### LINKED ISSUE #8417: Native Anthropic/OpenAI/Gemini chat-model factories ignore chat-model.api.default-timeout
## Problem

The native (v2) chat-model factories are inconsistent in how they honor the configured default timeout `camunda.connector.agenticai.aiagent.chat-model.api.default-timeout` (default `PT3M`):

- **Bedrock** (`BedrockConverseChatModelFactory`) resolves the timeout via `ChatModelProviderSupport.deriveTimeoutSetting(...)`, which falls back to the configured `default-timeout` when the request supplies no (or a non-positive) timeout.
- **Anthropic** (`AnthropicChatModelFactory`), **OpenAI** (`OpenAiChatModelFactory`), and **Gemini** (`GeminiChatModelFactory`) instead do `connection.timeouts() != null ? connection.timeouts().timeout() : null` and hand `null` straight to the vendor SDK, so they fall back to each SDK's own default and silently ignore the operator-configured `default-timeout`.

So a direct v2 user who sets `chat-model.api.default-timeout` gets it applied on Bedrock but ignored on Anthropic/OpenAI/Gemini.

## Scope note

The v1 → v2 rewrite path (#8409) already compensates at the mapping boundary: `V1ToV2ProviderConfigurationMapperImpl` resolves the configured default via `deriveTimeoutSetting` for every provider, so rewritten legacy jobs keep their previous timeout. This issue is only about **direct v2 configurations**, which the mapper does not touch.

## Expected

All native factories honor the configured `chat-model.api.default-timeout` consistently (as Bedrock already does) when the request carries no positive timeout — e.g. route Anthropic/OpenAI/Gemini through `deriveTimeoutSetting` as well.

## Notes

- Gemini has extra non-positive-timeout handling (it deliberately falls back to the SDK default on `PT0S`/negative); any fix should preserve that intent while still applying the configured default for the "unset" case.
- Unreleased (8.10), so no backward-compatibility constraint.

nikonovd:
## Summary

- Anthropic, OpenAI, and Gemini native (v2) chat-model factories passed an unset or non-positive timeout straight to the vendor SDK builder, silently falling back to the SDK's own default instead of `camunda.connector.agenticai.aiagent.chat-model.api.default-timeout`.
- Bedrock already resolved this correctly via `ChatModelProviderSupport.deriveTimeoutSetting`; the other three native factories now use it too, wired through `AgenticAiNativeProvidersConfiguration`.
- Scoped to direct v2 configurations only. The v1→v2 rewrite mapper already routes every provider through `deriveTimeoutSetting`, so rewritten legacy jobs were never affected.

## Related Issues

closes #8417

## Checklist

- [x] Backport labels are added if these code changes should be backported.


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/gemini/GeminiChatModelFactory.java
```diff
@@ -93,31 +103,22 @@ public boolean supports(ChatModelConfiguration configuration) {
...
     final var model = (GeminiChatModelConfiguration) configuration;
     final var connection = model.googleGemini();
-    final var configuredTimeout =
-        connection.timeouts() != null ? connection.timeouts().timeout() : null;
-    // a non-positive configured timeout (e.g. PT0S, or a negative FEEL result) falls back to
-    // the SDK default rather than being passed through - see toGeminiTimeoutMillis, which
-    // would otherwise clamp it to an unusable 1ms call timeout
     final var timeout =
-        configuredTimeout != null && configuredTimeout.isPositive() ? configuredTimeout : null;
+        deriveTimeoutSetting("Gemini model call", config, connection.timeouts(), LOGGER);
```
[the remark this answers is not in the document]
nikonovd: The decision was deliberate.

HanselIdes, line 107: Could we preserve an explicit expert-user escape hatch for “use the provider SDK default” rather than collapsing every non-positive or absent value into the Camunda default? The current change makes `null`, `PT0S`, and negative values all resolve to `chat-model.api.default-timeout` (normally `PT3M`), so there is no longer any way to leave the vendor timeout unset. That is particularly significant for Gemini, where the previous non-positive handling deliberately retained the SDK default, as noted in #8417.

I think the simple default is good, but the model should represent the expert choice explicitly and expose it through the element template rather than relying on an undocumented duration sentinel. For example:

| Input | Desired behavior |
| --- | --- |
| Timeout omitted | Apply configured Camunda default |
| Positive timeout | Apply explicit request timeout |
| Explicit “use provider default” option | Do not configure a timeout; preserve provider SDK behavior |

Before merging, could we document the **before and after behavior** in this form and verify the escape hatch for every provider/backend combination: Anthropic API/custom/Bedrock Mantle, OpenAI API/custom/Foundry, Gemini API/Vertex AI, and Bedrock Converse? Please also cover direct v2 input and the v1→v2 rewrite path, since the mapper currently materializes the Camunda default and may erase the distinction before the factory sees it.

HanselIdes (changes requested):
It seems like we're taking away expert user freedom to use the SDK default, against @maff's wishes (at least for Gemini)

> Gemini has extra non-positive-timeout handling (it deliberately falls back to the SDK default on `PT0S`/negative); any fix should preserve that intent while still applying the configured default for the "unset" case.


I like the idea of the escape hatch for expert users

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/gemini/GeminiChatModelFactory.java
nikonovd: SDK defaults were a problem beforehand, as they are default client implementations. Now in V2 we have streaming and it might not be a big issue anymore to have SDK defaults, as time-to-first-byte is smaller.

This PR applies the actual default behavior to all clients, I believe also the intent in the ticket was rather AI assistance putting too much effort in the task.

For me it'd be important to align all chat model behaviors equally as it was before and then think of expert escape hatches or timeout handling in general, as it'd introduce another layer of new behavior.

maff: I think the requirement stated in the original issue was overstating the need to keep the Gemini behavior (sorry about that!) - as far as I can tell, the "extra non-positive-timeout handling" means a request can run indefinitely which is something we should avoid for the connnector.

⏩ I'd agree that we should favor consistency here and apply the changes. The expert user escape hatch would basically be to set a custom duration.

maff (approved):
Pre-approving with the added comment (pending fixed merge conflicts).

connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/gemini/GeminiChatModelFactory.java
nikonovd: [ChatModelProviderSupport.deriveTimeoutSetting](https://github.com/camunda/connectors/blob/main/connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/chatmodel/provider/ChatModelProviderSupport.java#L28) filters for non-positive timeouts here and applies the configured default timeout per application properties (default `PT3M`).

From what I see by aligning here we won't go into infinite hanging by that.

HanselIdes: LGTM then - we can drop the requirement

HanselIdes (approved):
Reframing the "expert user escape hatch" as custom timeouts rather than provider SDK timeouts resolves my only concern.

PR #8912, opened 2026-09-15
fix(aws-sns): authorize inbound messages by verified topic

johnBgood:
## Description

Authorize inbound AWS SNS messages using the signature-verified `TopicArn` from the message body instead of relying on the caller-controlled HTTP header. The configured topic check remains before signature verification as a preliminary filter to reject obvious misses without certificate-fetch work, and is repeated against the verified topic before notification delivery or subscription confirmation.

Malformed topic ARN headers are rejected before region extraction. Header and verified-body values are not required to be equal, preserving supported relay behavior when both values satisfy the configured policy.

Regression coverage drives the webhook executable with offline cryptographically signed notifications and subscription confirmations, including structural verification that an unlisted subscription never invokes its callback URL.

## Related issues

Closes camunda/security-testing-findings#264

## Checklist

- [x] Backport labels are added for every maintained stable branch containing the AWS SNS connector.
- [x] No documentation update is required.


johnBgood:
Addressed the QA findings from camunda/team-qa-engineering run 34977787618:

- restored the configured-topic header check as a preliminary gate, preventing unlisted requests from reaching SNS certificate retrieval;
- validated SNS topic ARN structure before extracting the region, eliminating the malformed-header `ArrayIndexOutOfBoundsException`;
- retained the post-verification allow-list check against the signed body `TopicArn`;
- added a real offline-signed `SubscriptionConfirmation` regression that proves an unlisted topic does not invoke its callback URL.

The prefilter and signed-topic authorization tests pass on commit `902c7d89ed`.

Szik (approved):

Issue #8060, opened 2026-07-24
Agentic AI: Anthropic provider — Claude on Microsoft Foundry (Azure)


maff:
Anthropic provider — Claude on Microsoft Foundry (Azure) backend, using the Anthropic Messages wire format.

Auth: API key or Microsoft Entra ID (client credentials / managed identity). **HTTP proxy support mandatory.**

Backend of the Anthropic provider (#7218).

nikonovd:
closed with https://github.com/camunda/connectors/pull/8957

PR #8957, opened 2026-09-16
feat(agentic-ai): Claude on Microsoft Foundry for the native Anthropic provider

nikonovd:
## Description

The AI Agent can now run Anthropic (Claude) models on Microsoft Foundry (Azure AI Foundry).

A modeler picks one of three authentication strategies: API key, Entra ID client credentials, or Entra ID managed identity. Managed identity stays blocked on SaaS. Bean validation enforces that block on both providers.

One `FoundryAuthentication` model and one `FoundryCredentialResolver` now serve both providers. They replace the OpenAI-only `OpenAiFoundryCredentialResolver`.

The v2 element templates carry the new backend, for the task and sub-process flavors in both main and hybrid form. `docs/reference/native-providers.md` documents it. The shared Model field of both providers points Foundry users at their deployment name, which Foundry uses in place of the provider's model ID.

## Related issues

closes to #8060

## Checklist


connectors/agentic-ai/connector-agentic-ai/src/main/java/io/camunda/connector/agenticai/aiagent/model/request/v2/AnthropicChatModelConfiguration.java
```diff
@@ -241,6 +243,77 @@ public String toString() {
       }
     }
 
+    @TemplateSubType(id = FOUNDRY_ID, label = "Microsoft Foundry (Azure)")
```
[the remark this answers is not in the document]
nikonovd: Confirmed and fixed in 2b47956938. Microsoft's docs are explicit that the deployment name goes into the `model` field on both Foundry surfaces (`/anthropic/v1/messages` and the unified `/openai/v1`), and that it only defaults to the model ID.

Added a `tooltip` to the shared Model field on both providers (Anthropic and OpenAI, since both gained a Foundry backend in this PR): "On the Microsoft Foundry backend, enter the deployment name instead. It defaults to the model ID, but can be chosen freely when the deployment is created." Tooltip rather than `description` per the repo's field-hint rule (`tooltip` > `placeholder` > `description`), and a separate backend-conditional field would fragment the shared model group for a guidance-only problem.

Templates regenerated; no version bump, since guidance text does not change how the template applies.

Not touching the Bedrock backend, where the same catalog-vs-Bedrock-model-ID mismatch predates this PR.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/ProviderWireFormatSmokeTests.java
```diff
@@ -82,6 +83,7 @@ static Stream<ProviderWireFormatFixture> fixtures() {
         new OpenAiCompletionsV2WireFormatFixture(),
         new AnthropicMessagesV1WireFormatFixture(),
         new AnthropicMessagesV2WireFormatFixture(),
+        new AnthropicMessagesFoundryWireFormatFixture(),
```
maff: 💭 Not necessarily against adding this here, but what is the benefit of specifically testing anthropic on foundry here, given it just hits wiremock on another path? 🤔 So while if definitely makes sense cover it e2e on the real provider tests, does it add value to have a dedicated setup on these wire format tests? Then we probably should extend these to other scenarios as well (e.g. OpenAI on foundry).

maff (approved):
LGTM! 🚀 Non-blocking comment on the added e2e wire format tests.

connectors-e2e-test/connectors-e2e-test-agentic-ai/src/test/java/io/camunda/connector/e2e/agenticai/aiagent/wiremock/ProviderWireFormatSmokeTests.java
nikonovd: Agree, no dedicated fixture needed. Foundry's wire format is byte-identical to direct Anthropic — same Messages API, same request/response converters, same SSE streaming. Only differences are URL path (/anthropic prefix, appended by the Anthropic SDK's own FoundryBackend) and auth header selection (x-api-key for API-key auth, Authorization: Bearer for Entra ID) — both handled by the vendor SDK, no Camunda-side transform.

nikonovd:
@gbetances089 I assigned this to you for QA approval. Please check when you're free! 🙇‍♂️

gbetances089:
## Validation completed

We tested the Anthropic Microsoft Foundry API-key path with a real Foundry resource.

- Deployed the PR Connector Runtime image to a local Self-Managed Helm environment.
- Started an AI Agent process using the Anthropic Foundry backend, the configured Claude deployment, and the two Foundry values from Vault.
- The AI Agent jobs completed successfully. This confirms secret resolution, the Foundry endpoint format, API-key authentication, and the real model call.
- Ran the same focused real-provider CPT in GitHub Actions. It passed:
  https://github.com/camunda/connectors/actions/runs/35342312260

The API-key path is now covered by the real-provider test work.

Follow-up work for Entra client credentials, managed identity, and Claude Sonnet 5 quota/model availability is tracked in #9010.

gbetances089 (approved):

PR #8974, opened 2026-09-17
fix(agentic-ai): add No Auth option for OpenAI custom/compatible endpoint

### LINKED ISSUE #8870: Agentic AI: review no-auth and other gaps for custom/compatible OpenAI-style endpoints
Parent: #7211

__What should we do?__

Review the "Custom / compatible endpoint" backend of the OpenAI provider (and related custom-endpoint paths) in the new v2 model provider configuration for authentication gaps, and audit for other UX gaps in custom/compatible endpoint support.

Concretely, a self-hosted, no-auth OpenAI-compatible server (e.g. LM Studio) cannot currently be configured:

- With Provider `OpenAI` + Backend `Custom / compatible endpoint`, both **API endpoint** and **API key** are required fields with validation errors ("API key must not be empty") — there's no "No Auth" option, even though many local/self-hosted OpenAI-compatible servers (LM Studio, Ollama, llama.cpp server, etc.) don't require an API key at all.
- The separate, legacy LM Studio-specific `Provider Type` entry is a different code path with its own UX, which adds to the inconsistency between how "native" providers and generic custom/compatible endpoints are configured.

The current workaround is to select `OpenAI` + `Custom / compatible endpoint` and fill in a dummy API key value, but this isn't good UX for genuinely unauthenticated backends.

__Why should we do it?__

Custom/compatible OpenAI-style endpoints are meant to support arbitrary self-hosted and third-party OpenAI-API-compatible servers, several of which (LM Studio among them) don't require authentication. Forcing users to supply a fake API key is confusing and error-prone, and blocks a legitimate configuration. As part of this issue:

- Assess whether "No Auth" should be added as a valid authentication option for the Custom/compatible endpoint backend (and any other backends that plausibly need it).
- Audit the custom/compatible endpoint path more broadly for other configuration gaps relative to the native provider backends (e.g. defaults, validation messages, parity with the legacy LM Studio provider type).

__Context__

This originated from a user (Markus Hafellner) reporting in Slack that they could not configure LM Studio as a custom/compatible OpenAI endpoint in the new AI Agent connector template because of the required API key validation, together with screenshots of the Model Provider config panel showing the validation errors. Dmitri Nikonov, who works on this area, agreed this is a UX gap: "I agree this isn't good UX and No Auth should likely be a valid option."

maff:
## Summary
- Self-hosted OpenAI-compatible servers (LM Studio, Ollama, llama.cpp, etc.) often need no authentication at all, but the `custom`-backend's API key field was mandatory, blocking these configurations.
- Adds a `None` authentication choice to `OpenAiCustomEndpointAuthentication` (default), alongside existing `apiKey` and OAuth 2.0 client-credentials variants.
- Backed by a placeholder credential internally (openai-java's client builder requires some credential source to build at all — confirmed still true on the pinned SDK version), invisible in the element template/UI. Unauthenticated servers simply ignore the resulting header.
- Mirrors the existing Anthropic custom-backend UX (`None`/`API key`/`OAuth 2.0`), closing the parity gap called out in the issue.

Closes #8870

## Test plan
- [x] `OpenAiChatModelFactoryClientTest`: new WireMock-verified test asserting the placeholder `Authorization` header is sent for the `None` variant
- [x] `OpenAiChatModelFactoryTest` and `OpenAiChatModelFactoryClientTest` full suites green
- [x] Element templates regenerated (`mvn process-classes`), diff reviewed — "None" choice added, default flipped to `none`, no placeholder details leaked into the template


nikonovd (commented):
LGTM 🚀 

💭 I think we need a similar approach for Anthropic compatible backends, as most of the gateways also support their API protocol.

Please also check to set the correct backport label 😉

nikonovd (approved):
Forgot to approve!

PR #8976, opened 2026-09-17
test(http): annotate HttpTests with @SlowTest to skip it in the quick CI lane

maff:
## Summary

`HttpTests` (`connectors-e2e-test-http`) is a `@CamundaSpringProcessTest` (real Zeebe engine via testcontainers + Spring Boot). Every other `@CamundaSpringProcessTest` class in the repo carries `@SlowTest` (`io.camunda.connector.test.utils.annotation.SlowTest`), which disables the test when Maven is run with `-Dquickly` (see the CI "Build Connectors" step in `.github/workflows/TEST_FEATURE_BRANCH.yml`). `HttpTests` was the one exception, so it always ran even in the "quick"/fast CI lane.

Found while debugging an unrelated agentic-ai PR's CI failure — the quick lane job failed on a flaky/real-server error in `HttpTests` (multipart document parsing, HTTP header parse exception), which surfaced a real-engine e2e test running inside what's supposed to be the fast lane.

No issue number — none exists for this.

## Changes

- Added `@SlowTest` (+ import) to `HttpTests`, matching the placement used by sibling classes in the same module (`PollingConnectorTests`, `WebhookNotActivatedDocumentTests`, `WebhookPhysicalTenantScopingE2ETest`).
- Test-annotation only; no behavior change. Confirmed:
  - `mvn -pl connectors-e2e-test/connectors-e2e-test-http test -Dtest=HttpTests` → 16/16 pass (unchanged).
  - `mvn -pl connectors-e2e-test/connectors-e2e-test-http test -Dtest=HttpTests -Dquickly -DfailIfNoTests=false` → 16/16 skipped.
- Checked the rest of `connectors-e2e-test-http` for the same gap — the other 3 `@CamundaSpringProcessTest` classes already had `@SlowTest`.

## Test plan

- [x] `mvn test -Dtest=HttpTests` passes normally
- [x] `mvn test -Dtest=HttpTests -Dquickly` skips all 16 tests


johnBgood:
@maff This annotation was intentionally added. We considered this test to be important enough to always run. If there's flakiness, we should address that instead (I've seen a lot of 'Address already used' errors since I merged three tests into the same class).

maff:
Thanks for the feedback, closing if this a deliberate decision!

PR #8987, opened 2026-09-17
fix(webhook): bound inbound request body reads and add per-webhook rate limit

johnBgood:
## Summary

Fixes an unauthenticated memory-exhaustion / resource-exhaustion path in the inbound webhook endpoint, reported as `camunda/security-testing-findings#268` (medium severity, CWE-770 / CWE-789).

Previously, `InboundWebhookRestController.dispatch(...)` read the full request body via `getInputStream().readAllBytes()` **before** resolving the path to a registered webhook and before any authentication ran. Any `/inbound/{anything}` — registered or not — would have its entire body buffered into heap first, with no size limit applying to a direct stream read (Tomcat's `maxPostSize` and Spring's multipart limits do not apply here), and no inbound rate limiting existed anywhere in the runtime.

## Changes

- **Bounded body read**: `readAllBytes()` replaced with `readNBytes(maxRequestBodyBytes)` plus a one-byte overflow peek (same idiom already used in `ResponseMappers.java` for outbound HTTP responses). Oversized bodies are rejected with `413 Payload Too Large` before more than a single byte past the limit is ever buffered — this applies to every content type, closing the gap that Tomcat/Spring's existing limits don't cover.
  - New property: `camunda.connector.webhook.max-request-body-bytes`, default **10 MB** (matches Spring Boot's own multipart default).
- **Path resolved before the body is touched**: the connector lookup now happens first; an unregistered path returns `404` without reading any of the body at all.
- **Per-registered-webhook rate limit**: added using Guava's `RateLimiter` (already a `connector-runtime-spring` dependency), checked immediately after path resolution and before the body is read. Throttled requests get `429` without any body allocation.
  - Keyed by the resolved connector's `ExecutableId`, not the raw request path, so the limiter map can only grow to the number of *registered* webhooks — never to the number of distinct paths an unauthenticated caller can probe.
  - New properties: `camunda.connector.webhook.rate-limit.enabled` (default **true**) and `camunda.connector.webhook.rate-limit.permits-per-second` (default **1000** per webhook).
- All three rejection paths (`404`/`413`/`429`) return an empty body, consistent with the existing `WebhookSecurityException` convention of not leaking details on security-relevant rejections.

### On the 1000/s default
An initial, more conservative default (50/s, matching the existing JWKS-fetch rate limit precedent elsewhere in this codebase) broke the existing `WebhookControllerTestExceptionZeebeTest` suite, which legitimately fires ~16 near-simultaneous requests at one registered webhook. 1000/s passes that suite comfortably while remaining far below the flood volumes described in the finding; operators who know their expected per-webhook traffic can tune it down.

## Test plan

- [x] `InboundWebhookRestControllerTest` (connector-runtime-spring): **9/9 pass** — 5 pre-existing + 4 new (size boundary, oversized-body rejection without invoking the connector, unknown-path rejection proven via a request whose `getInputStream()` throws if ever called, and rate-limit throttling with an enabled/disabled variant).
- [x] `connector-runtime-spring` full suite: **574/575 pass**.
- [x] `spring-boot-starter-camunda-connectors` full suite (real Spring context, so `@Value` bindings are exercised, including `WebhookControllerTestExceptionZeebeTest`): **158/163 pass**.
- [x] `spotless:check` on both touched modules: clean.
- [ ] **1 unrelated failure across both runs**: `SearchQueryClientTest` and a handful of other Testcontainers-based tests (`WebhookControllerTestZeebeTest`, `ObjectMapperQualifierTest`, `JobRetriesIntegrationTest`, `RuntimeStartupWithConnectorsFromEnvVarsTests`, `RuntimeStartupWithConnectorsFromSpiTests`) fail to pull `camunda/camunda:8.11-SNAPSHOT` in this sandboxed CI-less environment (registry/network access, not a code issue) — pre-existing and reproducible on unrelated, untouched test classes, unaffected by this change.

closes camunda/security-testing-findings#268


johnBgood:
The camunda/security-testing-findings#268 fix in this PR has been superseded by #9012, which contains only the security-fix commit (`92ebb3c9b8`) cut cleanly from current `main`.

This branch (`relay/f336509b`) also carries 6 unrelated, unmerged commits (search-client-registry / import-scheduler work — `38dc6159d7`..`253b42a5e2`) that predate the security fix and now conflict with `main`. Those are untouched here and still need to be rebased by whoever owns that work; this comment doesn't close or modify this PR or its commits.

Recommend reviewing #9012 for the #268 fix instead of this PR.

johnBgood:
Closing this PR: it has been superseded by #9012, which contains only the camunda/security-testing-findings#268 fix, cut cleanly from current `main` and with all CI checks passing.

This PR should not be merged — it still bundles the security fix together with 6 unrelated, unmerged commits (search-client-registry / import-scheduler work on this branch) that now conflict with `main`. Those commits and this branch are left untouched by this close; only the PR itself is closed. Whoever owns that unrelated work can reopen this PR or open a fresh one from `relay/f336509b` once it's rebased.

PR #9005, opened 2026-09-18
test(agentic-ai): run Foundry CPT in CI

gbetances089:
Depends on #8957.

Adds an opt-in, internal-only CI workflow for the Anthropic Microsoft Foundry API-key CPT. It reads the approved QA values from Vault only when a trusted PR has the `real-llm-e2e` label, or when manually dispatched after the workflow reaches the default branch.

The workflow runs the smallest real-provider slice and does not expose credentials.

Do not merge before #8957; this branch is based on that feature PR.


gbetances089:
Closing this temporary standalone workflow approach. The Foundry real-provider CPT should instead be added to the existing capability-split CPT registry and `AI_AGENT_CPT_PR.yml` flow described in #8758, as tracked in #9010.

